Issue #1, opened 2024-03-05
Create a CI workflow to validate fmt, clippy, and tests on PR's


jcrossley3:
Existing tests are more integration than unit, i.e. they require postgres to run successfully

PR #3, opened 2024-03-05
Fix formatting and clippy errors

jcrossley3:
More work toward #1


bobmcwhirter (approved):

backend/api/src/system/package/mod.rs
```diff
@@ -927,330 +928,283 @@ mod tests {
         Ok(())
     }
 
-    #[cfg(test)]
-    mod tests {
```
jcrossley3, line 931: @bobmcwhirter This looks like more changes than it is. I just moved all the tests beneath a single `tests` module.

backend/common/src/purl.rs
```diff
@@ -128,15 +128,18 @@ impl Debug for Purl {
...
-impl From<&&str> for Purl {
-    fn from(value: &&str) -> Self {
-        PackageUrl::from_str(value).unwrap().into()
-    }
-}
+// impl From<&&str> for Purl {
+//     fn from(value: &&str) -> Self {
+//         PackageUrl::from_str(value).unwrap().into()
+//     }
+// }
```
jcrossley3, line 142: @bobmcwhirter I thought this looked weird. It's not used, but I forgot to delete it.

backend/common/src/purl.rs
```diff
@@ -128,15 +128,18 @@ impl Debug for Purl {
 
 impl From<&str> for Purl {
     fn from(value: &str) -> Self {
-        PackageUrl::from_str(value).unwrap().into()
```
jcrossley3, line 131: @bobmcwhirter We're using a clippy rule that doesn't like `unwrap`. We only allow them for tests. I don't know if deriving `Default` is The Right Thing in this case or not, but seemed harmless.

bobmcwhirter: I think the answer is TryFrom instead since failure is an option.

PR #4, opened 2024-03-05
Frontend skeleton

carlosthe19916:
Not sure how we will organize ourselves in terms of reviews and stuff like that but here is a first start of the UI.


jcrossley3 (commented):
I'm surprised there are >100 files. Are some of these external libraries? Is it typical that they're committed as a part of our project?

carlosthe19916:
> I'm surprised there are >100 files. Are some of these external libraries? Is it typical that they're committed as a part of our project?

The directory that contains the highest number is 'components'.  Currently it contains all components I've used in the POC but I think I can clean that directory but don't expect the number of files to be drastically reduced xD . The frontend has multiple small files rather than few but large files

carlosthe19916:
And no, there are no library's code. But as I said I'll clean it as much as I can for it to be a good enough skeleton.

jcrossley3:
> And no, there are no library's code. But as I said I'll clean it as much as I can for it to be a good enough skeleton.

No worries. I was just making sure you really created all those files. :smile:

.github/workflows/frontend.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: npm clean-install --ignore-scripts
+      - name: Lint sources
+        working-directory: frontend
+        run: npm run lint
+      - name: Build
+        working-directory: frontend
+        run: npm run build
+      - name: Test
+        working-directory: frontend
+        run: npm run test -- --coverage --watchAll=false
```
jcrossley3, line 38: Maybe use a default `working-directory` for all the steps?

```suggestion
    defaults:
      run:
        working-directory: backend

    steps:
      - uses: actions/checkout@v4
      - name: Use Node.js
        uses: actions/setup-node@v4
        with:
          node-version: 18.x
      - name: Verify package-lock.json
        run: ./scripts/verify_lock.mjs
      - name: Install
        run: npm clean-install --ignore-scripts
      - name: Lint sources
        run: npm run lint
      - name: Build
        run: npm run build
      - name: Test
        run: npm run test -- --coverage --watchAll=false
```

jcrossley3 (changes requested):

.github/workflows/frontend.yml
carlosthe19916: Thanks I didn't know about this trick!

carlosthe19916:
what about the current number now `100` sharp. Couldn't reduce it more xD

jcrossley3 (approved):

ctron (changes requested):
Welcome to the mess called JS …

What really worries me is the presence of a "server". We're trying to not split things up and keep it simply. We do we need another server? And why does it need to be NodeJS? I would expect that the only server we need is an `actix` (axum, warp, whatever) endpoint, serving those static files.

frontend/Dockerfile
```diff
@@ -0,0 +1,38 @@
+# Builder image
```
ctron, line 1: I guess this should be renamed to `Containerfile` …

carlosthe19916:
> Welcome to the mess called JS …

Honestly I don't see any mess. I smell personal bias in your comment, sorry.

> What really worries me is the presence of a "server". We're trying to not split things up and keep it simply. We do we need another server?

We need a server where the UI files will be served. Serving those files in the Backend server is just a bad practice and takes us back to a risk pattern where Backend and Frontend are tightly coupled.

A small good app would be composed by:
- DB
- Backend
- Frontend

Each component independent from the other and running in their own process.

You seems to be suggesting to reduce the components to:
- DB
- Backend (Frontend embedded here)

Not because it is possible it means it is advisable or efficient.


> And why does it need to be NodeJS?

The trustification repo was using Ngix. So we can debate about NodeJS vs Ngix vs Other. We can discuss about preferences and so on. But given the requirements I've taken into account i believe NodeJS serves better our use case.


> I would expect that the only server we need is an `actix` (axum, warp, whatever) endpoint, serving those static files.

As explained above, what you are suggesting is not just to `serve static files with axtic`. What you are really suggesting is to use the same server to serve UI and Backend together in the same server.

- This is bad practice (e.g. if using k8s, are we going to re-deploy the whole backend for each front-end change? Killing the backend just because the frontend needed to change a minor thing like colors is not reasonable). 
- Will serving backend and frontend in the same server will make our app faster or scalable?

The real question to me would be: `is serving the ui and backend in the same server` a requirement or just a preference?
- If it is a requirement then sure we can do it
- Otherwise I would advice to use good practices and separate frontend/backend

ctron:
> > Welcome to the mess called JS …
> 
> Honestly I don't see any mess. I smell personal bias in your comment, sorry.

Absolutely. I simply prefer a cleaner build system and a less cluttered structure of projects. But yes, that's my personal opinion and it should not have been part of the PR. Sorry for that.

Still, the presence of the server worries my a lot. As you said, we need some HTTP endpoint to serve the frontend. Trustification does this using NGINX, which is just used as a random, lightweight HTTP server, serving static files.

In Seedwing we embedded this into the binary. Removing the need for nginx. And the goal for trustify is to reduce the number of components, not increase them.

Additionally, the idea is that some non-technical person can run this on their laptop. Not even using Docker. Works great when we embed this into a server binary, But when those users also need to install a node JS runtime, I am not sure that's a good approach.

So it's ok for me that the server is JS. But for the backend, it is not.

carlosthe19916:
> > > Welcome to the mess called JS …
> > 
> > 
> > Honestly I don't see any mess. I smell personal bias in your comment, sorry.
> 
> Absolutely. I simply prefer a cleaner build system and a less cluttered structure of projects. But yes, that's my personal opinion and it should not have been part of the PR. Sorry for that.
> 
> Still, the presence of the server worries my a lot. As you said, we need some HTTP endpoint to serve the frontend. Trustification does this using NGINX, which is just used as a random, lightweight HTTP server, serving static files.
> 
> In Seedwing we embedded this into the binary. Removing the need for nginx. And the goal for trustify is to reduce the number of components, not increase them.
> 
> Additionally, the idea is that some non-technical person can run this on their laptop. Not even using Docker. Works great when we embed this into a server binary, But when those users also need to install a node JS runtime, I am not sure that's a good approach.
> 
> So it's ok for me that the server is JS. But for the backend, it is not.

That is a good point!
So the real question is:

`3 layers ("DB" + "Backend(rust) + Frontend(JS)")` vs `2 layers (DB + Backend(Based on rust and also serving UI))`

### 2 Layers:
Pros:
-  Easy to consume in a laptop (single command like `./runMyApp`)
Cons:
- Not good for k8s environment as Frontend and Backend are together in one (one dies, then both dies)

### 3 Layers:
Props:
- Good for k8s environments (frontend can die and won't affect backend)

Cons:
- Running in a laptop would need to follow 2 steps: ./runMyApp (for the backend) and then another command for starting the UI

The question to me is what is more important:
- Not to bother at all to a user in his laptop and allow him to run single commands but be weak in k8s environments OR
- Bother a bit to the user in his laptop but be strong and scalable in K8s environments


## What about the DB?
What are going to the DBs supported? If we will support e.g. Postgresql then wouldn't we need to use Docker to start the DB anyway? Therefore the idea of a single command for starting the whole app would not be real unless we will have a DB like SqlLite (good for a laptop but bad for k8s)

I fully understand the idea of making it simple for a person to start the app; however, I think just like the DB (`docker run postgresql`), the UI will need to be started as a separate process (`docker run frontend`)

Thanks for the questions as it really made think about what are the real general requirements.
- DB requirements (SqlLite vs other?)
- OIDC requirements (Keycloak vs other)
- Storage (Minio vs other)
- Frontend

Are we going to have an architecture for a laptop (e.g. sqlite + noAuth + noMinio but local filesystem) and a different one for k8s (e.g. postgresql + keycloak or other + minio or s3)?

My expectation, considering the times we have, is to have a single way of deploying everything in a laptop and in k8s. Then consider what is more important to us, the laptop user or a good k8s architecture?

Open to discuss it further

ctron:
> Not good for k8s environment as Frontend and Backend are together in one (one dies, then both dies)

That's not a real argument IMHO. If something dies, it gets restarted by k8s. If the backend dies, the frontend won't work either. Additionally, a nodejs runtime consumes much more memory and CPU than a static HTTP server. So that would actually be one more argument against it.

To my understanding the current goal is to have a single binary (not container). If there are technical reasons that speak against it, we should definitely consider those. But to my understanding the UI will be a static set of files which need to be served. So I don't see it.

About the DB, I think marco had the idea to use SQlite for this case. Storage will be the filesystem anyway. When it comes to OIDC something like `oxide-auth` could work.

carlosthe19916:
> > Not good for k8s environment as Frontend and Backend are together in one (one dies, then both dies)
> 
> That's not a real argument IMHO. If something dies, it gets restarted by k8s. If the backend dies, the frontend won't work either.

In a 3 layer arch, if the frontend dies, the backend can still live and continue serving anything running there like (rest apis, notifications being sent, running jobs, etc.). To me this is a real argument.

> Additionally, a nodejs runtime consumes much more memory and CPU than a static HTTP server. So that would actually be one more argument against it.

Let's define resource constrains and we can take it from there. E.g. the constrains is `X megabytes of memory, Y number of CPU, etc.` Just saying "it consumes more cpu and memory" does not make sense unless we have real numbers. resources need to be compare in terms of `how much i can afford vs how much the chosen tools consume` then evaluate tools from there. 

> 
> To my understanding the current goal is to have a single binary (not container). If there are technical reasons that speak against it, we should definitely consider those. But to my understanding the UI will be a static set of files which need to be served. So I don't see it.
> 
> About the DB, I think marco had the idea to use SQlite for this case. Storage will be the filesystem anyway. When it comes to OIDC something like `oxide-auth` could work.

So If I understand the single binary will serve a DB (sqllite) + backend (fronent embedded) into it. Sounds good, but it means we will need to follow a different approach when deploying it to k8s as we can not deploy sqllite in k8s (we need something like postgresql or equivalent), same thing to other components like storage (s3, minio, etc). Then are we ready to maintain and fully test 2 ways of deploying the UI? I would love to listen the opinion of the other members of the team as to me that sound bad. My opinion might not be popular in this regard, but think if we are turning extremes here

- One extreme: over engineer everything (everything is a microservice)
- Another extreme: everything is a single (literally a single) executable

That sounds like a second failure in architecture design to me.

ctron:
So again, unless I see a real reason to have a JavaScript backend, I am against adding one.

jcrossley3:
> So again, unless I see a real reason to have a JavaScript backend, I am against adding one.

FWIW, I tend to agree, but I like to think I have an open mind.

I will say that when @bobmcwhirter created the repo structure and included both `frontend/` and `backend/server/` directories, I was a bit taken aback, and @carlosthe19916 would be forgiven for taking that strong signal to include a nodejs server in this PR. :smile:

So let's enumerate the benefits. What is it buying me? How will it simplify my dev cycle? At this stage of the project, any arguments around future k8s deployments won't sway me. I'm only looking to add features fast. And I believe those features will require a fairly tight coupling of both ends, front and back. 

I could be wrong, though, or more likely correct now, but wrong later. Can we have our cake and eat it, too? Can we make it easy enough to serve the static frontend files both through the backend server or alternatively through its own?

What's the simplest thing that will work for us right now?

ctron:
I am open to some kind of "local dev mode" module, if that doesn't bother the backend in any way. But having the option today to serve this from a static endpoint too.

But having a `frontend/server` module in the architecture seems wrong.

helio-frota:
I like this :+1: 



TBQH I believe front-end should live in another repository, but we have improvements here at least.

And I do like these lines from Jim:

* How will it simplify my dev cycle?
* I'm only looking to add features fast.
* What's the simplest thing that will work for us right now?

carlosthe19916:
To me is really important productivity. Everyone working on their side of the project should be productive and not distracted by irrelevant things.

## Dev cycle
Mixing backend and frontend in dev mode will just kill the frontend side productivity. There 100 times better tools for dev mode in the JS world compared to Rust.

@ctron I think in regards to the Dev cycle you agree separating frontend and backend completely `I am open to some kind of "local dev mode" module`

That is a progress I think.

## Deployment cycle
Here is where we disagree. I'm proposing to deploy (production level) our artifacts in a way that the DB, Backend, And frontend run its own process (3 containers for instance). 

@jcrossley3 
> Can we have our cake and eat it, too? Can we make it easy enough to serve the static frontend files both through the backend server or alternatively through its own?

We can, I mean nothing is impossible. But to me it adds complexity and I like this quoting "Complexity kills productivity".
I would rather invest time doing one or the other but not both ways (at least not now that we want to be "fast and furious" releasing the next generation of Trustification). My experience tells me to go for what is proven to work and not to go in an adventure to discover what "might/might not" work in the short and also in the long term.

Guys, I really want to progress things as efficient as possible for everyone. If you tell me "having ui and backend together in a single server/process running is a requirement" I will do it, that's my job.

carlosthe19916:
Let's not turn this into a battle please. We all are a team, we just need to have some fair agreements. I'm open to change my mind. Should we have an internal chat about this? Also, I can definitely rely on your opinion when it comes to take arch decision as you are more years on the field so I trust your judgement.

bobmcwhirter:
Here's my random thoughts...

Can we get to a single process? Probably not, because forcing compatibility with sqlite instead of an external pgsql brings us to a lowest-common-denominator, which is suboptimal for such a data-intensive application.

Are 3-tier non-microservices deployments well understood? I think yes?

For dev cycle, whatever is most efficient would be my vote, and @carlosthe19916 knows how he's efficient.

I also think the backend folks can figure out a way to make their dev cycle work well enough.

And packaging-for-PM ease can hopefully be solved orthogonally, but PM will still need to run at least 2 containers, yeah? (pgsql & backend) so is 3 worse? (It may be).

So I'd prefer to work in the most straight-forward and efficient way for each end of the stick until/unless we discover bonafide usability issues.

ctron:
> And packaging-for-PM ease can hopefully be solved orthogonally, but PM will still need to run at least 2 containers, yeah? (pgsql & backend) so is 3 worse? (It may be).

I don't think we would need any container. Spin up the `.exe` (or whatever) and that's it. We've done that before.

And I didn't yet hear any reason why adding a whole new stack on the backend is necessary.

bobmcwhirter (approved):
Going from zero-to-one (nothing to something) is a great start. We can iterate and refine as we discover concrete problems.

PR #15, opened 2024-03-06
chore: stick with node lts

helio-frota:


bobmcwhirter (commented):
I have expertise in node versions. Defer to @carlosthe19916

frontend/package.json
```diff
@@ -32,8 +32,7 @@
     "server"
   ],
   "engines": {
-    "node": ">=18.14.2",
-    "npm": ">=9.5.0"
+    "node": "^20"
```
carlosthe19916, line 35: If we are updating the node version we should also update the npm version rather than just deleting it
e.g. `npm: >=10.2.0`

carlosthe19916 (changes requested):
@helio-frota please also change this file https://github.com/trustification/trustify/blob/main/frontend/Dockerfile with the appropriate image versions

carlosthe19916:
@helio-frota on another topic: thanks for upgrading versions. However, although important this should not be done on a regular basis (not every day or week) unless necessary. Let's try to focus on app features and let's leave what it works, "if it works let's don't touch it xD".

frontend/package.json
helio-frota: since npm comes with node it makes no much sense to have both but I'll add back :+1:

PR #21, opened 2024-03-07
test: tests trustify-cli help

helio-frota:
* Adds pretty_assertions as dev-dependency for easy cli-diff-tests-results visualization

### DIFF
--- a/backend/cli/Cargo.toml
+++ b/backend/cli/Cargo.toml
@@ -14,3 +14,6 @@ log = "0.4.19"
 serde_json = "1.0.104"
 thiserror = "1"
 tokio = { version = "1", features = ["full"] }
+
+[dev-dependencies]
+pretty_assertions = "1.4.0"
--- a/backend/cli/tests/cli.rs
+++ b/backend/cli/tests/cli.rs
@@ -1,0 +1,28 @@
+use pretty_assertions::assert_eq;
+use std::env;
+use tokio::process::Command;
+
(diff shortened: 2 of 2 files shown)


helio-frota:
/me trying to fix vscode that failed with fmt and clippy :+1:

helio-frota:
motivations/reasons I found:

* pretty_assertions as `dev-dependency` will not hurt (in case used) but we can remove it, not required 
* cargo-llvm-cov ignores `tests` hard-coded
* the result of the coverage tests makes more sense ( see screenshots )
* avoid big files, separating tests in different files
* I learned separated tests on `Command-Line Rust` book :D

bobmcwhirter (commented):
I'm not sure testing for textual output at this level is overly beneficial, but I'd welcome others' thoughts.

Some test strategy (or lint?) ensuring that everything has some description in the Clap structures/derives, mayhap more useful.

e.g., we don't really care the exact wording, test-wise, but we *do* care that we're not leaving big empty holes which should have some documentation/words in them.

ctron:
I 100% agree with that: https://github.com/trustification/trustify/pull/21#pullrequestreview-1925002592

helio-frota:
thanks for the feedback !

I also agree with the comments about the test-function-it-self :+1: 

The most important parts for me in this PR are:

* `tests` directory to satisfy the coverage tool
* the (optional) pretty_assertions dev-dependency

helio-frota:
Thanks all for the extra info!

Issue #23, opened 2024-03-07
Backend - Disable sqlx debug log


helio-frota:
https://www.sea-ql.org/SeaORM/docs/install-and-config/debug-log/

PR #26, opened 2024-03-08
chore: lower sqlx logging to "debug"

ctron:
The number of logs that sqlx creates is a bit too spammy for my taste, at least on the "info" level. I think it makes sense to keep them in general. Maybe even lower them to "trace".


helio-frota:
@ctron thanks for this

Related to #23 and #22 

>  Maybe even lower them to "trace".

Please do this and then we solve 2 things at the same time :+1: 

* current PR sqlx debug level don't shows seaorm logs and continue with unreadable log

debug

* If we change it to trace we can see seaorm readable log only ( to solve #23 )

trace

helio-frota (changes requested):
Please change to trace :+1:  -- comment added thanks

helio-frota:
extra:

>  solve 2 things at the same time 

1) if we turn off slqx, we will no longer see `elapsed = 2.816096ms`
2) with sqlx `trace` we keep seeing the seaorm and we solve the #23

bobmcwhirter (approved):

Issue #19, opened 2024-03-07
Don't spam CI reports


ctron:
I find the warnings spamming the CI reports massively annoying and distracting.

My expectation would be that if there warnings like that, a PR won't get merged before they are resolved. Or, we disable them as we don't seem to care.

Right now, reviewing a PR and getting dozens of unrelated warnings is not helping.

jcrossley3:
Man, they really bother me, too! Pedantry FTW!

ctron:
Closed by #28

PR #28, opened 2024-03-08
ci: disable npm lint, as we don't make use of its results right now

ctron:
As mentioned in #19 I find it pretty hard to work with PRs because of those unrelated warnings, which show up on every PR. Even though no changes have been made on those files.

My proposal here is to either merge this PR and disable this step for now, until we find some time to fix those warnigns. Or find a switch for "warnings as errors" and fail the build in the case of violations.

Spamming PRs and not using the outcome to improve the code seems counterproductive.

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -29,8 +29,8 @@ jobs:
         run: ./scripts/verify_lock.mjs
       - name: Install
         run: npm clean-install --ignore-scripts
-      - name: Lint sources
-        run: npm run lint
+#      - name: Lint sources
+#        run: npm run lint
       - name: Build
         run: npm run build
       - name: Test


carlosthe19916 (approved):

helio-frota:
> unrelated warnings

this is one of the reasons I mentioned to move UI to a diff repository

but the reality is that we have more reasons not only in CI bits ( if we want/need to make the thing more professional way )
example: https://github.com/KaotoIO/kaoto-ui/tree/main/.github/workflows

carlosthe19916:
> > unrelated warnings
> 
> this is one of the reasons I mentioned to move UI to a diff repository
> 
> but the reality is that we have more reasons not only in CI bits ( if we want/need to make the thing more professional way ) example: https://github.com/KaotoIO/kaoto-ui/tree/main/.github/workflows

I agree with you, the UI could live in another repository but perhaps its not the right time yet to do that. We need to collect things like this, see if it makes us slower or faster, what are the things that distract people (e.g. backend should not be distracted by UI work). Eventually evaluate things. Let's be patient in this topic (although I'm not the right person to even mention "patience" as I also need to learn that xD)

PR #40, opened 2024-03-08
Initial foray at integration embedded PostgreSQL.

bobmcwhirter:
Initially only applies for #[cfg(test)] types of targets, to convert "integration tests" into "unit tests".

The special system constructor returns a handle to the temporary database process, which *must be held onto* through the duration of a given test, as having it Drop'd causes the DB to shutdown.

Note: this includes assigning it to _ or _db, hence using an actually named (but other unused) variable.

### DIFF
--- a/.github/workflows/backend.yaml
+++ b/.github/workflows/backend.yaml
@@ -44,7 +44,7 @@ jobs:
       run: rm -Rf $HOME/.theseus/
 
     - name: Test
-      run: cargo test -- --test-threads=1 --nocapture
+      run: cargo test -- --nocapture
       env:
         RUST_LOG: debug
         DB_USER: postgres
--- a/.github/workflows/backend.yaml
+++ b/.github/workflows/backend.yaml
@@ -40,9 +40,11 @@ jobs:
       run: cargo check
     - name: Clippy
       run: cargo clippy --all-targets --all-features -- -D warnings -D clippy::unwrap_used
+    - name: PreTestCleanUp
+      run: rm -Rf $HOME/.theseus/
 
     - name: Test
-      run: cargo test
+      run: cargo test -- --nocapture
       env:
         RUST_LOG: debug
         DB_USER: postgres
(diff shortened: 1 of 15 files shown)


ctron:
I absolutely like that. If running tests, can they still run in parallel?

bobmcwhirter:
Currently not on CI but I'm sure we can sort that out

bobmcwhirter:
I think ultimately we need to upstream a PR to the postgresql-embedded crate for 2 reasons:

1) It supports `Clone` but also has `Drop` implementation which... shuts everything down. Wrapping in an Arc on our end helps that case.
2) Multithreaded-testing results in a race on the install/init the base pgsql. Some flavor of lockfile should possibly be involved.

bobmcwhirter:
(after a first run locally, when pgsql is installed/init'd, you can run tests concurrently no problem)

.github/workflows/backend.yaml
```diff
@@ -55,9 +40,12 @@ jobs:
       run: cargo check
     - name: Clippy
       run: cargo clippy --all-targets --all-features -- -D warnings -D clippy::unwrap_used
+    - name: PreTestCleanUp
+      run: rm -Rf $HOME/.theseus/
 
     - name: Test
-      run: cargo test
+      run: cargo test -- --test-threads=1 --nocapture
```
mrizzi, line 47: Testing this locally for me it worked successfully with just `cargo test`. Am I wondering if, after your latest changes, `--test-threads=1` is still required or it can be removed.

mrizzi (commented):
Just a comment based on local testing.

.github/workflows/backend.yaml
bobmcwhirter: Lemme try.

bobmcwhirter: I suspect it's only a problem on a perfectly-clean system that has never run the tests before (such as a CI worker).

mrizzi (approved):
@bobmcwhirter thanks a lot :+1:

PR #50, opened 2024-03-11
chore: align sbom importer with csaf importer

ctron:


bobmcwhirter (approved):
We should percolate all the DbStrategy stuff up (not now; just making a note) so we can import into a managed non-external DB for the PM use-case, some day.

PR #52, opened 2024-03-11
GET /advisories

carlosthe19916:
Hi tried to put my hands into generating REST endpoints for the advisories. Some questions came to my mind:

- Do we really need the following data inside the Paginated Object?

```rust
pub num_pages: u64,
pub prev_page: Option<Paginated>,
pub next_page: Option<Paginated>, 
```
From the REST point of view, whoever is making a request can infer all those 3 fields based on data being send and the first response of the server.

- At the moment, we save (in the DB) limited data about `Advisory` . We will definitely need to extract more fields out of the CSAF files. I could not find any code where we are extracting data from CSAF files and then map them to the DB or object. Is something like that already existing somewhere or it needs to be implemented? 

```rust
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    pub identifier: String,
    pub location: String,
    pub sha256: String,
}
```

I'm not expecting this PR to be merged as it does not contain the fields I expect to expose through the REST endpoints, but I just wanted to open this PR for my questions to make more sense.

### DIFF
--- a/backend/api/Cargo.toml
+++ b/backend/api/Cargo.toml
@@ -26,6 +26,7 @@ async-trait = "0.1.74"
 lenient_semver = "0.4.2"
 cpe = "0.1.3"
 postgresql_embedded = { version = "0.6.2", features = ["blocking", "bundled", "tokio" ] }
+serde = { version = "1.0.197", features = ["derive"] }
 
 
 [dev-dependencies]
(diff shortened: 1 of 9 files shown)


carlosthe19916:
Or perhaps starting on "Advisories" was a bad idea as other abstractions like "Packages", "CVEs", "SBOMs" might be better implemented at the moment? xD

bobmcwhirter:
We can certainly modify how pagination works. That was mostly a spike for me to understand how SeaORM handles paginating, along with my own ignorance of what's needed on the front-end.  I somewhat come from the school of HATEOAS, and generally prefer for links to be sent within the response, instead of the client having to calculate and construct URLs from patterns.

But that's not a strongly-held opinion.

bobmcwhirter:
ref: https://en.wikipedia.org/wiki/HATEOAS

Issue #51, opened 2024-03-11
Promote `backend/*` to the root of the repo


jcrossley3:
In light of #47 

Having a `Cargo.toml` at the root simplifies the dev cycle. 

Further, let's put everything that's not a cargo workspace beneath a single directory. Do we like `etc/`, `var/` or something else?

bobmcwhirter:
https://github.com/trustification/trustify/pull/53

PR #53, opened 2024-03-11
Reorganize `backend` to the root.

bobmcwhirter:
Move `deploy` and `test-data` under `etc/`
Remove `clients` until we need it.
Rename `cli` to `trustd` to indicate it's the daemon-ish server cli, not a client cli. Rename `system` to `graph` to be marginally more descriptive. Add a bonafide `Graph` type instead of an alias around `Arc<InnerSystem>` with appropriate `Deref` impl.


jcrossley3 (approved):
Awesome, thanks!

jcrossley3:
Fixes #51

PR #54, opened 2024-03-12
chore: work towards adding embedded oxide-auth OIDC server

ctron:


bobmcwhirter:
*YES PLEASE*

Before PTO (and after) I was looking at OIDC stuff and had some thoughts, but no success.

1) Our current `new_auth!(...)` wraps even the swagger-ui HTML serving, so we can't seem to load up the UI to even reach the [authorize] button.
2) I'd tried an `OidcStrategy` CLI arg group like the `DbStrategy`, thinking that if `::Managed` we do what you're doing here, with a static set of users ("admin", "user") and a static set of secrets/etc "admin-sercret", "user-secret") absolutely for demo-purposes/PM-use-case-only.
3) We're soon going to need to always have auth enabled, but (2) could replace --devmode and --auth-disabled by being predictable?

ctron:
I would love to have a simple OIDC server without all the fancy stuff, just a single username/password (plus backend clients/service accounts). Bonus points for having that embeddable into a Rust application. Then again, that doesn't seem to exist.

oxide-auth could be used to build that. However, it's missing some parts that we would need to implement ourselves. My feeling is it brings ~75%.

But I am not sure that's the right priority right now.

bobmcwhirter:
Yah, I agree that a full-but-simple OIDC would be lovely, but possibly not within priority scope right now.

A not-quite-full-but-simple OIDC that suffices to substitute for `--devmode` and/or `--auth-disabled` would be within scope, I suspect.

e.g., a server that is not unlike the preconfigured example from oxide, where we're can remove `--auth-disabled` entirely, and a small finite set of accounts (and credentials) are statically-defined to allow for:

1. predictable testing with predictable tokens and refresh tokens
2. an easy Day-0 out-of-the-box experience for someone just wanting to kick the tires.  

eg

```
Just run:

$ trustd

And then surf to localhost:8080/swagger-ui and login with admin/admin or user/user 
```

PR #55, opened 2024-03-12
PostgreSQL embedded: tests use system temp dir for installation

mrizzi:
Investigating the failure in https://github.com/trustification/trustify/actions/runs/8234967143/job/22517985508?pr=40#step:9:33, it turned out it was related to multi tests concurrently extracting PostgreSQL to the same folder.

Installing tests to extract PostgreSQL in folder within the system temporary directory enables tests to be executed concurrently.

The drawback with this implementation is that `into_path` consumes `TempDir` hence it won't be automatically deleted at the end of the test but considering the installation directory is within the system temporary directory, I think this could be considered acceptable in the context of tests execution.

### DIFF
--- a/.github/workflows/backend.yaml
+++ b/.github/workflows/backend.yaml
@@ -41,6 +41,6 @@ jobs:
     - name: Clippy
       run: cargo clippy --all-targets --all-features -- -D warnings -D clippy::unwrap_used
     - name: Test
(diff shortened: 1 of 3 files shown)


mrizzi:
@bobmcwhirter I haven't permissions to add you as reviewer but I would like to have your review please.

bobmcwhirter (approved):

PR #60, opened 2024-03-12
chore: port over infrastructure stuff

ctron:
It's really just porting it over to trustify, and adapting it to the database backend.

Authentication/Authorization is present, but not used yet.

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,0 +1,1 @@
+allow-unwrap-in-tests = true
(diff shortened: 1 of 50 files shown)


ctron:
Just to be clear, I think there's room for improvement:

* how the configuration works (clap arguments only at that point)
* I still think it might make sense to give this it's own crate

dejanb:
@ctron I think this is definitely useful beyond trustification and deserves its own crate

jcrossley3 (approved):

PR #61, opened 2024-03-12
ci: we no longer require nettle

ctron:


ctron:
https://github.com/trustification/trustify/blob/2eff76243aeac8abf8848f065dc871e06c704272/importer/Cargo.toml#L15-L16

I came from weird future-usage and nettle-friendly defaults from  sequoia PGP.

PR #64, opened 2024-03-12
Enable feature to fix build within child workspace

jcrossley3:
Build works fine from root, but fails from beneath `graph` and `common` workspaces without enabling the `native-tls` feature for `reqwest` crate.


helio-frota:
Good catch !

```
error[E0599]: no method named `danger_accept_invalid_hostnames` found for struct `ClientBuilder` in the current scope
  --> common/src/reqwest.rs:40:10
```

bobmcwhirter (approved):
Thanks! I also hit this.

PR #62, opened 2024-03-12
Begin work on CVE vuln parsing and ingesting.

bobmcwhirter:

PR #65, opened 2024-03-13
Move ingest csaf to ingestors & upload csaf /POST

carlosthe19916:
- Took workspace/modules organization from https://github.com/trustification/trustify/pull/62
- The idea is to reuse the current code for importing CSAF on the importer and also on the server rest endpoints
- Currently the `Advisory` table is only storing `id, identifier, location, and sha256`. Eventually we will need to extend those fields to the ones we will expose through the REST endpoint. As a POC, in this PR I'm only adding the `title` field to an advisory but eventually I plan to add more fields for them to be exposed trough our rest apis
- This PR contains a REST POST method to upload csaf files reusing the code already existing

I admit you have the expertise in regards of how to best accomplish this so I would appreciate some feedback. If this PR was too premature we can close it no problem. I'm just trying to slowly populate the data that eventually the UI will need. 

```mermaid
graph TD;
Terminal-->importer;
UI-->server;

importer-->ingestors;
server-->ingestors;
ingestors-->graphh;
graphh-->DB;
```


server/src/server/read/advisory.rs
```diff
@@ -0,0 +1,48 @@
...
+    files: Vec<TempFile>,
+}
+
+#[utoipa::path(responses((status = 200, description = "Upload a file")))]
+#[post("/advisories")]
+pub async fn upload_advisory(
+    state: web::Data<AppState>,
+    MultipartForm(form): MultipartForm<UploadForm>,
+) -> Result<impl Responder, Error> {
+    let temp_file = form.files.first().ok_or(Error::BadRequest {
```
ctron, line 33: I really think we should avoid any filesystem interaction like that for uploading a file. I am also not sure if we should use multipart forms for uploading. What not follow the approach we used for streaming data: https://github.com/trustification/trustification/blob/0585db52cdcfd2ec4f58484777706b406b1ec40c/bombastic/api/src/server.rs#L317-L343

ctron (changes requested):
A few minor things. What worries me is the use of temporary files for uploading. Especially considering we will run in a container. I think the approach we've been using so far (streaming data right to the target) worked file.

ingestors/src/advisory/csaf.rs
```diff
@@ -0,0 +1,43 @@
...
+use crate::advisory::AdvisoryMetadata;
+
+pub async fn ingest(
+    system: &Graph,
+    csaf: Csaf,
+    sha256: &str,
+    location: &str,
+) -> anyhow::Result<i32> {
+    let document_category = &csaf.document.category;
+    if !matches!(document_category, Category::Vex) {
```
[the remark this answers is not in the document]
carlosthe19916: I was assuming there was a good reason since the `importer` is already not importing files that don't have the  `vex` category. So I was just doing a copy/paste to not disrupt previous assumptions/decisions. I'll be glad to remove that restriction as long as I don't remove any decision/assumption taken previously.

https://github.com/trustification/trustify/blob/4fd2dec25c977f34edad29074504146f09735785/importer/src/csaf/mod.rs#L132-L135

server/src/server/mod.rs
```diff
@@ -14,6 +15,12 @@ pub enum Error {
     System(graph::error::Error),
     #[error(transparent)]
     Purl(#[from] packageurl::Error),
+    #[error(transparent)]
+    Io(std::io::Error),
+    #[error("Invalid request {msg:?}")]
```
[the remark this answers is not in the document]
carlosthe19916: Done, thanks!

server/src/server/read/advisory.rs
carlosthe19916: I remove the multipart thing and replace it by `payload: web::Payload` just as we do in "trustification".

ingestors/src/advisory/csaf.rs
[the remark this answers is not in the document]
ctron: Yea, that was there just to keep an initial import simple, as we don't use it so far. Or at least for RH data. But I don't think it makes sense to reject non-vex documentation at the core.

server/src/server/read/advisory.rs
```diff
@@ -0,0 +1,32 @@
...
+use trustify_ingestors as ingestors;
+
+#[utoipa::path(responses((status = 200, description = "Upload a file")))]
+#[post("/advisories")]
+pub async fn upload_advisory(
+    state: web::Data<AppState>,
+    payload: web::Payload,
+    content_type: Option<web::Header<ContentType>>,
+) -> Result<impl Responder, Error> {
+    let payload_bytes = payload.to_bytes().await?;
```
ctron, line 21: I think we should put a not here that we need to fix this later. As this basically allocates memory twice the size of the document (the in memory blob and the parsed structured). I think it would make sense to (later) find a way how to parse the document from a stream.

carlosthe19916: I added a TODO note.

ingestors/src/advisory/csaf.rs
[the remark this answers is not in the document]
carlosthe19916: Ok, now we are accepting all CSAF files regardless of its category

server/src/server/read/advisory.rs
```diff
@@ -0,0 +1,48 @@
...
+use csaf::Csaf;
+use std::fs;
+
+use crate::server::Error;
+use crate::AppState;
+
+use sha2::{Digest, Sha256};
+use trustify_ingestors as ingestors;
+
+pub fn sha256(val: &str) -> String {
```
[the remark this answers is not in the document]
carlosthe19916: Fixed

server/Cargo.toml
```diff
@@ -8,6 +8,8 @@ trustify-auth = { path = "../common/auth", features = ["actix"] }
 trustify-graph = { path = "../graph" }
 trustify-common = { path = "../common"}
 trustify-infrastructure = { path = "../common/infrastructure" }
+trustify-entity = { path = "../entity"}
+trustify-formats = { path = "../ingestors"}
```
ctron, line 12: I think this should be named `trustify-formats` to align the directory name?

carlosthe19916: I'm following what we already have into the `main` branch. The `main` branch has a directory `ingestors/` whose cargo module name is `trustify-formats`.

https://github.com/trustification/trustify/blob/a84adf048dfe98df3b3a54c12ac84a602e935934/ingestors/Cargo.toml#L2

I'm probably missing something, what should I change? the directory name or the line 12 `trustify-formats = { path = "../ingestors"}`

ctron: I don't know TBH. I was just confused by the fact I didn't see any use of `trustifty_ingestors`, but instead there is a `trustify_formats` crate which doesn't seem to exist in the filesystem.

I think names should be aligned unless there are good reasons not to. And if that's not the case, we should fix it.

It looks like this is all about ingesting, so naming it `trustify-ingestors` sounds reasonable to me.

ctron (approved):

server/Cargo.toml
carlosthe19916: I fully agree with you. I renamed the package name to `trustify-ingestors` for it to be aligned to the directory name. I assume it was a mistake to have it that way (main branch) in the first place

PR #68, opened 2024-03-13
build(fix): Enable feature to fix build within child workspace

helio-frota:
Same situation as #64 

```
error[E0432]: unresolved import `tokio::select`
 --> common/infrastructure/src/health/checks/local.rs:9:5
  |
9 | use tokio::select;
  |     ^^^^^^^^^^^^^ no `select` in the root

error: cannot determine resolution for the macro `select`
   --> common/infrastructure/src/health/checks/local.rs:102:17
    |
102 |                 select! {
    |                 ^^^^^^
    |
    = note: import resolution is stuck, try simplifying macro imports
```


jcrossley3 (approved):
Thanks!

PR #76, opened 2024-03-14
Add `title` and `description` to Vulnerability models and entities.

bobmcwhirter:
Provide a handy method for getting a new db transaction. CVE record importer uses transactions and sets title/description. A few small renames of system->graph.

### DIFF
--- a/entity/src/vulnerability.rs
+++ b/entity/src/vulnerability.rs
@@ -7,6 +7,8 @@ pub struct Model {
     #[sea_orm(primary_key)]
     pub id: i32,
     pub identifier: String,
+    pub title: Option<String>,
+    pub description_en: Option<String>,
 }
 
 #[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
--- a/entity/src/vulnerability.rs
+++ b/entity/src/vulnerability.rs
@@ -1,4 +1,4 @@
-use crate::advisory_vulnerability;
+use crate::{advisory_vulnerability, vulnerability_description};
 use sea_orm::entity::prelude::*;
 
 #[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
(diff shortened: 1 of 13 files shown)


helio-frota (approved):
I approved understanding ~37% of this code :+1:

entity/src/vulnerability.rs
```diff
@@ -7,6 +7,8 @@ pub struct Model {
     #[sea_orm(primary_key)]
     pub id: i32,
     pub identifier: String,
+    pub title: Option<String>,
+    pub description_en: Option<String>,
```
jcrossley3, line 11: The `_en` feels wrong as a field name suffix. Feels hard-coded. Would it make more sense as a LANG env var?

ingestors/src/cve/loader.rs
```diff
@@ -30,23 +30,43 @@ impl<'g> CveLoader<'g> {
...
+            .set_title(cve.containers.cna.title.clone(), Transactional::Some(&tx))
+            .await?;
 
+        vulnerability
+            .set_description_en(
+                cve.containers
+                    .cna
+                    .descriptions
+                    .iter()
+                    .find(|e| e.lang == "en")
```
jcrossley3, line 50: ibid

jcrossley3 (commented):
Mmm, bite-size yum!

entity/src/vulnerability.rs
bobmcwhirter: Fwiw it's lang from the CVE not user.

jcrossley3: But the `"en"` is hardcoded in the filter below. If it's from the CVE, does that imply an additional `pub lang: String,` field? That also feels a little overkilly to me, but if we need to support multiple langs...

helio-frota: >  is hardcoded in the filter below
good catch!

bobmcwhirter: Yes I reckon there could be multiple because yes there is a lang property. 

But it could be anything or multiple I reckon.

ctron (approved):

bobmcwhirter:
@jcrossley3 ready for re-review.

jcrossley3 (approved):
Is it better? More complex but more robust? More correct?

entity/src/vulnerability_description.rs
```diff
@@ -0,0 +1,29 @@
...
+use sea_orm::entity::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
+#[sea_orm(table_name = "vulnerability_description")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
+    pub vulnerability_id: i32,
+    pub lang: String,
+    pub description: String,
```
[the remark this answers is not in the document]
bobmcwhirter: Yah, if those previous Options were None, then we'd simply now just not create a row in the table.

PR #78, opened 2024-03-15
Add `aggregated_severity` + `current_release_date` to Advisory entity

carlosthe19916:
Adding some metadata to the Advisory entity
- Purpose: those fields are part of the the main table that list advisories in the UI.

### DIFF
--- a/entity/src/advisory.rs
+++ b/entity/src/advisory.rs
@@ -1,3 +1,4 @@
+use chrono::{DateTime, Utc};
 use sea_orm::entity::prelude::*;
 
 #[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
@@ -9,6 +10,10 @@ pub struct Model {
     pub location: String,
     pub sha256: String,
     pub title: Option<String>,
+    pub aggregate_severity: Option<String>,
+
+    /// Holds the date when the current revision of this document was released.
+    pub current_release_date: Option<DateTime<Utc>>,
 }
 
 #[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
--- a/entity/src/advisory.rs
+++ b/entity/src/advisory.rs
@@ -10,10 +10,10 @@ pub struct Model {
     pub location: String,
     pub sha256: String,
     pub title: Option<String>,
-    pub aggregate_severity: Option<String>,
+    pub severity: Option<String>,
 
     /// Holds the date when the current revision of this document was released.
-    pub current_release_date: Option<DateTime<Utc>>,
+    pub release_date: Option<DateTime<Utc>>,
 }
 
 #[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
--- a/entity/Cargo.toml
+++ b/entity/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2021"
 sea-orm = { version = "0.12", features = [ "sqlx-postgres", "runtime-tokio-rustls", "macros", "with-json" ] }
 serde_json = "1"
 tokio = { version = "1.30.0", features = ["full"] }
+chrono = "0.4.35"
 
 [dev-dependencies]
 anyhow = "1.0.72"
--- a/graph/Cargo.toml
+++ b/graph/Cargo.toml
@@ -28,6 +28,7 @@ lenient_semver = "0.4.2"
 cpe = "0.1.3"
 postgresql_embedded = { version = "0.6.2", features = ["blocking", "bundled", "tokio" ] }
 tempfile = "3"
+chrono = "0.4.35"
 
 
 [dev-dependencies]
--- a/graph/src/graph/advisory/csaf/mod.rs
+++ b/graph/src/graph/advisory/csaf/mod.rs
@@ -6,6 +6,7 @@ use crate::graph::error::Error;
 use csaf::{document::Category, Csaf};
 use sea_orm::ActiveValue::Set;
 use sea_orm::{ActiveModelTrait, TransactionTrait};
+use std::rc::Rc;
 use trustify_common::db::Transactional;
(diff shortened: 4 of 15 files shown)


bobmcwhirter:
What is `aggregated_severity` aggregated from?

carlosthe19916:
> What is `aggregated_severity` aggregated from?
- `aggregated_severity` is a defined field from the CSAF file itself, so the owner/vendor of the file defines it.

Screenshot from 2024-03-15 16-27-54

- Also technically, it should be generated by the aggregation of scores of all the `vulnerabilities` within that file. What is the exact formula? I don't know
Screenshot from 2024-03-15 16-33-27

bobmcwhirter:
Ah, thanks.

Yeah, I'd think if aggregate is aggregated (and calculable) perhaps we don't store it directly, but rather calculate from the bits within?  I dunno. @ctron has more knowledge around CSAF than I.

entity/src/advisory.rs
```diff
@@ -9,6 +10,10 @@ pub struct Model {
     pub location: String,
     pub sha256: String,
     pub title: Option<String>,
+    pub aggregate_severity: Option<String>,
```
ctron, line 13: Short term I think that's ok. Long term we must also have the namespace alongside that. "medium" from Red Hat might mean something different than "medium" from Acme Corp.

entity/src/advisory.rs
```diff
@@ -9,6 +10,10 @@ pub struct Model {
...
+
+    /// Holds the date when the current revision of this document was released.
+    pub current_release_date: Option<DateTime<Utc>>,
```
ctron, line 16: IIRC current release date is a mandatory field, we should map it accordingly.

ctron (changes requested):
Small change. I'll leave it you if you want to add the namespace or a note about it's absence.

entity/src/advisory.rs
carlosthe19916: Thanks for pointing this out. It allowed me to make 2 observations:
- @bobmcwhirter you wrote this block of code where we are ingesting a CVE file into the `Advisory` model (db table) https://github.com/trustification/trustify/blob/main/ingestors/src/cve/loader.rs#L54-L56 . There is also a test that points to this file https://github.com/trustification/trustify/blob/main/etc/test-data/mitre/CVE-2024-28111.json
  - My question is what exactly are we going to ingest under the `advisory` abstraction (model/table)? If the current `main` branch is correct then: @ctron I'm afraid that although `current_release_date` is a mandatory field for a CSAF file, it does not even exist on CVE files.

```rust
        self.graph
            .ingest_advisory(cve.cve_metadata.cve_id(), location, sha256, &tx)
            .await?;
```

- The same applies to OSV files https://github.com/trustification/trustify/blob/main/ingestors/src/advisory/osv/loader.rs#L48-L50 . We are ingesting them as an advisory.
  - This is the file ingested as an advisory https://github.com/trustification/trustify/blob/main/etc/test-data/osv/RUSTSEC-2021-0079.json 

I know `trustify` won't necessarily be exactly the same as `Trustification` but in Trustification we strictly stored `CSAF` files under the `Advisory` model. Am I right @ctron ?

May I ask for clarification on this regard please. The image below is what an advisory mean for Trustification. And advisory has `aggregated severity` + `Title` + `a set of Vulnerabilities associated to it` + `information about products affected by that advisory`

Screenshot from 2024-03-17 12-28-55

entity/src/advisory.rs
carlosthe19916: I see. Just a further question just to understand or open a discussion:

`severity` can mean something different for VendorA and for VendorB. Does it mean that:
- Option1: regardless of how VendorA and VendorB see `severity` we are going to respect what is written is the file itself and therefore render in the UI (or expose through the api) the exact `severity` defined in the file?
  - If we will respect what is written in the file perhaps the `namespace` does not add any value? Perhaps a field `vendor` is what would make sense to clearly express that the file belongs to certain vendor?
- Option2: because "medium" for Red Hat might might mean "critical" from Acme Corp (I changed your example to make the case). Does it mean we are going to come up with our own system for `severity` so all vendors are aligned to our system of categorizing severities?

I could've just left the namespace for "later" but I wanted to ask just for my understanding. If I'm not making any sense just ignore my message xD

entity/src/advisory.rs
carlosthe19916: Alright. It seems, Advisories will store CSAF + CVEs. Therefore `metadata` information should be defines as something that all of them have and not specific to a single schema. So far, in this PR I've defined 3 things that CSAF + CVES or any other file should have:
```rust
    pub title: Option<String>,
    pub severity: Option<String>,
    /// Holds the date when the current revision of this document was released.
    pub release_date: Option<DateTime<Utc>>,
```

- I renamed `aggregated_severity` to `severity` to not be CSAF specific.
- I renamed `current_release_data` to `release_date`.
- I'm forcing the metadata to be set when an advisory is instantiated (inserted into db).
- This PR is implementing the code for setting the metadata for CSAF files, but I'll open another PR to populate the same metadata for CVEs

Let me know what you think about it.

carlosthe19916:
@bobmcwhirter @ctron I would appreciate a review of this PR xD

entity/src/advisory.rs
bobmcwhirter: This is looking nicer for sure.

fwiw, a single Advisory can contain information about multiple Vulnerabilities.

Currently any score is on the Advisory, but perhaps should be on the join of Advisory+Vulnerabililty, to accomodate more of the CSAF model.

carlosthe19916:
Closing this PR as many changes are going on in the repository. It might be easier to open a new PR rather than fixing conflicts xD

PR #81, opened 2024-03-15
fix: add test for optimistic locking and fix it

ctron:
The reason for updating to Rust 1.76.0 was the introduction of `inspect_err`, which I find super helpful (also in this case):

```rust
operation().
    .inspect_err(|err| {
        log::info!("err: {err:?}");
    })
```

### DIFF
--- a/modules/importer/src/test.rs
+++ b/modules/importer/src/test.rs
@@ -1,14 +1,16 @@
 #![cfg(test)]
 
 use super::model::ImportConfiguration;
-use actix_web::http::StatusCode;
-use actix_web::{test, App};
+use actix_web::{
+    http::{header, StatusCode},
+    test, App,
+};
 use serde_json::json;
 use trustify_common::db::Database;
 
 #[actix_web::test]
 async fn test_default() {
-    env_logger::init();
+    let _ = env_logger::try_init();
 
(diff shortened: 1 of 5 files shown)


modules/importer/src/test.rs
```diff
@@ -1,14 +1,16 @@
...
+    http::{header, StatusCode},
+    test, App,
+};
 use serde_json::json;
 use trustify_common::db::Database;
 
 #[actix_web::test]
 async fn test_default() {
-    env_logger::init();
+    let _ = env_logger::try_init();
```
jcrossley3, line 13: Curious why this is better?

jcrossley3 (approved):
I expected to see `inspect_err` called somewhere in here. :smile:

modules/importer/src/test.rs
ctron: Because otherwise tests will fail. Having `init` will run that methods twice (for two different tests) and the second one will fail. That's why the CI just failed, because the logger was already initialized by the first test.

ctron:
> I expected to see `inspect_err` called somewhere in here. 😄

I used that, but just for debugging. I removed it afterwards.

However, it allows one to more easily inject that into the call flow. Vs ripping it out and handling it with a match or if.

PR #82, opened 2024-03-15
feat: add server for processing imports

ctron:


jcrossley3 (commented):
Is it weird to reference "pagination" in the backend? That seems like it belongs in the frontend. On the back, it's just cursors and batch sizes, right?

Are we correctly separating concerns?

ctron:
A real cursor would mean that someone needs to keep the DB connection alive. And that all further requests go to the same pod (once we have them). I don't think that's a viable process.

The alternative to that is to describe the current state (serialize the cursor), and we the request comes in, deserialize it, and create an appropriate query. And I think that's what we are doing here. Just in the form of `page_size/page_number`. And there is support for exactly that in SeaORM.

I am not sure what the alternative would be. But I am open to suggestions.

ctron:
I think it would be ok to reduce this to "limit" and "offset". Dropping all the other stuff. However, I just re-used what was there. Then again, SeaORM pagination (on the backend) works with `page_size` and `page`. So, why not leverage that?

ctron:
Anyway, if that's the only point of discussion on this PR, and give the fact that it didn't add pagination, just re-used it, I would propose we merge this PR, and tackle pagination next. As it should be the same for all APIs, and we currently have a common struct/pattern for this, it shouldn't be too difficult to refactor this at this point (but we also shouldn't wait for much longer).

ctron:
@jcrossley3 I started a discussion about pagination: https://github.com/trustification/trustify/discussions/91

bobmcwhirter (approved):

ctron:
Merged by #92

PR #84, opened 2024-03-15
Enable rough parsing and ingesting of OSV-schema advisory documents.

bobmcwhirter:
JIRA: TC-1083


jcrossley3 (approved):
Did we bring that `schema.rs` over from somewhere?

bobmcwhirter:
Yah, it came from trustification.git, and previously altered from an osv crate that seemed out-of-date and not quite suitable for our purposes.

PR #86, opened 2024-03-15
Replace a few more Into<Purl>'s hither and yon.

bobmcwhirter:
Add type constraints to make using transactions a bit easier. A `Transactional::Some(_)` owns its transaction, reducing the things you need to hang onto.
A `Transactional::Some(_)` is provided ready-to-eat from Graph::transaction(). A `()` can be used for non-transactional as shorthand for `Transactional::None`.


jcrossley3 (approved):
Glad you did this! All the anonymous lifetimes were a tad smelly.

PR #92, opened 2024-03-19
feat: revamp pagination according to our discussion

ctron:
As it's based on PR #82, only pay attention to the most recent commit.


jcrossley3 (commented):
I wouldn't have expected to see the introduction of both `Limiter` and `Paginated` to implement this feature. So it seems like a lot of code on our side to take advantage of a feature already provided by SeaORM. I worry we may have ported over some cruft, but if you think this is as simple as we can get, I defer.

carlosthe19916:
I can only give an opinion about pagination.  I can see the `PaginatedResults` now only has 2 fields `total`, `items` which is good IMHO

PR #96, opened 2024-03-19
chore: proper log level in auth

helio-frota:


ctron (changes requested):
I don't think we should raise that as an error in the logs.

Assuming someone sends us unverifiable tokens. We should just ignore them, without spamming our logs with errors.

helio-frota:
> We should just ignore them

makes sense :+1:  thanks

PR #98, opened 2024-03-19
Add CVSS3.x scoring to advisories.

bobmcwhirter:
Add related CVSS enums and entities/models to the DB.

### DIFF
--- a/common/src/cvss3.rs
+++ b/common/src/cvss3.rs
@@ -1,0 +1,358 @@
+use std::fmt::{Display, Formatter};
+use std::str::FromStr;
+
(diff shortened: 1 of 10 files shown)


ctron (approved):
I think having the CVSS stuff in a dedicated crate, or contributed to the `cvss` crate might make sense: https://github.com/rustsec/rustsec/tree/main/cvss

bobmcwhirter:
I agree probably pulling into a crate. 

The upstream crate seems... elderly `Updated: almost 2 years ago`

And CVSS4 is now a thing we should address.

Next PR I'll at least shuffle it up to another crate within our tree perhaps?

PR #102, opened 2024-03-19
Cvss4

bobmcwhirter:


cvss/src/cvss4.rs
```diff
@@ -0,0 +1,510 @@
...
+            "{}",
+            match self {
+                Self::Network => 'N',
+                Self::Adjacent => 'A',
+                Self::Local => 'L',
+                Self::Physical => 'P',
+            }
+        )
+    }
+}
```
jcrossley3, line 119: Minor nit, but I like the local `use` to remove some of the match branch noise. (Ignore me if you were specifically using `Self::` for some other reason).

You have other places in this PR where you could do this, too. If you do it, only do it because you like it, not because I like it. :)

```suggestion
impl Display for AttackVector {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        use AttackVector::*;
        write!(
            f,
            "{}",
            match self {
                Network => 'N',
                Adjacent => 'A',
                Local => 'L',
                Physical => 'P',
            }
        )
    }
}
```

jcrossley3 (approved):

cvss/src/cvss4.rs
bobmcwhirter: That's a good idea, I just forget about use within fns.

lemme ammend.

PR #104, opened 2024-03-20
Add OpenAPI and Swagger UI

ctron:


ctron:
There is still room for improvement, like types with a type argument. But it's a start.

bobmcwhirter (approved):

PR #113, opened 2024-03-21
feat: add some filesystem storage

ctron:


jcrossley3:
Forgive my ignorance, but why do we need this?

ctron:
> Forgive my ignorance, but why do we need this?

Wouldn't we want to store the content of the SBOMs and advisories we get asked to stored? Assuming we manage SBOMs, I think a use case would be to retrieve that document again at some point.

Additionally, if we decide to make changes in the way we extract information, we would be able to re-process those documents.

ctron:
Storing a lot of stuff in the database isn't the best idea IMO. For one, you'd consume RDS instance resources, vs S3 (which is much cheaper).

Second, I know that "can" store BLOBs. However, I also know that it's not performing well, compared to a filesystem.

ctron:
I still don't like the retrieval part (besides that I still need to implement it for SBOMs).

Right now, it uses the primary key of the advisory (the id). I think that's ok. However, that also requires that the user can find that ID of an advisory. I think there should be a second function: fetch advisory by "document ID" and "source".

jcrossley3 (commented):
My review is the last commit of #114 

But I don't particularly love it. And I thought we were going to have the default impl be the database, but I could be misremembering that.

jcrossley3:
Closing in light of #114, which includes these commits and resolves the conflicts resulting from #116

PR #114, opened 2024-03-25
Introduce storage module

jcrossley3:
This includes the commits from #113 and resolves the conflicts incurred from #116

I modified the `StorageBackend` a tad, breaking out the "faux sync" methods into a helper class.

### DIFF
--- a/modules/importer/src/server/csaf/storage.rs
+++ b/modules/importer/src/server/csaf/storage.rs
@@ -6,7 +6,7 @@ use csaf_walker::validation::{
 use parking_lot::Mutex;
 use std::sync::Arc;
 use tokio_util::io::ReaderStream;
-use trustify_module_ingestor::service::IngestorService;
+use trustify_module_ingestor::service::{advisory::Format, IngestorService};
 
 #[derive(Debug, thiserror::Error)]
 pub enum StorageError {
@@ -40,7 +40,11 @@ impl ValidatedVisitor for StorageVisitor {
         let location = doc.context.url().to_string();
 
         self.ingestor
-            .ingest(&location, ReaderStream::new(doc.data.as_ref()))
+            .ingest(
+                &location,
+                Format::CSAF,
+                ReaderStream::new(doc.data.as_ref()),
+            )
             .await
             .map_err(|err| StorageError::Storage(err.into()))?;
 
--- a/modules/importer/src/server/csaf/storage.rs
+++ b/modules/importer/src/server/csaf/storage.rs
@@ -1,14 +1,12 @@
 use crate::server::report::ReportBuilder;
 use async_trait::async_trait;
-use csaf_walker::{
-    retrieve::RetrievedAdvisory,
-    validation::{ValidatedAdvisory, ValidatedVisitor, ValidationContext, ValidationError},
+use csaf_walker::validation::{
+    ValidatedAdvisory, ValidatedVisitor, ValidationContext, ValidationError,
 };
 use parking_lot::Mutex;
-use std::io::BufReader;
 use std::sync::Arc;
-use trustify_module_graph::graph::Graph;
-use trustify_module_ingestor::service::advisory::csaf::loader::CsafLoader;
+use tokio_util::io::ReaderStream;
+use trustify_module_ingestor::service::IngestorService;
 
 #[derive(Debug, thiserror::Error)]
 pub enum StorageError {
@@ -19,7 +17,7 @@ pub enum StorageError {
 }
 
 pub struct StorageVisitor {
-    pub system: Graph,
+    pub ingestor: IngestorService,
     /// the report to report our messages to
     pub report: Arc<Mutex<ReportBuilder>>,
 }
@@ -38,19 +36,13 @@ impl ValidatedVisitor for StorageVisitor {
         _context: &Self::Context,
         result: Result<ValidatedAdvisory, ValidationError>,
     ) -> Result<(), Self::Error> {
-        self.store(&result?.retrieved).await?;
-        Ok(())
-    }
-}
+        let doc = result?;
+        let location = doc.context.url().to_string();
 
-impl StorageVisitor {
-    async fn store(&self, doc: &RetrievedAdvisory) -> Result<(), StorageError> {
-        let loader = CsafLoader::new(&self.system);
-
-        loader
-            .load(doc.url.as_str(), BufReader::new(doc.data.as_ref()))
+        self.ingestor
+            .ingest(&location, ReaderStream::new(doc.data.as_ref()))
             .await
-            .map_err(|e| StorageError::Storage(e.into()))?;
+            .map_err(|err| StorageError::Storage(err.into()))?;
 
         Ok(())
     }
--- a/modules/importer/src/server/mod.rs
+++ b/modules/importer/src/server/mod.rs
@@ -10,14 +10,16 @@ use std::time::Duration;
 use time::OffsetDateTime;
 use tokio::time::MissedTickBehavior;
 use trustify_common::db::Database;
+use trustify_module_storage::service::dispatch::DispatchBackend;
 
 /// run the importer loop
-pub async fn importer(db: Database) -> anyhow::Result<()> {
-    Server { db }.run().await
+pub async fn importer(db: Database, storage: DispatchBackend) -> anyhow::Result<()> {
+    Server { db, storage }.run().await
 }
 
 struct Server {
     db: Database,
+    storage: DispatchBackend,
 }
 
 impl Server {
--- a/modules/ingestor/src/endpoints/advisory.rs
+++ b/modules/ingestor/src/endpoints/advisory.rs
@@ -20,7 +20,7 @@ pub struct UploadAdvisoryQuery {
         (status = 400, description = "The file could not be parsed as an advisory"),
     )
 )]
-#[post("/advisories/{advisory_format}")]
+#[post("/advisories")]
 /// Upload a new advisory
 pub async fn upload_advisory(
     service: web::Data<IngestorService>,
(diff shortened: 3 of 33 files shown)


modules/importer/src/server/csaf/storage.rs
```diff
@@ -38,19 +36,17 @@ impl ValidatedVisitor for StorageVisitor {
...
-
-        loader
-            .load(doc.url.as_str(), BufReader::new(doc.data.as_ref()))
+        let doc = result?;
+        let location = doc.context.url().to_string();
+
+        self.ingestor
+            .ingest(
+                &location,
+                Format::CSAF,
```
bobmcwhirter, line 45: Typing while I read, this may be answered below but...

Can we link the Format::CSAF/OSV etc to the actual path param in the REST API, instead of being stringly-typed as I left it?

modules/importer/src/server/mod.rs
```diff
@@ -10,14 +10,16 @@ use std::time::Duration;
 use time::OffsetDateTime;
 use tokio::time::MissedTickBehavior;
 use trustify_common::db::Database;
+use trustify_module_storage::service::dispatch::DispatchBackend;
 
 /// run the importer loop
-pub async fn importer(db: Database) -> anyhow::Result<()> {
-    Server { db }.run().await
+pub async fn importer(db: Database, storage: DispatchBackend) -> anyhow::Result<()> {
```
bobmcwhirter, line 16: Why is a variable named `storage` of a type named `DispatchBackend`?

modules/ingestor/src/endpoints/advisory.rs
```diff
@@ -13,85 +19,104 @@ use trustify_module_graph::graph::Graph;
...
-#[post("/advisories/{advisory_format}")]
+#[post("/advisories")]
 /// Upload a new advisory
 pub async fn upload_advisory(
-    graph: web::Data<Graph>,
-    req: HttpRequest,
-    path: web::Path<String>,
+    service: web::Data<IngestorService>,
     payload: web::Payload,
+    web::Query(UploadAdvisoryQuery { location, format }): web::Query<UploadAdvisoryQuery>,
```
bobmcwhirter, line 31: Moving to query params instead of path param?  I'm fine with that. We should probably start a `Tenets` document or discussion to ensure we keep consistent and how we decide which option to choose.

modules/ingestor/src/endpoints/advisory.rs
```diff
@@ -13,85 +19,104 @@ use trustify_module_graph::graph::Graph;
...
-        Ok(HttpResponse::Created().json(advisory_id))
-    } else {
-        Err(Error::BadRequest {
-            msg: "Unsupported advisory format".to_string(),
-            status: StatusCode::EXPECTATION_FAILED,
-        })
-    }
+    let fmt = format
+        .map(|f| Format::from_str(&f))
+        .unwrap_or(Ok(Format::CSAF))?;
```
bobmcwhirter, line 35: Default to CSAF format? If someone POSTs in a RANDOFORMAT, which is not actually CSAF, then we'll end up with a bunch of "this is invalid CSAF" errors instead of "We don't support RANDOFORMAT" yeah?

modules/ingestor/src/endpoints/advisory.rs
```diff
@@ -13,85 +19,104 @@ use trustify_module_graph::graph::Graph;
...
+        let response = test::call_service(&app, request).await;
+        log::info!("response: {response:?}");
+
+        assert!(response.status().is_success());
+
+        Ok(())
+    }
+
+    #[test_log::test(actix_web::test)]
+    async fn upload_osv() -> Result<(), anyhow::Error> {
```
bobmcwhirter, line 100: nice!

modules/ingestor/src/service/advisory/csaf/mod.rs
```diff
@@ -1,2 +1,80 @@
...
+            .await
+            .map_err(|err| Error::Storage(anyhow!("{err}")))?;
+        let sha256 = hex::encode(digest);
+
+        let storage = SyncAdapter::new(self.storage.clone());
+        let reader = storage
+            .retrieve(sha256.clone())
+            .await
+            .map_err(Error::Storage)?
+            .ok_or_else(|| Error::Storage(anyhow!("file went missing during upload")))?;
```
bobmcwhirter, line 33: For bonus points, do we want to re-hash to also ensure file contents didn't change during the shuffling?

modules/ingestor/src/service/advisory/csaf/mod.rs
```diff
@@ -1,2 +1,80 @@
...
+
+        let result = match fmt {
+            Format::CSAF => {
```
bobmcwhirter, line 36: Not for this PR, but when we add a 3rd format, probably smart to shuffle into a `fn` on `Format` itself to handle the delegation to the appropriate loader?

modules/storage/src/service/dispatch.rs
```diff
@@ -0,0 +1,69 @@
...
+/// data.
+///
+/// NOTE: Right now we only have one type (filesystem), but the goal is to have an additional one
+/// soon (e.g. S3)
+#[derive(Clone, Debug)]
+pub enum DispatchBackend {
+    Filesystem(FileSystemBackend),
+}
+
+impl StorageBackend for DispatchBackend {
```
bobmcwhirter, line 21: Ahhh, `DispatchBackend` is-a form of storage. Ignore previous.

modules/storage/src/service/fs.rs
```diff
@@ -0,0 +1,225 @@
...
+/// layer.
+#[derive(Clone, Debug)]
+pub struct FileSystemBackend {
+    content: PathBuf,
+}
+
+const NUM_LEVELS: usize = 2;
+
+impl FileSystemBackend {
+    pub async fn new(base: impl Into<PathBuf>) -> anyhow::Result<Self> {
```
bobmcwhirter, line 50: See my pending PR for `trustd`. We should maybe default to `.trustify/...whatever` for the Simple Happy PM Use-Case?

bobmcwhirter: (not this PR though)

modules/storage/src/service/fs.rs
```diff
@@ -0,0 +1,225 @@
...
+
+        // finalize the digest
+
+        let digest = digest.finalize();
+
+        // create the target path
+
+        let hash = hex::encode(digest);
+        let target = level_dir(&self.content, &hash, NUM_LEVELS);
+        create_dir_all(&target).await.map_err(StoreError::Backend)?;
```
bobmcwhirter, line 118: Should we sprinkle some `from` derives or impl `From<..>` to avoid having to `map_err(...)`?

bobmcwhirter (commented):
Everything I've commented on is just nit-picky. 

I defer to your discretion as to what you want to address in this PR or in follow-ups.

I also will assume the burden of rebasing my subsequent PR so you can get this one landed.

bobmcwhirter (approved):
I meant to tick the [approve] box.

modules/ingestor/src/endpoints/advisory.rs
bobmcwhirter: https://github.com/trustification/trustify/discussions/122

modules/importer/src/server/mod.rs
jcrossley3: I dunno, honestly. There is a comment left by @ctron in `dispatch.rs` but I didn't fully grok it.

modules/ingestor/src/endpoints/advisory.rs
jcrossley3: Cool. I don't have a strong opinion either way, I was just reusing the Query struct.

modules/ingestor/src/endpoints/advisory.rs
jcrossley3: My logic may be flawed. How should it work?

modules/ingestor/src/service/advisory/csaf/mod.rs
jcrossley3: I think being able to successfully retrieve it ensures that, right?

modules/ingestor/src/service/advisory/csaf/mod.rs
jcrossley3: Agree.

modules/storage/src/service/dispatch.rs
jcrossley3: I reply to these in order. :smile:

modules/ingestor/src/endpoints/advisory.rs
bobmcwhirter: Maybe `.ok_or(Error::InvalidFormat)?` which will turn a None into a throw error, or a Some into the underlying value without unwrap.  Sorta the err-variant of `unwrap_or(...)`

modules/storage/src/service/fs.rs
jcrossley3: Probably. I expect much of the storage module to evolve.

modules/importer/src/server/csaf/storage.rs
jcrossley3: I assume this question will be answered by #122 ?

modules/ingestor/src/service/advisory/csaf/mod.rs
bobmcwhirter: I dunno tbh. I think it means we managed to find some file at the hash-named location. Belt-and-suspenders would then verify, I guess, if the file retrieved by hash, when hashed, hashes to the hash of the hash-based file.

Yo dawg, I heard you liked hash, so I put some hash in your hash so you can hash while you hash.

jcrossley3: I see what you mean, verify we got what we put.

modules/storage/src/service/fs.rs
jcrossley3: Should this go in #121 ?

modules/ingestor/src/endpoints/advisory.rs
jcrossley3: I think my logic is correct: I'm mapping and unwrapping an `Option` containing a `Result`. So the `unwrap_or` is returning an `Err` if the format param is invalid. The `Ok(Format::CSAF)` is only returned if `format` is `None`.

See #126

modules/importer/src/server/csaf/storage.rs
bobmcwhirter: Ahh. Kk!

PR #116, opened 2024-03-26
Move CSAF-knowledge out of Graph entirely.

bobmcwhirter:
Build a CSAF ingestor capable of integrating some CSAF data into the graph. Adjust the CSAF importer/walker to use the ingestor. Ammend the endpoints to support multiple (csaf and osv) advisory inputs for the time being.


modules/ingestor/src/endpoints/advisory.rs
```diff
@@ -1,38 +1,54 @@
...
-        (status = 200, description = "Upload a file"),
-        (status = 400, description = "The file could not be parsed as a CSAF document"),
-    )
+tag = "ingestor",
+request_body = Vec < u8 >,
+responses(
+(status = 200, description = "Upload a file"),
+(status = 400, description = "The file could not be parsed as an advisory document"),
+)
 )]
```
jcrossley3, line 16: broken indentation

modules/ingestor/src/service/advisory/csaf/loader.rs
```diff
@@ -0,0 +1,82 @@
...
+
+        let advisory = self
+            .graph
+            .ingest_advisory(&advisory_id, location, sha256, &tx)
+            .await?;
+
+        for vuln in csaf.vulnerabilities.iter().flatten() {
+            let id = match &vuln.cve {
+                Some(cve) => cve,
+                None => continue,
```
jcrossley3, line 42: I kinda hope this doesn't work like it looks like you want it to. Is it not weird that an assignment might continue the loop? Is it just me? If not, I'd prefer we filter the iterator instead.

modules/ingestor/src/service/advisory/osv/loader.rs
```diff
@@ -22,10 +22,12 @@ impl<'g> OsvLoader<'g> {
         &self,
         location: L,
         record: R,
-    ) -> Result<(), Error> {
+    ) -> Result<String, Error> {
```
jcrossley3, line 25: Nice. I always feel like `Result<(), ...>` is wrong, or at least benefits from more thought.

jcrossley3 (changes requested):
I think I like the loader

modules/ingestor/src/service/advisory/csaf/loader.rs
```diff
@@ -36,43 +36,46 @@ impl<'g> CsafLoader<'g> {
...
+        for vuln in csaf
+            .vulnerabilities
+            .iter()
+            .flatten()
+            .filter(|e| e.cve.is_some())
+        {
+            // we filtered above, so should always be Some(_) but let's
+            // still avoid a blind unwrap.
+            if let Some(id) = &vuln.cve {
+                let advisory_vulnerability = advisory.link_to_vulnerability(id, &tx).await?;
```
jcrossley3, line 48: Typically, combinations of `is_some` and `unwrap` are a smell that `filter_map` should be used, e.g. `.filter_map(|e| e.cve)` but because you require both `vuln.cve` and later `vuln.product_status`, I'm guessing you need to return a tuple? To me, nested for loops are a smell, in general.

modules/ingestor/src/service/advisory/csaf/loader.rs
```diff
@@ -0,0 +1,85 @@
...
+                        log::debug!("{}", package.to_string());
+                        //advisory_vulnerability
+                            //.ingest_affected_package_range(package, Transactional::None)
+                            //.await?;
+                    }
+
+                     */
+                }
+            }
+        }
```
jcrossley3, line 81: How do you feel about this instead?

```suggestion
        for (id, ps) in csaf.vulnerabilities.iter().flatten().filter_map(|e| {
            e.cve
                .as_ref()
                .and_then(|id| e.product_status.as_ref().map(|ps| (id, ps)))
        }) {
            let advisory_vulnerability = advisory.link_to_vulnerability(&id, &tx).await?;

            for r in ps.fixed.iter().flatten() {
                for purl in resolve_purls(&csaf, r) {
                    let package = Purl::from(purl.clone());
                    advisory_vulnerability
                        .ingest_fixed_package_version(package, &tx)
                        .await?;
                }
            }
            for r in ps.known_not_affected.iter().flatten() {
                for purl in resolve_purls(&csaf, r) {
                    let package = Purl::from(purl.clone());
                    advisory_vulnerability
                        .ingest_not_affected_package_version(package, &tx)
                        .await?;
                }
            }
            for _r in ps.known_affected.iter().flatten() {
                /*
                for purl in resolve_purls(&csaf, r) {
                    let package = Purl::from(purl.clone());
                    log::debug!("{}", package.to_string());
                    //advisory_vulnerability
                        //.ingest_affected_package_range(package, Transactional::None)
                        //.await?;
                }

                 */
            }
        }

```

jcrossley3 (commented):
Maybe...

modules/ingestor/src/service/advisory/csaf/loader.rs
bobmcwhirter: I'd thought about that, but only until I realized that `product_status` is only the bits we're using _right now_ but there is a lot of potential bits within the CSAF document we ultimately need to trawl through.

bobmcwhirter: Also, if there's a CVE and no product-status, we still want to link to the vulnerability, not only if both are non-None.

jcrossley3: Cool. I'm all out of suggestions!

jcrossley3 (approved):
Looks good!

PR #121, opened 2024-03-28
Rework `trustd` for The PM Use-Case.

bobmcwhirter:
By executing `trustd` without any arguments whatsoever, you will end up with a running and bootstrapped system using the embedded-pgsql crate. Postgress will be installed under $PWD/.trustify/... and migrations performed. Additionally, importer loops will start.


jcrossley3 (approved):
I like it!

bobmcwhirter:
I'll rebase tomorrow.

PR #137, opened 2024-04-04
feat: allow preventing a reset of the database

ctron:
This adds the ability to keep the database structure (and with that the data) during restarts.

I understand it makes sense for tests to use ephemeral databases. But when testing locally, IMHO restarting the server process should not drop all the data.

This happens as we "refresh" all migrations, which removes all tables, and thus the data.

One can choose now between:

* bootstrap: re-create database
* refresh: re-create schema
* default: simply run missing migrations

That changes the default and if you're modifying (not adding) you would need to provide `--creation refresh` when starting. But I think that's the exception case.

### DIFF
--- a/common/src/db/mod.rs
+++ b/common/src/db/mod.rs
@@ -101,6 +101,15 @@ enum DbStrategy {
     Managed(Arc<(PostgreSQL, TempDir)>),
 }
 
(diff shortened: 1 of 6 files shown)


jcrossley3 (approved):
Handy!

PR #140, opened 2024-04-04
feat: add search

ctron:
This add some very basic search. Currently it is possible to search advisories by title. But it's a start and should be a blueprint for adding more search capabilities.

This doesn't yet do any "full text search", but a plain and simple `ILIKE`.

Note: This also requires a PostgreSQL database. IIRC we already used some PSQL specific things before (like types?). However this uses the `ILIKE` operator, which does case insensitive comparisons.

### DIFF
--- a/modules/search/Cargo.toml
+++ b/modules/search/Cargo.toml
@@ -1,0 +1,27 @@
+[package]
+name = "trustify-module-search"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+trustify-common = { path = "../../common" }
+trustify-entity = { path = "../../entity" }
+
+actix-web = { workspace = true }
+anyhow = { workspace = true }
+csaf = { workspace = true }
+log = { workspace = true }
+reqwest = { workspace = true }
+sea-orm = { workspace = true }
+serde = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true }
+sikula = { workspace = true, features = ["sea-orm"] }
+thiserror = { workspace = true }
+time = { workspace = true }
+tokio = { workspace = true, features = ["full"] }
+utoipa = { workspace = true, features = ["actix_extras"] }
+
+[dev-dependencies]
+test-log = { workspace = true, features = ["env_logger", "trace"] }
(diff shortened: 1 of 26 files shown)


modules/search/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+
+actix-web = { workspace = true }
+anyhow = { workspace = true }
+csaf = { workspace = true }
+log = { workspace = true }
+reqwest = { workspace = true }
+sea-orm = { workspace = true }
+serde = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true }
+sikula = { workspace = true, features = ["sea-orm"] }
```
jcrossley3, line 19: I'm a little confused: what exactly is sikula buying us now that we have SQL?

ctron: I would not want to expose SQL to the API endpoints. That's an internal thing. Sikula defines, on the API level, what the user can search for.

jcrossley3: Sure, but that API led to some confusion in our first iteration, and I don't think its design was the result of any collaboration with a PM or a salesperson. :smile: 

At this stage, I'm not sure we want the mapping of user input to SQL to be encapsulated within an external library. I don't want to get into a cycle in which PR's to this repo amount to library version updates in `Cargo.toml`. I might call that "premature organization".

ctron: To my understanding we need some form of search. And the question is: how should that look like? And how can we easily implement that.

Right now this approach gives us a clear definition of what can be searched and how that will behave. We can change the HTTP side of things for handing that over in a different fashion between the frontend and the backend, if that's a concern.

But I still think we need a very clear definition of what/how we can search. Spreading that apart in the code just by using "names" and constructs like `HashMap<String,String>` will make the pretty complex.

If there's a better way to do (and implement) this. Let's see a replacement PR for this one and compare it.

jcrossley3: @ctron can you add some unit tests to this PR just to help me verify parity?

jcrossley3:
Closing in light of #234

PR #142, opened 2024-04-08
An alternative query language POC

jcrossley3:
An alternative to #140 that doesn't use sikula, somewhat based on the proposal described in #9. Any field of an Entity can be used for either filtering or sorting without needing to introduce new abstractions.


jcrossley3:
A matrix chat betwixt @jcrossley3 @bobmcwhirter and @carlosthe19916 resolved that the `q` parameter will take the form of `{text}(&{field}{op}{value})*`. The `{text}` portion will become an OR clause of LIKE operations for every string field in a given entity. The other filters will be AND'd with that OR clause to further constrain the results.

carlosthe19916 (changes requested):
- The swagger description mentions the REST endpoint should be `/api/v1/search/advisory/advisory` . That is not accurate since the real endpoint is `/api/v1/search/advisory`

Screenshot from 2024-04-09 09-55-00

- The following commands do not work:
```shell
curl -v http://localhost:8080/api/v1/search/advisory
curl -v http://localhost:8080/api/v1/search/advisory?q=something
```

I get an error:
```json
{"error":"SearchSyntax","message":"query syntax error: Invalid field name: ''"}
```

But this one does work:
```
curl -v http://localhost:8080/api/v1/search/advisory?q=something&sort=id
```

jcrossley3:
> * The swagger description mentions the REST endpoint should be `/api/v1/search/advisory/advisory` . That is not accurate since the real endpoint is `/api/v1/search/advisory`
> 
> * The following commands do not work:
> 
> ```shell
> curl -v http://localhost:8080/api/v1/search/advisory
> curl -v http://localhost:8080/api/v1/search/advisory?q=something
> ```

Fixed, thanks!

carlosthe19916 (approved):
Looks good to me. If I find bugs or things to enhance while integrating it with the UI I can open issues and solve them independently.

ctron:
I agree that it looks simpler. However, I also think that it actually it too simplistic:

* There's no abstraction between the search and the database structure.
  * Whenever we make changes to the underlying database structure, that has a direct impact up to the frontend.
  * There's also no way to search for something that's not available as a database field.
  * There's no way to prevent the user from searching certain fields, which might give access to fields the user must not have access to.
* From my previous experience with regular expression in such cases, it always turns into a mess at some point. Aside from the fact they aren't the most performant way of dealing with this.
  * It's already using some "encode"/"decode" scheme to prevent running into certain cases. Which, as we're processing user data, might lead to other side effects (assume the user provides `=\x07` as an input character).
* The mechanism isn't powerful enough
  * There seems to be no way to restrict the full text search to a number of fields. I think this will have a massive performance impact.
  * I don't seem to be able to search for things like "ssl but not boringssl"

I am sure this can all be fixed and added. But in the end, that just feels like we're re-creating what is already there. And I don't see the benefit in this. So instead of re-inventing the wheel, I think we should build on top something that we know works. Instead of ditching that and re-creating it as part of repository that we will never re-use, and that doesn't have a clear scope of a functionality like that.

bobmcwhirter:
@jcrossley3 @ctron -- Y'all should pop some beers, open the camera, and have a hearty chat with each other.

jcrossley3:
> @jcrossley3 @ctron -- Y'all should pop some beers, open the camera, and have a hearty chat with each other.

I like beer and @ctron, but the whole reason both PR's were created was for others to decide. Our chats usually amount to me saying "[YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)!" and Jens saying "but we might". I will always lose that argument.

I'll try to address his points above, though.

jcrossley3:
> I agree that it looks simpler. However, I also think that it actually it too simplistic:

Well, darn. IMO, the pinnacle of achievement is to build something that appears simple while not being simplistic. It's very hard to do, and I'm sure I've missed some edge cases, but "simplistic" stings a little. :cry: 

> * There's no abstraction between the search and the database structure.

I see this as a positive. The fewer the abstractions, the simpler the solution, and the better chance it has to evolve quickly. The `Filter` struct is parameterized by an `EntityType`, though, so the search is "logically constrained" by the database structure.
   
>   * Whenever we make changes to the underlying database structure, that has a direct impact up to the frontend.

As long as those changes are made to our entities, the compiler will catch those potential impacts to the frontend.

>   * There's also no way to search for something that's not available as a database field.

That's true, but I understood that to be a given. Per @bobmcwhirter anything we'd search for will be in the database.

>   * There's no way to prevent the user from searching certain fields, which might give access to fields the user must not have access to.

I understand that's a requirement, but it wasn't in scope for this POC.

> * From my previous experience with regular expression in such cases, it always turns into a mess at some point. Aside from the fact they aren't the most performant way of dealing with this.

I'm not sure how to respond to this, other than to :smile: and say, "I like regular expressions"

I don't love performance arguments without real data to support them.

>   * It's already using some "encode"/"decode" scheme to prevent running into certain cases. Which, as we're processing user data, might lead to other side effects (assume the user provides `=\x07` as an input character).

That was just a way to support escaping special characters (`&` and `|`) in search queries. I didn't think a user would ever ring a bell to search for an advisory. :smile: 

> * The mechanism isn't powerful enough
>   
>   * There seems to be no way to restrict the full text search to a number of fields. I think this will have a massive performance impact.

You might be right. You might also be wrong.

But I concede including the `sha256` field in every where clause is dumb. The mandate I was given was that "full text search" included every field of type `String`. That mandate will certainly evolve.

>   * I don't seem to be able to search for things like "ssl but not boringssl"

`ssl&title!=boringssl`

> I am sure this can all be fixed and added. But in the end, that just feels like we're re-creating what is already there. 

Where? Where is a concise listing of example queries we support? Can you add a [test like this](https://github.com/trustification/trustify/blob/24c6165b218fa79b03fdf53aa2718f0d2e13e6aa/modules/search/src/query.rs#L241-L304) to #140 so that we can compare apples to apples? 

> And I don't see the benefit in this. 

None? Absolutely no benefit? Like, zero? Really?

> So instead of re-inventing the wheel, I think we should build on top something that we know works. Instead of ditching that and re-creating it as part of repository that we will never re-use, and that doesn't have a clear scope of a functionality like that.

Honestly, I think you're being harsh. The wheels are different. We don't know if either will be re-used. Both are inspired by commonly used examples in the real world, so that's a form of "re-use". And this PR is re-using the sea-orm metadata to reduce the number of required abstractions. I think there is real power in that kind of re-use.

I appreciate your detailed review.

bobmcwhirter:
🍺 🍺 📷 🗣️

ctron:
> > I agree that it looks simpler. However, I also think that it actually it too simplistic:
> 
> Well, darn. IMO, the pinnacle of achievement is to build something that appears simple while not being simplistic. It's very hard to do, and I'm sure I've missed some edge cases, but "simplistic" stings a little. 😢

Ok, maybe I got some words wrong here. What is trying to say was: it looks simpler than the other approach. But only for now. And it also misses a few things that I think we need.

> > So instead of re-inventing the wheel, I think we should build on top something that we know works. Instead of ditching that and re-creating it as part of repository that we will never re-use, and that doesn't have a clear scope of a functionality like that.
> 
> Honestly, I think you're being harsh.

Maybe. I am sorry if you took that personally. That was never my intention.

However, I still think we should spend our time elsewhere. I think search is already solved by #140.

ctron:
> > I think search is already solved by #140.
> 
> I'm confused by "already solved". #140 only makes it "possible to search advisories by title", right? Nothing else.

Yes, because I didn't want to waste time implementing all the others before having consensus.

> 
> To search any other fields of any other models will require many more PR's, correct? Is that the "elsewhere" in which you think we should be spending our time?
> 
> In contrast, this PR already supports the very query you cited as an example of it not being "powerful enough", i.e. `ssl&title!=boringssl`

Right that works, I missed that part. How would it work with a query that we currently support: (product = A OR product = B OR product = C) AND year >= 2023 AND year < 2024 AND (servery = low OR severity = medium).

> I want to understand your aversion to this.

Because I am 100% we will run into the case where we need to search for stuff that is not in the database 1:1. Then we start working with SQL expressions/formulas, virtual tables and stuff like that. We still would not have a clear concept on how to prevent the user from accessing certain fields or to prevent an impact on the frontend when making changes in DB structure.

I just think that a solution should be simple on the frontend, but support more complex scenarios on the backend. And maybe that's just me. Just to allow future improvements without the need to refactor too much. And yes, that should have a limit.

"Elsewhere" is the core functionality. Correlation between sbom and vex. I think #140 provides the search we require right now (minus some more fields and unit tests). But it doesn't create technical debt that we need to clear out later. Which I believe this PR would add.

However, of people agree on that this PR is the right forward, please merge it.

jcrossley3:
> How would it work with a query that we currently support: (product = A OR product = B OR product = C) AND year >= 2023 AND year < 2024 AND (servery = low OR severity = medium).

`product=A|B|C&year>=2023&year<2024&severity=low|medium`

> However, of people agree on that this PR is the right forward, please merge it.

Is @carlosthe19916 enough people? :stuck_out_tongue_winking_eye:

jcrossley3:
> Because I am 100% we will run into the case where we need to search for stuff that is not in the database 1:1. Then we start working with SQL expressions/formulas, virtual tables and stuff like that. We still would not have a clear concept on how to prevent the user from accessing certain fields or to prevent an impact on the frontend when making changes in DB structure.

Before I merge this, I want to acknowledge on the record how important I think this is and that I'm not meaning to sound dismissive of it. At this point in time, I don't have a good general solution for these things, and I expect we'll need to introduce abstractions to account for them.

PR #154, opened 2024-04-12
Wire up /api/v1/vulnerability with filter & pagination.

bobmcwhirter:
Test for the above.
Add IntoParams to SearchOptions and Paginated to make re-use easier for Utoipa.


modules/graph/src/endpoints/vulnerability.rs
```diff
@@ -32,3 +83,125 @@ pub async fn affected_products(
...
+
+        let response: PaginatedResults<vulnerability::Model> =
+            test::call_and_read_body_json(&app, request).await;
+
+        assert_eq!(2, response.items.len());
+
+        Ok(())
+    }
+
+    #[actix_web::test]
```
jcrossley3, line 170: I'm really digging the `test-log` crate you found. Can you change this to `#[test_log::test(actix::test)]` so you can change the `println!` below to `log::debug!` maybe?

modules/graph/src/endpoints/vulnerability.rs
```diff
@@ -32,3 +83,125 @@ pub async fn affected_products(
...
+                .app_data(web::Data::from(graph.clone()))
+                .configure(crate::endpoints::configure),
+        )
+        .await;
+
+        let uri = "/api/v1/vulnerability?q=Bob";
+
+        let request = TestRequest::get().uri(uri).to_request();
+
+        println!("{:#?}", request);
```
jcrossley3, line 198: This one.

jcrossley3 (changes requested):
Assuming it works... nice! :smile:

modules/graph/src/endpoints/vulnerability.rs
```diff
@@ -1,7 +1,56 @@
+use crate::endpoints::Error;
 use crate::graph::Graph;
 use actix_web::{get, web, HttpResponse, Responder};
+use trustify_common::model::{Paginated, PaginatedResults};
+use trustify_entity::vulnerability::Model;
+use trustify_module_search::model::SearchOptions;
+//("q" = Option<String>, Query, description = "Search filter"),
```
ctron, line 7: I assume this can be removed.

entity/src/vulnerability.rs
```diff
@@ -1,7 +1,8 @@
 use crate::{advisory_vulnerability, vulnerability_description};
 use sea_orm::entity::prelude::*;
+use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
+#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel, Deserialize, Serialize)]
```
ctron, line 5: If I understand correctly, then we're directly exposing the database structure here. I would really want to avoid that.

Even though we only have three fields: having id and identifier is already confusing. Exposing any internal database ID doesn't make much sense IMHO.

ctron (changes requested):
I would prefer some abstraction between the database model and the model we use on the HTTP endpoint.

entity/src/vulnerability.rs
bobmcwhirter: Ah, good catch. I'll create a DTO.

ctron (approved):

jcrossley3 (approved):

PR #163, opened 2024-04-15
fix: don't treat all possible query values as strings

jcrossley3:
Need to figure out how sophisticated we want to be with dates.


ctron:
May I ask again, why are we spending time on re-solving all those problems?

jcrossley3:
> May I ask again, why are we spending time on re-solving all those problems?

No :wink:

bobmcwhirter (approved):

PR #167, opened 2024-04-16
feat: allow searching for SBOMs

ctron:


modules/search/src/model.rs
```diff
@@ -19,6 +19,10 @@ pub struct SearchOptions {
 pub struct FoundAdvisory {
```
bobmcwhirter, line 19: I've similar DTOs (AdvisorySummary, AdvisoryDetails) in the graph models for DTOing.

modules/search/src/service.rs
```diff
@@ -84,4 +84,42 @@ impl SearchService {
                 .collect(),
         })
     }
+
+    // `filters` should be of the form, "full text search({field}{op}{value})*", e.g.
+    // "some text&published>=2020/11/11&location=localhost&severity=low|high&modified=true"
+    pub async fn search_sboms<'a>(
```
bobmcwhirter, line 90: I'd hoped this would just be `graph.sboms(...same params...)` which could return the entire collection or a filtered subset.

modules/search/src/service.rs
```diff
@@ -84,4 +84,42 @@ impl SearchService {
...
+        &self,
+        filters: String,
+        sort: String,
+        paginated: Paginated,
+    ) -> Result<PaginatedResults<FoundSbom>, Error> {
+        let mut select = sbom::Entity::find()
```
bobmcwhirter, line 96: I've also used this pattern, and while not on your shoulders, I feel like we are repeating ourselves a lot, and perhaps @jcrossley3 could get us some more ergonomics around using search/sort.

modules/search/src/endpoints.rs
```diff
@@ -36,3 +40,23 @@ async fn search_advisories(
         service.search_advisories(q, sort, paginated).await?,
     ))
 }
+
+#[utoipa::path(
+    context_path = "/api/v1/search",
+    tag = "search",
+    params(
+        ("q", Query, description = "The query expression"),
```
bobmcwhirter, line 48:
```
params(
    SearchOptions
)
```

will bring all the docs for both `q` and `sort` and simplify the writing of each search-able endpoint.

Similar with `Paginated`.

ctron:
> Per conversation with @jcrossley3 yesterday, I think we decided `/api/v1/sbom` is the collection of all sboms, and to search within, `/api/v1/sbom?q=...` and avoid an `/api/v1/search/whatever` parallel set of endpoints.

I kept it aligned with the current pattern. We can of course refactor, but maybe it makes sense to do this outside of this PR. Either before or after merging this one.

> Also, I'd hoped to move any format-specific document-parsing and weaving into the graph out of graph crate and over to ingestor.

Makes sense, but right now that's tricky, because the tests of `graph` actually parse SPDX SBOMs. We can would need to either duplicate parsing code, shuffle tests around into a another crate, or come up with some alternate idea. Again not sure if we want to pack this into this PR.

modules/search/src/service.rs
bobmcwhirter: Not necessarily `locate_sbom(...)` but perhaps a new `.sboms(...)` method to do the fetch/filter.

ctron: Hm, I can't seem to find a method with that name.

bobmcwhirter: Yeah, we probably need to write it.

bobmcwhirter:
Also, agree with subsequent PR to shuffle locations of things.

ctron:
Superseded by https://github.com/trustification/trustify/pull/169

Issue #168, opened 2024-04-16
DRY up the query code


jcrossley3:
I've also used this pattern, and while not on your shoulders, I feel like we are repeating ourselves a lot, and perhaps @jcrossley3 could get us some more ergonomics around using search/sort.

_Originally posted by @bobmcwhirter in https://github.com/trustification/trustify/pull/167#discussion_r1567327786_

PR #170, opened 2024-04-16
Add `graph.sboms(query, paginated, tx).await`

bobmcwhirter:
Limited to querying/filtering only the Sbom entity proper, which only includes location and hash.


jcrossley3 (approved):
I'll factor that duplicated logic out soon, i.e. #168

ctron (commented):
Not a strong opinion, but to me the `Graph` already feels bloated and doesn't have a clear scope. Searching documents doesn't seem graph-related to me. And why do SBOMs get in there but not advisories?

But feel free to merge it anyway.

bobmcwhirter:
Where would you suggest putting the filtering? And the graph does hold advisory tables akin to SBOM tables yeah?

ctron:
> Where would you suggest putting the filtering? And the graph does hold advisory tables akin to SBOM tables yeah?

I am not sure we have a "graph" owning things. I think we have a big entity model, and that's good. And then we have aspects working on that model, "graph" being one of them. But we also have "ingestion", "importer", "storage", … and maybe something like "analytics", "proposals", "notifications", in the future. "search" feels like one of those aspects with a good scope.

I don't think we can just stuff it all into `Graph`, for me, that is too complex already.

bobmcwhirter:
I guess the question is... does a given thing (ingestors, etc) work against the DB entities directly? Or do they would against the `Graph`?

The current ingestors operate against `Graph`, which is a pretty finite set of operations.

Do we feel good about moving more advanced read-type activities against the entities to another module? Can we avoid repetition of stuff?

ctron:
I think it's a trade off/balance we need to strike. I personally think that without separating at least a bit, it will get ugly pretty soon (if not already). And I agree, we need to be careful not to repeat stuff. Divide and conquer.

ctron:
So if `Graph` is SBOM only, then the name feels wrong (too generic). My assumption was that there will be advisories/vulnerabilities int here too.

If its not SBOM only, then advisory search should go in there too. If it is SBOM only, then why is SBOM search in graph, but advisory search in "search"?

I think the rule of extracting "on top read only stuff" is a good start!

Issue #173, opened 2024-04-16
Make db::DbStrategy go away


bobmcwhirter:
In reality, with the exception of `for_test(...)` we don't use `db::DbStrategy::Managed(Arc<...>)`, while we *do* use `config::DbStrategy::Managed` but ultimately manage it from the `trustd` wrapper around the DB. The Db we construct and give to Graph is `External` from its POV, which makes complete sense.

I suspect we should rework the `for_test(...)` logic to be more test-fixure-y, and less ingrained in Db/Graph logic.

jcrossley3:
Fixed by #174

Issue #135, opened 2024-04-04
Pull up infrastructure to `trustd`


ctron:
Since a recent change the `infrastructure` part is no longer wrapping the lifecycle of the application, but only of the `http` part. That should be changed.

Whatever `trustd` run should be wrapped with the infrastructure stuff. That would also eliminate the need to use `println`.

bobmcwhirter:
closable?

ctron:
Not yet.

PR #174, opened 2024-04-16
fix: make PM mode work as a sane person might expect

### LINKED ISSUE #171: Determine if PM-Mode retains the DB across executions
In theory, $PWD/.trustify/... should be retained across runs and not blow away the DB each time by default.

jcrossley3:
Fixes #171

And resolves #185 in favor of subcommands:

- `trustd api`
- `trustd db create|migrate|refresh`
- `trustd`

The latter no-arg command is "PM mode", which fires up an embedded PostgreSQL instance (installed beneath `.trustify/`) and then runs `trustd db migrate` and `trustd api --auth-disabled`.


jcrossley3:
FYI, i put those `unwrap`'s in there (the ones clippy is complaining about) to force a panic because the app was just silently halting. Need to figure out the right place to init the env_logger, because putting it in trustd's main causes a WARN later when the infra configures its tracing.

What's the right way to do things around here?

bobmcwhirter:
I think at that point either eprintln and/or expect()

ctron:
> What's the right way to do things around here?

https://github.com/trustification/trustify/issues/135

Like here: https://github.com/trustification/trustification/blob/04a3554fa95b500ce2208c7d8ed97a96fa47788e/spog/api/src/lib.rs#L97-L112

ctron:
> We always CREATE_DATABASE and ignore "already exists" errors.

Assuming we want to run that in production at some point, we would have database creation as well as running migrations outside of the main process, in order to limit the permissions the actual database user requires.

So we at least would need a mode which just accepts an existing database, assumes that the structure is present, and does not try to modify anything beyond actual data.

jcrossley3:
Good points. I guess we need to ensure the "create db and ignore errors" behavior is only for "managed" db's. Which kind of conflicts with #173 I think.

I don't have a clear solution in mind yet, so open to ideas.

bobmcwhirter:
I agree that in Production we wouldn't want `trustd` to be creating DBs.

I'm not sure I agree that running migrations should be outside of `trustd`'s control though.

Thinking back to my Ruby On Rails days, migrations were always in-process at boot time and that seems to have worked?

trustd/src/main.rs
```diff
@@ -84,21 +85,23 @@ impl Trustd {
...
                 temporary: false,
                 installation_dir: db_dir.clone(),
+                data_dir,
                 ..Default::default()
             };
 
             let mut postgresql = PostgreSQL::new(PostgreSQL::default_version(), settings);
-            postgresql.setup().await?;
-            postgresql.start().await?;
+            postgresql.setup().await.unwrap();
```
bobmcwhirter, line 100: I'd at least `.expect("postgres did not install")` or something friendlier than a straight-up panic with a stack.

bobmcwhirter:
wrt "create db and ignore errors" maybe needs to float up out of `Db` and into the `trustd` wrapper/setup bits.

We either take an external config, or manage/bootstrap/ignore-errors up a db, and then ctor the `Db` with whichever database we ultimately ended up with booting up in PM-mode-or-not. From the `Db` perspective, they're all external. Separation of responsibility that `Db` just ... is a database. 

Externalize the optional management of the aforementioned database.

ctron:
> I'm not sure I agree that running migrations should be outside of `trustd`'s control though.
> 
> Thinking back to my Ruby On Rails days, migrations were always in-process at boot time and that seems to have worked?

The downside of this is, that the user connecting to the database would need some "admin" permission set. Which the application itself would not need.

Following the idea of "limiting access" … it feels better to not give a process, which receives connections from the internet, admin permissions to the DB.

So the process we've in place for GUAC is: init container -> runs migrations, actual container -> runs code … the latter one has a limited set of permissions. And that works quite well.

ctron:
wrt `panic`, `expect` … just a few lines above this code we use `?`, which I do prefer. So why break that pattern?

bobmcwhirter:
Ah, I'd not realized that `CREATE TABLE` and other migratory actions required higher permissions than simple user. 

Agreed on the bifurcated migration-vs-runtime.

bobmcwhirter:
@jcrossley3 so I think we can still make `db::DbStrategy` go away, and pass in some `bool` for `attempt_migrations` for the  PM-mode use-case, else we skip that branch in the init-container/production case.

jcrossley3:
@ctron With my last commit to this PR, I think I have the `Infrastructure` in trustd now, but I didn't have a really good example to guide me -- the `trust` cli in the old repo lacks `Infrastructure`, too. And I'm still seeing unexpected warnings when the server initializes its own `Infrastructure` -- I feel like that should only be done once, but I didn't know how to make the `InitData::new` stuff work without a context.

Here are the steps to reproduce to see the output I'm getting. I'm seeing correct logging now -- no println's -- and my exit code is correct, but I don't think I'm doing it correctly, because I have 2 "Default to INFO logging" lines: one for `trustd` and one for the `server` it starts.

```
[jim@fedora trustify]$ rm -rf ./.trustify/

[jim@fedora trustify]$ cargo run --bin trustd -- --auth-disabled
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/trustd --auth-disabled`
Default to INFO logging; use RUST_LOG environment variable to change
[2024-04-17T21:06:10.358Z INFO  trustify_infrastructure::infra] Infrastructure endpoint is disabled
[2024-04-17T21:06:10.358Z INFO  trustd] setting up managed DB
[2024-04-17T21:06:15.830Z ERROR trustd] Error: deadline has elapsed

[jim@fedora trustify]$ cargo run --bin trustd -- --auth-disabled
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/trustd --auth-disabled`
Default to INFO logging; use RUST_LOG environment variable to change
[2024-04-17T21:06:22.342Z INFO  trustify_infrastructure::infra] Infrastructure endpoint is disabled
[2024-04-17T21:06:22.342Z INFO  trustd] setting up managed DB
[2024-04-17T21:06:22.446Z INFO  trustd] postgresql installed under "/home/jim/src/trustification/trustify/.trustify/postgres"
[2024-04-17T21:06:22.446Z INFO  trustd] running on port 45441
Default to INFO logging; use RUST_LOG environment variable to change
Error initializing logging: SetLoggerError(())
[2024-04-17T21:06:22.447Z WARN  trustify_server] Authentication is disabled
[2024-04-17T21:06:22.447Z INFO  trustify_common::db] connect to postgres://trustify:trustify@localhost:45441/trustify
[2024-04-17T21:06:22.461Z ERROR trustd] error Connection Error: error returned from database: database "trustify" does not exist
    
    Caused by:
        0: error returned from database: database "trustify" does not exist
        1: error returned from database: database "trustify" does not exist
        2: database "trustify" does not exist
```

trustd/src/main.rs
```diff
@@ -25,119 +23,52 @@ pub enum Command {}
...
-                },
-                Err(err) => {
-                    log::error!("fundamental error {:?}", err);
+    async fn run_command(self) -> anyhow::Result<ExitCode> {
+        match self.command {
+            Some(Command::Api(run)) => run.run().await,
+            Some(Command::Db(run)) => run.run().await,
+            None => {
+                let Some(Command::Db(mut db)) =
+                    Trustd::parse_from(["trustd", "db", "migrate"]).command
```
bobmcwhirter, line 52: So we're taking "no arguments" (and only the no-arg case?) and re-feeding it back in implying a set of arguments?

trustd/src/main.rs
```diff
@@ -1,19 +1,17 @@
...
+mod db;
 
 #[allow(clippy::large_enum_variant)]
 #[derive(clap::Subcommand, Debug)]
-pub enum Command {}
+pub enum Command {
+    /// Run the API server
+    Api(trustify_server::Run),
+    /// Manage the database
+    Db(db::Run),
```
bobmcwhirter, line 13: I think we'd also want the `Importer` loop command?

trustd/src/main.rs
```diff
@@ -25,119 +23,52 @@ pub enum Command {}
...
-    pub swagger_ui_oidc: SwaggerUiOidcConfig,
 }
 
 impl Trustd {
     async fn run(self) -> ExitCode {
         match self.run_command().await {
             Ok(code) => code,
             Err(err) => {
-                log::error!("Error: {err}");
+                eprintln!("Error: {err}");
```
bobmcwhirter, line 33: Hah, I suggested `eprintln!(...)`

trustd/src/main.rs
```diff
@@ -25,119 +23,52 @@ pub enum Command {}
...
+                db.run().await?;
+
+                let api = Trustd::parse_from([
+                    "trustd",
+                    "api",
+                    "--auth-disabled",
+                    "--db-port",
+                    &postgres.settings().port.to_string(),
+                ]);
+                Ok(tokio::task::spawn_local(api.run()).await?)
```
bobmcwhirter, line 69: We'd probably want to embody this different, if we're also spawning the importer loop and wtf knows what else in the future?

trustd/src/main.rs
jcrossley3: Yes, `None` => PM mode. In fact, that's a good name for a fn to put that in.

trustd/src/main.rs
jcrossley3: Possibly. Out of scope for this PR, though. I just want to make PM mode make sense.

trustd/src/main.rs
jcrossley3: The importer loop is already "running". It's data driven, so if a PM adds an importer via the instructions in the README, it'll run each time they fire up trustd, until they delete it, because now the data won't be destroyed with each run.

trustd/src/main.rs
ctron: I think this would be wrapped by the `Infrastructure` stuff. Or, alternatively, errors should bubble up to `main`, and there we could do the "catch all" error reporting.

jcrossley3: Or are you suggesting we *only* have the `Infrastructure` stuff in `trustd` and nowhere else? Because that would involve a lot of rework in `server/src/lib.rs`.

jcrossley3: Recounting here the results of an external chat:

- have the infrastructure in the actual (sub-)command
- the top level main would log terminal errors using eprintln
- other, more cli like commands, can use env_logger

In a nutshell:

- cli sub-commands use env_logger
- server style commands use infrastructure

modules/fetch/src/endpoints/advisory.rs
```diff
@@ -79,9 +81,10 @@ mod test {
     use trustify_module_ingestor::graph::advisory::AdvisoryInformation;
     use trustify_module_ingestor::graph::Graph;
 
+    #[test_context(TrustifyContext, skip_teardown)]
     #[test(actix_web::test)]
-    async fn all_advisories() -> Result<(), anyhow::Error> {
-        let db = Database::for_test("api_all_advisories").await?;
+    async fn all_advisories(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
```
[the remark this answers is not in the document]
jcrossley3: sure thing

trustd/src/db.rs
```diff
@@ -0,0 +1,91 @@
...
+    Create,
+    Migrate,
+    Refresh,
+}
+
+impl Run {
+    pub async fn run(self) -> anyhow::Result<ExitCode> {
+        init_tracing("db-run", Tracing::Disabled);
+        use Command::*;
+        match self.command {
```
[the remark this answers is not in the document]
jcrossley3: No and no. See the last line of `Database::bootstrap`: `Self::new(database).await?.migrate().await`

I couldn't think of a use case where I'd want to bootstrap a DB and *not* migrate it. But if you think we need that, we can revisit.

trustd/src/main.rs
```diff
@@ -25,125 +23,54 @@ pub enum Command {}
...
-                    }
-                    log::error!("\t{err}");
-                }
-
-                ExitCode::FAILURE
-            }
+    async fn run(self) -> anyhow::Result<ExitCode> {
+        match self.command {
+            Some(Command::Api(run)) => run.run().await,
+            Some(Command::Db(run)) => run.run().await,
```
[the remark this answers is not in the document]
jcrossley3: I feel like you've asked about importer loops a lot, so I'm guessing my answers aren't punching through.

Currently, the API server will create a "loop" in response to a request like this:
```
http POST localhost:8080/api/v1/importer/redhat-sbom sbom[source]=https://access.redhat.com/security/data/sbom/beta/ sbom[keys][]=https://access.redhat.com/security/data/97f5eac4.txt#77E79ABE93673533ED09EBE2DCE3823597F5EAC4 sbom[disabled]:=false sbom[onlyPatterns][]=quarkus sbom[period]=30s sbom[v3Signatures]:=true
```
And you can see it like so:
```
http GET localhost:8080/api/v1/importer
```
Assuming you have https://httpie.io/ installed, try running those commands after firing up PM mode. Will that functionality suit the use cases you have in mind for importer loops? If not, what do we need to change?

trustd/src/main.rs
```diff
@@ -25,125 +23,54 @@ pub enum Command {}
...
+    let Some(Command::Db(mut db)) = Trustd::parse_from(["trustd", "db", "migrate"]).command else {
+        unreachable!()
+    };
+    let postgres = db.start().await?;
+    if postgres.database_exists(&db.database.name).await.is_err() {
+        db.command = db::Command::Create;
     }
+    db.run().await?;
+
+    let api = Trustd::parse_from([
```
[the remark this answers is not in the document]
jcrossley3: Interesting. I do love it, because it mirrors what the user would do. But I agree it'd be nice for the compiler to complain if we subsequently change the CLI.

Interestingly, my first pass was as you suggest, but I found myself wanting to implement `Default` for things that are already kind of implemented in the form of clap annotations that I didn't want to replicate. And I wanted to take advantage of that clap config, hence `parse_from`.

It's definitely a tradeoff, but for something like "PM mode", I was ok with it. Better than ok, actually, but if you insist...

common/src/db/test.rs
```diff
@@ -0,0 +1,41 @@
...
+        let config = crate::config::Database {
+            username: "postgres".into(),
+            password: "trustify".into(),
+            host: "localhost".into(),
+            name: "test".into(),
+            port: postgresql.settings().port,
+        };
+        let db = crate::db::Database::bootstrap(&config).await.unwrap();
+
+        TrustifyContext { db, postgresql }
```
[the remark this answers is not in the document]
jcrossley3: Partly. `DbStrategy`'s primary benefit was keeping the PostgreSQL instance in scope for the duration of a test. Which is what the `TrustifyContext` now does.

bobmcwhirter (approved):

PR #176, opened 2024-04-17
Serve UI in trustify 

carlosthe19916:
@bobmcwhirter you mentioned the UI being served by the trustd server. Here is a working example of it. 
- The current PR allows the UI to be served by trustd but the UI can still live alone without the need of trustd (e.g a container of only the UI can be created). So the UI can live in standalone mode but also within trustd.
- This PR implies moving back https://github.com/trustification/trustify-ui to this repository. I don't know if that is a good idea, in the past there were complains about it therefore the UI was moved to its own repository. What is your opinion?

I opened this PR mainly to show your idea in action. I'll defer the decision of where the UI code should live. So no need to merge this PR until we are happy with it.

### How to test this PR
- Checkout this PR
- Make sure you have NodeJS 20 installed and npm >9.5.0
- Execute:
```
cargo run --bin trustd -- --auth-disabled
```

Open http://localhost:8080


bobmcwhirter:
I love the idea but I suspect we can find a way so you can keep UI in your own repo. I'll ponder a bit.

ctron:
I think it's great to see that this is within reach! I really love that.

When it comes to having the code in the same repository, then the only thing that was bothering me in the past was the fact that there had been numerous warnings spamming the CI logs. So making a change on the backend side brought up a lot of unrelated warnings from the UI side. If that's solved (and stays solved in general) then I think it would be good to have both parts in the same repository. But that is just me. I understand people having a different preference.

bobmcwhirter:
So, you *know* I'm a monorepo kinda guy.

@carlosthe19916 -- what's your view on where it should live? Do you have a strong opinion? If merged back in, can we ensure that CI logs are quite clean?

carlosthe19916:
> So, you _know_ I'm a monorepo kinda guy.
> 
> @carlosthe19916 -- what's your view on where it should live? Do you have a strong opinion? If merged back in, can we ensure that CI logs are quite clean?

- I personally have no preference on **_where_** the code lives.
- I do have a preference in terms of being able to merge my own PRs in the `ui/` directory. Not sure what configuration to put in place in Github so I would be able to merge PRs without approvals (only on the `ui` directory). The other option is that the team could be willing to review UI code continuously so I don't wait long periods of time before getting things done. Could we relax the rules for the `ui` directory? If possible, let's move the UI here :)

ctron:
> * I do have a preference in terms of being able to merge my own PRs in the `ui/` directory. Not sure what configuration to put in place in Github so I would be able to merge PRs without approvals (only on the `ui` directory). The other option is that the team could be willing to review UI code continuously so I don't wait long periods of time before getting things done. Could we relax the rules for the `ui` directory? If possible, let's move the UI here :)

I think that points a big flaw. I am ok with just pushing to main or even force pushing every now and then. However, that shouldn't be the norm. We are actually required to have a review process in place by our internal policies. And for good reason!

If we do not have enough team members to do even a basic review, then I think we have a problem. Just ignoring that problem by not doing reviews feels completely wrong.

bobmcwhirter:
Would @gildub be a suitable reviewer? He has some react experience yeah?

jcrossley3:
From the commentary, it seems obvious we should keep the two repos separate for now. Everyone has mentioned potential downsides to doing so, and the only barely recognizable upside I see is "you know I'm a monorepo kinda guy".

bobmcwhirter:
@jcrossley3 well, the original motivation was "having trustd" serve it.  But, I also think we can sort that across repos, somehow.

gildub:
I think a monorepo approach makes sense for a project organized with full stack teams working across the board (likely broken down by services) otherwise having separate repos are easier to maintain UI dedicated team.
So maybe we need to discuss the most adapted team organization we want/need ?

In the meantime, I can review UI PRs.

helio-frota:
> otherwise having a separate repo is easier to maintain by an UI dedicated team.

> In the meantime, I can review UI PRs.

I volunteer to also review UI :+1:

carlosthe19916:
@bobmcwhirter @jcrossley3 @ctron I have updated this PR so the UI code still lives in its own repository.
There is still more things to do but this should be a good enough first attempt to have the PM mode working (PM mode does not have authentication).

To test this PR you only need to run

```shell
cargo run --bin trustd
```

Just like it is a requisite to have Rust installed, it is also a requisite to have NodeJs 20 installed in your computer. You can use  https://github.com/nvm-sh/nvm for installing NodeJS.

jcrossley3:
Sorry, but I'm an idiot, and I require more hand-holding. Can you include specific commands at the top of the README for seeing the UI? All I see is "Error" when I visit http://localhost:8080. You can assume I have `git` and `nvm` installed, but what idempotent commands can I run to ensure I actually see something when I try for the first time?

You can fill a barn with what I don't know about git submodules and nvm. My `ui/` directory is empty, so I'm sure I'm missing something.

bobmcwhirter:
Ah submodules strike again!

jcrossley3:
I don't think I want this. I'd prefer to treat the two repos separately. Serving the UI from the api server is a cool deployment feature, and I'm down for adding whatever tooling is necessary for packaging that, but IMO it's an overall negative for development.

If I get outvoted, we need to at least fix `build.rs` so that it doesn't think it needs to build `trustd` even when no source changed.

trustd/build.rs
```diff
@@ -34,7 +34,9 @@ fn main() {
 }
 
 fn clone_ui() -> io::Result<ExitStatus> {
-    Command::new(GIT_CMD).args(["submodule", "update"]).status()
+    Command::new(GIT_CMD)
+        .args(["submodule", "update", "--init", "--recursive"])
```
jcrossley3, line 38: Ah! I figured this out on my own, not realizing you were attempting to prevent me from having to do that.

carlosthe19916:
> Sorry, but I'm an idiot, and I require more hand-holding. Can you include specific commands at the top of the README for seeing the UI? All I see is "Error" when I visit http://localhost:8080. You can assume I have `git` and `nvm` installed, but what idempotent commands can I run to ensure I actually see something when I try for the first time?
> 
> You can fill a barn with what I don't know about git submodules and nvm. My `ui/` directory is empty, so I'm sure I'm missing something.

- I have added some minor changes to the README.md file
- The idea of this PR is for the user to need only 1 command to start everything, and that command should be `cargo run --bin trustd`. If that is not working, then that was not this PR's goal and I should change "something" for making it work.

I know you already confirmed you have NVM but could you double check you have NodeJS 20?
```
node -v
v20.11.1
```

- In theory you should not even need to clone the submodules as it should be done by the code here https://github.com/trustification/trustify/pull/176/files#diff-d6cffee37c35a86ae367b378bde922d3a28bd32512b9d5ed3bb81ffba7027492R37

@jcrossley3 you spotted a mistake on my side. The command for cloning the git submodule was not correct at all, I fixed it in my last commits

Sorry for the mistake, but could you test it again please?

trustd/build.rs
carlosthe19916: yep, the idea is not requiring the user to execute additional commands

jcrossley3: This change fixed my issue, thanks!

trustd/build.rs
```diff
@@ -0,0 +1,78 @@
+use std::path::Path;
+use std::process::{Command, ExitStatus};
+use std::{fs, io};
+
+static UI_DIR: &str = "../ui";
+static UI_DIR_SRC: &str = "../ui/src";
```
jcrossley3, line 6:
```suggestion
```
This path doesn't exist, so cargo builds each time.

trustd/build.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[cfg(windows)]
+static NPM_CMD: &str = "npm.cmd";
+#[cfg(not(windows))]
+static NPM_CMD: &str = "npm";
+
+fn main() {
+    println!("Build Trustify - build.rs!");
+
+    println!("cargo:rerun-if-changed={}", UI_DIR_SRC);
```
jcrossley3, line 23:
```suggestion
    println!("cargo:rerun-if-changed={}", UI_DIR);
```
This is what you meant, I think, and cargo only builds when something beneath there changes.

carlosthe19916:
> but IMO it's an overall negative for development.
I agree.

jcrossley3 (changes requested):
I still don't think I want this, but if we must, we need to bring the log output back.

bobmcwhirter:
Let's see if we can't crate it like the swagger UI and use as a dependency instead of git shenanigans. Say shenanigans one more time.

bobmcwhirter:
Then random jackholes like me won't need NPM.

jcrossley3:
> Then random jackholes like me won't need NPM.

You're expecting those same random jackholes to install `rustup`, correct? `nvm` is the `rustup` analog, and both seem fairly jackhole-proof to me.

jcrossley3:
But I agree, a crate would be nice.

bobmcwhirter:
Yah I'm wrong. Source means we'd all still compile it locally including NPM. 

Which is still better than submodules.

bobmcwhirter:
Closing, solved differently.

PR #181, opened 2024-04-17
refactor the filtering logic into a single `filtering` fn

### LINKED ISSUE #168: DRY up the query code

jcrossley3:
Fixes #168

### DIFF
--- a/modules/graph/src/graph/sbom/mod.rs
+++ b/modules/graph/src/graph/sbom/mod.rs
@@ -58,9 +58,7 @@ impl Graph {
     ) -> Result<PaginatedResults<SbomContext>, Error> {
         let connection = self.connection(&tx);
 
-        let SearchOptions { sort, q } = search;
-
-        let limiter = sbom::Entity::find().filtering(&q, &sort)?.limiting(
+        let limiter = sbom::Entity::find().filtering(search)?.limiting(
             &connection,
             paginated.offset,
             paginated.limit,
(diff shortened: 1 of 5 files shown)


modules/graph/src/graph/sbom/mod.rs
```diff
@@ -60,22 +60,11 @@ impl Graph {
...
-                .collect::<Result<Vec<_>, _>>()?
-                .iter()
-            {
-                select = select.order_by(s.field, s.order.clone());
-            }
-        }
-        select = select.order_by_desc(sbom::Column::Id);
-
-        let limiter = select.limiting(&connection, paginated.offset, paginated.limit);
+        let limiter = sbom::Entity::find().filtering(&q, &sort)?.limiting(
```
bobmcwhirter, line 63: Useful to just take the `SearchOptions` instead of a pair of Strings?

modules/search/src/query.rs
```diff
@@ -14,22 +17,53 @@ use time::{Date, OffsetDateTime};
 // Public interface
 /////////////////////////////////////////////////////////////////////////
 
-pub struct Filter<T: EntityTrait> {
+pub trait Query<T: EntityTrait> {
+    fn filtering(self, q: &str, sort: &str) -> Result<Select<T>, Error>;
+}
+
+impl<T: EntityTrait> Query<T> for Select<T> {
+    fn filtering(self, q: &str, sort: &str) -> Result<Self, Error> {
```
bobmcwhirter, line 25: Pair of strings which can easily be typed backwards?

bobmcwhirter (changes requested):

modules/graph/src/graph/sbom/mod.rs
jcrossley3: Good call! Done!

Issue #161, opened 2024-04-15
Hash more than just SHA256


bobmcwhirter:
Some day SHA256 will be broken maybe, so let's go ahead and hash SHA384 and whatever else seems sensible.

bobmcwhirter:
@helio-frota if you're interested, you could add SHA384 here:

https://github.com/trustification/trustify/blob/main/modules/ingestor/src/service/hashing.rs

crsche:
Hey @bobmcwhirter, could I give this issue a shot. I'm a bit new to group contribution, but I'll see what I can do.

bobmcwhirter:
Right on, thanks! Should be fairly straight-forward (famous last words). Just fork, branch, submit a PR and I'll be glad to review it.

bobmcwhirter:
Closed implemented by #182

PR #182, opened 2024-04-18
feat(ingestor): default to sha384 hashing and allow for algorithm customization

crsche:
## Changes

- Added the option to use any `ring::digest::Algorithm` for `trustify_module_ingestor::service::hashing::HashingRead` - defaulting to SHA384 (previously SHA256, see #161)

- Changed associated functions to match `ring::digest::Digest` behavior - mainly `.finish()`, that takes ownership of main to prevent any misuse.

- Added tests for hashing / reading.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5671,6 +5671,7 @@ dependencies = [
  "humantime",
  "log",
  "packageurl",
+ "rand",
  "reqwest",
  "ring 0.17.8",
  "sbom-walker",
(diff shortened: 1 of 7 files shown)


crsche:
I added SHA384 and SHA512 to `HashingRead`, but I didn't change usage anywhere—SHA256 is still the default. Should I make it so SHA384 is the default for loaders?

crsche:
I added SHA384 and SHA512 to `HashingRead`, but I didn't change usage anywhere—SHA256 is still the default. Should I make it so SHA384 is the default for loaders?

bobmcwhirter:
We probably should adjust the loaders (and the underlying entities/db schema) a bit, but I think that'd be a fair activity for some subsequent PR. Keeping this one tight and focused is perfect.

bobmcwhirter (approved):

bobmcwhirter:
Thank you for your contribution, should be merging shortly.

PR #184, opened 2024-04-19
Remove the terribly too clever usage of `From<...>` for the various

bobmcwhirter:
graph Contexts.

In preparation to split `graph` into better-organized parts.


bobmcwhirter:
@gildub finally removing the complexity you noticed around the `.into()` and `From<...>` cleverness for creating contexts.

jcrossley3 (approved):

PR #192, opened 2024-04-22
Checkpoint moving a lot of `graph` into `ingestor` who uses it the most.

bobmcwhirter:


bobmcwhirter:
Yes, it's large and I apologize. And actually churned up more things I'd like to clean up, but outside of the scope of *this* particular PR.

modules/ingestor/src/graph/sbom/mod.rs
```diff
@@ -44,7 +43,7 @@ pub struct SbomInformation {
 }
 
 impl From<()> for SbomInformation {
```
jcrossley3, line 45: This seems weird. Where/why is this used?

modules/ingestor/src/graph/sbom/tests.rs
```diff
@@ -45,7 +44,7 @@ async fn query_sboms() -> Result<(), anyhow::Error> {
         )
         .await?;
 
-    let other_sbom = system
+    let _other_sbom = system
```
jcrossley3, line 47: Do we have to assign these? Seems like a smell.

modules/storage/Cargo.toml
```diff
@@ -5,9 +5,9 @@ edition = "2021"
 publish = false
 
 [dependencies]
-trustify-common = { path = "../../common" }
-trustify-entity = { path = "../../entity" }
-trustify-module-graph = { path = "../graph" }
+trustify-common = { workspace = true }
+trustify-entity = { workspace = true }
+#trustify-module-fetch = { path = "../fetch" }
```
jcrossley3, line 10: Did you mean for this to remain?

jcrossley3 (approved):
Really too big to review, but I trust you.

modules/storage/Cargo.toml
bobmcwhirter: Nope!

modules/ingestor/src/graph/sbom/tests.rs
bobmcwhirter: also, just carried over, except Clippy started to complain. I dunno why.  Took the easy path to quiet Clippy.

PR #194, opened 2024-04-23
refactor: switch from env_logger to tracing logger

ctron:
Instead of pushing tracing events to log, and from log to env_logger, this consumes log into tracing, and then logs those with trace spans in mind.

This mostly means that trace spans only get logged when they fail.

It still uses the RUST_LOG variable, and defaults to INFO.

On top of that, there's color:

image


ctron:
> Trying it locally, I see some log messages with color, but some without, specifically those from `trustify_module_graph`. Is that expected?

No, that sounds strange. Can you share more details? Screenshot maybe?

ctron:
I moved this back to draft, as I am actually missing some log records. I need some more time to check this. Still, a screenshot would be good.

ctron:
I found the reason for the missing log events. I also re-added the target, I think it's useful (the module/crate). However, I still don't see any non-colored events.

ctron:
The issue with the missing logs is now solved. I can't reproduce the other (non colored logs).

ctron:
This PR also contains a fix for the server not starting up with auth-disabled when no SSO server is running.

bobmcwhirter:
Is this an artifact of `test-log` not using the same logger?

ctron:
Ah yes, that's a bug in test-log: https://github.com/d-e-s-o/test-log/pull/48 …let's see if that gets merged.

Basically there are two frameworks mixed: log and tracing.

ctron:
However, this scope of this PR also is only the binary, not the tests. And yes, we should have the same behavior for tests IMO. But that requires the patch/fix on the test_log side.

jcrossley3 (approved):

PR #199, opened 2024-04-24
docs: first try for some design docs

ctron:
This is a first test on how we could do design docs as part of the code. At least, closer to the code.

I think it works reasonably well. My expectation would be that the docs might not be fleshed out, might have some gaps in less relevant areas, but must be correct!

 I am not sure about the file system structure. Those docs could also go into their respective modules folders.

Rendered version: https://github.com/ctron/trustify/blob/feature/design_docs_1/docs/design/importer.md

### DIFF
--- a/docs/design/importer.md
+++ b/docs/design/importer.md
@@ -1,0 +1,88 @@
+# Importer
+
+The importer is responsible for importing documents into the system.
+
+Module level documentation: [modules/importer/README.md](../../modules/importer/README.md)
+
+## Entities
+
+```mermaid
+erDiagram
+
+IMPORTER only one to zero or more IMPORTER_REPORT : owns
+
+IMPORTER {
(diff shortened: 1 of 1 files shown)


PhilipCattanach:
@ctron - That's really good. It presents very nicely. 
The link to the ReadMe provides the ability to add some supporting narrative to complement the diagrams.
I really like the Entity, Service and Sequence diagrams 
Thank you!
P.S. I thought Bob's observation about  maintaining this information in lock step with all of the other assets in the project was right on the money. I think it will be beneficial to the entire team.

jcrossley3 (approved):
Very nice! Things I like:

- it's text!
- it's in the repo!
- it's rendered automatically by github

Thanks!

carlosthe19916 (approved):

PR #207, opened 2024-04-26
chore: auto-migrate db when using --devmode

ctron:


server/src/lib.rs
```diff
@@ -115,6 +115,11 @@ impl InitData {
         };
 
         let db = db::Database::new(&run.database).await?;
+
+        if run.devmode {
+            db.migrate().await?;
+        }
+
```
jcrossley3, line 122: Is this the answer to my question above? That it saves you a re-assignment. I can buy that.

```suggestion

        if run.devmode {
            db = db.migrate().await?;
        }

```

jcrossley3 (approved):
Asked and answered :wink:

server/src/lib.rs
ctron: Yes, that was the reason. `migrate` didn't seem to consume `self`, so I didn't want to `.clone().migrate()`, but just use `&self`.

PR #209, opened 2024-04-26
run tests against EXTERNAL_TEST_DB if set

jcrossley3:
Use at your own risk! :)

### DIFF
--- a/common/src/config.rs
+++ b/common/src/config.rs
@@ -1,3 +1,4 @@
+use std::env;
 use std::fmt::{Display, Formatter};
 use std::path::PathBuf;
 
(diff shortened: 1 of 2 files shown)


common/src/config.rs
```diff
@@ -22,6 +23,23 @@ pub struct Database {
     pub name: String,
 }
 
+// TODO: figure out how to make clap use this and remove the redundant
+// #[arg(...)]'s above
+impl Default for Database {
```
ctron, line 28: I think something around that could work:

```rust
fn test() {
    use clap::Parser;

    #[derive(clap::Parser, Debug, Clone)]
    pub struct Cli {
        #[command(flatten)]
        database: Database,
    }

    let _ = Cli::try_parse().unwrap().database;
}
```

You could also use the `clap::Parser` derive directly on the `Database` struct. I don't know if that has any side-effects.

jcrossley3: What would the `#[arg(...)]` lines become in the `Database` struct itself?

ctron: > What would the `#[arg(...)]` lines become in the `Database` struct itself?

They wouldn't change.

jcrossley3: Then I'd still have those defaults in two places? I'd like the `Default` impl to be the only place.

jcrossley3: Sorry, I don't understand what you're suggesting. I just want the defaults in a single place. And the only risk of panic is if `DB_PORT` isn't a number, which is probably fine.

ctron: I was suggesting to replace the hand crafted `std::env::var` code to be replaced by the clap generated parser. Removing that from the `Default` trait into some `fn` which clearly states that "this can panic".

jcrossley3: Hmm. Relying on clap macros to specify default values for structs used in a non-CLI context seems wrong to me.

helio-frota: > in a non-CLI 

yeah :+1:

jcrossley3: @ctron with this last change, I'm using `default_value_t = Self::default().username`. are you ok with that?

ctron (approved):
Let's merge this and I come up with a PR that shows what I had in mind.

PR #215, opened 2024-04-29
Improve performance when ingesting

ctron:
Baseline:

```
2024-04-29T13:53:32.751036Z  INFO trustify_module_ingestor::graph::sbom::spdx::tests::perf: parse: 2s 452ms 964us 857ns    
2024-04-29T13:53:32.751045Z  INFO trustify_module_ingestor::graph::sbom::spdx::tests::perf: ingest 1: 2ms 384us 614ns    
2024-04-29T13:53:32.751052Z  INFO trustify_module_ingestor::graph::sbom::spdx::tests::perf: ingest 2: 8s 766ms 4us 52ns    
2024-04-29T13:53:32.751059Z  INFO trustify_module_ingestor::graph::sbom::spdx::tests::perf: commit: 622us 181ns    
2024-04-29T13:53:32.751065Z  INFO trustify_module_ingestor::graph::sbom::spdx::tests::perf: query: 4ms 384us 992ns    
```

This PR:
```
2024-04-29T13:52:13.437803Z  INFO trustify_module_ingestor::graph::sbom::spdx::tests::perf: parse: 2s 473ms 401us 902ns    
2024-04-29T13:52:13.437813Z  INFO trustify_module_ingestor::graph::sbom::spdx::tests::perf: ingest 1: 2ms 352us 324ns    
2024-04-29T13:52:13.437821Z  INFO trustify_module_ingestor::graph::sbom::spdx::tests::perf: ingest 2: 1s 667ms 844us 146ns    
2024-04-29T13:52:13.437828Z  INFO trustify_module_ingestor::graph::sbom::spdx::tests::perf: commit: 635us 571ns    
2024-04-29T13:52:13.437835Z  INFO trustify_module_ingestor::graph::sbom::spdx::tests::perf: query: 4ms 257us 590ns 
```

This is accomplished using the following approach:

* Using predictable primary keys
* Use upserts to work around the "select else insert" pattern
* Batch the creation of package/version/qualifier entries

This also treats missing version components as an empty string. I don't think this should be a problem, but let's see.

## Predictable keys

Instead of letting the database create keys, which would requires reading those back (which doesn't work well with upserts), we use predictable keys. A v5 UUID based on the primary (unique) information. A v5 UUID is basically a digest/hash, but can be processed by postgres as a dedicated type.

This allows to create a predictable primary key for each of the package, package_version, qualified_package entity.

## Upserts

Instead of the "select else insert" pattern, which requires two dependent SQL statements to create a new entry, we use `INSERT … ON CONFLICT DO NOTHING`. As we now do know the primary key, we can ignore the fact that we don't get the primary key from the database, as we already know it.

## Batched operations

As we now can create (or ignore an existing) entry in a single step, we can also batch inserts using `insert_many`.

The downside with this is, that we do need to self-chunk the batches, as we otherwise might run into: https://github.com/SeaQL/sea-orm/discussions/710 … there's a workaround in there for Postgres, which my might apply later on.


common/src/purl.rs
```diff
@@ -15,13 +16,62 @@ pub enum PurlErr {
...
     pub ty: String,
     pub namespace: Option<String>,
     pub name: String,
     pub version: Option<String>,
-    pub qualifiers: HashMap<String, String>,
+    pub qualifiers: BTreeMap<String, String>,
+}
+
+const NAMESPACE: Uuid = Uuid::from_bytes([
+    0x37, 0x38, 0xb4, 0x3d, 0xfd, 0x03, 0x4a, 0x9d, 0x84, 0x9c, 0x48, 0x9b, 0xec, 0x61, 0x0f, 0x06,
```
bobmcwhirter, line 29: Is this arbitrary or some particular magic?

bobmcwhirter (approved):

PR #218, opened 2024-04-30
refactor: focus on clap for getting configuration

ctron:
This keeps the focus on clap for defining arguments and defaults. In order to not conflict with actual command line arguments to the test binary, it parses with an explicit empty list of arguments, still considering the env-vars.

Also, it restricts the part that can panic into the `external_test_db` function, which is not `pub`.

### DIFF
--- a/common/src/db/test.rs
+++ b/common/src/db/test.rs
@@ -1,6 +1,7 @@
-use std::env;
-
+use crate::config::Database;
+use clap::{ArgMatches, FromArgMatches, Parser};
 use postgresql_embedded::{PostgreSQL, Settings};
+use std::env;
 use tempfile::TempDir;
 use test_context::{test_context, AsyncTestContext};
 use tracing::{info_span, instrument, Instrument};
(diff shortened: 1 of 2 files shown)


common/src/db/test.rs
```diff
@@ -11,14 +12,36 @@ pub struct TrustifyContext {
     tempdir: Option<TempDir>,
 }
 
+/// collect database information for the external test database.
+///
+/// **NOTE:** This may panic in case where the [`Database`] arguments cannot be parsed.
+#[allow(clippy::expect_used)]
+fn external_test_db() -> Database {
+    #[derive(clap::Parser)]
+    struct Cli {
```
jcrossley3, line 21: This seems like such an abuse of clap to me. The computer running the program won't care, but the humans maintaining this code might.

jcrossley3 (commented):
I don't love it. It's exactly what I imagined from your description of it in #209

ctron:
I don't like panicking `Default` implementations, and calling `Default::default()` just to fill a clap default value.

I think this re-uses the existing logic.

ctron:
> I hate that less than I hate constructing a `clap::Parser` just to read env vars, though. Neither alternative is ideal.

I think calling clap for parsing env-var and telling it to skip processing OS args is more ideal than setting a value of `Default::default().foo`.

> I don't like that either, and view having to do it as a weakness of clap.

If you don't like that either, then I don't think going for that approach is a reasonable choice. And taking this to the authors of clap, I would expect two responses: a) there's a way to do what you want already (which is this PR) and b) what's the alternative you're suggesting?

jcrossley3:
> what's the alternative you're suggesting?

Easy!

```
pub struct Run {
    #[command(flatten)]
    #[arg(default_value_t)]
    pub database: Database,
...
```

jcrossley3:
FYI, I already referenced my preferred solution and the clap authors response [in a comment](https://github.com/trustification/trustify/blob/857a8aa2e9934eacdcf5d821751017f27bf6ee8b/common/src/config.rs#L26-L30) for posterity.

helio-frota:
from my part: I don't know :+1: 
praise: it looks like so a decision thing..and Jim is good deciding things :+1:

jcrossley3:
Closing in light of #238

PR #220, opened 2024-04-30
Allow fetching all packages from the SBOM

ctron:
As we don't seem to have an "describing packages", refactor the function to allow retrieving all packages.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5877,6 +5877,7 @@ dependencies = [
  "tokio",
  "trustify-common",
  "trustify-cvss",
+ "utoipa",
  "uuid",
 ]
 
--- a/common/src/db/limiter.rs
+++ b/common/src/db/limiter.rs
@@ -1,6 +1,6 @@
 use sea_orm::{
     ConnectionTrait, DbErr, EntityTrait, FromQueryResult, Paginator, PaginatorTrait, QuerySelect,
-    Select, SelectModel, Selector, SelectorTrait,
+    Select, SelectModel, SelectTwo, SelectTwoMany, SelectTwoModel, Selector, SelectorTrait,
 };
 use std::future::Future;
 use std::marker::PhantomData;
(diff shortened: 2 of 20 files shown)


ctron:
The idea is to have "all packages in the context of the SBOM". Unrelated to any relationships. The idea is browse through the SBOM content.

I am also working on the next function, which does: give me related packages (in the context of an sbom). Then again, those two functions seem related. So my next try is to make them one, at least on the service level. Using:

```rust
pub enum Which {
    /// All packages of the SBOM
    All,
    /// Top level packages of the SBOM (with no relationshhip
    TopLevel,
    /// Search relationships, given the left side and an optional relationship type
    Left(Purl, Option<Relationship>),
    /// Search relationships, given the right side and an optional relationship type
    Right(Purl, Option<Relationship>),
}
```

That could also get a `DescribesSbom` variant. then again, that would go into a completely different table. Let's see.

bobmcwhirter:
All packages within an SBOM should only be there because the SBOM either describes them or they are related to a described package, I think yeah?

Assuming we ingest all of that (we may be missing tables/etc) then grabbing left/right of relationships, and everything described, and unique-ing it shoud but... good?

ctron:
> All packages within an SBOM should only be there because the SBOM either describes them or they are related to a described package, I think yeah?

I don't think so. To my understanding "document describes" was an extension, that also wasn't well documented. It is not a requirement.

Searching for all packages in the quarkus SBOM, which don't have any relationship, I now get:

```json

{
    "items": [
        {
            "purl": "pkg://maven/com.redhat.quarkus.platform/quarkus-bom@3.2.11.Final-redhat-00001?repository_url=https://maven.repository.redhat.com/ga/&type=pom"
        }
    ],
    "total": 1
}
```

Which feels correct. Taking a look at the "document describes" part, there is nothing. Still, that SBOM has a valid package, named `pkg://maven/com.redhat.quarkus.platform/quarkus-bom@3.2.11.Final-redhat-00001?repository_url=https://maven.repository.redhat.com/ga/&type=pom`, which has a relationship to all the others. That sounds good.

ctron:
I updated the PR to allow searching for: root level packages (as described above), all packages (belonging to the sbom) … as well as searching for relationships inside the sbom.

Unifying the search packages and search relationships calls doesn't seem possible. Mainly because they actually return different data. In the context of relationships, it returns the relationship type too. And making the return format depending on the input search arguments feels wrong.

ctron:
Also see: https://github.com/spdx/spdx-spec/issues/395 … It's still not really clear to me: https://spdx.github.io/spdx-spec/v2.3/relationships-between-SPDX-elements/

bobmcwhirter (approved):

PR #222, opened 2024-05-01
Bump postgresql-embedded to 0.9.1

jcrossley3:
With this change, we default the database username to "postgres" since `pg::setup` no longer creates a superuser account for the username passed in its `Settings`.


common/src/config.rs
```diff
@@ -32,7 +32,7 @@ impl Default for Database {
     fn default() -> Self {
         const DEFAULT_PORT: u16 = 5432;
         Database {
-            username: env::var("DB_USER").unwrap_or("trustify".into()),
+            username: env::var("DB_USER").unwrap_or("postgres".into()),
```
bobmcwhirter, line 35: We changing credentials?

bobmcwhirter (approved):

common/src/config.rs
jcrossley3: IDU. We're changing the default username.

PR #229, opened 2024-05-03
chore: upgrade dependencies

ctron:
This updates dependencies to ones using http 1 (over http 0.2), reducing the number of duplicates, like reqwest.


helio-frota (approved):
praise: that is good I remember those 2 dependencies with update issues
http and biscuit :tada: 
https://gist.github.com/helio-frota/dab8edd2b1862bcbb6567e2cb1d29e1e#http

PR #232, opened 2024-05-03
Reduce noise from test-data paths

jcrossley3:


modules/ingestor/src/endpoints/advisory.rs
```diff
@@ -75,15 +72,9 @@ mod tests {
     async fn upload_default_csaf_format(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
         let db = ctx.db;
         let (storage, _temp) = FileSystemBackend::for_test().await?;
-
         let app = test::init_service(App::new().configure(|svc| configure(svc, db, storage))).await;
+        let payload = include_str!("../../../../etc/test-data/cve-2023-33201.json");
```
helio-frota, line 76: this macro is interesting!
https://doc.rust-lang.org/std/macro.include_str.html

helio-frota (approved):

PR #233, opened 2024-05-03
feat: detect advisory format and location

### LINKED ISSUE #144: Upload Advisory should only require the file, everything else should be determined by the server
## Current status:

https://github.com/trustification/trustify/blob/52b2cc9dfaa325ffcfdaee30e38371e444443b1f/modules/ingestor/src/endpoints/advisory.rs#L14C1-L26C23

```rust
#[utoipa::path(
    tag = "ingestor",
    request_body = Vec <u8>,
    params(
        ("format" = String, Query, description = "Format of the submitted advisory document (`csaf`, `osv`, ...)"),
        ("location" = String, Query, description = "Source the document came from"),
    ),
    responses(
        (status = 201, description = "Upload a file"),
        (status = 400, description = "The file could not be parsed as an advisory"),
    )
)]
#[post("/advisories")]
```

## Problem
- Format & location query parameters are required to upload a file

It means that the UI can not just take a file and upload it. It requires the user fill a form to specify the `format` of the file he is uploading, which is in my opinion prone to error and not user friendly. Imagine yourself uploading a file using CURL but you can not just take a file and upload it, but you have to determine which is the format of it. The problem is for humans, probably not for other apps

## Proposed solution
- The FILE should be the only required thing, everything else should be determined by the server
  - For the `location` query parameter: Technically the UI can be deployed in a different domain than the Backend, There can be multiple UI domains pointing to the same backend. `location` should be determined in the backend side as the relative path of the rest endpoint used e.g. `/advisories` 
  - for the `format` query parameter. There are JSON Schemas for each of the formats supported. Why we can just use the schemas to validate to which one the file belongs to? e.g. User uploads a file => the server compares the file to each of the supported json schemas => the format is determined by the server as the schema to which the file matches. That should be it. The JSON Schemas are
  

- CSAF Schema: https://github.com/oasis-tcs/csaf/blob/master/csaf_2.0/json_schema/csaf_json_schema.json
- OSV Schema https://github.com/ossf/osv-schema/blob/main/validation/schema.json
- CVE
Schema: https://github.com/CVEProject/cve-schema/blob/master/schema/v5.0/CVE_JSON_5.0_schema.json

jcrossley3:
Resolves #144

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5989,6 +5989,7 @@ dependencies = [
  "actix-web",
  "anyhow",
  "bytes",
+ "bytesize",
  "chrono",
  "cpe",
  "csaf",
--- a/modules/ingestor/Cargo.toml
+++ b/modules/ingestor/Cargo.toml
@@ -14,6 +14,7 @@ trustify-module-storage = { workspace = true }
 actix-web = { workspace = true }
 anyhow = { workspace = true }
 bytes = { workspace = true }
+bytesize = { workspace = true }
 chrono = { workspace = true, features = ["serde"] }
 cpe = { workspace = true }
 csaf = { workspace = true }
--- a/modules/ingestor/src/endpoints/advisory.rs
+++ b/modules/ingestor/src/endpoints/advisory.rs
@@ -1,23 +1,10 @@
 use crate::service::{advisory::Format, Error, IngestorService};
 use actix_web::{get, post, web, HttpResponse, Responder};
-use std::str::FromStr;
-
-#[derive(Clone, Debug, serde::Deserialize)]
-pub struct UploadAdvisoryQuery {
-    /// The source of the document.
-    ///
-    /// Only the base source, not the full document URL.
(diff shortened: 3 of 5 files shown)


carlosthe19916 (approved):
@jcrossley3 Look good to me!
If we could take this opportunity to move the "upload" endpoints under the "/api" base path, it would be great. But it can also be done in a separate PR.
Screenshot from 2024-05-06 10-20-28

jcrossley3:
> @jcrossley3 Look good to me! If we could take this opportunity to move the "upload" endpoints under the "/api" base path, it would be great. 

Agree. I tried, but then ran into #235

carlosthe19916:
> > @jcrossley3 Look good to me! If we could take this opportunity to move the "upload" endpoints under the "/api" base path, it would be great.
> 
> Agree. I tried, but then ran into #235

Let's move the endpoints to "/api/v1" in a separate PR then so we don't block this one.

Issue #158, opened 2024-04-15
Documents having their own document ID need to be grouped


ctron:
For documents which have their own document ID and which can change over time (like CSAF documents), there must be a way to get that latest document by its document ID. When searching, only the latest document must be searched and returned (unless we want an additional API to search through the history).

jcrossley3:
Still relevant?

ctron:
No this was actually fixed by https://github.com/trustification/trustify/pull/845 for CSAF docs. And for SBOMs, we need a better definition of what "latest" means.

Issue #157, opened 2024-04-15
Tables storing documents should have a single ID field based on the document digest


ctron:
Instead of having a database ID field and a digest field, there should be a single digest field. The digest should be the same as used for identifying the document in the storage/filesystem.

jcrossley3:
Please re-open and elaborate if this is still relevant

PR #234, opened 2024-05-03
Remove search module, porting query stuff to fetch

jcrossley3:


modules/ingestor/src/endpoints/advisory.rs
```diff
@@ -37,27 +37,6 @@ pub async fn upload_advisory(
     Ok(HttpResponse::Created().json(advisory_id))
 }
 
-#[utoipa::path(
```
ctron, line 40: I think it's ok to remove it from here, but it should end up somewhere else. I can't find that, did I overlook it?

ctron (changes requested):
It looks to me like the download (sbom, advisories) endpoint have simply been dropped. If that's the case, I am not ok with that.

modules/ingestor/src/endpoints/advisory.rs
jcrossley3: It's [already in the fetch module](https://github.com/trustification/trustify/blob/bc2c73fde8fd0b433c9facd784afb556846f5dee/modules/fetch/src/endpoints/advisory.rs#L37-L51), so not a part of this PR

modules/ingestor/src/endpoints/sbom.rs
```diff
@@ -30,24 +30,3 @@ pub async fn upload_sbom(
...
-#[utoipa::path(
-    tag = "ingestor",
-    responses(
-        (status = 200, description = "Download a an SBOM", body = Vec<u8>),
-        (status = 404, description = "The document could not be found"),
-    )
-)]
-#[get("/sboms/{id}")]
-/// Download an SBOM
-pub async fn download_sbom(
```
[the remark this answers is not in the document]
jcrossley3: I thought this was in the fetch module, too, but I don't see it now. :)

I think it makes sense for it to live there, I think, if we're ok with keeping all the GET's in fetch.

modules/ingestor/src/endpoints/advisory.rs
ctron: Yea, there's a version with the digest, which I am not sure we use much. I think it makes sense having both versions with the `id`.

jcrossley3: I'll leave these in for now, but we need to agree on what `xxx` is in this request: `/api/v1/advisory/{xxx}`. Is it the entity's `id` field or its `sha256` field? Or do we need an api for both?

Complicating matters further, these will also work:

- `/api/v1/advisory?q=id%3Dfoo`
- `/api/v1/advisory?q=sha256%3Dfoo`

ctron: I would go for the `id`.

bobmcwhirter: But we'd prefer to not leak IDs. And sha can he deterministic.

ctron (approved):

modules/ingestor/src/endpoints/advisory.rs
jcrossley3: Yeah, I thought we agreed IDs wouldn't be a part of the public API.

Potentially relevant: #157 #158

Issue #239, opened 2024-05-06
PM mode local time logs


helio-frota:
As PM-mode is a thing to run locally, probably it makes more sense to print the local time
https://github.com/trustification/trustify/issues/225#issuecomment-2093686896

**Note:** After some attempts to get this by adding [this](https://rust-lang-nursery.github.io/rust-cookbook/development_tools/debugging/config_log.html#include-timestamp-in-log-messages) to the [main](https://github.com/trustification/trustify/blob/main/trustd/src/main.rs#L64)  (probably not the best location to add this, btw)

I noticed:

We need to enable `features = ["clock"]` 
`chrono = { version = "0.4.35", default-features = false, features = ["clock"] }`

We can print the local time without issues **(1)**

`println!("{}", Local::now());`

And we get this error `Error initializing logging: SetLoggerError(())` from `tracing.rs` **(2)**

![2024-05-06_10-29](https://github.com/trustification/trustify/assets/6443576/a63a0e43-faef-4259-968b-21ad66bc0577)

jcrossley3:
Please don't add anything logger specific to `main`, which delegates all logger config to the `infrastructure` module. See the `init_tracing` calls in `db.rs`.

I suspect it's fine to configure local time there, but I defer to @ctron

helio-frota:
ok thanks for the extra information @jcrossley3 :+1:

ctron:
As @jcrossley3 said, the initialization of the logger/tracing happens in the infrastructure module. There are two variants, one when "tracing" is enabled, and all information is sent to an opentelemetry endpoint (in which case there's no logging on the console).

Or when "tracing" it not enabled, then everything goes to the console using this setup:

https://github.com/trustification/trustify/blob/b04474e306597e85df5bca9b8c964b75b037e285/common/infrastructure/src/tracing.rs#L147-L165

"Tracing" might not be the right term anymore, was "tracing" was "Tracing via Jaeger", which now actually is "Tracing via OpenTelemetry". "Tracing" itself is always present, it just might get routed to the console instead.

The output to the console happens via `tracing_subscriber::fmt::layer()`, which has some options. But I am not sure it has a timestamp format option. One thing to find out.

helio-frota:
@ctron thank you for detailed information :+1:

PR #243, opened 2024-05-07
chore: using local time for PM-mode

helio-frota:
Related to #239

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5761,6 +5761,7 @@ version = "0.3.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ad0f048c97dbd9faa9b7df56362b8ebcaa52adb06b498c050d2f4e32f90a7a8b"
 dependencies = [
+ "chrono",
  "matchers",
  "nu-ansi-term",
  "once_cell",
--- a/common/infrastructure/Cargo.toml
+++ b/common/infrastructure/Cargo.toml
@@ -29,7 +29,7 @@ serde = { workspace = true, features = ["derive", "rc"] }
 serde_json = { workspace = true }
 tokio = { workspace = true, features = ["time", "signal", "macros"] }
 tracing-opentelemetry = { workspace = true }
-tracing-subscriber = { workspace = true, features = ["env-filter", "ansi", "tracing-log"] }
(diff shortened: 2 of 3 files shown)


helio-frota:
```console
2024-05-07T08:24:27.486380625-03:00  INFO actix_server::accept: accept thread stopped
2024-05-07T08:24:27.486452376-03:00  INFO actix_server::worker: shutting down idle worker
2024-05-07T08:24:27.486502522-03:00  INFO actix_server::worker: shutting down idle worker
➜  trustify git:(localtime) date
Tue May  7 08:24:29 AM -03 2024
```

ctron:
Hm, I am unsure about this.

For PM mode, that might make sense.

For actual deployments, using UTC in any case makes more sense. Because otherwise it will become pretty hard to correlate events.

I am ok adding this. I am also ok having this as a default when launching from the command line. But I think there should be some "UTC log timestamps" switch.

helio-frota:
yeah makes no sense for deployments.

thanks for the extras :+1: 

I got what you mean.. `init_no_tracing` is not only for PM-mode... which may end up deploying with disable traces and not using UTC. that is bad.

helio-frota:
>  But I think there should be some "UTC log timestamps" switch.

yes, or even better way of think about this --> a local (PM-mode) and server (deployment) something-that-we-can-switch

helio-frota:
**question:** currently it is only possible to enable traces (programmatically) via compiled code ?

if yes I'm ok :+1:

helio-frota:
**note:** keeping your previous comment in mind I'm not OK with this PR.

```
There are two variants, one when "tracing" is enabled, and all information is sent to an opentelemetry endpoint (in which case there's no logging on the console).

Or when "tracing" it not enabled, then everything goes to the console....
```

Please correct me if I'm wrong, we have the following:

a) deployment :heavy_check_mark: 
b) deployment with traces :heavy_check_mark: 
c) pm-mode :heavy_check_mark: 
d) pm-mode with traces - is a thing that makes no sense other than for DEV purposes ( to check if things are working fine with traces etc ) :no_entry_sign:


So with this PR we fix **c)** and break **a)**

ctron:
Yes, I think that sums it up. "PM mode with tracing" is nothing we should focus on.

I guess having a simple `--force-utc-logs` might work.

helio-frota:
TBQH I'm more in favor of closing this PR and the related issue, than adding functionality that probably no one will use.

PR #245, opened 2024-05-07
Start on defining words.

bobmcwhirter:


docs/glossary.md
```diff
@@ -0,0 +1,51 @@
...
+This may be simply in reference to the vulnerability _as it exists in source-code form_.
+
+Other, more-involved stakeholders (product vendors, upstream project owners) may issue *additional* advisories.
+These opinions may be in reference to _concrete_ shipped products, contextualized to how the vulnerable code is _actually used_.
+
+### Package
+
+A package is an atomic artifact or component.
+Packages may be addressed using pURLs.
+A package may be described by an SBOM describing how it is created.
```
jcrossley3, line 32: An example here would be good, because I'm still fuzzy on the distinction of Package vs Product as both can be comprised of a collection of *[insert correct plural term here]* described by an SBOM.

docs/glossary.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Product
+
+A product is a _named collection of packages_ for a concrete shippable thing.
+
+Products may be addressed using CPEs or some other future identification method.
```
jcrossley3, line 38: Define CPE?

jcrossley3 (approved):

docs/glossary.md
```diff
@@ -0,0 +1,86 @@
...
+For instance, "All versions of RHEL 8.2.013, regardless of platform", or if more fully-specified, could imply "All versions of RHEL 8.x on AArch64".
+
+> **NOTE** 
+> CPEs are somewhat contentious, and used enough for us to not ignore, but not used enough to be a pivotal definition of "product" for any users of Trustify.
+
+### Artifact.
+
+> **NOTE**
+> CPEs are somewhat contentious, and used enough for us to not ignore, but not used enough to be a pivotal definition of "product" for any users of Trustify.
+> Not currently modelled
```
alanfx, line 73: Don't these notes look very similar? I get the feeling that CPEs are somewhat contentious, but are they also repetitive? Otherwise thanks for defining words, @bobmcwhirter!

ctron:
I really like that definition. I think it make a lot things much clearer. And now I also do understand what you had in mind with the CVE record.

Issue #237, opened 2024-05-06
GET /api/v1/vulnerability does not work


carlosthe19916:
`curl /api/v1/vulnerability` does not work. It gives back the following error:

```json
{"error":"Database","message":"Query Error: Relation to is not HasOne"}
```

helio-frota:
note: I'm not sure if this is a bug or an ongoing implementation... 
https://github.com/trustification/trustify/blob/main/entity/src/advisory_vulnerability.rs#L19-L25
Either way I uncommented these lines and nothing happened.

This is also failing for "get-by-id"

```console
2024-05-06T11:36:50.419952Z DEBUG sea_orm::driver::sqlx_postgres: SELECT "vulnerability"."id" AS "A_id", "vulnerability"."identifier" AS "A_identifier", "vulnerability"."title" AS "A_title", "advisory"."id" AS "B_id", "advisory"."identifier" AS "B_identifier", "advisory"."location" AS "B_location", "advisory"."sha256" AS "B_sha256", "advisory"."published" AS "B_published", "advisory"."modified" AS "B_modified", "advisory"."withdrawn" AS "B_withdrawn", "advisory"."title" AS "B_title" FROM "vulnerability" LEFT JOIN "advisory_vulnerability" ON "vulnerability"."id" = "advisory_vulnerability"."vulnerability_id" LEFT JOIN "advisory_vulnerability" ON "vulnerability"."id" = "advisory_vulnerability"."vulnerability_id" WHERE "vulnerability"."identifier" = '2' ORDER BY "vulnerability"."id" ASC
2024-05-06T11:36:50.426552Z DEBUG actix_web::middleware::logger: Error in response: Database(Query(SqlxError(Database(PgDatabaseError { severity: Error, code: "42712", message: "table name \"advisory_vulnerability\" specified more than once", detail: None, hint: None, position: None, where: None, schema: None, table: None, column: None, data_type: None, constraint: None, file: Some("parse_relation.c"), line: Some(463), routine: Some("checkNameSpaceConflicts") }))))
2024-05-06T11:36:50.427019Z  INFO actix_web::middleware::logger: ::1 "GET /api/v1/vulnerability/2 HTTP/1.1" 500 138 "-" "curl/8.2.1" 0.009620
```

bobmcwhirter:
Well, crap. Looking.

PR #248, opened 2024-05-07
Fix for #237

bobmcwhirter:
Adjust relationships between vulnerability and advisory join table. Add tests for each endpoint.


bobmcwhirter:
Fix for #237

jcrossley3 (commented):
Personally, I prefer delete to comment.

PR #250, opened 2024-05-08
Attempt to use trustify-ui as a crate dependency.

bobmcwhirter:
Pending merging (and un-forking) of https://github.com/trustification/trustify-ui/pull/13

### DIFF
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -10,7 +10,6 @@ use actix_web::{
     web::Json,
     HttpRequest, HttpResponse, Responder,
 };
-use actix_web_static_files::ResourceFiles;
 use anyhow::Context;
 use futures::FutureExt;
 use std::fmt::Display;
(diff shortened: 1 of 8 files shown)


server/src/lib.rs
```diff
@@ -144,8 +146,20 @@ impl InitData {
...
+
         let storage = DispatchBackend::Filesystem(FileSystemBackend::new(storage).await?);
 
+        // TODO: where/how should we configure these details?
+        let ui = UI {
+            version: env!("CARGO_PKG_VERSION").to_string(),
+            auth_required: String::from("false"),
+            oidc_server_url: String::from("http://localhost:8180/realms/trustify"),
+            oidc_client_id: String::from("trustify-ui"),
+            oidc_scope: String::from("email"),
```
carlosthe19916, line 158: I believe it should be populated with the values of `AuthConfigArguments` https://github.com/trustification/trustify/blob/main/common/auth/src/auth.rs#L15

But we can do it in another PR

carlosthe19916 (approved):

jcrossley3 (approved):
Works really well!!

jcrossley3:
Maybe update the title? No longer forked, correct?

server/src/lib.rs
bobmcwhirter: Okay, I'll work with @ctron in a subsequent PR.

bobmcwhirter:
> Maybe update the title? No longer forked, correct?

Correct!

PR #256, opened 2024-05-09
Thicken up the Vulnerability DTOs.

bobmcwhirter:
Hydrate details into a Vulnerability even if ingested after first light reference. Add some error-reporting if format auto-detection doesn't. Ensure `vulnerabilities` field is included, even if an empty array.

### DIFF
--- a/modules/ingestor/src/service/cve/cve_record/v5.rs
+++ b/modules/ingestor/src/service/cve/cve_record/v5.rs
@@ -1,5 +1,5 @@
-use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
+use time::OffsetDateTime;
 
 #[derive(Serialize, Deserialize, Debug)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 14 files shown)


modules/ingestor/src/service/format.rs
```diff
@@ -38,17 +38,37 @@ impl<'g> Format {
...
+            Err(e) => {
+                potential_errors.push(format!("if cve: {}", e));
+            }
+        }
+
+        Err(Error::UnsupportedFormat(format!(
+            "unknown :: {:?}",
+            potential_errors
+        )))
     }
```
jcrossley3, line 69: There's probably performance opportunities in here, e.g. use `RawValue` for everything in the doc except the field[s] required to recognize the format. That way, we'd save a crapton of memory allocation and we'd know exactly what format the errors applied to. Also, we should probably order them by most common expected.

We can do that later, obviously.

modules/ingestor/src/service/cve/cve_record/v5.rs
```diff
@@ -24,6 +24,34 @@ impl CveMetadata {
...
+            CveMetadata::Rejected(inner) => inner.date_updated,
+        }
+    }
+
+    pub fn date_rejected(&self) -> Option<OffsetDateTime> {
+        match self {
+            CveMetadata::Published(_) => None,
+            CveMetadata::Rejected(inner) => inner.date_rejected,
+        }
+    }
```
jcrossley3, line 54: I still don't understand why the `CveMetadata` enum exists.

jcrossley3 (approved):

modules/ingestor/src/service/format.rs
bobmcwhirter: pb https://github.com/trustification/trustify/issues/257

modules/ingestor/src/service/cve/cve_record/v5.rs
jcrossley3: Yeah, it would be clearer (to me) if the common fields were factored out of each.

PR #258, opened 2024-05-09
Move raw sbom/advisory endpoints beneath storage module

### LINKED ISSUE #255: /api/v1/advisory/{id}/download needs to take a sha256 instead of an id
The user can find the sha256 from our api, but not the id. and storage needs the sha anyway.

jcrossley3:
Fixes #255

We fetch them by their digest only, since database id's aren't exposed to users. It's worth noting that the implementations for returning either sbom or advisory are virtually identical, since storage only cares about a digest, no matter what kind of doc it refers to.

So we might just consider replacing these with a single `/api/v1/download/{sha256}` api. That way, we'd avoid having to do this...

We removed the web::scope's for sbom's and advisories in the fetch module because having them seemed to prevent the storage modules endpoints from being configured. I'm not exactly sure why.


ctron:
I am still not convinced we have a good understanding on "ids".

For storage I think it makes sense to store "by digest".

For the database, I am not so convinced:

* How do we handle the case where the user want the "most recent document by it's (document) id"? (e.g. the csaf id, or the spdx namespace).
* How do we handle the case where two different people upload the same SBOM, but both individuals can't see each others data due to some ACL?

bobmcwhirter (approved):

PR #259, opened 2024-05-10
test: migrator tests

helio-frota:
Related to #127 

Please let me know if makes sense


jcrossley3 (commented):
What exactly are you testing?

helio-frota:
@jcrossley3 `Test for forward and backwards and foreward migrations to ensure full drop-stuff.`

https://github.com/trustification/trustify/issues/127

bobmcwhirter (approved):

helio-frota:
@jcrossley3 PR updated thanks !

migration/tests/tests.rs
```diff
@@ -0,0 +1,33 @@
...
+async fn test_migrations(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
+    let db = ctx.db;
+
+    let conn_or_tx = db.connection(&());
+    let conn = match conn_or_tx {
+        ConnectionOrTransaction::Connection(conn) => conn,
+        _ => {
+            panic!("Error while getting db connection");
+        }
+    };
```
jcrossley3, line 19: You don't really need this...

migration/tests/tests.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    let db_ref = &db;
+    let migrations = Migrator::get_applied_migrations(db_ref).await?;
+    // 'Migrator.up' was called in bootstrap function when using TrustifyContext.
+    // At this point we already have migrations.
+    assert!(migrations.len() > 1);
+
+    Migrator::refresh(conn).await?;
```
jcrossley3, line 27: ...you can just call this:

```suggestion
    db.refresh().await?;
```

jcrossley3 (commented):
Much better, thanks!

migration/tests/tests.rs
helio-frota: I guess I don't know other way to do.

helio-frota: oh , yeah we refactored!
you're awesome!

jcrossley3 (changes requested):

migration/tests/tests.rs
```diff
@@ -0,0 +1,25 @@
...
+use test_log::test;
+use trustify_common::db::test::TrustifyContext;
+
+#[test_context(TrustifyContext, skip_teardown)]
+#[test(tokio::test)]
+async fn test_migrations(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
+    let db = ctx.db;
+
+    let db_ref = &db;
+    let migrations = Migrator::get_applied_migrations(db_ref).await?;
```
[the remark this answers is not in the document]
helio-frota: perfect!

migration/tests/tests.rs
```diff
@@ -0,0 +1,24 @@
...
+#[test(tokio::test)]
+async fn test_migrations(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
+    let db = ctx.db;
+
+    let migrations = Migrator::get_applied_migrations(&db).await?;
+    // 'Migrator.up' was called in bootstrap function when using TrustifyContext.
+    // At this point we already have migrations.
+    assert!(migrations.len() > 1);
+
+    db.refresh().await?;
```
jcrossley3, line 18: I'm assuming you've removed a `down` function in one of our migrations locally and run this test just to make sure it fails? I'm assuming line 18 will panic, yes?

helio-frota: ok about the expected behavior I would say -- yes.

```
       START             trustify-migration::tests test_migrations

running 1 test
Error: Migration Error: We Don't Do That Here
test test_migrations ... FAILED

failures:

failures:
    test_migrations

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.58s

        FAIL [   1.585s] trustify-migration::tests test_migrations
   Canceling due to test failure
------------
     Summary [   1.585s] 1 test run: 0 passed, 1 failed, 74 skipped
        FAIL [   1.585s] trustify-migration::tests test_migrations
error: test run failed
```

But the missing `down` AFAIK is the problem and this test will break the CI build and etc.. so we can fix the migration.

helio-frota: that is a bit weirdo test because probably we can call this as a cross-cutting-concern test instead of unit/integration/business-related test...

jcrossley3: > But the missing `down` AFAIK is the problem and this test will break the CI build and etc.. so we can fix the migration.

Yes, that's the exact behavior we want. CI should fail if we forgot a `down` migration.

jcrossley3 (approved):

PR #264, opened 2024-05-10
Add an `organization` entity with name, optional CPE key and maybe a …

bobmcwhirter:
…website.

Perform best effort analysis of advisories to ingest and link to appropriate organization. Provide optional `?issuer={name}` to advisory upload endpoint for cases where it cannot be determined from file content.

CVE loader ignores any issuer, as all CVEs are indeed from CVE Project/MITRE. CSAF loader ignores any issuer, as CSAF documents include issuer information within. OSV issuer can take some guesses based on content, but not consistently.


migration/src/m0000022_create_organization.rs
```diff
@@ -0,0 +1,51 @@
...
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        // Replace the sample below with your own migration scripts
+        manager
+            .create_table(
+                Table::create()
+                    .table(Organization::Table)
+                    .if_not_exists()
```
jcrossley3, line 14: In light of #127, I'm inclined to make it a practice to avoid these guards, because it allows you to omit the `down` fn. It's a minor nit, but I like simple, correct migrations, with as little code as possible. CI will whinge if you forget the `down` now.

migration/src/m0000022_create_organization.rs
```diff
@@ -0,0 +1,51 @@
...
+
+        Ok(())
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .drop_table(
+                Table::drop()
+                    .table(Organization::Table)
+                    .if_exists()
```
jcrossley3, line 37: ibid

modules/fetch/src/endpoints/advisory.rs
```diff
@@ -93,6 +93,7 @@ mod test {
         service
             .ingest(
                 "unit-test",
+                Some("Capt Pickles Industrial Conglomerate".to_string()),
```
jcrossley3, line 96: :taco:

modules/ingestor/src/endpoints/advisory.rs
```diff
@@ -14,11 +24,12 @@ use tokio_util::io::ReaderStream;
 /// Upload a new advisory
 pub async fn upload_advisory(
     service: web::Data<IngestorService>,
+    web::Query(UploadParams { issuer }): web::Query<UploadParams>,
     bytes: web::Bytes,
 ) -> Result<impl Responder, Error> {
     let fmt = Format::from_bytes(&bytes)?;
     let payload = ReaderStream::new(&*bytes);
-    let advisory_id = service.ingest("rest-api", fmt, payload).await?;
+    let advisory_id = service.ingest("rest-api", issuer, fmt, payload).await?;
```
jcrossley3, line 32: So, I was uncomfortable with this first param, "rest-api", which used to be the `location` until we decided that wasn't really useful in #233. So it seems like maybe you're bringing the location concept back in the form of an issuer? If that's the case, let's get rid of the hard-coded "rest-api" arg. Unless you think it's useful. I don't.

modules/ingestor/src/service/advisory/osv/loader.rs
```diff
@@ -22,12 +24,15 @@ impl<'g> OsvLoader<'g> {
     pub async fn load<L: Into<String>, R: Read>(
         &self,
         location: L,
+        issuser: Option<String>,
```
jcrossley3, line 27: You're making us look bad compared to AI.

```suggestion
        issuer: Option<String>,
```

jcrossley3 (changes requested):

modules/ingestor/src/endpoints/advisory.rs
bobmcwhirter: "remove location" I think should be a separate PR. Here, the issuer is *optional* and only used for OSV where that information is not necessarily derivable from the contents.

jcrossley3 (approved):
Excellent formatting!

PR #266, opened 2024-05-13
set up Multi OS ci workflow 

carlosthe19916:


carlosthe19916:
This PR served just to realize that the current CI does not work on MacOS and Windows. But as it is not a priority, I am closing this PR and similar work can be done in the future if we find it apropiate

PR #270, opened 2024-05-13
chore: drop as it seems to be deprecated and causes warnings

ctron:


ctron:
When using `podman compose` (with the `docker compose` backend), this gives a warning.

jcrossley3 (approved):

PR #271, opened 2024-05-13
Rework SBOM package structure

ctron:
Rework the SBOM package structure to align more with SPDX.

This includes:

* [x] Having "packages" (independent of PURLs) as part of the SBOM
* [x] Have references (0+) to PURLs and CPEs for each package
* [x] Drop the "describes" tables and use regular "describes" or "described by" relationships
* [ ] Rename a few "packages" to "purls", because that's what they are

This PR does not yet include:

* Allowing out-of-SBOM namespace relationships


ctron:
As tests are passing, I am removing the `draft` status on this one.

There are some cleanups missing. But I also think it would be good if the PR gets a review.

integration-tests/src/sbom/test/perf.rs
```diff
@@ -1,17 +1,51 @@
...
 use trustify_common::db::{test::TrustifyContext, Transactional};
+use trustify_common::model::Paginated;
+use trustify_module_fetch::model::sbom::SbomPackage;
+use trustify_module_fetch::service::FetchService;
+use trustify_module_ingestor::graph::{
+    sbom::spdx::{parse_spdx, Information},
+    Graph,
+};
+
+/// remove all relationships having broken references
```
bobmcwhirter, line 17: This will end up going into some report we capture yeah?

bobmcwhirter (approved):
Mostly just trusting you. It'll all come out in the wash.

PR #272, opened 2024-05-13
Put UI behind a default feature.

bobmcwhirter:
To build w/o the ui:

    cargo build -p trustify-trustd --no-default-features


ctron (approved):
I can very it fails with defaults and works with `--no-default-features`.

PR #275, opened 2024-05-13
fix: ensure order using primary keys, not hardcoded id fields

### LINKED ISSUE #267: "filtering" requires an 'id' field
I ran into the issue of:

```
Error: query syntax error: Entity missing Id field
```

* It would be helpful if there would be the name of the entity that's missing that field.
* Why is that required?

jcrossley3:
Fixes #267

All columns in composite keys will be added to the statement if not already present.

### DIFF
--- a/modules/fetch/src/query.rs
+++ b/modules/fetch/src/query.rs
@@ -2,8 +2,8 @@ use human_date_parser::{from_human_time, ParseResult};
 use regex::Regex;
 use sea_orm::sea_query::{extension::postgres::PgExpr, ConditionExpression, IntoCondition};
 use sea_orm::{
-    ColumnTrait, ColumnType, Condition, EntityTrait, Iden, Iterable, Order, QueryFilter,
-    QueryOrder, QueryTrait, Select, Value,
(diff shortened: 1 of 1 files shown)


bobmcwhirter (approved):
I await `else_matters`

bobmcwhirter:
Is it easy enough for me to provide a default sort per entity in our Service, if not user-provided?

jcrossley3:
> I await `else_matters`

Took me a second. Thanks for noticing. It was funnier when I was comparing `actual` to a string. Maybe I'll go back to that for the entertainment value.

jcrossley3:
> Is it easy enough for me to provide a default sort per entity in our Service, if not user-provided?

I don't know. I would expect the UI to drive the ordering.

jcrossley3:
> Is it easy enough for me to provide a default sort per entity in our Service, if not user-provided?

I might've misunderstood your question. You can certainly add an `.order_by` to your statement prior to calling `.filtering`. There are tests ensuring that `.filtering` doesn't muck with existing ordering in the statement and doesn't add redundant ordering. I'm always scared by potential performance impacts of being cavalier with ORDER BY clauses.

PR #289, opened 2024-05-14
First stage of refactoring.

bobmcwhirter:
Top-level crate (not module) peer of `entity` holding DTO `model`
  - head
  - summary
  - details

Only moved vuln/advisory, leaving SBOM in place to avoid causing @ctron rebase issues.


jcrossley3 (approved):

bobmcwhirter:
@jcrossley3 one note: I moved SearchOptions stuff to common/db, but then had to comment-out the tests since they use entity which uses common which uses entity which....

Can you help me think of a better place for it? Maybe within `entity`? Or a different test strategy?

Else, `fetch` still exists, with @ctron's work on SBOMs, but if this merges, I'll sort through that also into `modules/sbom`.

common/src/db/query.rs
```diff
@@ -342,6 +342,7 @@ fn maintain_order<T: EntityTrait>(stmt: Select<T>) -> Select<T> {
 // Tests
 /////////////////////////////////////////////////////////////////////////
 
+/*
```
jcrossley3, line 345: This is unacceptable, sir.

jcrossley3 (changes requested):
If we think we have to delete tests, we're doing it wrong. If there's no possible way to avoid your circular deps, then let me change the tests to use their own `Entity`.

But I don't love having to do that.

bobmcwhirter:
If it works with entity lemme move it to entity.

jcrossley3:
> If it works with entity lemme move it to entity.

#292 should allow you to move it anywhere

jcrossley3 (approved):

Issue #291, opened 2024-05-14
Default filtering adds columns to select, breaking "group by"


ctron:
Having a select with a custom selected set of columns (`.select_only()`), applying the user filter, seems to add more columns to the select, breaking "group by" expression, as those columns get selected, but not aggregated.

jcrossley3:
Can you give me an example? This works:
```rust
    #[test(tokio::test)]
    async fn custom_select() -> Result<(), anyhow::Error> {
        let query = advisory::Entity::find()
            .select_only()
            .column(advisory::Column::Id)
            .filtering(SearchOptions {
                q: "location=foo".into(),
                sort: "".into(),
            })?
            .build(sea_orm::DatabaseBackend::Postgres)
            .to_string();

        assert_eq!(
            r#"SELECT "advisory"."id" FROM "advisory" WHERE "advisory"."location" = 'foo' ORDER BY "advisory"."id" DESC"#,
            query
        );

        Ok(())
    }
```

ctron:
This won't:

```rust
#[test(tokio::test)]
async fn custom_select() -> Result<(), anyhow::Error> {
    let query = advisory::Entity::find()
        .select_only()
        .column(advisory::Column::Location)
        .group_by(advisory::Column::Location)
        .filtering(Default::default())?
        .build(sea_orm::DatabaseBackend::Postgres)
        .to_string();

    assert_eq!(
        r#"SELECT "advisory"."location" FROM "advisory" GROUP BY "advisory"."location" ORDER BY "advisory"."id" DESC"#,
        query
    );

    Ok(())
}
```

True, the test will pass. But the database will report the following error:

```
ERROR: column "advisory.id" must appear in the GROUP BY clause or be used in an aggregate function
```

jcrossley3:
I see. So I'm thinking if there's a GROUP BY clause, then I should leave it up to the caller to get the ORDER BY clause correct, and I should leave it alone. Make sense?

jcrossley3:
Or is that I shouldn't add the primary key ORDER BY unless it's SELECT'ed?

ctron:
That seems to get complicated. I guess the latter (only add primary keys which are selected) makes sense.

Still, I guess all of this will break when the user makes use of `sha256`, which is a valid column, but not selected.

jcrossley3:
[...]

PR #295, opened 2024-05-15
don't add primary key ORDER BY clause when GROUP BY is present

jcrossley3:
Fixes: #291

### DIFF
--- a/common/src/db/query.rs
+++ b/common/src/db/query.rs
@@ -320,13 +320,16 @@ fn envalue(s: &str, ct: &ColumnType) -> Result<Value, Error> {
 }
 
 fn maintain_order<T: EntityTrait>(stmt: Select<T>) -> Select<T> {
-    let binding = T::default();
-    let table = binding.table_name();
     let s = stmt.build(sea_orm::DatabaseBackend::Postgres).to_string();
+    if s.contains(" GROUP BY ") {
+        return stmt; // you're on your own, grouper!
+    }
     let orderby = match s.rsplit_once(" ORDER BY ") {
         Some((_, v)) => v,
         None => "",
     };
+    let binding = T::default();
+    let table = binding.table_name();
     T::PrimaryKey::iter().fold(stmt, |stmt, pk| {
         let col = pk.into_column();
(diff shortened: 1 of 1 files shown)


jcrossley3:
> So if I would filter for "GROUP BY", the sorting would be gone?

Yes. Without having access to the statement's internals, I'm not sure how to distinguish the text in the filter from an actual GROUP BY clause.

So maybe we don't need this PR. Because all the Entity's columns need to be included in the GROUP BY clause anyway.

ctron:
> So maybe we don't need this PR. Because all the Entity's columns need to be included in the GROUP BY clause anyway.

Maybe, taking the following example: https://github.com/trustification/trustify/blob/0d0d5a414f058570d841ed50b5544d2378db2290/modules/fetch/src/service/sbom.rs#L131-L145

The `sbom_id` is part of the primary key, but doesn't need to be returned, so it's not selected.

Technically, it could be added. But that would mean that whenever someone wants to use filtering, the statements have the crafted in way that they select all PKs and group them. Because: just in case.

jcrossley3:
The only reason the primary key ordering is added is to maintain consistent paginated results. Do we really need that? Is there a better way? I'm open to ideas.

ctron:
> The only reason the primary key ordering is added is to maintain consistent paginated results. Do we really need that? Is there a better way? I'm open to ideas.

that's actually an aspect of the "limiter". So I would say it shouldn't be a concern for search.

Search might override those settings.

jcrossley3:
How does the limiter ensure the ordering? Isn't it possible the same item might appear in multiple batches?

ctron:
> How does the limiter ensure the ordering? Isn't it possible the same item might appear in multiple batches?

It doesn't. It requires you have a stable order. That's why in most (if not all) there is an explicit order statement. Because only the query know what might make sense as a default order.

jcrossley3:
Ok, cool. I'm happy to assume the callers of `.filtering` will ensure the stable order. That'll simplify a lot!

jcrossley3:
I added the fix to #299 in order to avoid rebase conflicts do to the new names.

PR #296, opened 2024-05-15
Ensure dates are included in fully-ingested vulnerabilities.

bobmcwhirter:
Ensure dates are serialized in a non-array format as expected by the UI. Tests for the above.
Also, include the assertions we bothered to calculate.

### DIFF
--- a/model/src/advisory/details/advisory_vulnerability.rs
+++ b/model/src/advisory/details/advisory_vulnerability.rs
@@ -260,9 +260,7 @@ impl AdvisoryVulnerabilitySummary {
                     summaries.push(AdvisoryVulnerabilitySummary {
                         head,
                         cvss3_scores,
-                        assertions: AdvisoryVulnerabilityAssertions {
-                            assertions
-                        }
+                        assertions: AdvisoryVulnerabilityAssertions { assertions },
                     });
                 }
             }
--- a/model/src/advisory/details/advisory_vulnerability.rs
+++ b/model/src/advisory/details/advisory_vulnerability.rs
@@ -260,7 +260,9 @@ impl AdvisoryVulnerabilitySummary {
                     summaries.push(AdvisoryVulnerabilitySummary {
                         head,
                         cvss3_scores,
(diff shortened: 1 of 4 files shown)


bobmcwhirter:
nb: ingesting an advisory only may result in not having dates present on vulnerability. To get vulnerability information, you *must* ingest a NIST NVD CVERecord advisory.

carlosthe19916 (changes requested):
- I can see `GET /api/v1/advisory/{id}` contains now "title", "modified" and "published".
- But `GET /api/v1/vulnerability` does does not contain "title", "modified" and "published" yet. Could we also include the new fields on `GET /api/v1/advisory`? I assume we are using the same DTOs but for some reason those fields are not being populated

carlosthe19916:
wait, my apologies. I misread the response from the backend. If I call `GET /api/v1/vulnerability/CVE-2022-2668` then I get:

```json
{
    "identifier": "CVE-2022-2668",
    "advisories": [
        {
            "identifier": "RHSA-2022:6787",
            "sha256": "2e58c99f88c47c9d4ffee0f6707800c99ae95fb89a56bb88cf9d074798a22c56",
            "issuer": "Red Hat Product Security",
            "published": "2022-10-04T15:53:11Z",
            "modified": "2024-05-02T15:55:56Z",
            "title": "Red Hat Security Advisory: Red Hat Single Sign-On 7.5.3 security update",
            "cvss3_scores": [],
            "assertions": {}
        }
    ]
}
```

I can see the format of dates are correct, I remember the dates were correct even after this PR. 
I was assuming this PR would add  "title", "modified" and "published" to the Vulnerability body buy I was wrong, sorry

carlosthe19916 (approved):

bobmcwhirter:
Ifff a CVERecord from NVD for a given vuln has been ingested, then the dates/title *will* appear. We only treat advisories from NVD as authoritative for CVE vulnerability details. Confusing, yes, but I think it's working per design?

PR #297, opened 2024-05-15
feat: provide some default/sample data for PM mode

ctron:
Also, add a description field.

@carlosthe19916 the description field might be useful for the UI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6512,7 +6512,9 @@ dependencies = [
  "trustify-auth",
  "trustify-common",
  "trustify-infrastructure",
+ "trustify-module-importer",
  "trustify-server",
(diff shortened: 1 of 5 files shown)


bobmcwhirter (approved):
I like this.  I also wonder if it should be... just always default-y?

e.g, in a migration always set up importers we can predict, maybe with disabled=true on some/all/most?

ctron:
The downside of this would be that we would need to handcraft those inserts. This way we can use the service. But yes, that would be possible.

bobmcwhirter:
Yah, we need to avoid using the SeaORM entities from the migrations, for sure.

Perhaps we codify `bootstrap` a bit more for "first run" setup?

PR #299, opened 2024-05-15
Some convenience fn's for query, some renaming, and docs

### LINKED ISSUE #291: Default filtering adds columns to select, breaking "group by"

jcrossley3:
Also fixes #291

### DIFF
--- a/common/src/db/query.rs
+++ b/common/src/db/query.rs
@@ -17,38 +17,72 @@ use utoipa::IntoParams;
 // Public interface
 /////////////////////////////////////////////////////////////////////////
 
-#[derive(
-    Clone,
-    Default,
-    Debug,
-    serde::Deserialize,
-    serde::Serialize,
-    utoipa::ToSchema,
-    utoipa::IntoParams,
-)]
-#[serde(rename_all = "camelCase")]
-pub struct SearchOptions {
-    /// The search filter
-    #[serde(default)]
-    pub q: String,
-    #[serde(default)]
-    /// Sort options
-    pub sort: String,
+/// Convenience function for creating a search query
+///
+/// ```
+/// use trustify_common::db::query::q;
+///
+/// let query = q("foo&bar>100").sort("bar:desc");
+///
+/// ```
+pub fn q(s: &str) -> Query {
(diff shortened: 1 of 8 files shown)


common/src/db/query.rs
```diff
@@ -17,38 +17,72 @@ use utoipa::IntoParams;
...
+    /// [Column](sea_orm::EntityTrait::Column) variant.
+    ///
+    /// Both `{query}` and `{value}` may contain `|`-delimited
+    /// alternate values that will result in an OR clause. Any `|` or
+    /// `&` within a query/value should be escaped with a backslash,
+    /// e.g. `\|` or `\&`.
+    ///
+    /// `{op}` should be one of `=`, `!=`, `~`, `!~, `>=`, `>`, `<=`,
+    /// or `<`.
+    pub fn q(s: &str) -> Self {
```
bobmcwhirter, line 53: so if I call .sort() first, and then q(), my sort gets trampled by ""?

bobmcwhirter (approved):
My only concern is the ordering and mutation of the query builder thing.

common/src/db/query.rs
bobmcwhirter: "Don't do that" is an acceptable answer, maybe.

jcrossley3: `.sort()` takes a `self` but `.q()` doesn't. There's no trampling because there's a field for each. I just figured `q` was the more likely used field. You could always just `Query {sort: "foo".into(), q: "".into() }`. It's ugly, but it works.

I'm open to better ideas if you have 'em.

jcrossley3: Alternatively, if you only needed sorting, `q("").sort("foo")`

jcrossley3: We could also introduce a top-level `sort`, in the same way `query::q()` calls `query::Query::q()`. So if you find yourself wanting to only sort a lot, we could create `query::sort()`

PR #300, opened 2024-05-15
Add support for Organization-related endpoints.

bobmcwhirter:
- Includes advisories issued by an organization.
- Simplify some of the from_entity ish to avoid unnecessary Result<Option<T>>
- Clean up API paths on openapi, etc, a bit.

### DIFF
--- a/modules/advisory/src/endpoints/mod.rs
+++ b/modules/advisory/src/endpoints/mod.rs
@@ -9,7 +9,8 @@ pub fn configure(config: &mut web::ServiceConfig, db: Database) {
     let service = AdvisoryService::new(db);
     config
         .app_data(web::Data::new(service))
-        .service(web::scope("/api/v1/advisory").service(all).service(get));
+        .service(all)
+        .service(get);
 }
 
 #[derive(OpenApi)]
@@ -32,7 +33,6 @@ pub struct ApiDoc;
 
 #[utoipa::path(
     tag = "advisory",
-    context_path = "/api/v1/advisory",
     params(
         Query,
         Paginated,
@@ -41,7 +41,7 @@ pub struct ApiDoc;
         (status = 200, description = "Matching vulnerabilities", body = PaginatedAdvisorySummary),
     ),
 )]
-#[get("")]
+#[get("/api/v1/advisory")]
 pub async fn all(
     state: web::Data<AdvisoryService>,
     web::Query(search): web::Query<Query>,
@@ -52,7 +52,6 @@ pub async fn all(
 
 #[utoipa::path(
     tag = "advisory",
-    context_path = "/api/v1/advisory",
     params(
         ("sha256", Path, description = "SHA256 of the advisory")
     ),
@@ -61,7 +60,7 @@ pub async fn all(
         (status = 404, description = "Matching advisory not found"),
     ),
 )]
-#[get("/{sha256}")]
+#[get("/api/v1/advisory/{sha256}")]
 pub async fn get(
     state: web::Data<AdvisoryService>,
     sha256: web::Path<String>,
--- a/modules/advisory/src/endpoints/mod.rs
+++ b/modules/advisory/src/endpoints/mod.rs
@@ -32,6 +32,7 @@ pub struct ApiDoc;
 
 #[utoipa::path(
     tag = "advisory",
+    context_path = "/api/v1/advisory",
     params(
         Query,
         Paginated,
@@ -51,6 +52,7 @@ pub async fn all(
 
 #[utoipa::path(
     tag = "advisory",
+    context_path = "/api/v1/advisory",
     params(
         ("sha256", Path, description = "SHA256 of the advisory")
     ),
@@ -59,7 +61,7 @@ pub async fn all(
         (status = 404, description = "Matching advisory not found"),
     ),
 )]
-#[get("{sha256}")]
+#[get("/{sha256}")]
 pub async fn get(
     state: web::Data<AdvisoryService>,
     sha256: web::Path<String>,
--- a/modules/organization/src/endpoints/mod.rs
+++ b/modules/organization/src/endpoints/mod.rs
@@ -1,0 +1,75 @@
+use crate::service::OrganizationService;
+use actix_web::{get, web, HttpResponse, Responder};
+use trustify_common::db::query::Query;
+use trustify_common::db::Database;
+use trustify_common::model::Paginated;
+use utoipa::OpenApi;
+
+pub fn configure(config: &mut web::ServiceConfig, db: Database) {
+    let service = OrganizationService::new(db);
+    config
+        .app_data(web::Data::new(service))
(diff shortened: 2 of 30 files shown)


carlosthe19916 (commented):
I am reviewing PRs mainly based on the output generated (I trust you in regards of the quality of the code).

- I can see a new field `issuer` on each advisory. Awesome! Thanks for that
- Not sure if it was originated by your PR but I can see some Vulnerabilities have now `severity=""` (empty severity). I thought, severity was a mandatory field and not an optional one. Is this a bug or it is expected? In my opinion each vulnerability must have a severity set and it must not be an optional field, but you are closer to the details of it.

Screenshot from 2024-05-16 11-03-26

bobmcwhirter:
Yah I have a TODO for that. 

Be aware a vuln has no inherent severity and will be blank if there's no advisory yet. 

I'll amend with code and tests demonstrating the various cases.

carlosthe19916:
> Be aware a vuln has no inherent severity and will be blank if there's no advisory yet.

- I see. So by design a Vulnerability might not have a severity assigned right? That sounds odd, I was expecting all Vulnerabilities to have a severity. If that is how things are designed, may I suggest that instead of an "empty string", for vulnerabilities without severity, we use "null"? If you see the image I shared above `severity=""`, but `severity=null` looks better.
- If this is something to be done in another PR I am fine merging this one and then addressing the severity in a separate PR. So as you prefer.

modules/organization/src/endpoints/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+use trustify_common::db::query::Query;
+use trustify_common::db::Database;
+use trustify_common::model::Paginated;
+use utoipa::OpenApi;
+
+pub fn configure(config: &mut web::ServiceConfig, db: Database) {
+    let service = OrganizationService::new(db);
+    config
+        .app_data(web::Data::new(service))
+        .service(web::scope("/api/v1/organization").service(all).service(get));
```
jcrossley3, line 12: I'm not sure exactly what the right way is, but I wish we could come up with a standard convention for arranging endpoints. I know we had issues with `web::scope` with advisories, so I might suggest we remove this and...

modules/organization/src/endpoints/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+        trustify_model::organization::OrganizationDetails,
+        trustify_model::organization::PaginatedOrganizationSummary,
+    )),
+    tags()
+)]
+pub struct ApiDoc;
+
+#[utoipa::path(
+    tag = "organization",
+    context_path = "/api/v1/organization",
```
jcrossley3, line 30: ...and remove all the `context_path` attributes in the `utoipa::path`'s...

modules/organization/src/endpoints/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+    params(
+        Query,
+        Paginated,
+    ),
+    responses(
+        (status = 200, description = "Matching organizations", body = PaginatedAdvisorySummary),
+    ),
+)]
+#[get("")]
```
jcrossley3, line 39: ...and fully qualify the paths here, e.g. 

```suggestion
#[get("/api/v1/organization")]
```
I'm not sure if this is best, but I'd like us to be consistent.

modules/advisory/src/endpoints/mod.rs
```diff
@@ -32,6 +32,7 @@ pub struct ApiDoc;
 
 #[utoipa::path(
     tag = "advisory",
+    context_path = "/api/v1/advisory",
```
jcrossley3, line 35: Didn't we decide not to do this?

modules/organization/src/endpoints/mod.rs
bobmcwhirter: I think the new modularization "solves" the web::scope issue for now.

You can ask the organization module to weave itself in. It's the only thing that can own /api/v1/organization, and can choose to /api/v2.3/organization also w/o having to muck with the server end of the stick.

modules/organization/src/endpoints/mod.rs
bobmcwhirter: If we remove the context_path from utoipa, then the swagger-ui/openapi stuff shows *the wrong path*, since it's ignorant of the web::scope, seemingly?

modules/organization/src/endpoints/mod.rs
bobmcwhirter: Ah, maybe you're exactly right.

Lemme give that a whirl.

modules/advisory/src/endpoints/mod.rs
bobmcwhirter: I DON'T KNOW ANYTHING ANY MORE

bobmcwhirter:
@jcrossley3 you're right, as usual.

carlosthe19916:
@bobmcwhirter I hate to block this PR but:
- If I hit `GET /api/v1/advisory` then I get `severity` for each vulnerability.

Screenshot from 2024-05-16 17-07-35

- But if I hit `GET /api/v1/vulnerability` then I get `aggregated_severity` for each Vulnerability

Screenshot from 2024-05-16 17-08-51

Can we be consistent in the naming?

carlosthe19916 (approved):
Once we will be able to have consistent naming `severity` vs `aggrevated_severity` for vulnerabilities we can merge this PR IMHO!

bobmcwhirter:
Oh, that's exactly intentional.

A `vulnerability` is just a name. It has no particular inherent severity or score, ever.

An `advisory` may express an opinion about a vulnerability, with a score/severity. In fact, a single `advisory` may express multiple opinions about multiple `vulnerability` within its context. When you load an `advisory`, the `vulnerabilities` under it are *its own opinions (and no other)* about the vulnerability.  

From the `advisory` end of the stick, the context is *only and exactly that advisory*. So any `vulnerability` underneath it is filtered by that context of that particular advisory.

From the same `advisory` context, the `average_*` data is simply the average of *all of its own opinions* about a `vulnerability`, without accounting for any other advisory's opinions.

When you start from `vulnerability` roots, though, the *average* is an actual average, of *all advisories* and *all of their opinions* about the vulnerability, which is why it's `average_*`.

jcrossley3 (approved):

ctron:
Judging from the CSAF spec 2.0: https://docs.oasis-open.org/csaf/csaf/v2.0/cs01/csaf-v2.0-cs01.html#3212-document-property---aggregate-severity

> Aggregate severity (aggregate_severity) of value type object with the mandatory property text and the optional property namespace is a vehicle that is provided by the document producer to convey the urgency and criticality with which the one or more vulnerabilities reported should be addressed. It is a document-level metric and applied to the document as a whole — not any specific vulnerability.

I would interpret this is an information covering the document itself, not it's vulnerabilities, nor their average.

PR #309, opened 2024-05-16
chore: adding a basic bench using pm-mode

helio-frota:


helio-frota:
converted to a draft as I don't know if we should merge it on the same repository.

PR #317, opened 2024-05-20
There's probably a better way

jcrossley3:
How do you test actions without merging them first?


bobmcwhirter:
fwiw, I've found that actions in a PR run, or at least changes-to-an-existing action do.

bobmcwhirter:
see if your change is PRing here maybe? https://github.com/trustification/trustify/actions

bobmcwhirter (approved):

PR #320, opened 2024-05-21
Rename/refactor ingestor & fetch modules quite a bit.

bobmcwhirter:
- Move storage endpoints to fetch documents into the document module
- Move storage endpoint tests to the document module
- Change the endpoints for downloading documents, also allowing more RESTful and future expansion to other digest'ing of keys.
- Adjust all naming


ctron (approved):
I am not that happy with the change of `ingestor` -> `document`. I hope we can improve on that in the future.

bobmcwhirter:
I can make the Ingestor a DocumentIngestor mayhap?

bobmcwhirter:
Damnit

modules/advisory/src/endpoints/mod.rs
```diff
@@ -76,5 +92,72 @@ pub async fn get(
...
+#[utoipa::path(
+    tag = "advisory",
+    request_body = Vec <u8>,
+    params( UploadParams ),
+    responses(
+    (status = 201, description = "Upload a file"),
+    (status = 400, description = "The file could not be parsed as an advisory"),
+    )
+)]
+#[post("/api/v1/document/advisory")]
```
jcrossley3, line 112: I'm unsure about the `document` path segment. Why is it necessary?

modules/sbom/src/endpoints/sbom.rs
```diff
@@ -117,3 +121,70 @@ pub async fn related(
...
+    request_body = Vec <u8>,
+    params(
+    ("location" = String, Query, description = "Source the document came from"),
+    ),
+    responses(
+    (status = 201, description = "Upload an SBOM"),
+    (status = 400, description = "The file could not be parsed as an advisory"),
+    )
+)]
+#[post("/api/v1/document/sbom")]
```
jcrossley3, line 144: ibid

jcrossley3 (commented):
I like the module/service distinction.

ctron:
I don't like it, sorry. What's the difference between a top level serve vs a top level module, which also has a service.

What will happen when we decide we need an endpoint for a service, do we rename it to module? I am not sure there's a requirement that a "module" must have an endpoint. I think that's ok.

I like the idea of the `sbom` module. That seems self-contained enough, and might pull in additional (internal) dependencies to implement its functionality.

bobmcwhirter:
What if... all services in services/ and the endpoints in endpoints/ 

I never quite knew what "modules" meant.

ctron (approved):

PR #329, opened 2024-05-22
ci: add a multi-arch release workflow

ctron:
This makes the release build multi-arch capable for the following targets:

* MacOS 12 for Intel
* MacOS 14 for Mx
* Windows
* Linux (GNU & MUSL) (amd64 & aarch64)

The release will look like this: https://github.com/ctron/trustify/releases/tag/v0.0.0-alpha.1

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -9,68 +9,241 @@ permissions:
   contents: write
 
 jobs:
-  build-and-upload:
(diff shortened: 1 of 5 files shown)


ctron:
Checked it works for MacOS 14 (Mx) and Linux amd64 GNU

bobmcwhirter (approved):

PR #330, opened 2024-05-22
Push advisory and sbom keys towards multi-variant HashKey.

bobmcwhirter:
FS Storage only (currently) supports SHA256, which is good as that's the only hash we're currently storing in the database.


modules/storage/src/service/fs.rs
```diff
@@ -149,8 +150,30 @@ impl StorageBackend for FileSystemBackend {
...
+                    std::io::ErrorKind::Unsupported,
+                    HashKeyError::UnsupportedAlgorithm(inner.clone()),
+                ));
+            }
+            _ => {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::InvalidInput,
+                    HashKeyError::UnsupportedAlgorithm("unknown".to_string()),
+                ));
+            }
```
jcrossley3, line 174: I'm not entirely sure how to deal with the non-exhaustiveness of this, but I think the error should contain the algorithm rather than the digest, right? Something like this...

```suggestion
            _ => {
                let algo = String::from(match hash_key.to_string().split_once(':') {
                    Some((v, _)) => v,
                    _ => "unknown",
                });
                return Err(std::io::Error::new(
                    std::io::ErrorKind::Unsupported,
                    HashKeyError::UnsupportedAlgorithm(algo),
                ));
            }
```

jcrossley3 (changes requested):

jcrossley3 (approved):

PR #332, opened 2024-05-23
feat: add an embedded OIDC server

ctron:


ctron:
It works. I just not sure how to expose this to the user. It should be disabled by default, that's why I put it behind a feature flag, which is off by default. Having the feature flag present during compilation, you also need `--embedded-oidc during startup.

The full command to start is:

```
cargo run -p trustify-trustd --features garage-door -- api --devmode --embedded-oidc
```

However, for the PM mode we want this on I guess, by default. Maybe it makes sense to compile two binaries? One insecure, one not.

bobmcwhirter (commented):
Also-- is there a way to communicate to the UI that the embedded OIDC is being used, so we can perhaps slap a `Hold up, you're insecure! NOT FOR PRODUCTION USE` banner on every page?

ctron:
The PR now also has two binaries (`trustd` and `trustd-pm`, the latter includes the embedded OIDC server).

Also, I added a container build, which uses the default `trustd` only.

jcrossley3 (changes requested):
How is this going to impact my life as a developer?

Answer in the form of some doc (e.g. README.md) in the PR, please.

ctron:
> How is this going to impact my life as a developer?
> 
> Answer in the form of some doc (e.g. README.md) in the PR, please.

By default that feature is not enabled and the information in the readme still applies.

I agree it might make sense to add this to some installation/getting started document. But I would avoid spamming the main readme with a lot of options.

ctron:
@jcrossley3 I added some docs. Linking them from the main readme.

modules/fundamental/src/advisory/endpoints/mod.rs
```diff
@@ -1,26 +1,34 @@
...
 
-    config
-        .app_data(web::Data::new(advisory_service))
-        .service(all)
-        .service(get)
-        .service(upload)
-        .service(download);
+    config.app_data(web::Data::new(advisory_service)).service(
+        web::scope("/api/v1/advisory")
+            .wrap(new_auth(auth))
```
jcrossley3, line 26: Why do we have to do this here? Can't we inject the auth in the code that calls `configure`? I'd like to keep the api paths DRY.

modules/fundamental/src/advisory/endpoints/mod.rs
```diff
@@ -44,6 +52,7 @@ pub struct ApiDoc;
 
 #[utoipa::path(
     tag = "advisory",
+    context_path = "/api/v1/advisory",
```
jcrossley3, line 55: This is brittle and error-prone, imo.

modules/fundamental/src/advisory/endpoints/mod.rs
```diff
@@ -52,7 +61,7 @@ pub struct ApiDoc;
         (status = 200, description = "Matching vulnerabilities", body = PaginatedAdvisorySummary),
     ),
 )]
-#[get("/api/v1/advisory")]
+#[get("")]
```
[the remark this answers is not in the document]
jcrossley3: I prefer seeing the whole path next to the handler.

modules/fundamental/src/advisory/endpoints/test.rs
```diff
@@ -59,7 +59,7 @@ async fn all_advisories(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
     let (storage, _) = FileSystemBackend::for_test().await?;
 
     let app = actix_web::test::init_service(
-        App::new().configure(|config| configure(config, db, storage.clone())),
+        App::new().configure(|config| configure(config, db, storage.clone(), None)),
```
jcrossley3, line 62: What is the extra parameter buying us?

modules/fundamental/src/endpoints.rs
```diff
@@ -8,19 +10,20 @@ pub fn configure(
     config: &mut web::ServiceConfig,
     db: Database,
     storage: impl Into<DispatchBackend>,
+    auth: Option<Arc<Authenticator>>,
 ) {
     let storage = storage.into();
 
     let ingestor_service = IngestorService::new(Graph::new(db.clone()), storage.clone());
     config.app_data(web::Data::new(ingestor_service));
```
jcrossley3, line 18: Can't we inject auth into the config before we call all the configure's, obviating the need to pass it?

jcrossley3 (changes requested):

modules/fundamental/src/advisory/endpoints/mod.rs
ctron: I am not sure how that should work. The scope needs to be wrapped with the auth-middleware. Nothing else.

modules/fundamental/src/advisory/endpoints/mod.rs
ctron: It you have a better idea, I am open for suggestions.

modules/fundamental/src/advisory/endpoints/test.rs
ctron: That we can authenticate only a scope vs all of the server endpoints. Which would include the UI, which actually is there to trigger then authentication.

modules/fundamental/src/endpoints.rs
ctron: I don't know how. I am open to suggestions.

modules/fundamental/src/organization/endpoints/mod.rs
```diff
@@ -1,16 +1,24 @@
...
 
-pub fn configure(config: &mut web::ServiceConfig, db: Database) {
+pub fn configure(config: &mut web::ServiceConfig, db: Database, auth: Option<Arc<Authenticator>>) {
     let service = OrganizationService::new(db);
-    config
-        .app_data(web::Data::new(service))
-        .service(all)
-        .service(get);
+    config.app_data(web::Data::new(service)).service(
+        web::scope("/api/v1/organization")
```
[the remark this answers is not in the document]
jcrossley3: I did a bit of digging, but couldn't come up with a simple solution. I was hoping to hang the auth middleware off the App, with an ability to exclude certain paths, e.g. the UI.

Let's go with your single scope idea, but only as much as we have to, i.e. `web::scope("/api")`. That way, every utoipa `context-path` is the same, and therefore hopefully a little less error-prone: `"/api"`.

That means every actix method macro will start with a `/v1/...`. That should at least make adding `/v2` handlers straightforward when the time comes.

ctron:
@jcrossley3 I updated the scope to use a unified `/api` scope.

modules/fundamental/src/advisory/endpoints/test.rs
```diff
@@ -59,7 +59,8 @@ async fn all_advisories(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
     let (storage, _) = FileSystemBackend::for_test().await?;
 
     let app = actix_web::test::init_service(
-        App::new().configure(|config| configure(config, db, storage.clone())),
+        App::new()
+            .service(web::scope("/api").configure(|config| configure(config, db, storage.clone()))),
```
jcrossley3, line 63: This is only necessary due to the "/api" prefix in the test uri below, correct? Alternatively, we could remove the "/api" prefix there, I guess.

ctron: True. However, that would mean we have two sets of URLs. One for testing, one for the actual server. Which I think we should avoid.

jcrossley3: Maybe. But within the context of a single test, adding a scope seems like unnecessary noise to me. I hope we can trust actix to resolve the path in the same way I trust DNS to resolve the hostname.

Issue #254, opened 2024-05-09
Implement Search of Packages endpoint


carlosthe19916:
Currently there is no way of searching packages. I can see there are 2 endpoints implemented already for packages but "Searching packages" is the first endpoint the UI will need to start the flow.

![Screenshot from 2024-05-09 11-36-59](https://github.com/trustification/trustify/assets/2582866/542034eb-e28f-410a-b543-a24bad8a7afe)

bobmcwhirter:
pending some refactor by @ctron

carlosthe19916:
@jcrossley3 @bobmcwhirter Based on this comment https://github.com/trustification/trustify/pull/336#issuecomment-2127965499 my impression was that the "Search packages" was already available but when I see the swagger UI I can see only:

![Screenshot from 2024-06-06 10-47-50](https://github.com/trustification/trustify/assets/2582866/1f998f16-4bc6-467b-a4c7-65bdd07837ff)

So which is the endpoint to get the "list of all packages with appropriate pagination"? it seems that is missing. And that is the very first thing I need to start the whole flow in the UI (without it all the other endpoints although valuate and not useful at all)

carlosthe19916:
Closing this issue as it was resolved by https://github.com/trustification/trustify/pull/390

PR #336, opened 2024-05-23
First swath at /api/v1/package/***.

bobmcwhirter:


jcrossley3:
Will this resolve #254 ? If so, can you write as much in the description?

bobmcwhirter:
Yes! I think.

jcrossley3 (commented):
First I've heard of "ecosystems". Is that just an OSV thing? Will this impact our goal of parity with v1?

What's an ecosystem and how do they relate to packages? Maybe include some description in the glossary within this PR?

PR #343, opened 2024-05-30
Add migrations for status and package assertion.

bobmcwhirter:
This PR re-works all of the assertions stuff by primarily...

removing the old table-per-assertion crap that didn't work.
adding version-range table (with associated version-scheme [semver, maven, pypi, etc])
adding a status table to avoid hard-coding and allow expansion in the future.

Currently surfaced through ingesting advisories to create rows in the tables.
And then pulled out for *only* `/api/v1/package/{uuid}` currently while we evaluate the shape of the DTOs.

Will definitely expand where package status can appear later (package, package-version, advisory, etc...)

But the plumbing is in place.

**ALSO**

This removes quite a few of the read-side operations from `Graph` as they shouldn't be there.


migration/src/m0000331_create_status.rs
```diff
@@ -0,0 +1,95 @@
...
+        .await?;
+        insert(db, "fixed", "Fixed", "Vulnerabililty is fixed").await?;
+        insert(
+            db,
+            "recommended",
+            "Recommended",
+            "Vulnerabililty is fixed & recommended",
+        )
+        .await?;
+
```
jcrossley3, line 54: Why not just a rusty enum for these? Will we need to add more later without publishing a new release?

migration/src/m0000335_create_version_scheme.rs
```diff
@@ -0,0 +1,193 @@
...
+        .await?;
+
+        insert(
+            db,
+            "gem",
+            "Rubygems",
+            Some("https://guides.rubygems.org/patterns/#semantic-versioning"),
+        )
+        .await?;
+
```
jcrossley3, line 74: These make less sense as a rusty enum, I suppose, but it still feels kinda gross hardcoding things in SQL rather than Rust.

migration/src/m0000345_create_version_comparison_fns.rs
```diff
@@ -0,0 +1,61 @@
...
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .get_connection()
+            .execute_unprepared(r#"drop function semver_gt"#)
+            .await?;
+
+        manager
+            .get_connection()
+            .execute_unprepared(r#"drop function semver_gte"#)
+            .await?;
```
jcrossley3, line 27: At the risk of pointing out a problem to which I have no practical suggestion...

I don't love referring to the names of all the functions created in the bag of sql. It feels brittle. I get why you want to define all the functions in a single sql. I'm inclined to suggest you drop all of them in another sql, but that extra level of indirection doesn't really solve the problem of one file referring to things created in another. I'd just as soon see the names in the migration, like you have it.

Ideally, the `up()` would create all its functions in some kind of namespace and the `down()` could just obliterate the namespace (and all it contains)...

...but I dunno how to do that... so...

...nevermind.

migration/src/m0000350_remove_old_assertion_tables.rs
```diff
@@ -0,0 +1,236 @@
...
+            )
+            .await?;
+
+        Ok(())
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .create_table(
+                Table::create()
```
jcrossley3, line 60: I for one appreciate this level of effort, but when an `up()` consists mostly of `.drop_table`'s and its `down()` a bunch of `.create_table`'s -- especially for a feature that doesn't actually work -- I might be inclined to forget the `down()` as long as the old migrations drop their tables conditionally, i.e. if they exist.

Maybe. I'm not sure. Just looks weird. Forget I mentioned it.

jcrossley3 (approved):
Migrations are hard. Nice tests.

modules/fundamental/src/advisory/service/test.rs
```diff
@@ -268,9 +269,31 @@ async fn single_advisory(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
...
+            "fixed",
+            VersionInfo {
+                scheme: "semver".to_string(),
+                spec: VersionSpec::Exact("1.2.3".to_string()),
+            },
+            (),
+        )
         .await?;
 
+    println!("A");
```
ctron, line 295: Log?

modules/fundamental/src/package/endpoints/test.rs
```diff
@@ -250,6 +254,8 @@ async fn package(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
     let response: QualifiedPackageDetails =
         actix_web::test::call_and_read_body_json(&app, request).await;
 
+    println!("{:#?}", response);
```
ctron, line 257: Also log?

modules/fundamental/src/package/model/details/qualified_package.rs
```diff
@@ -44,6 +54,36 @@ impl QualifiedPackageDetails {
                 .ok_or(Error::Data("underlying package missing".to_string()))?
         };
 
+        /*
```
ctron, line 57: I would prefer not to add more commented out code.

migration/src/m0000331_create_status.rs
bobmcwhirter: While there are some intrinsic statuses, I don't deign to know all statuses customers might want to apply, in the future. So leaving the architecture open to extension at runtime.

I waffle about using the slug as the actual PK.

migration/src/m0000335_create_version_scheme.rs
bobmcwhirter: This could be rusty, but the SQL also has to know the name of the version scheme to select the correct `cmp` function.

modules/fundamental/src/package/model/details/qualified_package.rs
```diff
@@ -44,6 +54,36 @@ impl QualifiedPackageDetails {
...
+                version_range::Column::LowInclusive,
+                version_range::Column::HighVersion,
+                version_range::Column::HighInclusive,
+            ])
+            .left_join(package::Entity)
+            .join(JoinType::LeftJoin, package::Relation::PackageVersions.def())
+            .left_join(version_range::Entity)
+            .filter(package_status::Column::PackageId.eq(package.id))
+            .filter(SimpleExpr::FunctionCall(
+                Func::cust("version_matches".into_identity().into_iden())
```
ctron, line 80: It's possible to create `Iden`s which can then be re-used. This prevents spreading identifiers throughout the code: https://github.com/trustification/trustify/blob/1d00670178f628f3bd1b803886d03e68a1de44ad/common/src/db/func.rs#L4-L14

migration/src/m0000345_create_version_comparison_fns.rs
bobmcwhirter: You're right. Imma be writing more (for other version scheme `cmp`) so lemme try to make a more-clear way (helper functions or something) to make this more directly comprehensible.

ctron (changes requested):
It ok in general. I think there are some minor things that I either don't understand or that should be cleaned up before merging.

I also noticed a bunch of tests that seem to be stripped down. Not sure that's the intention. If it is, ok.

entity/src/version_range.rs
```diff
@@ -0,0 +1,41 @@
+use crate::package_status;
+use sea_orm::entity::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
+#[sea_orm(table_name = "version_range")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
```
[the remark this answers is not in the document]
bobmcwhirter: Yes, I believe so.

entity/src/version_range.rs
```diff
@@ -0,0 +1,41 @@
...
+    pub version_scheme_id: String,
```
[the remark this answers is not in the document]
bobmcwhirter: Ah, this i want keyed on string/text, not i32 or uuid, so will remedy.

I want to be able to pass `version_range.*` without any complex joins in order to determine if a version matches the range. This means knowing some predictable version-scheme key, and avoiding a Uuid->scheme_name lookup is bonus.

modules/fundamental/src/advisory/service/test.rs
bobmcwhirter: Yes yes.

modules/fundamental/src/package/model/details/qualified_package.rs
bobmcwhirter: Right on.

modules/fundamental/src/package/model/details/qualified_package.rs
bobmcwhirter: Good point. Will do.

bobmcwhirter:
@ctron tests stripped down because they were... the wrong tests? bad tests? Basically they were relying on a lot of read-centric operations on Graph that shouldn't be there.

Ultimately, I think we need to probably go about testing ingestors differently, to ensure ingested data is as-expected, probably from the fundamental side of the stick.

bobmcwhirter:
@jcrossley3 adjusted migrations in this PR to be completely self-contained, not importing enums from others.

Also made their enums private to avoid escapage.

migration/src/m0000345_create_version_comparison_fns/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .get_connection()
+            .execute_unprepared(include_str!("is_numeric.sql"))
```
jcrossley3, line 11: I was gonna suggest we introduce a `sql/` dir somewhere to hold all these and then noticed you created a module dir to contain them. That's good, too.

jcrossley3 (approved):

entity/src/version_range.rs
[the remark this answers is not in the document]
ctron: So is it an "id"?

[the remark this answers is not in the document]
ctron: Maybe it makes sense to add a comment explaining the idea?

ctron (approved):

PR #346, opened 2024-05-31
Fix wall of notifications when adding multiple reviewers to a PR

ctron:
Right now, when adding multiple reviewers on a PR, this triggers a repeated invocation of the event handler (one for each reviewer). It also send out an event for each reviewer, each time. So the number of messages sent is x².

This change fixes this and updates a few github actions in the process.

### DIFF
--- a/.github/scripts/map-users/main.mjs
+++ b/.github/scripts/map-users/main.mjs
@@ -12,13 +12,8 @@ export class Mapper {
   requestedReviewers() {
     let users = [];
 
-    for (let user of this.payload?.pull_request?.requested_reviewers) {
-
-      const login = user?.login;
-      if (login === undefined) {
-        continue;
-      }
-
+    const login = this.payload?.requested_reviewer?.login;
+    if (login !== undefined) {
       users.push(login);
     }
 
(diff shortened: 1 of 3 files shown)


.github/scripts/map-users/main.mjs
```diff
@@ -12,13 +12,8 @@ export class Mapper {
...
 
-    for (let user of this.payload?.pull_request?.requested_reviewers) {
-
-      const login = user?.login;
-      if (login === undefined) {
-        continue;
-      }
-
+    const login = this.payload?.requested_reviewer?.login;
+    if (login !== undefined) {
```
bobmcwhirter, line 16: `undefined is not a function`

.github/scripts/map-users/test.mjs
```diff
@@ -1,12 +1,14 @@
 import * as fs from 'fs';
 import { channelMessage, Mapper } from "./main.mjs";
 
-const github = JSON.parse(fs.readFileSync('payload.json'))?.github;
+// const github = JSON.parse(fs.readFileSync('payload.json'))?.github;
+// const output = channelMessage(github.event)
 
-const output = channelMessage(github.event)
+const payload = JSON.parse(fs.readFileSync('payload.json'));
```
bobmcwhirter, line 7: `[object Object]`

bobmcwhirter (approved):

.github/scripts/map-users/main.mjs
ctron: true … truthy? … or something like that: 
image

PR #350, opened 2024-06-03
Fix some build issues

ctron:
This change fixes some build issues:

* openssl is not part of ubi-micro, this will be installed now too
* postgresql-embedded is forced into "bundled" mode, which doesn't work with offline builds. It is still the default, but there's a way to opt-out now.
* I forgot to uptick the versions during the last release, this fixes it.


dejanb (approved):

bobmcwhirter (commented):
Why can't I approve this?image

PR #351, opened 2024-06-03
Fix PM mode auth setting when using embedded OIDC

ctron:
Also: uptick to next release version


helio-frota:
any chance to add a readme command to copy/paste and run ?

helio-frota (approved):
thanks :+1:

PR #360, opened 2024-06-04
Add GraphQL API to trustify server

gildub:
Initial integration of GraphQL service using async-graphql providing the following :

- A `/graphql` endpoint on existing Actix server.
- Graphiql  user interface available when running the browser with `*:8080/graphql` URL
- A `trustify-module-graphql` member defined in 'modules/graphql` which contains the GraphQL Schema
- The initial available queries are :
```rust
type RootQuery {
	getAdvisoryById(id: Int!): Advisory!
	getAdvisories: [Advisory!]!
	getOrganizationByName(name: String!): Organization!
	getSbomById(id: UUID!): Sbom!
	getSbomsByLocation(location: String!): [Sbom!]!
	getVulnerabilityById(identifier: String!): Vulnerability!
	getVulnerabilities: [Vulnerability!]!
}
```
Note : The schema is available by adding `println!("{}", &schema.sdl());`

I changed advisory entity field `advisory.issuer_id` to `advisory.organization_id`. The reason is because I initially thought graphql schema mapping was automatically looking up the related entity (organization) for a foreign key. 
That's the way some GraphQL frameworks work. This doesn't seem relevant with async-graphql, meanwhile I kept the change because it makes more sense. In case we keep issuer_id as foreign key then it would make sense to name the associated entity "Issuer".

### DIFF
--- a/entity/Cargo.toml
+++ b/entity/Cargo.toml
@@ -7,8 +7,15 @@ edition = "2021"
 trustify-common = { workspace = true }
 trustify-cvss = { workspace = true }
 
+async-graphql = { version = "7.0.5", features = ["uuid", "time"] }
 cpe = { workspace = true }
-sea-orm = { workspace = true, features = ["sqlx-postgres", "runtime-tokio-rustls", "macros", "with-json", "postgres-array"] }
+sea-orm = { workspace = true, features = [
+    "sqlx-postgres",
+    "runtime-tokio-rustls",
+    "macros",
+    "with-json",
+    "postgres-array",
+] }
 serde = { workspace = true }
 serde_json = { workspace = true }
 time = { workspace = true }
@@ -17,4 +24,4 @@ utoipa = { workspace = true }
 [dev-dependencies]
 anyhow = { workspace = true }
 log = { workspace = true }
-test-log = { workspace = true, features = ["log", "trace"]}
+test-log = { workspace = true, features = ["log", "trace"] }
--- a/entity/src/advisory.rs
+++ b/entity/src/advisory.rs
@@ -16,7 +16,7 @@ use trustify_common::db;
 pub struct Model {
     #[sea_orm(primary_key)]
     pub id: Uuid,
-    #[graphql(name="name")]
+    #[graphql(name = "name")]
     pub identifier: String,
     pub issuer_id: Option<i32>,
     pub location: String,
--- a/entity/src/advisory.rs
+++ b/entity/src/advisory.rs
@@ -57,7 +57,7 @@ pub enum Relation {
         belongs_to = "super::organization::Entity"
         from = "Column::IssuerId"
         to = "super::organization::Column::Id")]
-    Organization,
+    Issuer,
 
     #[sea_orm(has_many = "super::cvss3::Entity")]
     Cvss3,
@@ -74,7 +74,7 @@ pub enum Relation {
 
 impl Related<organization::Entity> for Entity {
     fn to() -> RelationDef {
-        Relation::Organization.def()
+        Relation::Issuer.def()
     }
 }
 
--- a/entity/src/advisory.rs
+++ b/entity/src/advisory.rs
@@ -18,7 +18,7 @@ pub struct Model {
     pub id: Uuid,
     #[graphql(name="name")]
     pub identifier: String,
-    pub organization_id: Option<i32>,
+    pub issuer_id: Option<i32>,
     pub location: String,
     pub sha256: String,
     pub published: Option<OffsetDateTime>,
@@ -55,7 +55,7 @@ impl Model {
 pub enum Relation {
     #[sea_orm(
         belongs_to = "super::organization::Entity"
-        from = "Column::OrganizationId"
+        from = "Column::IssuerId"
         to = "super::organization::Column::Id")]
     Organization,
 
--- a/entity/src/advisory.rs
+++ b/entity/src/advisory.rs
@@ -16,6 +16,7 @@ use trustify_common::db;
 pub struct Model {
     #[sea_orm(primary_key)]
     pub id: Uuid,
+    #[graphql(name="name")]
     pub identifier: String,
     pub organization_id: Option<i32>,
     pub location: String,
--- a/entity/src/advisory.rs
+++ b/entity/src/advisory.rs
@@ -2,16 +2,22 @@ use crate::{
     advisory_vulnerability, affected_package_version_range, cvss3, fixed_package_version,
     not_affected_package_version, organization, vulnerability,
 };
+use async_graphql::*;
 use sea_orm::entity::prelude::*;
+use std::sync::Arc;
 use time::OffsetDateTime;
+use trustify_common::db;
 
-#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
+#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel, SimpleObject)]
+#[graphql(complex)]
+#[graphql(concrete(name = "Advisory", params()))]
 #[sea_orm(table_name = "advisory")]
+
 pub struct Model {
     #[sea_orm(primary_key)]
     pub id: Uuid,
     pub identifier: String,
-    pub issuer_id: Option<i32>,
+    pub organization_id: Option<i32>,
     pub location: String,
     pub sha256: String,
     pub published: Option<OffsetDateTime>,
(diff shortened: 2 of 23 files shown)


entity/src/advisory.rs
```diff
@@ -2,16 +2,22 @@ use crate::{
...
+#[graphql(complex)]
+#[graphql(concrete(name = "Advisory", params()))]
 #[sea_orm(table_name = "advisory")]
+
 pub struct Model {
     #[sea_orm(primary_key)]
     pub id: i32,
     pub identifier: String,
-    pub issuer_id: Option<i32>,
+    pub organization_id: Option<i32>,
```
bobmcwhirter, line 20: I think having it named 'issuer_id' is still preferred. An organization does issue it, but conceivably there could be other organizations (performing other roles) involved.

e.g., issuer, publisher, etc.

Does the GraphQL require the table name in the fk field?

bobmcwhirter (changes requested):

entity/Cargo.toml
```diff
@@ -7,8 +7,15 @@ edition = "2021"
 trustify-common = { workspace = true }
 trustify-cvss = { workspace = true }
 
+async-graphql = { version = "7.0.5", features = ["uuid", "time"] }
 cpe = { workspace = true }
-sea-orm = { workspace = true, features = ["sqlx-postgres", "runtime-tokio-rustls", "macros", "with-json", "postgres-array"] }
+sea-orm = { workspace = true, features = [
```
ctron, line 12: I would prefer keeping dependencies in a single line. I know in this case it can be hard to read. However, having set of single lines allows one to easily have IntelliJ sort all lines.

An alternative could be to pull out such dependencies out of this block and just add it later. Similar to: https://github.com/trunk-rs/trunk/blob/c160ed3ff7c98a94ab00f14acbbd198268a92525/examples/yew-tls/Cargo.toml#L13-L18

entity/src/advisory.rs
gildub: This is a common practice in GraphQL to have foreign key being labelled as "<Name of Table/Entity_id> because this quasi systematically is going to be used as the key for looking-up the record, especially in one-to-one relationship. But there are no such technical constraint for async-graphql, so we can restore the issuer_id.

entity/Cargo.toml
gildub: @ctron, 

I'm happy to use either approach. 
Meanwhile this is done automatically by Rust-analyzer which I believe is now the default for many IDEs.
To me it seems to be an IntelliJ issue and the main reason for that is a very long import line, whatever would be a configured limit of 80, 120, etc number of characters, is not sustainable.

carlosthe19916:
@gildub just pasting my question/comment from our last meeting (where you did a demo)

>It seems what is exposed are the Model/Entities directly. Does it mean the DB must be designed in a way in which the Model/Entity is 100% ready for client consumption? Currently the REST approach is exposing DTOs not the Model/Entity directly

And by Model/Entity I meant the Database tables. This should be something to consider and be evaluated

entity/src/advisory.rs
gildub: @bobmcwhirter,  I have restored `advisory.issuer_id`.

gildub:
@carlosthe19916, 

We can expose only what we need to.

For example we can filter out any entity field (or database column) with help `#[graphql(skip)]` attribute.
Also we can rename any field using `#[graphql(name="blah") ]`.
Alternatively we can replace (or override) some field by wrapping them (usually when doing some calculation on top or aggregating).  

Beyond that we can define structs to create new GraphQL objects which we populate with whatever data that don't have to be directly corresponding to an Entity table. For instance the data could be partly be the result of some entity data or the combination of several sources.

gildub:
@bobmcwhirter, 
I assumed there are no used case for using different Actix-web servers for each Rest and Graphql APIs.  To be revisited if needed.
The other assumption is the GraphQL endpoint, we'll it's more to get started but it's ultimately a team  decision to be taken.

As we have /api for the REST API, do we want to change that to not confuse it with GraphQL's API. Or maybe that's not a subject and /api is by default REST.
Also how to name the GraphQL endpoint  ?

bobmcwhirter:
Let's share the same actix http server. No reason to add more open ports.

And I'd be fine with `/graphql/v1` (or without version if you think that's better) in the URL space.

gildub:
@bobmcwhirter, versioning vs evolution is the question we need to answer.

GraphQL can use both but has a tendency to use evolution.
With evolution :
- A new field can be added without conflict because previous client don't have it yet.
- When changing an existing field, it can be marked as deprecated (feature supported by async-graphql) and replace with a new field name.

The evolution approach allows to not use versioning but has cons.
- The schema could end up polluted with lots of @deprecated field.
- The field name is less "neat" (`name` is the old name replaced by `newName`).

One solution to overcome those issues is to use semantic versioning at the object and/or field levels.
This is describe here https://blog.logrocket.com/versioning-fields-graphql/

But at the end it means a developer must actively introspect the schema to check if the schema was upgraded and if fields re are deprecated.

There are solution as well with the help of extensions, also supported by async-graphql.

So by using evolution and combining semantic versioning we can't get best of both world.
If we use that latter approach then we don't need `/graphql/v1`.
What do you think ?

entity/src/advisory.rs
```diff
@@ -20,13 +27,37 @@ pub struct Model {
...
+}
+
 #[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
 pub enum Relation {
     #[sea_orm(
         belongs_to = "super::organization::Entity"
         from = "Column::IssuerId"
         to = "super::organization::Column::Id")]
-    Issuer,
+    Organization,
```
bobmcwhirter, line 60: Would still prefer to not rename this relationship.

modules/ingestor/src/graph/sbom/mod.rs
```diff
@@ -186,7 +186,7 @@ impl Graph {
             .map(|sbom| SbomContext::new(self, sbom)))
     }
 
-    async fn locate_many_sboms<TX: AsRef<Transactional>>(
+    pub async fn locate_many_sboms<TX: AsRef<Transactional>>(
```
bobmcwhirter, line 189: Most of these fn's you are using from `ingestor` should actually end up in some `FooService` in the `fundamental` module.

bobmcwhirter: clarifying: this isn't your problem. But we should probably move them out of `ingestor` (or provide similar in `fundamental` such as `SbomService`) for usage by GraphQL stuff.

modules/ingestor/src/graph/vulnerability/mod.rs
```diff
@@ -94,6 +94,19 @@ impl Graph {
             .await?
             .map(|cve| VulnerabilityContext::new(self, cve)))
     }
+
+    pub async fn get_vulnerabilities<TX: AsRef<Transactional>>(
+        &self,
+        tx: TX,
+    ) -> Result<Vec<VulnerabilityContext>, Error> {
+        Ok(entity::vulnerability::Entity::find()
```
bobmcwhirter, line 102: Likewise, see `AdvisoryService` in `fundamental`. It's used by the advisory REST endpoints, and could likewise be consumed by GraphQL.

server/Cargo.toml
```diff
@@ -12,8 +12,10 @@ trustify-module-importer = { workspace = true }
 trustify-module-ingestor = { workspace = true }
 trustify-module-storage = { workspace = true }
 trustify-module-ui = { workspace = true }
+trustify-module-graphql = { workspace = true }
 
 actix-web = { workspace = true }
+actix-cors = { workspace = true }
```
bobmcwhirter, line 18: humor: You're going to draw the wrath and ire of @ctron by touching CORS. :)

server/src/lib.rs
```diff
@@ -148,6 +152,12 @@ impl Run {
     }
 }
 
+async fn index_graphiql() -> Result<HttpResponse> {
```
bobmcwhirter, line 155: question: This provides both the playground UI for graphql *and* the actual endpoint?

server/src/lib.rs
```diff
@@ -266,6 +281,20 @@ impl InitData {
                             openapi::openapi(),
                             swagger_oidc.clone(),
                         ));
+                    svc.service(
```
bobmcwhirter, line 284: Can we move this endpoint setup to something like `modules/graphql/src/endpoints.rs` matching other modules?

server/src/lib.rs
```diff
@@ -330,7 +359,7 @@ async fn index(ci: ConnectionInfo) -> Result<Json<Vec<url::Url>>, UrlGenerationE
 
     result.extend(build_url(&ci, "/"));
     result.extend(build_url(&ci, "/openapi.json"));
-    result.extend(build_url(&ci, "/openapi/"));
+    result.extend(build_url(&ci, "/openapi.json"));
```
bobmcwhirter, line 362: question: seems to duplicate lines and remove the swagger UI url?

bobmcwhirter (changes requested):

server/src/lib.rs
gildub: Yes it does, I think it uses a combo of GET, empty query and client being a browser to determine it's not a query/mutation to serve the playground.

entity/src/advisory.rs
gildub: That was a left over, thanks. Done.

server/src/lib.rs
gildub: That's a mistake. Thanks !

server/Cargo.toml
gildub: @bobmcwhirter, I removed those permissive rules ! So far so good.

server/src/lib.rs
gildub: Done

modules/ingestor/src/graph/sbom/mod.rs
bobmcwhirter: Do you want to try to move to the relevant FooService (augmenting as necessary?) in this PR, or do that in a subsequent one?

gildub: @bobmcwhirter, I'd rather do it in a follow-up PR. It'll give me time to catch-up with you/@jcrossley3/@ctron so I can grasp the Entity/Ingestor/Fundamental layers roles.

modules/ingestor/src/graph/vulnerability/mod.rs
gildub: Same as above

PR #366, opened 2024-06-05
fix: client ids need to split by comma when coming from the env-var

ctron:


carlosthe19916 (approved):

common/auth/src/authenticator/config.rs
```diff
@@ -69,7 +69,7 @@ impl From<AuthenticatorConfigArguments> for Option<AuthenticatorConfig> {
 #[command(next_help_heading = "Authentication settings")]
 pub struct SingleAuthenticatorClientConfig {
     /// The client IDs to allow
-    #[arg(env = "AUTHENTICATOR_OIDC_CLIENT_IDS", long = "authentication-client-id", action = ArgAction::Append)]
+    #[arg(env = "AUTHENTICATOR_OIDC_CLIENT_IDS", long = "authentication-client-id", action = ArgAction::Append, value_delimiter = ',')]
```
carlosthe19916, line 72: Why do we need multiple clients? I though that now that trustify is a single server the only client required is the one used by the UI. I'm asking just because I am curious, I'll leave to you guys the final decision.

carlosthe19916 (approved):

common/auth/src/authenticator/config.rs
ctron: Because I am also using this from the command line, using `http`. Which I believe is a valid use case. For that you need a confidential/protected client, which works differently.

PR #386, opened 2024-06-06
Refactor SBOM ingestion to remove duplicated storage code

jcrossley3:
This DRY's up the SBOM ingestion and brings it in line with Advisory ingestion. The `ingest_sbom` function has been removed. All docs now go through the `ingest` method of the `IngestorService`, passing a `Format` variant used to select an appropriate loader to populate the database after storing the doc.


modules/ingestor/src/service/format.rs
```diff
@@ -39,8 +41,14 @@ impl<'g> Format {
                 let loader = CveLoader::new(graph);
                 loader.load(source, reader, checksum).await
             }
-            Format::SPDX => todo!(),
-            Format::CycloneDX => todo!(),
+            Format::SPDX => {
+                let loader = SPDXLoader::new(graph);
```
ctron, line 45: Just a nitpick  … It's `CveLoader`, but `SDPXLoader` and `CycloneDXLoader`. I would prefer to go for plain camel case. And I think Rust does too: https://rust-lang.github.io/rust-clippy/rust-1.78.0/index.html#/acron

ctron (approved):

PR #387, opened 2024-06-06
fix: add a git cli version

ctron:
Workaround for: https://github.com/rust-lang/git2-rs/issues/1057

### DIFF
--- a/.github/scripts/Containerfile
+++ b/.github/scripts/Containerfile
@@ -10,7 +10,7 @@ RUN mkdir /stage
 
 # install zlib and cleanup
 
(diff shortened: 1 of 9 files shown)


ctron:
I am not completely happy with this PR. The downside is, that this requires the `git` binary in order to import from OSV and CVE sources. Which in PM mode, we would need to require the user to bring.

On the other side, it doesn't segfault.

ctron:
I also checked `gix`, which is a pure Rust implementation for a git client. However, it lacks some features. Most notably the `reset` function. Maybe working around that limitation is an option.

bobmcwhirter (approved):

PR #396, opened 2024-06-10
Allow filtering and sorting from synthetic non-Entity columns.

bobmcwhirter:
This is prereq for sorting by calculated columns like `average_score`

### DIFF
--- a/common/src/db/query.rs
+++ b/common/src/db/query.rs
@@ -100,13 +100,13 @@ impl<T: EntityTrait> Filtering<T> for Select<T> {
         let mut result = if q.is_empty() {
             self
         } else {
-            self.filter(Filter::from_str(q, context.columns())?)
+            self.filter(Filter::parse(q, context.columns())?)
         };
 
         if !sort.is_empty() {
             result = sort
                 .split(',')
-                .map(|s| Sort::from_str(s, context.columns()))
+                .map(|s| Sort::parse(s, context.columns()))
                 .collect::<Result<Vec<_>, _>>()?
                 .into_iter()
                 .fold(result, |select, s| {
@@ -243,7 +243,7 @@ struct Filter {
 }
 
 impl Filter {
-    fn from_str<C: IntoColumns>(s: &str, context: C) -> Result<Self, Error> {
+    fn parse<C: IntoColumns>(s: &str, context: C) -> Result<Self, Error> {
         const RE: &str = r"^(?<field>[[:word:]]+)(?<op>=|!=|~|!~|>=|>|<=|<)(?<value>.*)$";
         static LOCK: OnceLock<Regex> = OnceLock::new();
         #[allow(clippy::unwrap_used)]
@@ -259,7 +259,7 @@ impl Filter {
                 operands: Operand::Composite(
                     encoded
                         .split('&')
-                        .map(|e| Filter::from_str(e, columns.clone()))
+                        .map(|e| Filter::parse(e, columns.clone()))
                         .collect::<Result<Vec<_>, _>>()?,
                 ),
             })
@@ -327,7 +327,7 @@ struct Sort {
 }
 
 impl Sort {
-    fn from_str<C: IntoColumns>(s: &str, context: C) -> Result<Self, Error> {
+    fn parse<C: IntoColumns>(s: &str, context: C) -> Result<Self, Error> {
         let columns = context.columns();
 
         let s = s.to_lowercase();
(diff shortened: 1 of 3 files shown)


common/src/db/query.rs
```diff
@@ -130,42 +145,210 @@ pub enum Error {
 // Internal types
 /////////////////////////////////////////////////////////////////////////
 
+#[derive(Default)]
+pub struct Columns {
+    columns: Vec<(ColumnRef, ColumnDef)>,
+}
```
jcrossley3, line 151: If I'm understanding correctly, you're expanding the public interface of this module to include `Columns`, so the comment above it is wrong, since "internal types" aren't public.

common/src/db/query.rs
```diff
@@ -130,42 +145,210 @@ pub enum Error {
...
-struct Filter<T: EntityTrait> {
-    operands: Operand<T>,
+struct Filter {
+    operands: Operand,
     operator: Operator,
 }
 
-struct Sort<T: EntityTrait> {
-    field: T::Column,
+impl Filter {
```
jcrossley3, line 200: It's hard for me to see what changed here. I think it'd be easier to see if you edited the old code directly, instead of commenting it. (I kinda hate seeing huge commented-out sections of code anywhere)

common/src/db/query.rs
```diff
@@ -188,6 +371,7 @@ impl<T: EntityTrait> From<Filter<T>> for ConditionExpression {
 // FromStr impls
 /////////////////////////////////////////////////////////////////////////
 
+/*
```
jcrossley3, line 374: I know you like this technique. I don't. :)

common/src/db/query.rs
```diff
@@ -420,18 +611,34 @@ mod tests {
...
-        assert!(Sort::<advisory::Entity>::from_str("location:desc").is_ok());
-        assert!(Sort::<advisory::Entity>::from_str("Location").is_ok());
-        assert!(Sort::<advisory::Entity>::from_str("Location:Asc").is_ok());
-        assert!(Sort::<advisory::Entity>::from_str("Location:Desc").is_ok());
+        assert!(Sort::from_str_with_entity::<advisory::Entity>("location").is_ok());
+        assert!(Sort::from_str_with_entity::<advisory::Entity>("location:asc").is_ok());
+        assert!(Sort::from_str_with_entity::<advisory::Entity>("location:desc").is_ok());
+        assert!(Sort::from_str_with_entity::<advisory::Entity>("Location").is_ok());
+        assert!(Sort::from_str_with_entity::<advisory::Entity>("Location:Asc").is_ok());
+        assert!(Sort::from_str_with_entity::<advisory::Entity>("Location:Desc").is_ok());
```
jcrossley3, line 619: To me, the new name is awkward, perhaps because it's replacing a `FromStr` impl? I can't decide if introducing some new type, e.g. `SortedEntity` with its own `from_str` is an improvement or not.

jcrossley3 (changes requested):
Clever idea, but I'm still fuzzy on how this will work with pagination. Won't this still be sorting each batch returned by the limiter instead of all the results matching the query?

Maybe a code snippet from the handler -- the caller of `filtering_with` -- interacting with the limiter would clear things up for me.

bobmcwhirter:
Ultimately, all this is doing is allowing referencing in sort/filter to columns outside of the `E::Column` set.

We should be able to sort/filter for any column involved in the query, not just those tied to a given Entity.

I don't suppose it'll interact poorly with pagination, as these columns still poke right into the SQL.

e.g.

```
SELECT advisory.location, CHARLEN(advisory.location) as location_len FROM advisory ORDER BY location_len
```

And normal offset/limit can still apply.

common/src/db/query.rs
```diff
@@ -126,77 +141,116 @@ pub enum Error {
...
+    }
 }
 
-struct Sort<T: EntityTrait> {
-    field: T::Column,
-    order: Order,
+impl IntoColumns for &Columns {
+    fn columns(&self) -> Columns {
+        (*self).clone()
+    }
```
jcrossley3, line 157: Why is this necessary?

jcrossley3:
@bobmcwhirter I have some concerns about excessive cloning. Because `Filter` and `Sort` aren't part of the public interface, I'm ok having them take a `&Columns` arg instead of an `IntoColumns` impl. Speaking of which, I think we can let that consume `self` to further simplify things.

I'm gonna approve this and push a subsequent PR along those lines.

jcrossley3 (approved):

Issue #383, opened 2024-06-06
Advisory search: sort by `average_severity` does not work


carlosthe19916:
## Some Context

Given the Advisories UI table (image below), I'd like to sort Advisories by Severity.

![Screenshot from 2024-06-06 13-40-02](https://github.com/trustification/trustify/assets/2582866/e4ef2fe6-e35b-48d1-a9ef-695fa9dd4057)

- The Advisory UI Table is populated by `GET /api/v1/advisory`. The JSON is similar to:

```json
{
    "items": [
        {
            "identifier": "CVE-2023-20862",
            "hashes": [
                "sha256:e6a2bc1084c77809af965dc4facb0c28aca67210f054eb8de4778190b8c6347e"
            ],
            "issuer": {
                "id": 2,
                "name": "Red Hat Product Security",
                "cpe_key": null,
                "website": null
            },
            "published": "2023-04-19T00:00:00Z",
            "modified": "2023-11-14T22:09:12Z",
            "title": "spring-security: Empty SecurityContext Is Not Properly Saved Upon Logout",
            "average_severity": "medium",... this is the field I want to SortBy
            "average_score": 6.3,
            "vulnerabilities": [
                {
                    "identifier": "CVE-2023-20862",
                    "non_normative": {
                        "non_normative": true,
                        "identifier": "CVE-2023-20862",
                        "title": "spring-security: Empty SecurityContext Is Not Properly Saved Upon Logout",
                        "discovered": "2023-07-31T00:00:00Z",
                        "released": "2023-04-19T00:00:00Z"
                    },
                    "severity": "medium",
                    "score": 6.3
                }
            ]
        },
    ],
    "total": 2595
}
```

## Here is the problem

When I try to search Advisories and order them by `average_severity` I get an error. The request I am executing is:

```
GET /api/v1/advisory?sort=average_severity:asc&limit=10&offset=0
```

And this is the exact error I get from the backend `400 Bad Request`:
```json
{
    "error": "Query error",
    "message": "query syntax error: Invalid field name for sort: 'average_severity'"
}
```

This might  be just me not knowing how to properly create the query for sorting, or it might be a real bug.

Could you help me with this please?

carlosthe19916:
On a side note. The query for sorting by `published` works so my queries are grammatically correct 

```
GET /api/v1/advisory?sort=published:asc&limit=10&offset=0
```

bobmcwhirter:
[...]

PR #399, opened 2024-06-11
Allow calculation of CVSS3 score using SQL functions, like a madman.

bobmcwhirter:
Provide enough escape-hatches around filterin/sorting to do my evil deeds. Do some evil deeds in SQL to allow sorting/filtering by synthetic `average_score` on advisories. Do even more evil, by writing entirely too many SQL functions. Lay in appropriate DOWN migration for cvss3 scoring functions.

### DIFF
--- a/modules/fundamental/src/advisory/service/mod.rs
+++ b/modules/fundamental/src/advisory/service/mod.rs
@@ -1,13 +1,19 @@
-use sea_orm::{ColumnTrait, EntityTrait, QueryFilter};
+use sea_orm::{
+    ColumnTrait, ColumnTypeTrait, EntityTrait, FromQueryResult, IntoIdentity, QueryFilter,
+    QuerySelect, QueryTrait,
+};
+use sea_query::{ColumnType, Func, FunctionCall, SimpleExpr};
+use time::OffsetDateTime;
+use uuid::Uuid;
 
 use crate::advisory::model::{AdvisoryDetails, AdvisorySummary};
 use crate::Error;
-use trustify_common::db::limiter::LimiterTrait;
-use trustify_common::db::query::{Filtering, Query};
+use trustify_common::db::limiter::LimiterAsModelTrait;
+use trustify_common::db::query::{Columns, Filtering, Query};
 use trustify_common::db::{Database, Transactional};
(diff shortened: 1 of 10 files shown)


bobmcwhirter:
@carlosthe19916 this should allow sorting by `average_score` (not severity, since that's just a string derived from the score).

Relates to https://github.com/trustification/trustify/issues/383

modules/fundamental/src/advisory/service/mod.rs
```diff
@@ -24,19 +30,79 @@ impl AdvisoryService {
...
-            paginated.offset,
-            paginated.limit,
-        );
+        let cvss3_score: FunctionCall = Func::cust("cvss3_score".into_identity());
+        let cvss3_score = cvss3_score.args([SimpleExpr::Custom("cvss3".to_string())]);
+
+        // To be able to ORDER or WHERE using a synthetic column, we must first
+        // SELECT col, extra_col FROM (SELECT col, random as extra_col FROM...)
+        // which involves mucking about inside the Select<E> to re-target from
+        // the original underlying table it expects the entity to live in.
```
jcrossley3, line 56: This is a little terrifying.

jcrossley3 (approved):
I feel a little dirty approving this. But I don't have any better ideas.

modules/fundamental/src/advisory/service/mod.rs
bobmcwhirter: Apparently it's normal! We so similar for `SELECT COUNT(*) FROM ( <original select> )` on the limiter stuff.

ctron:
So we are doing all this work just to avoid having an intermediate search model?

mrizzi:
Just a comment after having seen the PLSQL code.
Considering the efforts we're putting into having rustsec/CVSS library to fully cover CVSS 3.1 specs, i.e. https://github.com/rustsec/rustsec/pull/1198/, would be possible to calculate the score just once when the vectorString is ingested and then store the value in an column? too basic/not enough approach?

bobmcwhirter:
Marco: yes we should probably just store the computer score. 

Will still need the extra magic to sort and filter by the AVG() but it would reduce the work of the DB by some. 

I'll enhance.

PR #402, opened 2024-06-12
Look up documents before retrieving, improve id/key naming conventions

### LINKED ISSUE #394: GET /api/v1/advisory/urn:uuid:abcd/download does not work
Response from the backend 

```json
{
    "error": "Unsupported hash algorithm",
    "message": "Unsupported algorithm urn:uuid"
}
```

ctron:
Fix #394: Before downloading the actual content, we look up the sha256 digest from the database and use that to fetch the content from the storage. This is done for advisories, but also for SBOMs.

Also:
  * When uploading a document, return the `Id` of the database instead of some ID field from the document
  * Rename `HashOrUuidKey` to `Id`
  * Make it clearer what key/id means for the storage (`StorageKey`)

### DIFF
--- a/modules/fundamental/src/advisory/endpoints/mod.rs
+++ b/modules/fundamental/src/advisory/endpoints/mod.rs
@@ -140,12 +140,11 @@ pub async fn download(
     key: web::Path<String>,
 ) -> Result<impl Responder, Error> {
     let id = Id::from_str(&key).map_err(Error::HashKey)?;
-    let Some(hash_key) = advisory.fetch_advisory(id, ()).await?.and_then(|adv| {
-        adv.head
-            .hashes
-            .into_iter()
-            .find(|h| matches!(h, Id::Sha256(_)))
-    }) else {
+    let Some(hash_key) = advisory
+        .fetch_advisory(id, ())
+        .await?
+        .and_then(|adv| adv.head.find_sha256().cloned())
+    else {
         return Ok(HttpResponse::NotFound().finish());
     };
 
--- a/modules/fundamental/src/advisory/endpoints/mod.rs
+++ b/modules/fundamental/src/advisory/endpoints/mod.rs
@@ -139,8 +139,11 @@ pub async fn download(
     advisory: web::Data<AdvisoryService>,
     key: web::Path<String>,
 ) -> Result<impl Responder, Error> {
+    // the user requested id
     let id = Id::from_str(&key).map_err(Error::HashKey)?;
-    let Some(hash_key) = advisory
+
+    // look up document by id
+    let Some(id) = advisory
         .fetch_advisory(id, ())
         .await?
         .and_then(|adv| adv.head.find_sha256().cloned())
@@ -152,7 +155,7 @@ pub async fn download(
         .get_ref()
         .storage()
         .clone()
-        .retrieve(hash_key)
+        .retrieve(id.try_into()?)
         .await
         .map_err(Error::Storage)?
         .map(|stream| stream.map_err(Error::Storage));
--- a/modules/fundamental/src/advisory/endpoints/mod.rs
+++ b/modules/fundamental/src/advisory/endpoints/mod.rs
@@ -9,7 +9,7 @@ use std::str::FromStr;
 use tokio_util::io::ReaderStream;
 use trustify_common::db::query::Query;
 use trustify_common::db::Database;
-use trustify_common::hash::HashOrUuidKey;
+use trustify_common::id::Id;
 use trustify_common::model::Paginated;
 use trustify_module_ingestor::service::{Format, IngestorService};
 use trustify_module_storage::service::StorageBackend;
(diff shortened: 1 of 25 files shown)


ctron:
There's one catch. It relies on `fetch_advisory`, which is also limited and allows only retrieving by `Uuid` and `Sha256`. But I guess that another problem to fix.

bobmcwhirter (approved):

ctron:
The CI disagrees. I'll fix that tomorrow.

carlosthe19916 (approved):

carlosthe19916:
Tested and the REST API worked as expected, thanks!

modules/fundamental/src/advisory/endpoints/mod.rs
```diff
@@ -135,12 +135,26 @@ pub async fn upload(
...
+    };
+
+    let Some(hash) = adv.head.hashes.into_iter().find_map(|hash| match hash {
+        Id::Sha256(hash) => Some(hash),
+        _ => None,
+    }) else {
+        return Ok(HttpResponse::NotFound().finish());
+    };
+
+    let hash_key = Id::Sha256(hash);
```
jcrossley3, line 155: minor style nit. might warrant adding a `find_hash` method to `AdvisoryHead`

```suggestion
    let id = Id::from_str(&key).map_err(Error::HashKey)?;
    let Some(hash_key) = advisory.fetch_advisory(id, ()).await?.and_then(|adv| {
        adv.head
            .hashes
            .into_iter()
            .find(|h| matches!(h, Id::Sha256(_)))
    }) else {
        return Ok(HttpResponse::NotFound().finish());
    };
```

jcrossley3 (changes requested):

ctron:
I updated the PR and would value a second review.

The same issue was present for SBOMs. That should not also be fixed.

I also did refactor the storage part in a way that it is clear that there is a "storage key", which is opaque on the API. And there's a translation from the `Id` (and `Vec<Id>`). Which I is a better approach to the `find_sha256` function.

modules/fundamental/src/advisory/endpoints/test.rs
```diff
@@ -416,9 +417,9 @@ async fn upload_default_csaf_format(ctx: TrustifyContext) -> Result<(), anyhow::
         .to_request();
 
     let response = actix_web::test::call_service(&app, request).await;
-    let doc: Value = actix_web::test::read_body_json(response).await;
-    log::debug!("{doc}");
-    assert_eq!(doc.as_str(), Some("CVE-2023-33201"));
+    let id: Id = actix_web::test::read_body_json(response).await;
+    log::debug!("{id}");
+    assert!(matches!(id, Id::Uuid(_)));
```
jcrossley3, line 422: As a user, I took some comfort seeing the identifier of the doc I posted returned upon success. The opaque uuid gives me less comfort. Not for this PR, but do you think it's worth returning a struct with `uuid` and `identifier` fields upon success?

modules/fundamental/src/advisory/endpoints/mod.rs
```diff
@@ -135,16 +135,23 @@ pub async fn upload(
...
+        return Ok(HttpResponse::NotFound().finish());
+    };
 
-    let stream = service
+    let stream = ingestor
         .get_ref()
         .storage()
         .clone()
-        .retrieve(hash_key)
+        .retrieve(advisory.head.hashes.try_into()?)
```
jcrossley3, line 154: much cleaner!

modules/storage/src/service/fs.rs
```diff
@@ -94,7 +93,9 @@ impl StorageBackend for FileSystemBackend {
         // set up reader
 
         let mut stream = pin!(stream);
-        let mut digest = Sha256::new();
+        let mut sha256 = Sha256::new();
+        let mut sha384 = Sha384::new();
+        let mut sha512 = Sha512::new();
```
jcrossley3, line 98: I don't quite understand why storage needs to calculate all these for every write. It's only gonna use one of them, correct?

jcrossley3 (commented):
I like the `StorageKey`

modules/storage/src/service/fs.rs
ctron: Correct. But it returns them after the insert. My idea was to refactor in another step, passing those into the loaders, so that they might be able to use that info. Right now, they do this during the read. Which also works.

ctron:
@jcrossley3 unless you think the PR is getting too big, I could amend it tomorrow, adding the struct and trying to pass through the list of hashes.

jcrossley3:
> @jcrossley3 unless you think the PR is getting too big, I could amend it tomorrow, adding the struct and trying to pass through the list of hashes.

I do think it's getting pretty big. Can you change the title to be more descriptive and move the "Fix #394" into the description before merging?

ctron:
I continued work in https://github.com/trustification/trustify/pull/405

PR #405, opened 2024-06-14
Further refactor digest and ID handling

ctron:
This builds upon #402 and tries to improve the following:

* Re-use `Digests` and remove duplication
* Collect digests during storage and re-use when inserting document into the DB


ctron:
@jcrossley3 So I continue here to improve things.

ctron:
Right now the CI is expected to fail as refactoring the tests will be more effort. I'll do that if we can agree that this is the right approach.

ctron:
Pushing another change which implements the result struct suggested in #402

jcrossley3 (approved):

Issue #401, opened 2024-06-12
Advisory search: Filter by average_severity does not work


carlosthe19916:
We need to be able to filter our advisories based on their severity: "none,medium, high, critical"

the query could be something like:
```
average_severity=medium|high
```

bobmcwhirter:
@carlosthe19916 should be merged if you'd like to verify and close this issue.

carlosthe19916:
Thanks @bobmcwhirter . I verified, this has been solved in the main branch and it is working as expected

PR #406, opened 2024-06-14
Add a cvss3_severity enum (none, low, medium, high, critical)

bobmcwhirter:
Add a cvss3.severity column with that enum.
Calculate it in a migration.
Adjust Query API to support enums (Value->ValueOrSimpleExpr for envalue()) Add SQL function to calculate severity from score (for non-stored avg(score)) Add average_severity to the query.

### DIFF
--- a/common/src/db/query.rs
+++ b/common/src/db/query.rs
@@ -4,8 +4,8 @@ use sea_orm::entity::ColumnDef;
 use sea_orm::sea_query::{extension::postgres::PgExpr, ConditionExpression, IntoCondition};
 use sea_orm::{
     sea_query, ColumnTrait, ColumnType, Condition, EntityName, EntityTrait, Iden, IntoIdentity,
-    Iterable, Order, PrimaryKeyToColumn, QueryFilter, QueryOrder, QuerySelect, QueryTrait, Select,
-    Value,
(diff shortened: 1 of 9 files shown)


common/src/db/query.rs
```diff
@@ -403,9 +400,24 @@ impl FromStr for Operator {
 // Internal helpers
 /////////////////////////////////////////////////////////////////////////
 
+#[derive(Debug)]
+enum ValueOrSimpleExpr {
```
jcrossley3, line 404: I don't love the name, but it's private, so whatevs. I'd prefer something shorter, e.g. `V`, `Rhs`, `Expr`. Maybe `Expr::Value` and `Expr::Simple`? I dunno. If you like it as it is, I'm good.

jcrossley3 (approved):

common/src/db/query.rs
bobmcwhirter: I'm preferring verbosity these days personally. 'Expr' I'd find ambiguous with SeaORM's own Expr, 'V' feels terse, 'Rhs' maybe, but... eh?

bobmcwhirter:
@carlosthe19916 should fix https://github.com/trustification/trustify/issues/401

PR #409, opened 2024-06-17
Enable pg_stat_statements postgres extension when in pm-mode

JimFuller-RedHat:
Sets embedded postgres config to preload **pg_stat_statements** lib
```
  shared_preload_libraries=pg_stat_statements
```
so we can invoke
```
  CREATE EXTENSION IF NOT EXISTS pg_stat_statements;
```
when _pm-mode_ is engaged, which installs **pg_stat_statements** into **trustify** embedded pg database.

This is a minimal first effort - in the future we may want to do more - such as pass in env var to set pg configs, etc ... though as _pm-mode_ is purely developmental I stopped there.

**Note**: Bumped **postgresql_embedded**=0.10.1, **postgresql_archive**=0.10.1

### DIFF
--- a/common/src/db/mod.rs
+++ b/common/src/db/mod.rs
@@ -173,17 +173,18 @@ impl Database {
                 format!("DROP DATABASE IF EXISTS \"{}\";", database.name),
             ))
             .await?;
-
         let create_db_result = db
             .execute(Statement::from_string(
                 db.get_database_backend(),
                 format!("CREATE DATABASE \"{}\";", database.name),
             ))
             .await?;
-
         db.close().await?;
 
         let db = Self::new(database).await?;
+        let install_pg_stat_statements_ext_db_result = db
+            .execute_unprepared("CREATE EXTENSION IF NOT EXISTS \"pg_stat_statements\";")
+            .await?;
         db.migrate().await?;
 
         Ok(db)
--- a/trustd/src/db.rs
+++ b/trustd/src/db.rs
@@ -1,4 +1,5 @@
 use postgresql_embedded::PostgreSQL;
+use std::collections::HashMap;
 use std::env;
 use std::fs::create_dir_all;
 use std::process::ExitCode;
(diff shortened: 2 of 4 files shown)


common/src/db/mod.rs
```diff
@@ -167,6 +167,15 @@ impl Database {
...
 
+        let install_extension_db_result = db
+            .execute(Statement::from_string(
+                db.get_database_backend(),
+                format!(
+                    "CREATE EXTENSION IF NOT EXISTS \"{}\";",
+                    "pg_stat_statements"
+                ),
+            ))
+            .await;
```
ctron, line 178: I think this misses a `?` at the end. Which expands into something like:

```rust
let result = match result {
  Ok(result) => result,
  Err(err) => return Err(err),
}
```

Which is a little bit like "ok or throw". But only works in function which return a `Result`. Also the error of the expression must implement `Into` to the error type of the function's result.

When not using the result (not assigning it), I would expect `cargo check` to warn about this. In this case it looks to me as if it gets assigned (which eliminates the warning) but then doesn't evaluate the result. If you don't need to result value, you can just use `?` and don't assign it. That also counts as "used".

trustd/src/db.rs
```diff
@@ -67,12 +68,17 @@ impl Run {
...
+            "shared_preload_libraries".to_string(),
+            "pg_stat_statements".to_string(),
+        )]);
         let settings = postgresql_embedded::Settings {
             username: self.database.username.clone(),
             password: self.database.password.clone(),
             temporary: false,
             installation_dir: db_dir.clone(),
             timeout: Some(Duration::from_secs(30)),
+            configuration: db_conf,
```
ctron, line 81: In a case like that, you might want to consider using the same name for the variable as for the field. In which case you can collapse this into:

```rust
let configuration = ();
let settings = Settings {
  configuration,
};
```

I am not sure it makes sense here. Just a note.

ctron (changes requested):
Looks good to me, with one (possible) issue on the result of the execute call. But maybe I got that wrong.

trustd/src/db.rs
JimFuller-RedHat: I originally did that but lint complained ... will investigate again.

common/src/db/mod.rs
JimFuller-RedHat: ah I probably missed off, thx!

trustd/src/db.rs
JimFuller-RedHat: works now, thx

ctron (approved):

Issue #284, opened 2024-05-14
SBOM: get application "version" and spec "spdx|cyclonedx"


carlosthe19916:
- Trustification has a way of identifying the version of the application from which the SBOM was taken. See image below (the image belongs to Trustification):

![Screenshot from 2024-05-14 12-47-19](https://github.com/trustification/trustify/assets/2582866/76a5332b-2020-4a98-9bd2-ef294f0a715c)

- On the other hand, it would be beneficial to also have on a field the spec format to which the SBOM belongs to. e.g. spdx|cyclonedx

ctron:
I think what would also make sense is to somehow have the "namespace" (unique ID). As an SBOM may have the same name/version, it might be hard to distinguish those. However, we might decide to not show it as prominent as the other fields.

On the other side, it might be an alternative to group "by name", show only the most recent, and provide the user with a sub-query to see older versions too.

ctron:
Closed by https://github.com/trustification/trustify/issues/414

carlosthe19916:
@ctron this issue was closed as "done". which are the fields that are supposed to be used in for rendering the "version" of an SBOM? I can only see a field `described_by.version` and `described_by` is an array; Are you saying a single SBOM can have multiple versions? 

![image](https://github.com/trustification/trustify/assets/2582866/0801c153-c02c-4215-94a6-31667e43f60d)

ctron:
An SBOM doesn't always have a version. Not to be confused with the "version of the SBOM" or the "spec version of the SBOM".

CycloneDX can have a single defining component, which may have a version. SPDX has zero or more "describing" packages, which don't have a version, but which can have an external reference (like PURLs) which may have a version. That's also why some of our SBOMs have a version as part of the SBOM name.

Fun :)

I don't think the backend should make any modifications to the data when returning it. Just return what is known.

The UI can make this more user friendly on some overview screens (like tables). But should provide unaltered detail in more detailed screens.

So what could be done is to extract a single version from that response. If there's none, then there's none. If there's one, the situation is simple. If there are more than one, then the UI needs to get a little bit creating. I could imagine to show multiple, and then stop at some point with "and 5 more". Which will rarely be the case I guess.

Issue #414, opened 2024-06-17
Fetching all SBOMs actually returns all nodes of the SBOM


ctron:
That's due to a wrongly used relationship definition. As there can only be one relationship, we need to different way. SeaORM seems to solves this via "links".

I have a local fix for this, but want to tackle this in combination with #284.

ctron:
Closed by #424

PR #419, opened 2024-06-18
Add SBOM version, fix SBOM listing

ctron:
The main motivation of this PR is to fix issue #284, adding an SBOM version to the response.

For CycloneDX there can be a dedicated version of the main component of the SBOM. For SPDX, there can exist one or more packages describing the SBOM, which may include the version information.

So the information which can be provides is a "described_by" set of packages per SBOM, which contains CPEs, PURLs, a name and an optional version information. However, this set can also be empty.

During work in this issue I discovered another one (#414), where instead of returning all SBOMs, the endpoint returned all SBOM nodes. This is also fixed as part of this PR.

### DIFF
--- a/migration/src/m0000250_create_sbom_package.rs
+++ b/migration/src/m0000250_create_sbom_package.rs
@@ -14,6 +14,7 @@ impl MigrationTrait for Migration {
                     .if_not_exists()
                     .col(ColumnDef::new(SbomPackage::SbomId).uuid().not_null())
                     .col(ColumnDef::new(SbomPackage::NodeId).string().not_null())
+                    .col(ColumnDef::new(SbomPackage::Version).string())
                     .primary_key(
                         Index::create()
                             .col(SbomPackage::SbomId)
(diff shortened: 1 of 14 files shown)


ctron:
cc @carlosthe19916

migration/src/m0000250_create_sbom_package.rs
```diff
@@ -14,6 +14,7 @@ impl MigrationTrait for Migration {
                     .if_not_exists()
                     .col(ColumnDef::new(SbomPackage::SbomId).uuid().not_null())
                     .col(ColumnDef::new(SbomPackage::NodeId).string().not_null())
+                    .col(ColumnDef::new(SbomPackage::Version).string())
```
jcrossley3, line 17: Please create a new migration for this.

jcrossley3 (changes requested):

ctron:
Looks like this got merged in via #424. I'll create a follow up PR.

PR #424, opened 2024-06-18
feat: search SBOMs by PURL or package id

### LINKED ISSUE #413: Given a certain Package we should be able to get list of SBOMs contain that specif package
Given package A, I need to have the list of SBOMs that contain that package

This will support the use cases defined in the old Trustification

image

ctron:
Closes #413 

---

This builds on ~~#423~~ and #419 and should only be merged after those.

### DIFF
--- a/migration/src/m0000250_create_sbom_package.rs
+++ b/migration/src/m0000250_create_sbom_package.rs
@@ -14,6 +14,7 @@ impl MigrationTrait for Migration {
                     .if_not_exists()
                     .col(ColumnDef::new(SbomPackage::SbomId).uuid().not_null())
                     .col(ColumnDef::new(SbomPackage::NodeId).string().not_null())
+                    .col(ColumnDef::new(SbomPackage::Version).string())
                     .primary_key(
                         Index::create()
                             .col(SbomPackage::SbomId)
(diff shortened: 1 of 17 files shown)


carlosthe19916 (approved):
May I suggest considering renaming "related"?. The first time I saw it I asked "related to what?" it is not easy to relate the path to what actually the endpoint does.

Some suggestions:

- /api/v1/sbom/by-package
- /api/v1/sbom-by-package

Feel free to choose a better naming.

On the other hand there is also another endpoint /api/v1/sbom/{key}/related . We are using again the term `related` and it is not clear the purpose of the endpoint


Screenshot from 2024-06-19 09-02-00

ctron:
> May I suggest considering renaming "related"?. 

I wasn't happy with this either. I do like `/api/v1/sbom/by-package`, so I went with that.

migration/src/m0000250_create_sbom_package.rs
```diff
@@ -14,6 +14,7 @@ impl MigrationTrait for Migration {
                     .if_not_exists()
                     .col(ColumnDef::new(SbomPackage::SbomId).uuid().not_null())
                     .col(ColumnDef::new(SbomPackage::NodeId).string().not_null())
+                    .col(ColumnDef::new(SbomPackage::Version).string())
```
jcrossley3, line 17: Please create new migrations for things like this going forward.

PR #429, opened 2024-06-18
test: ensuring the embedded UI works

### LINKED ISSUE #426: Add a test ensuring the embedded UI works
There should be a test which ensures that the embedded UI can be opened (code = 2xx) without any authentication. It is not important that the UI is rendered or works. It should just check that it is served.

helio-frota:


modules/ui/src/endpoints.rs
```diff
@@ -5,3 +5,33 @@ use trustify_ui::{trustify_ui, UI};
...
+            version: "".to_string(),
+            auth_required: "".to_string(),
+            oidc_server_url: "".to_string(),
+            oidc_client_id: "".to_string(),
+            oidc_scope: "".to_string(),
+            analytics_enabled: "".to_string(),
+            analytics_write_key: "".to_string(),
+        };
+
+        let app = test::init_service(App::new().configure(|cfg| configure(cfg, &test_ui))).await;
```
bobmcwhirter, line 26: Hm, I suspect this may still miss if someone adds endpoints differently. Sorta needs to be an "integration test" I suspect, hitting actual HTTP from the actual running server.

bobmcwhirter (changes requested):

modules/ui/src/endpoints.rs
helio-frota: the issue says:

>  It is not important that the UI is rendered or works. It should just check that it is served.

> hitting actual HTTP from the actual running server.

I don't disagree but that is other issue then ? :+1:

helio-frota: >  I suspect this may still miss if someone adds endpoints differently. 

`let req = test::TestRequest::get().uri("/oooo").to_request();`

yeah it tests almost nothing 

`PASS [   0.010s] trustify-module-ui endpoints::test::test_ui_get`

helio-frota: better to close this :+1:

PR #431, opened 2024-06-18
docs: document the SBOM structure in the database

ctron:


docs/design/sbom.md
```diff
@@ -0,0 +1,84 @@
...
+features of CycloneDX. In the database, this is being represented by the following structure:
+
+```mermaid
+erDiagram
+    Sbom {
+        int sbom_id PK "inherited"
+        string id "inherited"
+        string name "inherited"
+        string namespace "inherited"
+        uuid sha256
```
jcrossley3, line 28: Seems like a lot of id's: sbom_id, id, and sha256. There's also a node_id and document_id in the entity. I'm confused.

ctron:
I updated the documentation. Indeed those IDs did not reflect that actual state of the DB structure. I hope it's better now.

jcrossley3 (approved):

PR #432, opened 2024-06-18
feat: support IS [NOT] NULL in queries

jcrossley3:
```
foo=null  ---> WHERE "foo" IS NULL
foo!=null ---> WHERE "foo" IS NOT NULL
foo=NuLl  ---> WHERE "foo" IS NULL
```


bobmcwhirter (approved):
NiceOrExcellent!

ctron:
I've the feeling we are repeating a mistake here: https://www.wired.com/story/null-license-plate-landed-one-hacker-ticket-hell/

jcrossley3:
> NiceOrExcellent!

Haha. Fwiw, it was one of your recent commit messages that inspired the new name.

PR #433, opened 2024-06-18
test: ensuring the embedded UI works

helio-frota:
I'm running PM-mode just to have access to DB.
And then running devmode with no auth and a different port and doing the basic requests

### DIFF
--- a/trustd/tests/tests.rs
+++ b/trustd/tests/tests.rs
@@ -1,0 +1,103 @@
+use regex::Regex;
+use std::process::Stdio;
+use std::{env, path::Path};
+use std::{fs, thread, time};
+use tokio::io::{AsyncBufReadExt, BufReader};
+use tokio::process::Command;
+
+#[tokio::test]
+async fn test_embedded_ui() {
+    // This test creates a 'trustify/trustd/.trustify' directory, so we need to remove it.
+    let trustify_dir = Path::new(".trustify");
+    if trustify_dir.exists() {
+        fs::remove_dir_all(trustify_dir).expect("Failed to remove '.trustify'");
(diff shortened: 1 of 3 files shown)


bobmcwhirter (approved):
This is awesome! Maybe make a test Context to make testing against a running instance in the future easier also?

helio-frota:
thanks!

> Maybe make a test Context to make testing against a running instance in the future easier also?

no idea if that is easy but sounds like a great idea :+1:

helio-frota:
pm-mode-only will be a bit easier than running 2 instances...(thinking on testContext)

the second trustd is running after (not in parallel with) pm-mode to avoid this error constraint "pg_type_typname_nsp_index" https://stackoverflow.com/questions/54351783/duplicate-key-value-violates-unique-constraint-postgres-error-when-trying-to-c

trustd/tests/tests.rs
```diff
@@ -0,0 +1,103 @@
...
+                .arg("--db-password")
+                .arg("trustify")
+                .arg("--db-port")
+                .arg(&pg_port)
+                .stdout(Stdio::piped())
+                .spawn()
+                .expect("Failed to start trustd devmode.");
+
+            // Wait for availability
+            let seconds = time::Duration::from_secs(15);
```
bobmcwhirter, line 68: Re-reviewing... can we replace this arbitrary sleep() with watching stdio or something to truck along as soon as the service is up?

Alternatively, instead of forking and executing, can we run main() or similar directly from within the test?

helio-frota: good, thanks I'll try :+1:

PR #439, opened 2024-06-19
feat: add filtering/query to "find by sbom package"

### LINKED ISSUE #438: /api/v1/sbom/by-package should do full pagination and filtering on the backend or let it to be done on the client
Sorry, I missed this point when I reviewed the original PR that introduced this change.

The current endpoint does:
- pagination (limit, offset)
- But does not filtering (there is no `q` query parameter). Let's say I want to filter all sboms whose name contain quarkus `/api/v1/sbom/by-package?q=quarkus&id=package_id`

I think pagination and filtering come as a package: ALL or NONE. If the server does pagination, then it should also do filtering. Otherwise, it should not do neither pagination nor filtering and let it to be done in the client side.

The decision of where to do filtering/pagination should depend on how many SBOMs we expect for a single package. IMHO it should be less than thousands, and pagination/filtering could be done in the client side. But I am open to disagreement.

Should we:
-  Add filtering to the backend?
- Or should we remove pagination (offset, limit) from the current endpoint and let it be done by the client?

image

ctron:
Closes #438

### DIFF
--- a/modules/fundamental/src/sbom/endpoints/mod.rs
+++ b/modules/fundamental/src/sbom/endpoints/mod.rs
@@ -99,6 +99,7 @@ struct AllRelatedQuery {
     tag = "sbom",
     context_path = "/api",
     params(
+        Query,
         Paginated,
         AllRelatedQuery,
     ),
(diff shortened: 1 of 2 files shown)


carlosthe19916 (approved):
Thanks!!!

jcrossley3 (approved):

Issue #422, opened 2024-06-18
UI broken since GraphQL was merged


ctron:
Merging #360 broke the embedded UI. There's a handler registered for `/` now, which requires authorization and seems to divert all traffic to the graphql endpoint.

ctron:
Should be closed by https://github.com/trustification/trustify/pull/423

PR #440, opened 2024-06-19
Ignoring embedded UI test

jcrossley3:
This test has problems:

- it takes too long (it never completes in my local env)
- it leaves .trustify dirs behind
- it shells out to a binary
- it uses a hardcoded port
- it makes requests of URL's that don't exist
- it's not clear that it affirms the fix for #422 

I recommend it be moved beneath `server/` and refactored to use TestRequest.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,4 +35,4 @@ jobs:
       - name: Test
         run: cargo test -- --nocapture
         env:
-          RUST_LOG: info
+          RUST_LOG: info,sqlx=error,sea_orm=error
--- a/common/src/db/mod.rs
+++ b/common/src/db/mod.rs
@@ -127,7 +127,7 @@ impl Database {
         let port = database.port;
         let db_name = &database.name;
         let url = format!("postgres://{username}:{password}@{host}:{port}/{db_name}");
-        log::info!("connect to {}", url);
+        log::debug!("connect to {}", url);
 
         let mut opt = ConnectOptions::new(url);
         opt.min_connections(16);
--- a/integration-tests/src/sbom/test/cyclonedx.rs
+++ b/integration-tests/src/sbom/test/cyclonedx.rs
@@ -42,7 +42,7 @@ async fn test_parse_cyclonedx(ctx: TrustifyContext) -> Result<(), anyhow::Error>
                 )
                 .await?;
 
-            log::info!("{:?}", packages);
+            log::debug!("{:?}", packages);
 
             assert_eq!(41, packages.total);
 
--- a/integration-tests/src/sbom/test/perf.rs
+++ b/integration-tests/src/sbom/test/perf.rs
@@ -20,7 +20,7 @@ async fn ingest_spdx_medium(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
                 .describes_packages(sbom.sbom.sbom_id, Default::default(), ())
                 .await?;
 
-            log::info!("{:#?}", described);
+            log::debug!("{:#?}", described);
             assert_eq!(1, described.items.len());
             assert_eq!(
                 described.items[0],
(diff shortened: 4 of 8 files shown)


helio-frota:
* it takes too long :heavy_check_mark:  -- true -- (it never completes in my local env)  -- That is probably due the arbitrary time then
* it leaves .trustify dirs behind -- :man_facepalming: probably needs another delete at the end
* it shells out to a binary :heavy_check_mark:
* it uses a hardcoded port :heavy_check_mark:  yeah not the best option in the planet but we need something diff than 8080
* it makes requests of URL's that don't exist -- I'm not following that part

* it's not clear that it affirms the fix for https://github.com/trustification/trustify/issues/422 :heavy_check_mark: -- yeah the test was based on this https://github.com/trustification/trustify/issues/426 issue

helio-frota (approved):

helio-frota:
> I recommend it be moved beneath server/ and refactored to use TestRequest.

That is a mock server 
https://github.com/helio-frota/trustify/blob/dea876a7f0d838ea121a4e112156bd7de29ea5ee/modules/ui/src/endpoints.rs#L28

I think I'm not using TestRequest correctly then, because it says `/foobar` returns status-code `200`
https://github.com/trustification/trustify/pull/429/files#r1644637254

jcrossley3:
> * it leaves .trustify dirs behind -- 🤦‍♂️ probably needs another delete at the end

No. You must ensure it leaves no turds behind, whether it fails or passes, which is harder to do than it sounds. Better to design your tests so that no filesystem is required, if possible.

> * it makes requests of URL's that don't exist -- I'm not following that part

Try it with curl/wget/httpie in a shell. Try all the URL's in your tests: 

- http://localhost:8080/sboms
- http://localhost:8080/packages
- http://localhost:8080/importers
- http://localhost:8080/vulnerabilities
- http://localhost:8080/foobar
- http://localhost:8080/oooo

You'll see that all of those url's all return the exact same thing: a blob of javascript.

But if you try these, you'll get different results:

- http://localhost:8080/api/v1/advisory
- http://localhost:8080/graphql/

And this is due to the way we [configure our routes](https://github.com/trustification/trustify/blob/773e957153100ddfea186a1b2e00b7be47054749/server/src/lib.rs#L264-L301), which is critically important knowledge for you as this test's author.

> * it's not clear that it affirms the fix for [UI broken since GraphQL was merged #422](https://github.com/trustification/trustify/issues/422) ✔️ -- yeah the test was based on this [Add a test ensuring the embedded UI works #426](https://github.com/trustification/trustify/issues/426) issue

I'm pretty sure #426 fell out of #422, but you can confirm that with @ctron

jcrossley3:
> Updated: ~That is a mock server~ https://github.com/helio-frota/trustify/blob/dea876a7f0d838ea121a4e112156bd7de29ea5ee/modules/ui/src/endpoints.rs#L28

I know. And it grok's Actix route configuration. You should be able to affirm our configuration with it.

> I think I'm not using TestRequest correctly then, because it says `/foobar` returns status-code `200` https://github.com/trustification/trustify/pull/429/files#r1644637254

See my above comment to understand why that is.

Do this: create a failing test with `TestRequest`. Revert #423 in your local environment and convince yourself that the bug exists, both by manually trying it in a browser and a shell, and then automatically with a unit test using `TestRequest`. Once you have that failing, restore the fix (#423) in your local environment. Your test should now pass. If it doesn't, you did something wrong.

helio-frota:
```
You'll see that all of those url's all return the exact same thing: a blob of javascript.

But if you try these, you'll get different results:

http://localhost:8080/api/v1/advisory
http://localhost:8080/graphql/
And this is due to the way we [configure our routes](https://github.com/trustification/trustify/blob/773e957153100ddfea186a1b2e00b7be47054749/server/src/lib.rs#L264-L301), which is critically important knowledge for you as this test's author.
```

When I started the test I was thinking (and continued) on the literal part `embedded UI works` the ui-crate. 

thanks for the extra info :+1:

Issue #376, opened 2024-06-05
To open the Swagger UI: /openapi/ works but /openapi does not work.


carlosthe19916:
- http://localhost:8080/openapi/ WORKS
- http://localhost:8080/openapi DOES NOT WORK (notice the slash character is not at the end of the url)

PR #441, opened 2024-06-20
fix: UI and APIs: Remove trailing slash from path

gildub:
Remove "/" for service definitions which was added as a suffix to involved endpoints.

Fix https://github.com/trustification/trustify/issues/376
Fix https://github.com/trustification/trustify/issues/422

### DIFF
--- a/modules/graphql/src/endpoints.rs
+++ b/modules/graphql/src/endpoints.rs
@@ -20,12 +20,12 @@ pub fn configure(svc: &mut web::ServiceConfig, db: Database, graph: Arc<Graph>) 
         .finish();
 
     svc.service(
-        web::resource("/")
+        web::resource("")
             .guard(guard::Post())
             .to(GraphQL::new(schema.clone())),
     );
 }
 
 pub fn configure_graphiql(svc: &mut web::ServiceConfig) {
-    svc.service(web::resource("/").guard(guard::Get()).to(index_graphiql));
+    svc.service(web::resource("").guard(guard::Get()).to(index_graphiql));
 }
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -279,27 +279,26 @@ impl InitData {
                                 trustify_module_graphql::endpoints::configure_graphiql(svc);
                             }),
                     );
-                    svc.app_data(web::Data::from(self.graph.clone())).service(
-                        web::scope("/api")
-                            .wrap(new_auth(self.authenticator.clone()))
-                            .configure(|svc| {
-                                trustify_module_importer::endpoints::configure(svc, db.clone());
+                    svc.app_data(web::Data::from(self.graph.clone()))
+                        .service(
+                            web::scope("/api")
+                                .wrap(new_auth(self.authenticator.clone()))
+                                .configure(|svc| {
+                                    trustify_module_importer::endpoints::configure(svc, db.clone());
 
-                                trustify_module_fundamental::endpoints::configure(
-                                    svc,
-                                    db.clone(),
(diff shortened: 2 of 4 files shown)


server/src/lib.rs
```diff
@@ -279,26 +279,33 @@ impl InitData {
...
+                                trustify_module_importer::endpoints::configure(svc, db.clone());
+
+                                trustify_module_fundamental::endpoints::configure(
+                                    svc,
+                                    db.clone(),
+                                    storage.clone(),
+                                );
+                                trustify_module_ui::endpoints::configure(svc, &self.ui);
+                            }),
+                    );
```
jcrossley3, line 308: Can you explain what your intent is here? Configuring the same endpoints within two different scopes seems wrong.

jcrossley3 (changes requested):

server/src/lib.rs
gildub: @jcrossley3, sorry I just simplified that. Please re-review.

gildub: @jcrossley3, the intent should be clearer from the latest.
I believe the embedded UI should be clearly distinct from any API services the backend is providing.

server/src/lib.rs
```diff
@@ -279,26 +279,27 @@ impl InitData {
...
+                                trustify_module_fundamental::endpoints::configure(
+                                    svc,
+                                    db.clone(),
+                                    storage.clone(),
+                                );
+                            }),
+                    );
+                    svc.app_data(web::Data::from(self.graph.clone())).service(
+                        web::scope("")
+                            .wrap(new_auth(self.authenticator.clone()))
```
jcrossley3, line 297: What problem are you addressing here? Why wrap the ui endpoints with auth that wasn't there before? What does that have to do with trailing slashes?

jcrossley3 (changes requested):
Can you include "Fixes #376" in the PR description so that it will be closed when this PR merges, please?

server/src/lib.rs
gildub: @jcrossley3, the trailing slash is involved in  `trustify_module_ui::endpoints::configure()` and part of the confusion (at least for me) is that the UI is currently defined (in main) as part of https://github.com/trustification/trustify/blob/main/server/src/lib.rs#L282.

jcrossley3: What are you fixing? What specific URL breaks before this change and works after it?

gildub: @jcrossley3, this is a reorganization fix I thought obvious. I can remove it.

jcrossley3: Forgive my stupidity. But I still want to know what you were fixing. It wasn't obvious to me.

gildub: @jcrossley3, please don't be stupid ! ;) 
It's me who is not capable of explaining in plain words what I thought made sense and after all it might be out of scope. So I'll try in a separate PR....

modules/graphql/src/endpoints.rs
```diff
@@ -20,12 +20,12 @@ pub fn configure(svc: &mut web::ServiceConfig, db: Database, graph: Arc<Graph>)
         .finish();
 
     svc.service(
-        web::resource("/")
+        web::resource("")
```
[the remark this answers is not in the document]
gildub: @jcrossley3, because we're using [web::scope](https://actix.rs/docs/url-dispatch#scoping-routes) it behaves as a prefix before the web::resource. That's actually the issue this PR is all about.

gildub:
https://github.com/trustification/trustify/pull/441

Issue #273, opened 2024-05-13
Can we get rid of the "location" parameter?


jcrossley3:
"remove location" I think should be a separate PR. Here, the issuer is *optional* and only used for OSV where that information is not necessarily derivable from the contents.

_Originally posted by @bobmcwhirter in https://github.com/trustification/trustify/pull/264#discussion_r1598774977_

ctron:
The idea of the "location" was to be able to trace where this document came from. With the idea of being able to delete all related documents when the source gets deleted.

I think this relates to the question: how do we handle the case when two different users upload the same SBOM/advisory, and those users should be isolated?

bobmcwhirter:
We did, yeah?

jcrossley3:
> We did, yeah?

No. It's still hard-coded for advisories and required for SBOM's. We should agree on its value. Are the use cases @ctron cites above real?

ctron:
I would say https://github.com/trustification/trustify/pull/442 closes this one.

Issue #160, opened 2024-04-15
For relevant documents, remove `location` as a defining part of a composite key


bobmcwhirter:
Just SHAs.

ctron:
What if the same document comes in from two different sources? For example: company A publishes an SBOM, and company B re-publishes that as part of their own solution. How could we distinguish those? Assuming I would want to clean up SBOMs from company B, how could I find out what came in from them?

ctron:
Because I just noticed this on the advisories: All functionality seems to work with an "identifier". Which seems to be the "document id" (e.g. RHSA…), and deriving information on that.

Dropping the digest from this leads to the ability of one source overriding statements of the other source.

Keeping the digest, seems to lead to the result that we have a lot of statements that are no longer relevant.

I believe that keeping the source/location as part of the document ID  would solve both cases.

bobmcwhirter:
Keep only SHA not dropping the digest.

bobmcwhirter:
Lemme drink some coffee and think some more

bobmcwhirter:
Yeah, so I think always keep the SHA digest. Either using it as the PK for all related statements or the current auto-increment PK (which is tied 1:1 to the SHA).

I do agree knowing the 'source' of some document could be useful, but may also be optional, in terms of "where the document was sitting prior to ingesting".

But also knowing the `source` of the document from it's... internal data? externally-applied signatures?

e.g., regardless *where in terms of location* the document came from (MySpace, USB drive), if it's signed by Red Hat, it's a Red Hat-sourced document.

ctron:
> e.g., regardless _where in terms of location_ the document came from (MySpace, USB drive), if it's signed by Red Hat, it's a Red Hat-sourced document.

I think we need a lot more clarity around our management of documents. When I delete a source (e.g. "CSAFs from Acme Inc"), how can I remove them from my inventory? If they published a Red Hat SBOM as part of their product, will that delete the Red Hat SBOM which came in from the RH source?

PR #442, opened 2024-06-20
feat: implements labels using jsonb

ctron:
Closes https://github.com/trustification/trustify/issues/273
Closes https://github.com/trustification/trustify/issues/160


migration/src/m0000030_create_sbom.rs
```diff
@@ -26,11 +26,11 @@ impl MigrationTrait for Migration {
                     .if_not_exists()
                     .col(ColumnDef::new(Sbom::SbomId).uuid().not_null())
                     .col(ColumnDef::new(Sbom::NodeId).string().not_null())
-                    .col(ColumnDef::new(Sbom::Location).string().not_null())
                     .col(ColumnDef::new(Sbom::DocumentId).string().not_null())
                     .col(ColumnDef::new(Sbom::Sha256).string().not_null())
                     .col(ColumnDef::new(Sbom::Published).timestamp_with_time_zone())
                     .col(ColumnDef::new(Sbom::Authors).array(ColumnType::String(None)))
+                    .col(ColumnDef::new(Sbom::Labels).json_binary().not_null())
```
jcrossley3, line 33: I think we're squarely in the phase of "no changes to existing migrations if possible". Instead of changing two migrations, can you create a new one with the necessary column changes to both tables?

jcrossley3 (changes requested):

ctron:
@jcrossley3 added the migration and left a gap for https://github.com/trustification/trustify/pull/444

migration/src/m0000030_create_sbom.rs
```diff
@@ -26,7 +26,14 @@ impl MigrationTrait for Migration {
                     .if_not_exists()
                     .col(ColumnDef::new(Sbom::SbomId).uuid().not_null())
                     .col(ColumnDef::new(Sbom::NodeId).string().not_null())
-                    .col(ColumnDef::new(Sbom::Location).string().not_null())
```
jcrossley3, line 29: I don't understand why any changes to existing migrations are required. Kinda defeats the purpose of migrations, no?

ctron: The change is, that it allows the us of the (now) deprecated `Sbom::Location` variant. That changes that line, but that only affects the compilation/check phase.

jcrossley3: I now understand why "allowing the use" is required: because we have migrations that import other migrations. IMHO, this is a confusing practice because each migration should be self-contained. Any `DeriveIden` enums should be private and only contain variants relevant to the migration in which they're defined.

ctron: That would require us to re-define identifiers locally. Which I believe contradicts the whole idea of such identifier definitions. But if that's what it takes to get the PR merged …

ctron:
I made the change anyway, I hope we can merge it now.

jcrossley3 (approved):

PR #451, opened 2024-06-24
Add a first re-ingest test, fix issues discovered by doing that

ctron:


bobmcwhirter (commented):
Reading the comments, I see the "problem" that two versions act like two sboms. 

I'm unclear if this PR somehow unifies them, or if it maintains them as two distinct SBOMs.

I think we decided they stay two distinct SBOMs, both available for the thing they describe? Perhaps sorted by... date? serial number? so we can see evolution over time if a vendor re-issues against the same SBOM repeatedly?

ctron:
The PR currently only tests the status quo. It should not change the behavior. Two different files will end up in two different SBOMs. Unless their are actual (byte-by-byte) copies.

Also see the discussion: #453

bobmcwhirter:
Excellent, that's the behaviour I was expecting/hoping for.

bobmcwhirter (approved):

ctron:
Merged as part of #452

PR #452, opened 2024-06-25
Add syft re-ingest test, fix discovered issues with files

ctron:
`syft` creates SBOMs which contain "files" and those have relationships. However, we current do not ingest those files, and so the relationships are considered "broken", failing the ingestion of such files.

This PR adds support for files in the SBOM graph and also adds a test for syft, making use of this.

This PR is based on PR #451.


migration/src/lib.rs
```diff
@@ -34,6 +34,7 @@ mod m0000315_create_cvss3_scoring_function;
 mod m0000320_create_cvss3_score_column;
 mod m0000325_create_cvss3_severity_column;
 mod m0000330_add_sbom_version;
+mod m0000350_add_sbom_file;
```
bobmcwhirter, line 37: You'll hafta to renumber this.

migration/src/m0000350_add_sbom_file.rs
```diff
@@ -0,0 +1,44 @@
+use crate::m0000030_create_sbom::SbomNode;
```
bobmcwhirter, line 1: I know it feels like repetition, but let's try for a while to put enums locally and not import from other mgirations.

migration/src/m0000350_add_sbom_file.rs
```diff
@@ -0,0 +1,44 @@
...
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .drop_table(Table::drop().table(SbomFile::Table).if_exists().to_owned())
+            .await?;
+        Ok(())
+    }
+}
+
+#[derive(DeriveIden)]
+pub enum SbomFile {
```
bobmcwhirter, line 40: Drop `pub` to avoid tempting folks to cross-migration import.

bobmcwhirter (changes requested):

PR #454, opened 2024-06-25
feat: add CWE for vulnerabilities

### LINKED ISSUE #281: Should/can generate Vulnerability's CWE as metadata?
Trustification exposes CWE in a column that exposes the list of vulnerabilities within a CSAF file. See image below:

Screenshot from 2024-05-14 10-36-57

- Let's define if we should also include CWE on Trustify
- CWE is also a column defined in the original mockups designed by UX

ctron:
Closes #281 

Should be merged after #442, #452 , #451.

### DIFF
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -42,6 +42,7 @@ mod m0000345_create_version_comparison_fns;
 mod m0000350_remove_old_assertion_tables;
 mod m0000355_labels;
 mod m0000360_add_sbom_file;
+mod m0000370_add_cwe;
 
 pub struct Migrator;
 
@@ -90,6 +91,7 @@ impl MigratorTrait for Migrator {
             Box::new(m0000350_remove_old_assertion_tables::Migration),
             Box::new(m0000355_labels::Migration),
             Box::new(m0000360_add_sbom_file::Migration),
+            Box::new(m0000370_add_cwe::Migration),
         ]
     }
 }
--- a/modules/fundamental/src/advisory/model/details/advisory_vulnerability.rs
+++ b/modules/fundamental/src/advisory/model/details/advisory_vulnerability.rs
@@ -52,15 +52,13 @@ impl AdvisoryVulnerabilityHead {
 
         if let Some(advisory_vuln) = advisory_vuln {
             Ok(AdvisoryVulnerabilityHead {
-                head: VulnerabilityHead::from_vulnerability_entity(vulnerability, tx).await?,
+                head: VulnerabilityHead::from_vulnerability_entity(vulnerability),
                 severity: score.severity().to_string(),
                 score: score.value(),
                 non_normative: VulnerabilityHead::from_advisory_vulnerability_entity(
                     &advisory_vuln,
                     vulnerability,
-                    tx,
-                )
-                .await?,
+                ),
             })
         } else {
             Err(Error::Data(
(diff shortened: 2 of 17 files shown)


migration/src/lib.rs
```diff
@@ -34,6 +34,8 @@ mod m0000315_create_cvss3_scoring_function;
 mod m0000320_create_cvss3_score_column;
 mod m0000325_create_cvss3_severity_column;
 mod m0000330_add_sbom_version;
+mod m0000350_add_sbom_file;
```
bobmcwhirter, line 37: These stacked PRs that encompass previous PRs are sometimes a challenge for me to suss out which part needs reviewing.

migration/src/m0000360_add_cwe.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+#[derive(DeriveIden)]
+pub enum Vulnerability {
+    Table,
+    Cwe,
+}
+
+#[derive(DeriveIden)]
+pub enum AdvisoryVulnerability {
```
bobmcwhirter, line 60: Drop `pub`

migration/src/m0000360_add_cwe.rs
```diff
@@ -0,0 +1,63 @@
...
+                    .to_owned(),
+            )
+            .await?;
+
+        Ok(())
+    }
+}
+
+#[derive(DeriveIden)]
+pub enum Vulnerability {
```
bobmcwhirter, line 54: Drop `pub`

modules/fundamental/src/advisory/model/details/advisory_vulnerability.rs
```diff
@@ -59,15 +59,13 @@ impl AdvisoryVulnerabilityHead {
 
         if let Some(advisory_vuln) = advisory_vuln {
             Ok(AdvisoryVulnerabilityHead {
-                head: VulnerabilityHead::from_vulnerability_entity(vulnerability, tx).await?,
+                head: VulnerabilityHead::from_vulnerability_entity(vulnerability),
```
bobmcwhirter, line 62: I think `tx` might be un-used currently, but was keeping the same pattern so that in the future (which may never arrive) it'd be available, and async, to allow for further scrobbling around in the DB to populate iff needed.

Possibly YAGNI. Not a strongly-held opinion.

modules/graphql/src/vulnerability.rs
```diff
@@ -27,6 +27,7 @@ impl VulnerabilityQuery {
                 published: vulnerability.vulnerability.published,
                 modified: vulnerability.vulnerability.modified,
                 withdrawn: vulnerability.vulnerability.withdrawn,
+                cwe: vulnerability.vulnerability.cwe,
```
bobmcwhirter, line 30: Do we know this is a 1-to-1 relationship? S'fine if so.

modules/ingestor/src/graph/vulnerability/mod.rs
```diff
@@ -28,6 +29,7 @@ impl VulnerabilityInformation {
             || self.published.is_some()
             || self.modified.is_some()
             || self.withdrawn.is_some()
+            || self.cwe.is_some()
```
bobmcwhirter, line 32: Yeah, I start to wonder if we should pass Option<FooInfo> instead of always FooInfo possibly composes of nothing but `None`.

Not your problem; I did this.

bobmcwhirter (changes requested):

ctron:
> These stacked PRs that encompass previous PRs are sometimes a challenge for me to suss out which part needs reviewing.

Yea, I understand that. It reduces the amount of stuff to rebase though.

I updated the PR and it should not be based purely on `main`.

modules/fundamental/src/advisory/model/details/advisory_vulnerability.rs
ctron: Yea, I saw that. Same as the fn being async. However, removing that I noticed that some code could be re-used.

modules/graphql/src/vulnerability.rs
ctron: To my understanding yes. But I might be wrong.

bobmcwhirter (approved):

PR #455, opened 2024-06-25
Bring forward status assertions to package-versions.

bobmcwhirter:
Also, bring towards Vulnerability details, including stanzas akin to Note: the set of possible statuses is unbound, but may include at least some intrinsic ones, such as "affected", "not_affected", and "fixed".

Versions are denoted as one of:

- A singular version without delimiters.
- A range, using [ and ] to denote *inclusive* and ( and ) as exclusive.
  - These can be mixed within a single range, to indicate that the range is half-open, possibly. (Inclusive of the low side, but exclusive to the high side)
- For any range, low or high may be absent, indicating unbounded/infinite.
- If no versions are provided, a `*` is provided, indicating _all versions_ are represented by that status.

```
      "statuses": {
        "affected": [
          {
            "package": {
              "uuid": "9adb7324-89b7-5fe0-a556-23218e39ebf3",
              "purl": "pkg://cargo/hyper"
            },
            "version": "[0.0.0-0,0.14.10)"
          }
        ],
        "fixed": [
          {
            "package": {
              "uuid": "9adb7324-89b7-5fe0-a556-23218e39ebf3",
              "purl": "pkg://cargo/hyper"
            },
            "version": "0.14.10"
          }
        ]
      }
    },
```

### DIFF
--- a/modules/fundamental/src/package/model/details/package_version.rs
+++ b/modules/fundamental/src/package/model/details/package_version.rs
@@ -1,9 +1,17 @@
+use crate::advisory::model::AdvisoryHead;
 use crate::package::model::{PackageHead, PackageVersionHead, QualifiedPackageHead};
+use crate::vulnerability::model::VulnerabilityHead;
 use crate::Error;
-use sea_orm::ModelTrait;
+use sea_orm::{
+    ColumnTrait, EntityTrait, LoaderTrait, ModelTrait, QueryFilter, QuerySelect, RelationTrait,
+};
+use sea_query::{Asterisk, Expr, Func, JoinType, SimpleExpr};
 use serde::{Deserialize, Serialize};
-use trustify_common::db::ConnectionOrTransaction;
-use trustify_entity::{package, package_version, qualified_package};
+use trustify_common::db::{ConnectionOrTransaction, VersionMatches};
+use trustify_entity::{
+    advisory, organization, package, package_status, package_version, qualified_package, status,
+    version_range, vulnerability,
+};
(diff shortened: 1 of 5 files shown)


modules/fundamental/src/package/model/details/package_version.rs
```diff
@@ -12,20 +20,24 @@ pub struct PackageVersionDetails {
...
     ) -> Result<Self, Error> {
-        let package = package_version
-            .find_related(package::Entity)
-            .one(tx)
-            .await?
-            .ok_or(Error::Data(
-                "underlying package missing for package-version".to_string(),
-            ))?;
+        let package = if let Some(package) = package {
+            package
```
jcrossley3, line 33: nit: `unwrap_or_else`?

jcrossley3 (approved):

modules/fundamental/src/package/model/details/package_version.rs
bobmcwhirter: Alas, can't or-else with an async closure, to load-if-None, yeah?

jcrossley3: Gotcha. The `if let` seems less clear (to me) than the `match` in this particular case, but like I said, a nit.

```rust
        let package = match package {
            Some(p) => p,
            None => package_version
                .find_related(package::Entity)
                .one(tx)
                .await?
                .ok_or(Error::Data("underlying package missing".to_string()))?,
        };
```

Issue #456, opened 2024-06-26
Bad CSAF ingestion performance


ctron:
I noticed a massive drop in performance when running the CSAF import process. I don't have numbers yet, but I'll investigate.

In the SQL logs I noticed at of a "select then insert" patters again.

ctron:
I tried to go back a bit in the git history to see if that's a recent change. Doesn't look like that's the case.

I started to work with the test data file `cve-2023-33201.json`. Which takes around 470ms to ingest. It looks like the majority of time goes into ingesting the vulnerabilities. Parsing, inserting, and committing the main entry takes around 70ms.

ctron:
The majority of time goes into `ingest_product_statuses`.

ctron:
I changed the title as I cannot provide any numbers. I just have the feeling it got much slower since some time ago.

JimFuller-RedHat:
@ctron is this still an issue ?

ctron:
Hm, good question. Could it be better. Probably, but we tried to optimize things already.

Still, importing that initial dataset still takes ~12 hours.

helio-frota:
Generating flamegraph with `cargo run --bin xtask --release generate-dump` for `ImporterConfiguration::Csaf(CsafImporter {` only, generated this after some 2 or 3 minutes:

at lest has some relation with previous comments

![2024-10-23_11-17](https://github.com/user-attachments/assets/f4ae371a-f941-4d20-9992-9160b04451c1)

Adding a zipped svg because when we click the preview svg in github comment, for some reason we can't navigate on flamegraph.. better to download unzip and click to open in the browser
[d.zip](https://github.com/user-attachments/files/17493339/d.zip)

helio-frota:
@ctron can we change the title , suggestion:

`try to prevent asyn-runtime from stealing cpu cycles that slow down csaf ingestion performance (up to an acceptable level)` ?

I was not able to "tokio::spawn_blocking"  [this](https://github.com/trustification/trustify/blob/ed72d871cd9a9d2afe6ce30380d0bbd5edf7b497/modules/ingestor/src/service/advisory/csaf/creator.rs#L47) due some weirdo lifetime things I'm not good yet.

But I noticed that it also helps a bit with tokio trying to poll causing more blocking situation

ctron:
Let's close this for now. I'd expect the database to be the bottleneck (or how we use the DB). But let's focus on more specific issues.

JimFuller-RedHat:
[...]

PR #458, opened 2024-06-26
Improve performance of CSAF ingestion

ctron:
With the performance test, this improves the import time from ~470ms to ~240ms


ctron:
Also see: #456

migration/src/m0000380_create_package_status_index.rs
```diff
@@ -0,0 +1,43 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .create_index(
```
bobmcwhirter, line 10: Nice, thanks!

modules/ingestor/src/graph/advisory/advisory_vulnerability.rs
```diff
@@ -31,7 +31,7 @@ pub enum Version {
 }
 
 impl VersionInfo {
-    fn into_active_model(self) -> version_range::ActiveModel {
+    pub fn into_active_model(self) -> version_range::ActiveModel {
```
bobmcwhirter, line 34: bonus points: if you can understand why implementing `IntoActiveModel` caused rustc to complain about duplicate impls, that'd be lovely. There was a trait I could use, but I couldn't make rustc happy, hence this one-off fn.

modules/ingestor/src/service/advisory/csaf/creator.rs
```diff
@@ -0,0 +1,147 @@
...
+        for r in ps.iter().flatten() {
+            for purl in resolve_purls(csaf, r) {
+                let package = Purl::from(purl.clone());
+
+                if let Some(version) = package.version.clone() {
+                    self.entries.push(PackageStatus {
+                        package,
+                        status,
+                        info: VersionInfo {
+                            scheme: "generic".to_string(),
```
bobmcwhirter, line 47: not your problem, just noting: there's a few schemes which happen to match the purl `type` so we could avoid `generic` perhaps, if we know it's a python, or maven, or rust (semver) type.

modules/ingestor/src/service/advisory/csaf/creator.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    pub async fn create(self, connection: &impl ConnectionTrait) -> Result<(), Error> {
+        let mut checked = HashMap::new();
+
+        let mut purls = PurlCreator::new();
+
+        for ps in &self.entries {
+            // ensure a correct status, and get id
+            if let Entry::Vacant(entry) = checked.entry(ps.status) {
+                entry.insert(Self::check_status(ps.status, connection).await?);
```
bobmcwhirter, line 75: Does this mean status rows can be implicitly added upon first sighting?  I think that's fine.

bobmcwhirter (approved):

modules/ingestor/src/service/advisory/csaf/creator.rs
ctron: Nope, that just inserts it into the hash map. The idea is to look it up only once, cache the `status.id` and re-use it a bit further down the function. Versus: looking it up for each entry.

modules/ingestor/src/graph/advisory/advisory_vulnerability.rs
ctron: That one was a bit weird, it seems to be coming from the blanket implementation of that trait in seaorm:

```rust
impl<A> IntoActiveModel<A> for A
where
    A: ActiveModelTrait,
{
    fn into_active_model(self) -> A {
        self
    }
}
```

Which basically means that you can't implement `IntoActiveModel` for anything that should become an `ActiveModel`. Which feels like an API bug to me.

It also doesn't seem possible to implement `From` or `Into` for that case due to similar conflicts.

ctron: Ok, that's not correct. Because in this case `A` would be `VersionInfo`, which doesn't implement `ActiveModelTrait`. And the compiler complains about `impl sea_orm::IntoActiveModel<trustify_entity::version_range::ActiveModel> for <trustify_entity::version_range::Entity as sea_orm::EntityTrait>::Model` being the conflict.

ctron: Ok, I give up on the bonus points and leave it a mystery.

Issue #466, opened 2024-06-27
GET /api/v1/package/{uuid} fails with some packages


carlosthe19916:
Steps to reproduce:
- Start a clean instance of trustify
- Enable only the "redhat-csaf" importer. Wait some minutes
- Search for packages like ose-powervs-cloud-controller-manager-rhel8 and try to fetch it by their uuid, we will see an error:

```json
{
    "error": "Database error",
    "message": "Query Error: error returned from database: invalid input syntax for type numeric: \"sha256:f579ee538e6f4bfa4d9c20bf2680d7fc3ec7a757a40637a743b7962187b4d6a1\""
}
```

https://github.com/trustification/trustify/assets/2582866/6e1012b5-f0e1-43a8-b1d0-9eb1284b4044

> Hint: it seems to happen with package whose version are something like sha256:cfb31631b19c2e9d2ef28bac6d4471c97e79976b0c9590032cf0206bfcd3e4f3 ; on the other hand, packages with versions like 1.2.3 work fine

ctron:
This looks like a mix of data types. "Packages" is most like "PURLs" (which we still label incorrectly). Actual packages have an ID which is a UUID (because we use UUIDs in many cases internally).

What looks like a "package version" here is, most likely, actually the "version" component of a "PURL" from a container. Same format, but unknown to our system.

The endpoint `/api/v1/package/{uuid}` does NOT work with packages, but with PURLs. The UUID is just accidentally a UUID, that should just be an "ID" field (which happens to be a UUID, but that's an implementation detail).

bobmcwhirter:
Can we verify if this is still an issue?

carlosthe19916:
> Can we verify if this is still an issue?

It is still happening. Unfortunately, it is tricky to reproduce. I needed to run the CVE 2024 importer and then the CSAF importer wait for some time and search for packages as the video below

https://github.com/user-attachments/assets/73aa35f1-b0ba-4032-ac2c-23dcc533d7f6

bobmcwhirter:
Your video seems to show `/api/v1/package/{id}` usage, and I no longer know what endpoint that is.

If it's SBOM packages, that'd be under `/api/v1/sbom/{id}/packages`.

For purls, it'd be /api/v1/purl/{id}

I'll try to reproduce...

Issue #461, opened 2024-06-26
Surface package status for packages in UI.


bobmcwhirter:
Added in https://github.com/trustification/trustify/commit/d8ca36d6a5af74b307c4087a4d4a99652a5c2058

From `/api/v1/package/{uuid}` or `/api/v1/package/version/{uuid}`

Packages (and package/version) now include advisories and status, which may include stuff such as:

```
      "statuses": {
        "affected": [
          {
            "package": {
              "uuid": "9adb7324-89b7-5fe0-a556-23218e39ebf3",
              "purl": "pkg://cargo/hyper"
            },
            "version": "[0.0.0-0,0.14.10)"
          }
        ],
        "fixed": [
          {
            "package": {
              "uuid": "9adb7324-89b7-5fe0-a556-23218e39ebf3",
              "purl": "pkg://cargo/hyper"
            },
            "version": "0.14.10"
          }
        ]
      }
    },
```

carlosthe19916:
- the `GET /api/v1/package/{uuid}` is used when the user enters to the details of a Package
- When we see the details of a package we need to know
  - Which are the VULNERABILITIES that affect the current package (in the image below is the CVE Tab)


![Screenshot from 2024-06-26 21-12-49](https://github.com/trustification/trustify/assets/2582866/f9a71863-40e9-4c3a-a179-adff0baf00a3)

What the current endpoint is generating is ADVISORIES and not VULNERABILITIES which IMO is not what we want.

This is a hard requirement for compatibility with the old Trustification. The image below belong to RHTPA and it renders VULNERABILITIES not Advisories

![image](https://github.com/trustification/trustify/assets/2582866/e6714939-bd63-4562-9072-8299e7e43bc4)

bobmcwhirter:
A vulnerability is attached by way of an advisory. The vulns are there. Under the advisory that claims they are relevant.

carlosthe19916:
@bobmcwhirter 
Given a PACKAGE X what we need to know is:
1. Which are the VULNERABILITIES that affect the PACKAGE X. 
1. Which are the SBOMs that contain the PACKAGE X. (this case is covered already)

These 2 things are part of the Famous three Use Cases that are the Core of Trustification's design.

![Screenshot from 2024-06-26 21-47-01](https://github.com/trustification/trustify/assets/2582866/7efbd555-639f-485a-a897-16594ed9bf60)

[...]

Issue #435, opened 2024-06-19
How to do GET /api/v1/package/{PURL}?


carlosthe19916:
- If I fetch data from `GET /api/v1/sbom/{key}/packages` then I get the following response:

```json
{
    "items": [
        {
            "id": "SPDXRef-008baaa8-8056-48c3-b019-24a82b3a4bf1",
            "name": "quarkus-smallrye-metrics",
            "purl": [
                "pkg://maven/io.quarkus/quarkus-smallrye-metrics@2.13.8.Final-redhat-00004?repository_url=https://maven.repository.redhat.com/ga/&type=jar"
            ]
        }
    ],
    "total": 880
}
```

- Each Purl is currently decomposed in the UI and then rendered in a table, like the image below:

![Screenshot from 2024-06-19 11-02-34](https://github.com/trustification/trustify/assets/2582866/7b5b0153-bc77-44e4-b48a-eac68f315edc)

- Here is the question. How can I use a single PURL and use it in `GET /api/v1/package/{PURL}`? currently in order to fetch a single package I need the `uuid` of a package, but in the example above all I have is a PURL

ctron:
I totally understand the confusion. I don't have an answer, but I can explain it.

The `{id}` (what you have as `{key}`), is actually the UUID of the SBOM. That returns "packages", which used to be PURLs, but today are actual SBOM packages/components. These have again an ID and may (or may not) have multiple CPEs and PURLs.

The ID of the package in this case is `SPDXRef-008baaa8-8056-48c3-b019-24a82b3a4bf1`. You can use this to fetch additional relationships, in the context of the SBOM. It is a "package/component of the SBOM" (SPDX: package, cyclonedx: component).

The `/api/v1/packages` endpoints target "PURLs", which used to be known as "packages". However, to me, PURLs are identifiers for packages. CPEs and hashes are other types of identifiers for packages.

If you want to learn more about "packages" (not in the context of an SBOM), then you can use those endpoints (from `/api/v1/packages`). There you have endpoints like:

* `/v1/package/type/{type}/{namespace_and_name:[^@]+}`
* `/v1/package/type/{type}/{namespace_and_name:[^@]+}@{version}`

There are also APIs which allow you to get information for those "packages" (actually PURLs) by their internal ID. The internal ID (which is a UUID) can actually be derived from the PURL. But I would **NOT (!!)** recommend to do that. I see this as an implementation detail that no one should build upon.

[...]

PR #468, opened 2024-06-27
refactor: make it clear we're working with PURLs

ctron:
As it feels the confusion happens quite often, even to me, I think we should rename things.

We have "packages", which have some meaning according to our glossary: https://github.com/trustification/trustify/blob/main/docs/glossary.md#package

> A package is an atomic artifact or component. Packages may be addressed using pURLs. A package may be described by an SBOM describing how it is created and its contents. A package may certainly contain other packages (e.g. shading one Java jar into another). A package may also be the sole member of a Product (UBI-8.0.13-x86.oci may be the singular package within the "UBI 8.0.13-x86" product). A package is one step more abstract than an artifact.

However, in the next section that definition actually calls out PURLs:

> Package URLs (pURLs) are possibly ambiguous names applied to packages. […]

So there are "packages", and "PURLs", which can be used to give "packages" a name.

In our current API and database structure, we have (because history) several entries, endpoints, and services which are named "packages", but actually deal with PURLs only.

This PR tries to clean up the user facing API (REST endpoints) by changing the base URL from `/api/v1/package` to `/api/v1/package/by-url`, when it comes to PURL based calls.

## Next steps

It would also be reasonable to rename functions, services, and database structures towards PURLs. But naming is hard, so let's start small.

## Alternatives

One alternative could be to use `/api/v1/purl`. However, those API still deal with packages, just by using PURLs as name. So keeping "packages" make sense to me.


JimFuller-RedHat:
**random aside** - in corgi land we sidestepped the overloading of the term 'package' by just using the equally imprecise term of 'component' which always has a 1:1 relationship with a single PURL. That gave us some 'wiggle' room if ever we had to assign a purl to something like a single script or something less then a package.

bobmcwhirter:
I think in my mind, a package _is_ something that can be referenced by a pURL.

A pURL is just a name for a pURL-addressible thing.

The ambiguity I think I was talking about in the glossary is more around... a pURL doesn't describe the binary bytes produced. Those would be `artifacts`.

eg, a specific build of `pkg:maven/org.apache/log4j@1.2.3` with a given hash of `sha256:<blah>` would be a non-ambiguous artifact, based on the package referenced by `pkg:maven/org.apache/log4j@1.2.3`.

"Bob McWhirter" is a name, somewhat ambiguous (there's at least one Canadian and one Arizonan sharing my name). This particular Bob who is typing now is the the artifact with my particular `sha256`.  I think?

All that being said, I'm not opposed to `by-url`, but that makes me wonder, is there any other non-url way to discuss a package, beyond our internal `uuid`?  Or do we consider that sufficient to bifurcate?

I think similarly, while other systems may store "URLs", they are ultimately speaking about _the thing behind the URL_, e.g., the endpoint, the webpage, the podcast you can download.

Naming is hard. Naming things that name things is even harder.

Do we work with names? Do we work with things, and just use names to refer to them?

I dunno.

bobmcwhirter:
I guess I may fall into the camp of "a package is anything that can be described by a pURL" which might as well just be a pURL. Maybe.

JimFuller-RedHat:
when you say 'But one package can have multiple PURLs assigned,' 

Do you mean a package can contain other packages which have purls or do you mean a single package can have multiple purls ?

If the later then  eg. a single unique id assigned to a single package creates a contract ... put another way if we do have other purls that need to map to a package I would suggest pointing to canonical purl (ultimately this is 'our purl ') rather then directly map to package. 

This is a subtle difference but we have already seen differences in who mints a purl ... sometimes we are the first person to do so, then sometime later upstream mints a purl. Also we should expect wholesale refactoring of purls a real possibility as that proto spec matures and upstream work more with it ... it becomes burdensome to have to update all associations in database, etc. when this happens.

From a REST API (conceptual model) pov we can make the association transparent ... eg the above applies more to the logical internal model to handle unknown change later on .

Lastly,  I could be easily convinced I am overthinking all this....

ctron:
In SPDX-land, a single package can have a list of external references, which can by of type `purl`: https://spdx.github.io/spdx-spec/v2.3/package-information/#721-external-reference-field … which leads to each package having a list of cpes or purls attached, if I get that right.

And I do recall a conversation where that might actually make sense. Publishing the same artifact to multiple maven repositories.

bobmcwhirter:
I know we're going into the weeds here, but I also think `pkg:maven/com.foo/bar@1.2.3` _implies_ `?repository_url=m2.maven.org` or whatever maven central's repo is.

So the same jar published elsewhere would have to include `?repository_url=somewhere.else`

JimFuller-RedHat:
+1 to external reference ... and maybe all this kind of stuff be represented as labels/tags ? maybe that is overloading generic thing too much and worth representing in core model as 'external-reference' ... that makes it clearly different then the canonical purl.

ctron:
I think the model that we have is actually quite good. We have base PURLs, versioned PURLs, qualified PURLs. Referencing those from SBOM packages/components. Also towards CPEs and hashes.

It's only the names that feel confusing.

carlosthe19916:
just to be sure. The current way the UI is fetching packages is:
-  First step: get the list of packages from `GET /api/v1/package`. Each item from the list obtained here has a field `uuid`. Then
- Second step: get a single package using `GET /api/v1/package/{uuid}`

Are those step still going to be valid? I would expect "yes" right?

jcrossley3:
> As it feels the confusion happens quite often, even to me, I think we should rename things.

Can you describe the confusion that happens quite often? What problems are we solving?

> This PR tries to clean up the user facing API (REST endpoints) by changing the base URL from `/api/v1/package` to `/api/v1/package/by-url`, when it comes to PURL based calls.

I don't see how this clarifies things. To me, `/api/v1/package/{purl}` is pretty clear. I'm asking for the package referenced by a purl.

I also don't understand the terms "base", "qualified" and "versioned". Will the consumers of our API understand them?

ctron:
> > As it feels the confusion happens quite often, even to me, I think we should rename things.
> 
> Can you describe the confusion that happens quite often? What problems are we solving?

* https://github.com/trustification/trustify/issues/435
* https://github.com/trustification/trustify/issues/461
* https://github.com/trustification/trustify/issues/466

I also did stumble over this myself when navigating the code. You see something that's named "package", but indeed it's a PURL.

> 
> > This PR tries to clean up the user facing API (REST endpoints) by changing the base URL from `/api/v1/package` to `/api/v1/package/by-url`, when it comes to PURL based calls.
> 
> I don't see how this clarifies things. To me, `/api/v1/package/{purl}` is pretty clear. I'm asking for the package referenced by a purl.

How would you ask then for a package by CPE?

> 
> I also don't understand the terms "base", "qualified" and "versioned". Will the consumers of our API understand them?

I don't know, but they already existed. And, they only make sense on the context of a PURL. Not in the context of a "package". So it might be that you are too confused by the naming.

ctron:
> just to be sure. The current way the UI is fetching packages is:
> 
>     * First step: get the list of packages from `GET /api/v1/package`. Each item from the list obtained here has a field `uuid`. Then
> 
>     * Second step: get a single package using `GET /api/v1/package/{uuid}`
> 
> 
> Are those step still going to be valid? I would expect "yes" right?

That depends on what you mean by "package". Package of an SBOM no. Package referenced by a CPE, no. Package referenced by a PURLs, yes.

ctron:
> Also, a general thought:
> 
>     * There must be a single way of fetching an Entity (single entity). I mean, having `/api/v1/package/by-purl/{key}` and `/api/v1/package/{key}` seems illogical to me.

Correct. But indeed what you get back is information coming from the PURL. Not from the package itself. Basically it splits the PURL into a lot of database entities and returns that.

> 
>     * I have no issues with multiple "search" endpoints (list of entities) `/api/v1/package?q=abc` and `/api/v1/package/by-purl?q=abc` where these endpoints return a paginated result.

That's why one idea was to just go for `/api/v1/purl`. Because basically you're only searching for PURLs. And they are references packages inside of SBOMs. And so they reference SBOMs. But you could also find an SBOM package by CPE or hash (not today, but hopefully in the future).

jcrossley3:
> That's why one idea was to just go for `/api/v1/purl`. Because basically you're only searching for PURLs.

What is the use case for that? To me, a purl is not a resource -- it's just the *name* of a resource, in this case a `Package`.

Do we have a need to return a paginated list of names?

jcrossley3:
Could some of the confusion be coming from the existence of both a `package` and an `sbom_package` table? I would expect an `sbom` record to relate to a `package` record somehow.

ctron:
> > That's why one idea was to just go for `/api/v1/purl`. Because basically you're only searching for PURLs.
> 
> What is the use case for that? To me, a purl is not a resource -- it's just the _name_ of a resource, in this case a `Package`.
> 
> Do we have a need to return a paginated list of names?

Looks like that. If you take a look at e.g. `PackageService::package_by_uuid`, which returns `PackageDetails`, that's simply a decomposed PURL, stored in tables. And it's absolutely not about "packages".


> Could some of the confusion be coming from the existence of both a `package` and an `sbom_package` table? I would expect an `sbom` record to relate to a `package` record somehow.

Right, the `package` table was there before. But it only stores PURLs, not packages. And `sbom` record relates to an `sbom_package`, which relates to zero or more `packages` (which indeed are PURLs). It also relates to one or more `cpes`, which is correct.

That's why I think it makes sense to rename services and entities too. But before doing all the work, I think we should agree on that.

jcrossley3:
> Right, the `package` table was there before. 

Before what? If you mean before we implemented [this](https://github.com/trustification/trustify/blob/main/docs/design/sbom.md), then I might question why the existing `package` table wasn't incorporated into that design.

> But it only stores PURLs, not packages. 

This sentence is confusing. A PURL is just a name. It's a logical concatenation of attributes of a package (the concept, not the table). A package will have those attributes whether a particular SBOM says it has a PURL/CPE/identifier or not, right?

From the above doc, "SBOM packages are an entity of their own, but may have zero or more PURLs or CPEs (or other identifiers)". That sentence betrays a lack of shared understanding, I think. In an abstract sense, a package will always have a PURL, because it can be constructed from the package's attributes. From a literal reading of the spec, a package may not have an identifier, but is that what we should be modeling in the database?

My intuition is that our database designers -- Jens and Bob -- need to agree on whether we're modeling entities in the abstract or according to a spec.

ctron:
> > Right, the `package` table was there before.
> 
> Before what? If you mean before we implemented [this](https://github.com/trustification/trustify/blob/main/docs/design/sbom.md), then I might question why the existing `package` table wasn't incorporated into that design.

Because it didn't fit.

> > But it only stores PURLs, not packages.
> 
> This sentence is confusing. A PURL is just a name. It's a logical concatenation of attributes of a package (the concept, not the table). A package will have those attributes whether a particular SBOM says it has a PURL/CPE/identifier or not, right?

I don't think it's the sentence that is confusing, but the structure in the database. The `package`, `versioned_package`, `qualified_package` table only store decomposed parts of PURLs. I see the reason for that. It's the naming that is confusing!

Actually the package does not. It comes from the context of the SBOM. The same package can be named by different SBOM by different names.

> In an abstract sense, a package will always have a PURL, because it can be constructed from the package's attributes. From a literal reading of the spec, a package may not have an identifier, but is that what we should be modeling in the database?

Again, in the database we model PURLs (with tables mentioned above). Not packages.

I think the database structure works just fine. We have (actual) SBOM packages, named by PURLs and CPEs. (any maybe others in the future). It simply is the naming.

bobmcwhirter:
While I think a package is "anything that can be address by a Package URL (pURL)" I don't hold it strongly enough, nor do I have enough about SBOM packages in my head to disagree with @ctron right now.

My vote would be to proceed (after we winnow down other outstanding package-related PRs.......) and see how we feel with the purl-centric naming.

modules/fundamental/src/package/endpoints/base.rs
```diff
@@ -8,32 +8,32 @@ use trustify_common::model::Paginated;
...
     ),
     responses(
-        (status = 200, description = "Details for the versionless base package", body = PackageDetails),
+        (status = 200, description = "Details for the versionless base PURL", body = PackageDetails),
     ),
 )]
-#[get("/v1/package/base/{uuid}")]
+#[get("/v1/package/by-purl/base/{id}")]
 pub async fn get(
     service: web::Data<PackageService>,
```
bobmcwhirter, line 19: Do we want to rename the service to PurlService?

bobmcwhirter (approved):

modules/fundamental/src/package/endpoints/base.rs
jcrossley3: If we do, I'd argue the api should become `#[get("/v1/purl/base/{id}")]`

ctron: Maybe not the service, but the functions and types that deal with PURLs. The main namespace/module still is `package`, which is ok.

ctron:
I updated the PR, renaming the package tables to "purl". Renaming the fields as well.

bobmcwhirter (approved):

PR #469, opened 2024-06-27
Reduce duplicate status ingestions.

bobmcwhirter:
In the status-builder, ignore qualifiers, impl PartialEq where necessary. In the vuln fetching, add a `distinct`.

### DIFF
--- a/modules/ingestor/src/service/advisory/csaf/creator.rs
+++ b/modules/ingestor/src/service/advisory/csaf/creator.rs
@@ -8,12 +8,13 @@ use crate::{
 use csaf::{definitions::ProductIdT, Csaf};
 use sea_orm::{ActiveValue::Set, ColumnTrait, ConnectionTrait, EntityTrait, QueryFilter};
 use sea_query::IntoCondition;
+use std::collections::HashSet;
 use std::{collections::hash_map::Entry, collections::HashMap};
 use trustify_common::{db::chunk::EntityChunkedIter, purl::Purl};
 use trustify_entity::{package_status, status, version_range};
 use uuid::Uuid;
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, Eq, Hash, PartialEq)]
 struct PackageStatus {
     package: Purl,
     status: &'static str,
@@ -23,7 +24,7 @@ struct PackageStatus {
 pub struct PackageStatusCreator {
     advisory_id: Uuid,
     vulnerability_id: i32,
-    entries: Vec<PackageStatus>,
+    entries: HashSet<PackageStatus>,
 }
 
 impl PackageStatusCreator {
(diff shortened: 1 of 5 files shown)


modules/ingestor/src/service/advisory/csaf/creator.rs
```diff
@@ -37,17 +38,22 @@ impl PackageStatusCreator {
     pub fn add_all(&mut self, csaf: &Csaf, ps: &Option<Vec<ProductIdT>>, status: &'static str) {
         for r in ps.iter().flatten() {
             for purl in resolve_purls(csaf, r) {
-                let package = Purl::from(purl.clone());
+                let mut package = Purl::from(purl.clone());
+                package.qualifiers.clear();
```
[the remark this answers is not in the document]
bobmcwhirter: wrt package-vs-purl, was going to wait for Jen's PR to settle (and the conversation around it).

I'm clearing the qualifiers, because we're ultimately tracking unqualified packages (purls), and the additional qualifiers break `PartialEq` by comparing against bits we shouldn't be carrying around at this point.

Basically, we have an `pkg:rpm/crap@1.2.3?arch=ppc` and another `pkg:rpm/crap@1.2.3?arch=c64` and those `?arch=...` are just noise for this process, so I strip them off, so we don't end up with `pkg:rpm/crap@1.2.3` in the queue _twice_.

modules/fundamental/src/vulnerability/service/test.rs
```diff
@@ -60,3 +60,29 @@ async fn statuses(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
...
+
+    ingestor
+        .ingest(("source", "test"), None, Format::CSAF, data)
+        .await?;
+
+    let vuln = vuln_service
+        .fetch_vulnerability("CVE-2024-5154", Transactional::None)
+        .await?;
+
+    println!("{}", serde_json::to_string_pretty(&vuln)?);
```
jcrossley3, line 85: Can you assert instead of print?

bobmcwhirter: It ended up being a duplicate test. Rm'd.

bobmcwhirter:
@jcrossley3 can I get a re-re when you have a moment, please?

modules/ingestor/src/service/advisory/csaf/creator.rs
```diff
@@ -37,17 +38,22 @@ impl PackageStatusCreator {
...
                         package,
                         status,
                         info: VersionInfo {
                             scheme: "generic".to_string(),
                             spec: VersionSpec::Exact(version),
                         },
-                    });
+                    };
+
+                    if !self.entries.contains(&status) {
```
ctron, line 54: Wouldn't it make sense to use a `HashSet` for this? To me it looks like as if all struct could implement `Hash` and `Eq`.

bobmcwhirter: Good idea. I'll give that a whirl.

ctron (approved):

PR #473, opened 2024-06-27
Ensure vulnerability.identifier is unique.

### LINKED ISSUE #471: After I upload 2 Advisory Files I ran into apparent duplicate Vulnerabilities
Steps to reproduce:

- Upload the following Advisories in an empty Trustify
  - [CVE-2021-32714.json](https://github.com/user-attachments/files/16019459/CVE-2021-32714.json)
  - [RUSTSEC-2021-0079.json](https://github.com/user-attachments/files/16019460/RUSTSEC-2021-0079.json)
- Fetch the list of Vulnerabilities from `GET /api/v1/vulnerability` and you will see the following response:

```json
{
    "items": [
        {
            "identifier": "CVE-2021-32714",
            "title": "Integer Overflow in Chunked Transfer-Encoding",
            "published": "2021-07-07T19:35:10Z",
            "modified": "2021-07-07T19:35:10Z",
            "cwe": "CWE-190",
            "average_severity": null,
            "average_score": null,
            "advisories": [
                {
                    "uuid": "urn:uuid:7e8868ec-170a-492f-a93c-d64130f4220f",
                    "identifier": "CVE-2021-32714",
                    "hashes": [
                        "sha256:53dbabe96c9e778941cff248c2607e927a2d0984d2fd8d81617dafb2bc53d420"
                    ],
                    "issuer": {
                        "id": 2,
                        "name": "GitHub_M",
                        "cpe_key": null,
                        "website": null
                    },
                    "published": "2021-07-07T19:35:10Z",
                    "modified": "2021-07-07T19:35:10Z",
                    "title": "Integer Overflow in Chunked Transfer-Encoding",
                    "severity": null,
                    "score": null
                }
            ]
        },
        {
            "identifier": "CVE-2021-32714",
            "average_severity": "critical",
            "average_score": 9.1,
            "advisories": [
                {
                    "uuid": "urn:uuid:b0ecbfbf-0b21-47ab-9226-cdb5b061323b",
                    "identifier": "RUSTSEC-2021-0079",
                    "hashes": [
                        "sha256:8297ce58d1a870b0ef2f1e3bb4a08e87b54696e5faa85ba1724978ef80d2d5b1"
                    ],
                    "issuer": {
                        "id": 1,
                        "name": "Rust Security Advisory Database",
                        "cpe_key": null,
                        "website": null
                    },
                    "published": "2021-07-07T12:00:00Z",
                    "modified": "2021-10-19T22:14:35Z",
                    "title": "Integer overflow in `hyper`'s parsing of the `Transfer-Encoding` header leads to data loss",
                    "severity": "critical",
                    "score": 9.1
                }
            ]
        }
    ],
    "total": 2
}
```

Both items in the response have the same `identifier`.
The problem is that the `identifier` is being used to fetch a single Advisory e.g. `/api/v1/vulnerability/CVE-2021-32714`.
The question is, which Vulnerability am I fetching if both have the same `identifier`?

The other entities are using something like `uuid`, not sure if that would be the solution but whatever the solution is, there should be a  unique key to identify a Vulnerability

bobmcwhirter:
Add an EXCLUSIVE lock while ingesting vuln identifiers to prevent us from getting into an error state neeedlessly.

Fixes #471

### DIFF
--- a/modules/ingestor/src/graph/vulnerability/mod.rs
+++ b/modules/ingestor/src/graph/vulnerability/mod.rs
@@ -5,7 +5,8 @@ use crate::graph::error::Error;
 use crate::graph::Graph;
 use sea_orm::ActiveValue::Set;
 use sea_orm::{
-    ActiveModelTrait, ColumnTrait, EntityTrait, ModelTrait, QueryFilter, QuerySelect, RelationTrait,
+    ActiveModelTrait, ColumnTrait, ConnectionTrait, EntityTrait, ModelTrait, QueryFilter,
+    QuerySelect, RelationTrait,
 };
 use sea_query::{Condition, JoinType};
 use std::fmt::{Debug, Formatter};
@@ -57,6 +58,10 @@ impl Graph {
         information: I,
         tx: TX,
     ) -> Result<VulnerabilityContext, Error> {
+        self.connection(&tx)
+            .execute_unprepared("LOCK TABLE vulnerability IN EXCLUSIVE MODE;")
+            .await?;
+
         // TODO: consider transforming into upsert
         let information = information.into();
         if let Some(found) = self.get_vulnerability(identifier, &tx).await? {
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -44,6 +44,7 @@ mod m0000355_labels;
 mod m0000360_add_sbom_file;
 mod m0000370_add_cwe;
 mod m0000380_create_package_status_index;
+mod m0000385_create_vulnerability_index_unique_constraint;
 
 pub struct Migrator;
 
@@ -94,6 +95,7 @@ impl MigratorTrait for Migrator {
             Box::new(m0000360_add_sbom_file::Migration),
             Box::new(m0000370_add_cwe::Migration),
             Box::new(m0000380_create_package_status_index::Migration),
+            Box::new(m0000385_create_vulnerability_index_unique_constraint::Migration),
         ]
     }
 }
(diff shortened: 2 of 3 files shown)


ctron:
Hm, that feels a bit drastic. Maybe necessary. Maybe there's a way to transform this indeed into an upsert, which should achieve the same? Anyway, I defer to @JimFuller-RedHat when it comes to DB stuff like that.

JimFuller-RedHat:
As long as the ORM produces something like the following sql:

```
ALTER TABLE your_table ADD CONSTRAINT constraint_name UNIQUE (your_column);
```
that adds a unique constraint on the column ... we should be good.

Agree with @ctron eg. would reccommend to avoid exclusive locking of the table ... this can cause all sorts of surprising performance issues in read/write scenarios - pg, by default, handles locking at various levels and should just 'do the right thing' with the defined unique constraint.

modules/ingestor/src/graph/vulnerability/mod.rs
```diff
@@ -57,6 +58,10 @@ impl Graph {
         information: I,
         tx: TX,
     ) -> Result<VulnerabilityContext, Error> {
+        self.connection(&tx)
+            .execute_unprepared("LOCK TABLE vulnerability IN EXCLUSIVE MODE;")
```
JimFuller-RedHat, line 62: should be enough to define unique constraint - are we observing otherwise (which is just another way of asking if we have a unit test that replicates existing collision) ?

bobmcwhirter:
Wrong solution.

Issue #434, opened 2024-06-19
/api/v1/sbom/{key}/packages?q=some_string does not return data


carlosthe19916:
I am trying to filter packages of an sbom by a string using
```
/api/v1/sbom/{key}/packages?q=some_string
```

and although I know the there should be results (as I am entering a valid string) I get zero results in the response.

> This endpoint does work as expected `/api/v1/package?q=some_string`

PR #474, opened 2024-06-27
feat: enable backslash to escape operators in filter expressions

jcrossley3:
Relates to #434 as a test for filtering packages was added

### DIFF
--- a/common/src/db/query.rs
+++ b/common/src/db/query.rs
@@ -527,7 +527,11 @@ fn encode(s: &str) -> String {
 }
 
 fn decode(s: &str) -> String {
-    s.replace('\x07', "&").replace('\x08', "|")
+    s.replace('\x07', "&")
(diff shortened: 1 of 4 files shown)


common/src/db/query.rs
```diff
@@ -527,7 +527,11 @@ fn encode(s: &str) -> String {
 }
 
 fn decode(s: &str) -> String {
-    s.replace('\x07', "&").replace('\x08', "|")
+    s.replace('\x07', "&")
+        .replace('\x08', "|")
+        .replace(r"\\", "\x08")
```
bobmcwhirter, line 532: What's our threshold for a parser?

bobmcwhirter (approved):

common/src/db/query.rs
jcrossley3: When its relative value is obvious?

PR #475, opened 2024-06-28
Allow setting labels for advisories and SBOMs

ctron:
Coming from @bobmcwhirter's suggestion of making the labelling a two-step process: https://matrix.to/#/!heUmIqxVhZejiSjzVu:matrix.org/$VtbkZ9oC8eeQqzDLlsCes5sMWQRiBZrNwnhf231UZZ4?via=matrix.org&via=dentrassi.de 

image

… here is the PR to allow settings labels. I think this makes sense in any way. We can still add the ability to set initial labels once we find the best way to pass them to the backend.

### DIFF
--- a/entity/src/labels.rs
+++ b/entity/src/labels.rs
@@ -11,6 +11,7 @@ use std::ops::{Deref, DerefMut};
     serde::Serialize,
     serde::Deserialize,
     sea_orm::FromJsonQueryResult,
+    utoipa::ToSchema,
 )]
 pub struct Labels(pub HashMap<String, String>);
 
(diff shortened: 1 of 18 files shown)


entity/src/labels.rs
```diff
@@ -31,6 +32,36 @@ impl Labels {
...
+        self
+    }
+
+    /// Apply a label update.
+    ///
+    /// This will apply the provided update to the current set of labels. Updates with an empty
+    /// value will remove the label.
+    pub fn apply(mut self, update: Labels) -> Self {
+        for (k, v) in update.0 {
+            if v.is_empty() {
```
bobmcwhirter, line 53: question: Given it's a String, `is_empty()` is the empty string, yeah?
Would `HashMap<String,Option<String>>` be more straight-forward? Allow for value-less labels? Just thinking how sometimes in other domains we hit things like "ensure $FOO_ENV is set to a non-empty string" and that results in "useless" values like `1` or `true` being used, even though their value isn't ever considered.

modules/fundamental/src/advisory/endpoints/test.rs
```diff
@@ -584,3 +586,54 @@ async fn download_advisory_by_id(ctx: TrustifyContext) -> Result<(), anyhow::Err
...
+
+/// Test setting labels
+#[test_context(TrustifyContext, skip_teardown)]
+#[test(actix_web::test)]
+async fn set_labels(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
+    with_upload(ctx, |result, app| {
+        Box::pin(async move {
+            // update labels
+
+            let request = TestRequest::patch()
```
bobmcwhirter, line 598: nice: PATCH

Roy would be proud.

bobmcwhirter (approved):

entity/src/labels.rs
ctron: Yes, I considered that as well. However, you would one properly serialize that using query parameters? That felt overly complex for having the ability to use an empty string as value.

If we feel that's super important. Ok. Otherwise maybe it's a good compromise.

bobmcwhirter: Nope, that's a valid response. Sally forth.

PR #479, opened 2024-06-28
Move vulnerability table to use the identifier as the natural primary…

bobmcwhirter:
… key.


ctron (changes requested):
Look good in general. What I don't like is renaming the column from `_id` to `_identifier`. The meaning seems the same. Yet we seem to only use `_id`. Unless there's no chance to get the migration right, I think we should use `_id`.

bobmcwhirter:
I waffled on that. We can move to _id

PR #488, opened 2024-07-02
refactor: use UUIDs as primary keys instead of integers

ctron:
Discussion: https://github.com/trustification/trustify/discussions/459

This ignores migrations and simply replaces the type. That's on purpose as explained in the discussion. It might be possible to do this actually, however that would involve:

* Disabling foreign key constraints being affected
* Deleting affected indexes
* Creating a new column with the UUID type
* Converting all data from i32 to UUID, using a UUID v5 with a namespace
  * Which requires loading a postgres extension
  * or doing the work in Rust, as part of the migration
* Re-creating foreign key constraints
* Re-creating indexes

And the same for the `down` migration. I just don't think it's worth the effort.

### DIFF
--- a/entity/src/advisory.rs
+++ b/entity/src/advisory.rs
@@ -18,7 +18,7 @@ pub struct Model {
     pub id: Uuid,
     #[graphql(name = "name")]
     pub identifier: String,
-    pub issuer_id: Option<i32>,
+    pub issuer_id: Option<Uuid>,
     pub sha256: String,
     pub sha384: Option<String>,
     pub sha512: Option<String>,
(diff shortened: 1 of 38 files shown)


bobmcwhirter:
I generally approve of this retroactive rewriting of migrations, having done an i32->UUID migration in the past, but would like to defer to @jcrossley3 if he has any opinions.

jcrossley3:
> I generally approve of this retroactive rewriting of migrations, having done an i32->UUID migration in the past, but would like to defer to @jcrossley3 if he has any opinions.

I don't, other than I'd personally like to see "BREAKS MIGRATIONS" in the first line of any commit message that does.

ctron:
I went with the `BREAKING-CHANGE` modifier (plus description) from conventional commits.

bobmcwhirter (approved):

PR #493, opened 2024-07-02
feat: ingest products from sboms

dejanb:
This PR will create organizations, products and products version entities from the packages that SBOM describes.


modules/ingestor/src/graph/sbom/spdx.rs
```diff
@@ -101,30 +156,6 @@ impl SbomContext {
 
         creator.create(&db).await?;
 
-        // prepare relationships
```
ctron, line 104: Why do we remove relationships?

ctron: Ah I see it was moved to the top.

dejanb: Yeah, it's just moved above packages, so we can use ids to act on the package processing

migration/src/m0000300_create_product_version.rs
```diff
@@ -55,7 +55,8 @@ impl MigrationTrait for Migration {
                     .table(ProductVersion::Table)
                     .name(INDEX_BY_PID_V)
                     .if_not_exists()
-                    .unique()
+                    //TODO what about duplicate version for the same product?
```
bobmcwhirter, line 58: I suppose if duplicate, just use the existing row?  upsert behaviour, instead of duplicates?

modules/ingestor/src/graph/product/mod.rs
```diff
@@ -35,10 +37,16 @@ impl<'g> ProductContext<'g> {
             version: Set(version.clone()),
         };
 
-        Ok(ProductVersionContext::new(
-            self,
-            model.insert(&self.graph.connection(&tx)).await?,
-        ))
+        //TODO: what about duplicate versions?
+        let product_version =
```
bobmcwhirter, line 41: So here's where we'd have graph.ingest_product(...) that would either fetch an existing (and wrap it in a Context) or insert a new (and wrap it in a Context).  Stick the de-dupe logic into graph::ingest_product().

"ingest" is basically "upsert" in graph-speak, confusingly.

migration/src/m0000300_create_product_version.rs
dejanb: Yeah, this uniqueness works fine in manual testing, but it makes some tests failure. It was just left here to investigate these failures further and see if it is a valid use case to have these duplicate values.

modules/ingestor/src/graph/product/mod.rs
dejanb: Added some logic to check for the existing version and update the sbom if needed. I hoped that I could have done it only on conflict, but it seems sea orm is not yet there.

modules/ingestor/src/graph/sbom/spdx.rs
```diff
@@ -82,6 +119,24 @@ impl SbomContext {
...
+
+            if product_packages.contains(&package.package_spdx_identifier) {
+                let pr = self
+                    .graph
+                    .ingest_product(
+                        package.package_name.clone(),
+                        ProductInformation {
+                            vendor: package.package_supplier.clone(),
+                        },
+                        (),
```
ctron, line 131: Shouldn't that be part of the transaction?

modules/ingestor/src/graph/sbom/cyclonedx.rs
```diff
@@ -80,6 +81,22 @@ impl SbomContext {
...
         if let Some(metadata) = &sbom.metadata {
             if let Some(component) = &metadata.component {
+                let pr = self
+                    .graph
+                    .ingest_product(
+                        component.name.clone(),
+                        ProductInformation {
+                            vendor: component.publisher.clone().map(|p| p.to_string()),
+                        },
+                        (),
```
ctron, line 91: Shouldn't that be part of the transaction?

migration/src/m0000300_create_product_version.rs
```diff
@@ -54,6 +54,7 @@ impl MigrationTrait for Migration {
                     .unique()
                     .col(ProductVersion::ProductId)
                     .col(ProductVersion::Version)
+                    .col(ProductVersion::SbomId)
```
ctron, line 57: I think the consensus right now is to not modify the original migration, but to create a new one, altering the table structure. To me it looks like that migration already existed before this PR.

ctron (changes requested):

modules/ingestor/src/graph/sbom/spdx.rs
dejanb: Fixed. Thanks for spotting it

migration/src/m0000300_create_product_version.rs
dejanb: Moved it to the new migration that drops and recreates the index

ctron: But doesn't this PR still contain the additional column?

dejanb: Forgot to remove it 🤦

ctron (approved):

PR #495, opened 2024-07-02
Move /api/v1/package to /api/v1/purl.

bobmcwhirter:
Rename DTOs, adjust fields to be purl-centric instead of package-centric. Allow `/api/v1/purl/{URL_ENCODED_PURL}` to work as well as `/api/v1/purl/{UUID}`.


jcrossley3 (commented):
I'm not convinced we need this resource, whether we call it "package" or "purl". I see it as a filter (of SBOM's and Vulnerabilities/Advisories) instead of a resource.

ctron (approved):
Yay!

PR #496, opened 2024-07-02
feat: using xz2 instead of rust-lzma

helio-frota:
To be able to update postgresql-embedded version in future

### DIFF
--- a/integration-tests/src/stream.rs
+++ b/integration-tests/src/stream.rs
@@ -1,17 +1,26 @@
 #![cfg(test)]
 
 use bytes::Bytes;
-use lzma::LzmaError;
 use std::convert::Infallible;
-
+use std::io::Error;
+use std::io::ErrorKind;
+use std::io::Read;
 use test_context::futures::{stream, Stream};
+use xz2::read::XzDecoder;
 
 /// Take a slice, xa decompress the data and return it as a stream.
 ///
 /// If the decompression fails, that will result in an error on the stream.
-pub fn xz_stream(data: &[u8]) -> impl Stream<Item = Result<Bytes, LzmaError>> {
-    let result = lzma::decompress(data).map(|data| data.into());
-    stream::once(async { result })
+pub fn xz_stream(data: &[u8]) -> impl Stream<Item = Result<Bytes, Error>> {
+    let mut result = XzDecoder::new(data);
+    let mut extracted_content = Vec::new();
+    match result.read_to_end(&mut extracted_content) {
+        Ok(_) => (),
+        Err(e) => {
+            Error::new(ErrorKind::Other, e);
+        }
+    }
+    stream::once(async { Ok(Bytes::from(extracted_content)) })
 }
 
 /// Create a stream from a static BLOB.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3398,6 +3398,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "lzma-sys"
+version = "0.1.20"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5fda04ab3764e6cde78b9974eec4f779acaba7c4e84b36eca3cf77c581b85d27"
+dependencies = [
+ "cc",
+ "libc",
+ "pkg-config",
+]
+
+[[package]]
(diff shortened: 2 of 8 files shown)


helio-frota:
in advance... the code I'm less confident is the function `pub fn xz_stream` (that is called by other functions...)
but the tests are passing so ... :shrug:

integration-tests/src/stream.rs
```diff
@@ -1,17 +1,20 @@
...
 /// Take a slice, xa decompress the data and return it as a stream.
 ///
 /// If the decompression fails, that will result in an error on the stream.
-pub fn xz_stream(data: &[u8]) -> impl Stream<Item = Result<Bytes, LzmaError>> {
-    let result = lzma::decompress(data).map(|data| data.into());
-    stream::once(async { result })
+pub fn xz_stream(data: &[u8]) -> impl Stream<Item = Result<Bytes, Error>> {
+    let mut result = XzDecoder::new(data);
+    let mut extracted_content = Vec::new();
+    let _ = result.read_to_end(&mut extracted_content);
```
ctron, line 16: I don't think we should ignore the error.

ctron (changes requested):
I am not sure why this is required. rust-lzma does its job and is about 1 year old.

xz2 on the other hand is over two years without a release. If there's a good reason to upgrade, then maybe to `liblzma`, which has a recent release, is a fork of `xz2` and seems to support newer versions.

helio-frota:
> I am not sure why this is required. 

When trying to update postgresql-embedded to 0.13
this situation came up https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key

> Primarily, Cargo requires that there is at most one package per links value. In other words, it is forbidden to have two packages link to the same native library.

as postgresql-embedded is using xz2 and we are using rust-lzma

helio-frota:
>  If there's a good reason to upgrade, then maybe to liblzma, which has a recent release, is a fork of xz2 and seems to support newer versions.

Thanks for the extra info :+1: 

When using liblzma we can see the same error, when updating postgresql-embedded

```
error: failed to select a version for `lzma-sys`.
    ... required by package `xz2 v0.1.7`
    ... which satisfies dependency `xz2 = "^0.1.7"` of package `postgresql_archive v0.13.0`
    ... which satisfies dependency `postgresql_archive = "^0.13.0"` of package `postgresql_embedded v0.13.0`
    ... which satisfies dependency `postgresql_embedded = "^0.13.0"` of package `trustify-trustd v0.1.0-alpha.10 (/home/heliofrota/Desktop/qqqqqq/trustify/trustd)`
versions that meet the requirements `^0.1.18` are: 0.1.20, 0.1.19

the package `lzma-sys` links to the native library `lzma`, but it conflicts with a previous package which links to `lzma` as well:
package `liblzma-sys v0.3.4`
    ... which satisfies dependency `liblzma-sys = "^0.3.4"` of package `liblzma v0.3.2`
    ... which satisfies dependency `liblzma = "^0.3.2"` of package `trustify-integration-tests v0.1.0-alpha.10 (/home/heliofrota/Desktop/qqqqqq/trustify/integration-tests)`
```

ctron:
That's a mess. Maybe there's a way to have `postgresql-embedded` without xz support? Via a feature?

helio-frota:
> Maybe there's a way to have postgresql-embedded without xz support? Via a feature?

Apparently currently no.
https://github.com/theseus-rs/postgresql-embedded/tree/0e317fb192db3f8b9e5e7e71cb3779a58e7deb4d/postgresql_archive#feature-flags

But maybe we could ask for the author to create that optional ?

This is being used here https://github.com/theseus-rs/postgresql-embedded/blob/0e317fb192db3f8b9e5e7e71cb3779a58e7deb4d/postgresql_archive/src/configuration/zonky/extractor.rs#L12

Apparently a new thing in case users want to use for tests https://github.com/zonkyio/embedded-postgres-binaries

ctron:
> But maybe we could ask for the author to create that optional ?

I think that makes sense. From what I saw, that change was only 3 days ago.

helio-frota:
closing!

https://github.com/theseus-rs/postgresql-embedded/issues/98#issuecomment-2206912376

PR #502, opened 2024-07-03
feat: allow canceling import runs

ctron:


ctron:
I think what might make sense for the UI is to no override the current state with the "disabled" badge. When the configuration is disabled, the importer might run up until the next point of checking. So it would be disabled but still running. At least for up to 1 minute.

bobmcwhirter:
Can we have a "stopping" state?

ctron:
> Can we have a "stopping" state?

Tricky. I would say: disabled + running => stopping.

The problem is that there is no state machine. Adding a state machine like that would mean to use op-locks in addition, in order to advance the state. And what happens when the job gets re-enabled during "stopping".

Right now its a job running, and a "disabled" flag. One is controlled by the server, the other by the user. And that's it.

carlosthe19916:
- From the REST point of view, how is the "cancel" a running import will work? I do not see new endpoints.
   - Or does it mean that I can use `PUT /api/importer/{name}` and set `state=cancel` or similar action? If this is the way, does it mean I can also do `PUT state=running|waiting,etc`?

ctron:
When you disable it, it will be canceled when it's running.

bobmcwhirter (approved):

PR #507, opened 2024-07-04
feat: add more data to the products API

dejanb:
This PR adds more data to product API, such as vendor and versions for the product

### DIFF
--- a/modules/fundamental/src/product/model/mod.rs
+++ b/modules/fundamental/src/product/model/mod.rs
@@ -2,12 +2,11 @@ use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 use uuid::Uuid;
 
-mod summary;
-pub use summary::*;
+pub mod summary;
 
 use crate::Error;
 use trustify_common::db::ConnectionOrTransaction;
-use trustify_entity::product;
+use trustify_entity::{product, product_version};
 
 #[derive(Serialize, Deserialize, Debug, Clone, ToSchema)]
 pub struct ProductHead {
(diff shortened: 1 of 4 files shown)


modules/fundamental/src/product/model/mod.rs
```diff
@@ -26,3 +25,36 @@ impl ProductHead {
         })
     }
 }
+
+#[derive(Serialize, Deserialize, Debug, Clone, ToSchema)]
+pub struct ProductVersionHead {
+    pub id: Uuid,
+    pub version: String,
+    pub sbom_id: Option<Uuid>,
```
ctron, line 33: We should add `#[serde(default, skip_serializing_if="Option::is_none")]`. Same for the others.

ctron (changes requested):

modules/fundamental/src/product/model/mod.rs
dejanb: Thanks, I missed that

modules/fundamental/src/product/model/summary.rs
```diff
@@ -1,15 +1,20 @@
...
 use trustify_common::paginated;
-use trustify_entity::product;
+use trustify_entity::{organization, product, product_version};
 use utoipa::ToSchema;
 
 #[derive(Serialize, Deserialize, Clone, Debug, ToSchema)]
 pub struct ProductSummary {
     #[serde(flatten)]
     pub head: ProductHead,
+    pub versions: Vec<ProductVersionHead>,
```
ctron, line 15: I think it makes sense adding the serde (default & skip) here too.

ctron (approved):

PR #512, opened 2024-07-05
Move TrustifyContext to its own module to avoid some circularity.

bobmcwhirter:
Add capabilities:
- Holds a Graph, the DB, a backend storage, and an ingestor
- Allow easy ingestion of docs from `etc/test-data/`


bobmcwhirter:
image

This is the primary change, if you can sort through the noise. Making it easier to ingest documents for test setup.

jcrossley3 (commented):
does this obviate the `ingest` helper in modules/fundamental/src/sbom/endpoints/test.rs, too? Looks like you could do the same thing you did for the advisory endpoints there.

bobmcwhirter:
It should do yeah.

modules/fundamental/src/advisory/endpoints/test.rs
```diff
@@ -375,10 +353,11 @@ async fn search_advisories(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
...
     // ingest some advisories
-    let data = include_bytes!("../../../../../etc/test-data/mitre/CVE-2024-27088.json");
-    let _id = ingest(&ingestor, data).await;
-    let data = include_bytes!("../../../../../etc/test-data/mitre/CVE-2024-28111.json");
-    let _id = ingest(&ingestor, data).await;
+    ctx.ingest_documents([
+        (Format::CVE, "mitre/CVE-2024-27088.json"),
+        (Format::CVE, "mitre/CVE-2024-28111.json"),
+    ])
+    .await?;
```
jcrossley3, line 360: Why pass the `Format`? Kinda defeats the whole purpose of `Format::from_bytes`, which is the code path that should be tested, imo.

jcrossley3 (changes requested):

bobmcwhirter:
I have a Stream. Maybe it's rewindable?

modules/fundamental/src/advisory/endpoints/test.rs
bobmcwhirter: Ingestor's ingest() takes a Format. We should make a variant that doesn't require it. try_ingest() maybe.

modules/fundamental/src/advisory/endpoints/test.rs
```diff
@@ -49,33 +44,19 @@ where
...
 
-async fn ingest(service: &IngestorService, data: &[u8]) -> IngestResult {
-    use trustify_module_ingestor::service::Format;
-    service
-        .ingest(
-            ("source", "unit-test"),
-            Some("Capt Pickles Industrial Conglomerate".to_string()),
-            Format::from_bytes(data).unwrap(),
-            ReaderStream::new(data),
-        )
```
jcrossley3, line 60: If you change the parameter of `ingest_documents` to a collection of &[u8]'s? You can use this logic to detect the `Format` and create a `Stream`...

jcrossley3, line 60: If you change the parameter of `ingest_documents` to a collection of &[u8]'s? You can use this logic to detect the `Format` and create a `Stream`...

modules/fundamental/src/advisory/endpoints/test.rs
```diff
@@ -375,10 +353,11 @@ async fn search_advisories(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
...
 
     // ingest some advisories
-    let data = include_bytes!("../../../../../etc/test-data/mitre/CVE-2024-27088.json");
-    let _id = ingest(&ingestor, data).await;
-    let data = include_bytes!("../../../../../etc/test-data/mitre/CVE-2024-28111.json");
-    let _id = ingest(&ingestor, data).await;
+    ctx.ingest_documents([
+        (Format::CVE, "mitre/CVE-2024-27088.json"),
+        (Format::CVE, "mitre/CVE-2024-28111.json"),
+    ])
```
jcrossley3, line 359: This could become:
```suggestion
    ctx.ingest_documents([
        include_bytes!("../../../../../etc/test-data/mitre/CVE-2024-27088.json"),
        include_bytes!("../../../../../etc/test-data/mitre/CVE-2024-28111.json"),
    ])
```

jcrossley3 (changes requested):

modules/fundamental/src/advisory/endpoints/test.rs
helio-frota: +1 compile time file content load

bobmcwhirter: Less worried about compile time. I hate the ../../../.. ish.

jcrossley3: Ok, let's make this work:
```
ctx.ingest_documents([
    "mitre/CVE-2024-27088.json",
    "mitre/CVE-2024-28111.json",
])

File: test-context/src/lib.rs [hunk 4]
```diff
@@ -0,0 +1,151 @@
...
+        for path in paths {
+            //ingestor.ingest((), None, Format::from_bytes(bytes), bytes)
+            let path = test_data.join(path);
+            let mut file = tokio::fs::File::open(path).await?;
+            let mut bytes = Vec::new();
+            file.read_to_end(&mut bytes).await?;
+
+            let format = Format::from_bytes(&bytes)?;
+
+            let stream = stream::iter([Ok::<_, Infallible>(Bytes::copy_from_slice(&bytes))]);
```
jcrossley3, line 43: I think you can just do this:

```suggestion
            let stream = ReaderStream::new(bytes.as_ref());
```

jcrossley3 (changes requested):
Can you remove the `ingest` fn in `modules/fundamental/src/sbom/endpoints/test.rs` as a part of this PR, too? Seems odd to exclude it.

jcrossley3:
> Can you remove the `ingest` fn in `modules/fundamental/src/sbom/endpoints/test.rs` as a part of this PR, too? Seems odd to exclude it.

@bobmcwhirter nevermind. after looking at it more closely, it doesn't appear to be a trivial change.

test-context/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+
+impl TrustifyContext {
+    pub async fn ingest_documents<'a, P: IntoIterator<Item = &'a str>>(
+        &self,
+        paths: P,
+    ) -> Result<(), anyhow::Error> {
+        let workspace_root = find_workspace_root()?;
+        let test_data = workspace_root.join("etc").join("test-data");
+
+        for path in paths {
```
jcrossley3, line 34: What would be handy (for the sbom test I asked you to change, in particular) would be to return a mapping of path to `IngestResult`.

bobmcwhirter: Yep, done!

Added results when multi-ingest (indexed by input path), and also a single simple route for a single-doc ingest.

modules/fundamental/src/advisory/endpoints/test.rs
bobmcwhirter: Done

test-context/src/lib.rs
bobmcwhirter: Thanks!

test-context/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+            results.insert(
+                path.to_string(),
+                self.ingest_document_inner(&test_data, path).await,
+            );
+        }
+
+        Ok(IngestionResults { results })
+    }
```
jcrossley3, line 59: I don't quite grok the need for introducing `IngestionResults`? I was expecting you'd return something like `paths.map(...).collect()` with a type of `Result<Vec<IngestResult>, _>` or some such.

jcrossley3 (commented):
Snazzy!

test-context/src/lib.rs
bobmcwhirter: Mostly so we could index by path, instead of positional. I can alterate it though. I have no preference.

bobmcwhirter: Simplified as just a Vec<..> per your suggestion.

test-context/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+        let workspace_root = find_workspace_root()?;
+        let test_data = workspace_root.join("etc").join("test-data");
+
+        let mut results = Vec::new();
+
+        for path in paths {
+            results.push(self.ingest_document_inner(&test_data, path).await?);
+        }
+
+        Ok(results)
```
jcrossley3, line 40: I always balk at the "iterate while pushing to another mutable container" pattern, but the futures complicate things. I came up with this, but I'm not sure I like it better:

```suggestion
        let futures = paths
            .into_iter()
            .map(|path| async { self.ingest_document_inner(&test_data, path).await });
        join_all(futures).await.into_iter().collect()
```

jcrossley3 (approved):

PR #515, opened 2024-07-09
Replace `integration-tests` module with conventional rust integs in their relevant modules

jcrossley3:
The `fundamental` module seemed appropriate for most of the integs.

This moves all the integ-specific test data to `etc/test-data` with the rest of it, and relies on the `TrustifyContext` instance passed to each test to find it, whether bytes or stream, and/or ingest it for the test.

This removed some redundancy from the integs (e.g. streams, xz) in light of recent features added to `TrustifyContext`.


bobmcwhirter (approved):

bobmcwhirter:
Thanks for this!

ctron (approved):
Yes, after the refactoring some time ago, that looks like a much improved structure.

PR #518, opened 2024-07-09
Percolate status CPE context into vulnerability endpoint.

bobmcwhirter:
Percolate status CPE context into vulnerability endpoint.
    Fully perform a query to find "related" SBOMs to a vulnerability
    via advisory status.

    Related is any status other than "not_affected" due to the extreme
    amount of noise produced by including that status.

    For an example of the current output at the time of this commit
    from the /api/v1/vulnerability/{identifier} endpoint, see

https://gist.github.com/bobmcwhirter/9bd690ec6f603d28e1c04376fbb8063d


bobmcwhirter:
A vulnerability detail response will now include...

A list of each SBOM related, along with the status of the *particular package* that caused that relation (e.g., "affected", "fixed") with the exception of "not_affected" because that felt noisy.

Happy to mutate the response to include more/less/different information after your review.

bobmcwhirter:
@jcrossley3 and while I didn't mention it in the description, I did sidebar link to the relevant GH issue.

image

bobmcwhirter:
@carlosthe19916 the relevant new stanza is about 10,000 lines down:

https://gist.github.com/bobmcwhirter/9bd690ec6f603d28e1c04376fbb8063d#file-foo-L10579-L10594

modules/fundamental/src/vulnerability/service/test.rs
```diff
@@ -47,3 +47,33 @@ async fn statuses(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
...
+async fn statuses_too(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
+    let service = VulnerabilityService::new(ctx.db.clone());
+
+    ctx.ingest_documents([
+        "cve/CVE-2024-29025.json",
+        "csaf/rhsa-2024-2705.json",
+        "spdx/quarkus-bom-3.2.11.Final-redhat-00001.json",
+        "spdx/quarkus-bom-3.2.12.Final-redhat-00002.json",
+    ])
+    .await?;
```
jcrossley3, line 62: Pretty!

jcrossley3 (approved):

bobmcwhirter:
I have further thoughts I'll stack tomorrow

PR #521, opened 2024-07-09
chore: using cargo build

helio-frota:
Hi folks I found this today during `cargo clean ; cargo build`

```
   Compiling trustify-trustd v0.1.0-alpha.11 (/home/heliofrota/Desktop/tc/trustify/trustd)
error: queries overflow the depth limit!
  |
  = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`trustd`)
  = note: query depth increased by 130 when computing layout of `tracing::instrument::Instrumented<{async block@trustify_module_importer::server::Server::run::{closure#0}::{closure#0}}>`

error: could not compile `trustify-trustd` (bin "trustd") due to 1 previous error
```

> cargo build catches all Rust compilation errors.
cargo check only catches some subset of the possible compilation errors.
https://rust-lang.github.io/rfcs/3477-cargo-check-lang-policy.html#decision

PR #523, opened 2024-07-09
Revert "chore: update telemetry components"

helio-frota:
This reverts commit b478c44582575a479421c57fe82b5ce4f630e085.

* `cargo build` works inside `trustd` dir, but not from workspace root https://gist.github.com/helio-frota/07c753afd5dceb4d65be14509798de9e
* For some reason it works with github actions https://github.com/trustification/trustify/pull/521#issuecomment-2218330154
* Able to build when commenting this line https://github.com/trustification/trustify/blob/main/server/src/lib.rs#L253
  * Comment from @jcrossley3: 
  > commit introduced some threshold that you stopped hitting when you commented out that line, but i'm not sure that's enough


ctron:
As neither me, nor the CI experiences that issue. I am not sure simply reverting that PR is the right approach.

One thing that might be worth trying (and maybe you already did) is a `cargo clean`. 

If it doesn't, then I think that simply following the instructions, raising that limit, makes sense. Indeed we are using a lot of macros, and that limit is a safeguard from Rust to not run into infinite macro recursions. However, if there's a need, then it is ok to raise that limit.

ctron (changes requested):
Let's try finding the cause for this and fixing it.

jcrossley3:
> As neither me, nor the CI experiences that issue. I am not sure simply reverting that PR is the right approach.

CI doesn't run `cargo build`. Did you? Breaks for me and @helio-frota every time.

jcrossley3:
> Let's try finding the cause for this and fixing it.

Sure, but why not do that *before* breaking the build?!?! Why update to a pre-release, anyway? What feature were we missing?

Let's revert this so we don't hold up others while trying to figure out why a pre-release dep broke our build?

helio-frota:
so folks I closed the temporary fix and we end up with revert or increase the limit.
then I decided to merge :sunglasses: 
in case problems we can re-re-revert or something, git supports that :+1:

jcrossley3 (approved):

Issue #526, opened 2024-07-10
advisory_vulnerability lost its primary key


ctron:
It looks like `advisory_vulnerability` lost its primary key in one of the migrations.

ctron:
Due to that fact, it also looks like as if ingesting the same CVE twice will lead to duplicate entries in that table. Similar to #525.

Issue #525, opened 2024-07-10
Re-ingesting a CVE creates additional descriptions


ctron:
Each time a CVE is ingested, it adds all the descriptions again.

PR #527, opened 2024-07-10
Add a test for re-ingesting CVEs and fixing the duplication issues.

ctron:
Add a test for re-ingesting CVEs and fixing the duplication issues.

Based on  #524 (sorry for stacked PRs).

Closes https://github.com/trustification/trustify/issues/526
Closes https://github.com/trustification/trustify/issues/525


modules/fundamental/tests/cve.rs
```diff
@@ -0,0 +1,84 @@
+#![allow(clippy::expect_used)]
+
+use anyhow::bail;
+use futures_util::stream;
+use std::convert::Infallible;
+use test_context::test_context;
+use test_log::test;
+use trustify_common::id::Id;
+use trustify_module_ingestor::service::{Format, IngestorService};
+use trustify_test_context::TrustifyContext;
```
jcrossley3, line 10: For a test to belong in this directory, I think it should import something from this module. This test should probably live beneath `modules/ingestor/tests/`.

modules/fundamental/tests/cve.rs
```diff
@@ -0,0 +1,84 @@
...
+
+#[test_context(TrustifyContext)]
+#[test(tokio::test)]
+async fn reingest_cve(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
+    let ingestor = IngestorService::new(ctx.graph.clone(), ctx.storage.clone());
+
+    let data = ctx.document_bytes("cve/CVE-2021-32714.json").await?;
```
jcrossley3, line 17: You might simplify this test by using `ctx.ingest_document[s]`. That'll return you the id's you need but saves you from creating your own `IngestorService`.

jcrossley3 (changes requested):

modules/fundamental/tests/cve.rs
ctron: Originally I was assuming to use stuff from `fundamental`, but it looks like that's not needed. So I'll move it.

ctron: Moved.

modules/fundamental/tests/cve.rs
ctron: Done.

jcrossley3 (approved):

PR #529, opened 2024-07-10
fix: temporarily avoiding error instrumentation

helio-frota:
For some reason with recent dep updates , when error instrumentation is enabled on server run function
it goes crazy

https://docs.rs/tracing/latest/tracing/attr.instrument.html

### DIFF
--- a/modules/importer/src/server/mod.rs
+++ b/modules/importer/src/server/mod.rs
@@ -59,7 +59,11 @@ struct Server {
 }
 
 impl Server {
-    #[instrument(skip_all, err)]
+    // TODO: revisit this later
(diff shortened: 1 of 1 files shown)


helio-frota:
I've made some other attempts with the macro, adding derive display and debug on the server struct but without success...

jcrossley3:
I think the "fix" is probably to add the `#![recursion_limit = "256"]` attribute to `trustd/src/main.rs` as it appears the latest opentelemetry requires more than the default limit of 128.

But I continue to question why we need to depend on a pre-release of opentelemetry. If we don't, then I'd prefer we continue using the latest stable version.

helio-frota:
closing to reduce the discussion related to revert and limit only :+1:

PR #532, opened 2024-07-11
Ingestion fixes, auto-create vulnerabilities using upsert

ctron:
This PR adds a few tests for re-ingesting documents. It also fixes issues alongside those. Most notably, it auto-creates vulnerabilities and switches the creation of those to an upsert model.

### DIFF
--- a/modules/ingestor/src/graph/organization.rs
+++ b/modules/ingestor/src/graph/organization.rs
@@ -6,6 +6,7 @@ use trustify_entity::organization;
 
 use crate::graph::{error::Error, Graph};
 
+#[derive(Clone, Debug)]
 pub struct OrganizationInformation {
     pub cpe_key: Option<String>,
     pub website: Option<String>,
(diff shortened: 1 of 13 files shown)


modules/ingestor/src/graph/organization.rs
```diff
@@ -40,10 +41,24 @@ impl<'g> OrganizationContext<'g> {
     }
 }
 
-impl super::Graph {
+impl Graph {
+    #[instrument(skip(self, tx), err)]
+    pub async fn get_organizations(
```
bobmcwhirter, line 46: related: I'd still argue we should be deleting get_plurals() and defer that ish to the appropriate fundamental Service.

dejanb (approved):

bobmcwhirter (approved):
We also need some changes in the fundamental VulnerabilityService to avoid the stupid-ass workaround I put there for synthesizing vulnerabilities from the advisory_vuln table when no row existed in the vuln table.

ctron:
@bobmcwhirter I updated the PR and dropped the special handling for the advisory_vulns.

modules/ingestor/src/graph/organization.rs
ctron: Created a tracking issue: https://github.com/trustification/trustify/issues/547

PR #542, opened 2024-07-12
Include vulnerability information in SBOM details.

bobmcwhirter:
`/api/v1/sbom/{id}` produces a fuller SBOM details which includes SBOM -> advisory -> vulnerability information, for all statuses, which may include `fixed`, `not_affected`, `known_affected` etc.

### DIFF
--- a/common/src/db/mod.rs
+++ b/common/src/db/mod.rs
@@ -4,6 +4,8 @@ pub mod chunk;
 pub mod limiter;
 pub mod query;
 
+pub mod multi_model;
+
 use anyhow::Context;
 use migration::{Migrator, MigratorTrait};
 use sea_orm::{
--- a/common/src/db/multi_model.rs
+++ b/common/src/db/multi_model.rs
@@ -1,0 +1,89 @@
+use sea_orm::{
+    ColumnTrait, DbErr, EntityTrait, FromQueryResult, IntoSimpleExpr, Iterable, QueryResult,
+    QuerySelect, Select, SelectModel, Selector,
+};
+use sea_query::{ColumnRef, SimpleExpr};
+
+pub trait ColumnsPrefixed: Sized {
+    fn try_columns_prefixed<C, I>(self, prefix: &str, cols: I) -> Result<Self, DbErr>
+    where
+        C: ColumnTrait,
+        I: IntoIterator<Item = C>;
+}
+
+impl<T: QuerySelect> ColumnsPrefixed for T {
+    fn try_columns_prefixed<C, I>(mut self, prefix: &str, cols: I) -> Result<Self, DbErr>
+    where
+        C: ColumnTrait,
+        I: IntoIterator<Item = C>,
+    {
+        for col in cols.into_iter() {
+            if let SimpleExpr::Column(col_ref) = col.into_simple_expr() {
+                match col_ref {
+                    ColumnRef::Column(name)
+                    | ColumnRef::TableColumn(_, name)
+                    | ColumnRef::SchemaTableColumn(_, _, name) => {
+                        let mut prefixed = String::new();
+                        prefixed.push_str(prefix);
+                        prefixed.push_str(&name.to_string());
+
+                        self = self.column_as(col, prefixed);
+                    }
+                    ColumnRef::Asterisk => {
+                        return Err(DbErr::Custom("Unable to prefix asterisk".to_string()))
+                    }
+                    ColumnRef::TableAsterisk(_) => {
+                        return Err(DbErr::Custom("Unable to prefix asterisk".to_string()))
+                    }
+                }
+            } else {
+                return Err(DbErr::Custom("Unable to prefix column".to_string()));
+            }
+        }
+        Ok(self)
+    }
+}
+
+pub trait SelectIntoMultiModel: Sized {
+    fn try_model_columns<E: EntityTrait>(self, entity: E) -> Result<Self, DbErr>;
+
+    fn try_into_multi_model<M>(self) -> Result<Selector<SelectModel<M>>, DbErr>
+    where
+        M: FromQueryResultMultiModel;
+}
+
+impl<E: EntityTrait> SelectIntoMultiModel for Select<E> {
+    fn try_model_columns<O: EntityTrait>(self, entity: O) -> Result<Self, DbErr> {
+        let name = entity.module_name();
+        let prefix = format!("{name}$");
+        self.try_columns_prefixed(&prefix, O::Column::iter())
+    }
+
+    fn try_into_multi_model<M>(self) -> Result<Selector<SelectModel<M>>, DbErr>
+    where
+        M: FromQueryResultMultiModel,
+    {
+        let select = M::try_into_multi_model(self)?;
+        Ok(select.into_model::<M>())
+    }
+}
+
+pub trait FromQueryResultMultiModel: FromQueryResult {
+    fn try_into_multi_model<E: EntityTrait>(select: Select<E>) -> Result<Select<E>, DbErr>;
+
+    fn from_query_result_multi_model<E: EntityTrait>(
+        res: &QueryResult,
+        entity: E,
+    ) -> Result<E::Model, DbErr> {
+        let name = entity.module_name();
+        let prefix = format!("{name}$");
+        E::Model::from_query_result(res, &prefix)
+    }
+
+    fn from_query_result_multi_model_optional<E: EntityTrait>(
+        res: &QueryResult,
+        entity: E,
+    ) -> Result<Option<E::Model>, DbErr> {
+        Ok(Self::from_query_result_multi_model(res, entity).ok())
+    }
+}
--- a/common/src/db/multi_model.rs
+++ b/common/src/db/multi_model.rs
@@ -23,10 +23,7 @@ impl<T: QuerySelect> ColumnsPrefixed for T {
                     ColumnRef::Column(name)
                     | ColumnRef::TableColumn(_, name)
                     | ColumnRef::SchemaTableColumn(_, _, name) => {
-                        let mut prefixed = String::new();
-                        prefixed.push_str(prefix);
-                        prefixed.push_str(&name.to_string());
-
+                        let prefixed = format!("{prefix}{}", name.to_string());
                         self = self.column_as(col, prefixed);
                     }
                     ColumnRef::Asterisk => {
--- a/entity/src/sbom.rs
+++ b/entity/src/sbom.rs
@@ -36,6 +36,37 @@ pub enum Relation {
     PackageRelatesToPackages,
 }
 
+pub struct SbomPurlsLink;
+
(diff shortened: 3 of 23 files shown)


bobmcwhirter:
Out for SBOM details now looks kinda like https://gist.github.com/bobmcwhirter/5e9b4965fa7a00e32ac95860830a28d8

bobmcwhirter:
Current output

https://gist.github.com/bobmcwhirter/95fcfbbe0a7b2f3659f3e46dfc0b17c4

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -109,16 +114,92 @@ impl SbomService {
...
+            .join(JoinType::Join, purl_status::Relation::Status.def())
+            .join(
+                JoinType::LeftJoin,
+                purl_status::Relation::VersionRange.def(),
+            )
+            .join(JoinType::LeftJoin, purl_status::Relation::ContextCpe.def())
+            .join(JoinType::Join, purl_status::Relation::Advisory.def())
+            .join(JoinType::Join, purl_status::Relation::Vulnerability.def())
+            .select_only()
+            .columns_prefixed("advisory", advisory::Column::iter())
```
ctron, line 170: To my understanding this seems to align (or must be aligned) with the `QueryCatcher`. Wouldn't it make sense to have an associated function on that struct, which performs all the `columns_prefixed` calls? So that logic lives in the same file?

modules/fundamental/src/sbom/model/mod.rs
```diff
@@ -36,11 +45,24 @@ pub struct SbomPackage {
...
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub purl: Vec<String>,
+    pub purl: Vec<SbomPackagePurl>,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub cpe: Vec<String>,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, ToSchema)]
+#[serde(untagged)]
+pub enum SbomPackagePurl {
```
ctron, line 55: I am not sure I am happy with that. I think it would be ok if that would be an input type (allowing more flexible input always seems good). But to my understanding, that's an output type. And I think in that case it would make sense for the caller to understand what can be expected as a return type.

Having a struct returning both purl and details would be fine I guess. But having either this or that, in the same array, might be quite complex.

entity/src/versioned_purl.rs
```diff
@@ -27,6 +27,20 @@ pub enum Relation {
...
+pub struct AdvisoriesLink;
+
+impl Linked for AdvisoriesLink {
+    type FromEntity = Entity;
+    type ToEntity = super::purl_status::Entity;
+
+    fn link(&self) -> Vec<LinkDef> {
+        vec![
+            Relation::BasePurl.def(),
+            //super::purl_status::Relation::BasePurl.def(),
```
ctron, line 39: If it's dead, sweep it under the rug.

entity/src/sbom.rs
```diff
@@ -36,6 +36,40 @@ pub enum Relation {
...
+    type FromEntity = Entity;
+    type ToEntity = super::base_purl::Entity;
+
+    fn link(&self) -> Vec<LinkDef> {
+        vec![
+            Relation::Packages.def(),
+            super::sbom_package::Relation::Purl.def(),
+            super::sbom_package_purl_ref::Relation::Purl.def(),
+            super::qualified_purl::Relation::VersionedPurl.def(),
+            //super::versioned_purl::Relation::BasePurl.def(),
```
ctron, line 66: Either it's dead (and should be gone), or be a "todo" for later.

common/src/db/mod.rs
```diff
@@ -263,3 +265,37 @@ impl DatabaseErrors for DbErr {
...
+}
+
+impl<T: QuerySelect> ColumnsPrefixed for T {
+    fn columns_prefixed<C, I>(mut self, prefix: &str, cols: I) -> Self
+    where
+        C: ColumnTrait,
+        I: IntoIterator<Item = C>,
+    {
+        for col in cols.into_iter() {
+            if let SimpleExpr::Column(col_ref) = col.into_simple_expr() {
```
ctron, line 283: To me it looks like that, if the input doesn't match the functions expectation, it will silently be dropped. That might lead to those runtime issues we'd like to avoid.

I think there should at least some some source level doc explaining this. Even better (from my perspective) would be to go for a `try_` style function, returning an error in a case like that.

ctron (changes requested):
Looks good in general. Some minor things I'd like to discuss or fix.

entity/src/versioned_purl.rs
bobmcwhirter: Remind me to be cautious if you ever acquire a Bob-sized rug...

common/src/db/mod.rs
bobmcwhirter: +1

bobmcwhirter:
@carlosthe19916 this PR is SBOM->vulns

carlosthe19916:
@bobmcwhirter I can not judge the code but can we fix this?

image

I was trying to figure out how/where the SBOM-Vuln fields are located

bobmcwhirter:
Yep, will do.

bobmcwhirter:
@carlosthe19916 Schema updated.

Also not, SBOM packages API is now returning our more-robust Purl object structures (along with the string purl), instead of just the string purl. Might affect the UI.

modules/fundamental/src/sbom/model/mod.rs
bobmcwhirter: Sorted the coalesce, removed `SbomPackagePurl` enum...

common/src/db/multi_model.rs
```diff
@@ -0,0 +1,89 @@
...
+        for col in cols.into_iter() {
+            if let SimpleExpr::Column(col_ref) = col.into_simple_expr() {
+                match col_ref {
+                    ColumnRef::Column(name)
+                    | ColumnRef::TableColumn(_, name)
+                    | ColumnRef::SchemaTableColumn(_, _, name) => {
+                        let mut prefixed = String::new();
+                        prefixed.push_str(prefix);
+                        prefixed.push_str(&name.to_string());
+
```
jcrossley3, line 29: Turns out I hate `mut` more than macros... who knew? :woman_shrugging: 
```suggestion
                        let prefixed = format!("{prefix}{}", name.to_string());
```

carlosthe19916:
@bobmcwhirter This is a sample (I removed numerous fields to just focus on one) of the response I get from `GET /api/v1/sbom/{id}`

```json
{
  "advisories": [
    {
      "status": [
        {
          "vulnerability_id": "CVE-2024-2700",
          "status": "not_affected",
          "context": {
            "cpe": "cpe:/a:redhat:quarkus:3.2:*:el8:*"
          },
          "packages": [
            {
              "id": "SPDXRef-734f59cd-d145-49a3-9016-bec9b93c05c2",
              "name": "quarkus-kubernetes-service-binding-spi",
              "version": "3.2.11.Final-redhat-00001",
              "purl": [ // Why the PURL is an array and not an object?
                {
                  "uuid": "36d91938-b758-5abe-93f8-3a9e4e6472d9",
                  "purl": "pkg://maven/io.quarkus/quarkus-kubernetes-service-binding-spi@3.2.11.Final-redhat-00001?repository_url=https://maven.repository.redhat.com/ga/&type=jar",
                  "base": {
                    "uuid": "218cb45a-f09c-5bfb-b66e-00a3a8fdae95",
                    "purl": "pkg://maven/io.quarkus/quarkus-kubernetes-service-binding-spi"
                  },
                  "version": {
                    "uuid": "d2a79cb3-1a05-5df8-be90-afc82e3d4792",
                    "purl": "pkg://maven/io.quarkus/quarkus-kubernetes-service-binding-spi@3.2.11.Final-redhat-00001",
                    "version": "3.2.11.Final-redhat-00001"
                  },
                  "qualifiers": {
                    "repository_url": "https://maven.repository.redhat.com/ga/",
                    "type": "jar"
                  }
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

Why the `advisories.status.packages.purl` is an array and not an object? The parent object already has the fields `name` & `version` which indicates to me that we are talking about a single package and a single Purl (not an array of purls). If there were more than one Purl don't you think that the field `advisories.status.packages.name` and `advisories.status.packages.version` would not be accurate?

bobmcwhirter:
I honestly don't know the answer to that. It's using the `SbomPackage` structure you'd get from `/api/v1/sbom/{id}/packages` I believe. @ctron may be able to say why it's an array. I suspect because an SBOM package may include multiple pURLs, but I'm unsure.

This is where I'm trying to be consistent, re-using existing SBOM package DTOs, so they could be similarly displayed regardless of the context.

But, that being said, I don't understand the plurality here either :)

bobmcwhirter:
The `name` and `version` of the parent package are the name and version of the _package_ itself, which may or may not include zero or more pURLs and/or CPEs, so while similar-appearing, are actually distinct bits of data.

bobmcwhirter:
I would venture to say, though, that if the `purl` field is an array, perhaps it should be pluraled to at least be `purls`.

carlosthe19916:
```json
{
  "advisories": [
    {
      "status": [
        {
           "packages": [
            {
              "id": "SPDXRef-734f59cd-d145-49a3-9016-bec9b93c05c2",
              "name": "quarkus-kubernetes-service-binding-spi",
              "version": "3.2.11.Final-redhat-00001", // THIS VERSION
              "purl": [
                {
                  "uuid": "36d91938-b758-5abe-93f8-3a9e4e6472d9",
                  "purl": "pkg://maven/io.quarkus/quarkus-kubernetes-service-binding-spi@3.2.11.Final-redhat-00001?repository_url=https://maven.repository.redhat.com/ga/&type=jar",                  
                  "version": {
                    "uuid": "d2a79cb3-1a05-5df8-be90-afc82e3d4792",
                    "purl": "pkg://maven/io.quarkus/quarkus-kubernetes-service-binding-spi@3.2.11.Final-redhat-00001",
                    "version": "3.2.11.Final-redhat-00001" // IS THIS VERSION, what if the array has more than one element?
                  },
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

I think the internal models are being leaked on the rest endpoints and that is the reason of having an array but I might be wrong.

I suggest leaving this PR as it is and later on in a separate PR polish this kind of things. Please don't let this thing to block this PR

ctron (approved):

ctron:
> I believe. @ctron may be able to say why it's an array. I suspect because an SBOM package may include multiple pURLs, but I'm unsure.

From the CycloneDX spec: https://cyclonedx.org/docs/1.6/json/#components_items_evidence

image


From the SPDX spec: https://spdx.github.io/spdx-spec/v2.3/package-information/#7211-description

image

PR #546, opened 2024-07-15
ci: generate cyclonedx sboms during the build

### LINKED ISSUE #534: Generate Good Quality SBOM for Trustify, continuously.
Bonus points if we generate both SPDX and CycloneDX.

ctron:
Also: use GITHUB_TOKEN for `cargo binstall` to avoid GH API rate limits

Closes #534

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -181,6 +181,8 @@ jobs:
         run: |
           cargo binstall cross -y
 
+      - run: mkdir -p upload
+
       - name: Build | Build
         shell: bash
         env:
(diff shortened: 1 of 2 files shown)


ctron:
Not sure it's a "good" SBOM. But it is one.

bobmcwhirter:
Seems there's some leakage of build machine details?

```
    {
      "type": "library",
      "bom-ref": "path+file:///Users/bob/repos/trustification/trustify/modules/importer#trustify-module-importer@0.1.0-alpha.11",
      "name": "trustify-module-importer",
      "version": "0.1.0-alpha.11",
      "scope": "required",
      "purl": "pkg:cargo/trustify-module-importer@0.1.0-alpha.11?download_url=file%253A%252F%252F..%25252Fmodules%25252Fimporter"
    },
```

Any way to ... make that less leaky?

ctron:
You mean that part?

```
"bom-ref": "path+file:///Users/bob/repos/trustification/trustify/modules/importer#trustify-module-importer@0.1.0-alpha.11",
```

Indeed, that feels a bit weird. But it's something we'd need to address with the cargo plugin. Which is fine, I contributed before.

ctron:
The question is, is that a blocker for us?

bobmcwhirter:
Nah, just a comment.

bobmcwhirter (approved):

PR #550, opened 2024-07-15
fix: remove non-normative data from advisory vulnerability

### LINKED ISSUE #543: /api/v1/vulnerabilities contains duplicate field names
The `non_normative` fields are mentioned in the parent json field already. See image below.

Screenshot from 2024-07-14 18-29-16

jcrossley3:
Fixes #543

If we have the CVE ingested, i.e. a Vulnerability record, we'll use its data and declare it "normative". Otherwise, we'll use the non-normative data from the advisory.

### DIFF
--- a/modules/fundamental/src/advisory/model/details/advisory_vulnerability.rs
+++ b/modules/fundamental/src/advisory/model/details/advisory_vulnerability.rs
@@ -10,7 +10,6 @@ use utoipa::ToSchema;
 pub struct AdvisoryVulnerabilityHead {
     #[serde(flatten)]
     pub head: VulnerabilityHead,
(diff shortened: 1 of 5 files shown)


bobmcwhirter:
What if multiple advisories?

jcrossley3:
> What if multiple advisories?

Since we're returning a list of advisories, I didn't think it relevant. If the Vulnerability doesn't exist, you get the Advisory.

jcrossley3:
I don't think the previous DTO handled multiple advisories anyway, did it?

bobmcwhirter:
Some do. We can sort that later.

bobmcwhirter (approved):

PR #557, opened 2024-07-17
Some minor improvements

ctron:
While working on other stuff, I made some minor improvements. Maybe it makes sense to merge those anyway.

The `strum` stuff isn't used in this PR, but in the other one that I am working on. One can do `Relationships::VARIANTS` to get all of them as an input to the search function.


jcrossley3 (approved):

ctron:
Hm, interesting. IIRC `strum` has the `derive` feature, which we set unconditionally. Not sure what goes wrong.

helio-frota:
build fixed after adding derive feature, thanks

PR #569, opened 2024-07-19
chore: avoiding relative path

helio-frota:
Using .cargo/config.toml creating an env var
called WORKSPACE_ROOT

### DIFF
--- a/modules/fundamental/src/advisory/endpoints/test.rs
+++ b/modules/fundamental/src/advisory/endpoints/test.rs
@@ -30,7 +30,7 @@ use trustify_module_ingestor::{
     model::IngestResult,
 };
 use trustify_module_storage::service::fs::FileSystemBackend;
-use trustify_test_context::TrustifyContext;
+use trustify_test_context::{document_bytes, TrustifyContext};
 use uuid::Uuid;
 
 async fn query<S, B>(app: &S, q: &str) -> PaginatedResults<AdvisorySummary>
@@ -384,7 +384,8 @@ async fn upload_default_csaf_format(ctx: TrustifyContext) -> Result<(), anyhow::
             .service(web::scope("/api").configure(|svc| configure(svc, db, storage)));
     }))
     .await;
-    let payload = include_str!("../../../../../etc/test-data/csaf/cve-2023-33201.json");
+
+    let payload = document_bytes("csaf/cve-2023-33201.json").await?;
 
     let uri = "/api/v1/advisory";
     let request = TestRequest::post()
@@ -410,7 +411,7 @@ async fn upload_osv_format(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
         web::scope("/api").configure(|svc| crate::endpoints::configure(svc, db, storage)),
     ))
     .await;
-    let payload = include_str!("../../../../../etc/test-data/osv/RUSTSEC-2021-0079.json");
+    let payload = document_bytes("osv/RUSTSEC-2021-0079.json").await?;
 
     let uri = "/api/v1/advisory";
     let request = TestRequest::post()
@@ -436,7 +437,7 @@ async fn upload_cve_format(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
         web::scope("/api").configure(|svc| crate::endpoints::configure(svc, db, storage)),
     ))
     .await;
-    let payload = include_str!("../../../../../etc/test-data/mitre/CVE-2024-27088.json");
+    let payload = document_bytes("mitre/CVE-2024-27088.json").await?;
 
     let uri = "/api/v1/advisory";
     let request = TestRequest::post()
@@ -493,7 +494,7 @@ async fn upload_with_labels(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
             .service(web::scope("/api").configure(|svc| configure(svc, db, storage)));
     }))
     .await;
-    let payload = include_str!("../../../../../etc/test-data/csaf/cve-2023-33201.json");
+    let payload = document_bytes("csaf/cve-2023-33201.json").await?;
 
     let uri = "/api/v1/advisory?labels.foo=bar&labels.bar=baz";
     let request = TestRequest::post()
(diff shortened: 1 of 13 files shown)


modules/fundamental/src/advisory/endpoints/test.rs
```diff
@@ -384,7 +384,11 @@ async fn upload_default_csaf_format(ctx: TrustifyContext) -> Result<(), anyhow::
             .service(web::scope("/api").configure(|svc| configure(svc, db, storage)));
     }))
     .await;
-    let payload = include_str!("../../../../../etc/test-data/csaf/cve-2023-33201.json");
+
+    let payload = include_str!(concat!(
+        env!("WORKSPACE_ROOT"),
+        "/etc/test-data/csaf/cve-2023-33201.json"
+    ));
```
jcrossley3, line 391: I'd prefer we use TrustifyContext, which is already capable of finding our test data:

```suggestion
    let payload = TrustifyContext::document_bytes("csaf/cve-2023-33201.json");
```

jcrossley3 (changes requested):
I appreciate the spirit and intent of this. But I'd like you to do it slightly differently.

First, I'd rather not add `WORKSPACE_ROOT` because we already have [`TrustifyContext::find_workspace_root`](https://github.com/trustification/trustify/blob/ea568e03aaaf2205db7d9b217c39d4bb0ae2a503/test-context/src/lib.rs#L91-L104)

And I'd like to remove not only all the `include_str!` calls, but all the `include_bytes!` calls as well. Do a grep for `etc/test-data` in our source tree for all the places we can refactor to use our new methods in `TrustifyContext`.

To do so, you'll need to convert its `document_bytes` and `document_stream` functions to not be methods, i.e. remove their `&self` parameter, which they don't use anyway. You'll also need to change calls to `self.document_bytes` in that file to `Self::document_bytes`.

Make sense?

modules/fundamental/src/advisory/endpoints/test.rs
jcrossley3: I'm not sure we even need `document_{bytes,stream}` to be associated with `TrustifyContext`. We could bring them outside its impl (just like `find_workspace_root`) and import them directly when needed, e.g.
```
use trustify_test_context::document_bytes;
let payload = document_bytes("csaf/cve-2023-33201.json");
```

helio-frota:
pr updated but we still have these remaining cases that I was not able to find a way to fix yet:

```
➜  trustify git:(workspace-root) rg "etc/test-data/"
modules/ingestor/src/service/sbom/spdx.rs
81:        let data = include_bytes!("../../../../../etc/test-data/ubi9-9.2-755.1697625012.json");

modules/ingestor/src/service/sbom/cyclonedx.rs
87:        let data = include_bytes!("../../../../../etc/test-data/zookeeper-3.9.2-cyclonedx.json");

modules/ingestor/src/service/advisory/cve/loader.rs
180:        let data = include_bytes!("../../../../../../etc/test-data/mitre/CVE-2024-28111.json");

modules/ingestor/src/service/advisory/osv/loader.rs
304:        let data = include_bytes!("../../../../../../etc/test-data/osv/RUSTSEC-2021-0079.json");

modules/ingestor/src/service/advisory/csaf/loader.rs
195:        let data = include_bytes!("../../../../../../etc/test-data/csaf/CVE-2023-20862.json");
276:        let data = include_bytes!("../../../../../../etc/test-data/csaf/rhsa-2024_3666.json");

modules/fundamental/src/advisory/endpoints/test.rs
532:const DOC: &[u8] = include_bytes!("../../../../../etc/test-data/csaf/cve-2023-33201.json");
```

modules/fundamental/src/advisory/endpoints/test.rs
helio-frota: this part is looking cool now, I added 2 basic tests just to make sure it works :+1:

jcrossley3:
> pr updated but we still have these remaining cases that I was not able to find a way to fix yet:

You may need to massage things a bit, but you should be able to do it. Here's an example:

```diff
modified   modules/ingestor/src/service/sbom/spdx.rs
@@ -78,7 +78,7 @@ mod test {
     async fn ingest_spdx(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
         let db = ctx.db.clone();
         let graph = Graph::new(db);
-        let data = include_bytes!("../../../../../etc/test-data/ubi9-9.2-755.1697625012.json");
+        let data = ctx.document_bytes("ubi9-9.2-755.1697625012.json").await?;
 
         let (storage, _tmp) = FileSystemBackend::for_test().await?;
 
@@ -88,8 +88,8 @@ mod test {
             .ingest(
                 ("source", "test"),
                 None,
-                Format::sbom_from_bytes(data)?,
-                stream::iter([Ok::<_, Infallible>(Bytes::from_static(data))]),
+                Format::sbom_from_bytes(&data)?,
+                stream::iter([Ok::<_, Infallible>(Bytes::copy_from_slice(&data))]),
             )
             .await
             .expect("must ingest");
```

helio-frota:
thanks that tip worked well, pr updated :+1: 

the last part is a constant, I don't know if make sense to refactor unless we don't need to create weirdo things like this: https://github.com/martelinho-de-ouro/moria-watcher/blob/main/common/src/vars.rs#L4-L13

```
➜  trustify git:(workspace-root) rg "etc/test-data/"
modules/fundamental/src/advisory/endpoints/test.rs
532:static DOC: &[u8] = include_bytes!("../../../../../etc/test-data/csaf/cve-2023-33201.json");
```

jcrossley3:
> the last part is a constant, I don't know if make sense to refactor unless we don't need to create weirdo things like this

That `with_upload` fn is essentially doing what the new `TrustifyContext::ingest_document` does, so it can be significantly simplified, since it can delegate most of what it does to the `ctx` it is passed.

I would turn DOC into a string:
```rust
const DOC: &str = "csaf/cve-2023-33201.json";
```
And wherever it's used, do with it what you've done everywhere else in this PR. Make sense?

helio-frota:
thanks again , pr updated :+1:

jcrossley3 (approved):

PR #574, opened 2024-07-20
Only log internal server errors, don’t send details to the API user.

### LINKED ISSUE #573: Internal Server Errors should be logged at the ERROR level and the API response given to the client should not contain the details.

chirino:
Fixes #573

### DIFF
--- a/modules/fundamental/src/error.rs
+++ b/modules/fundamental/src/error.rs
@@ -51,23 +51,12 @@ impl ResponseError for Error {
             Self::Purl(err) => {
                 HttpResponse::BadRequest().json(ErrorInformation::new("InvalidPurlSyntax", err))
             }
-            Self::Actix(err) => {
-                HttpResponse::InternalServerError().json(ErrorInformation::new("System Actix", err))
-            }
             Self::BadRequest { msg, status } => {
                 HttpResponse::build(*status).json(ErrorInformation::new("Bad request", msg))
             }
-            Self::Any(err) => HttpResponse::InternalServerError()
-                .json(ErrorInformation::new("System unknown", err)),
             Error::Ingestor(inner) => {
                 HttpResponse::BadRequest().json(ErrorInformation::new("Ingestor error", inner))
             }
-            Error::UnsupportedHashAlgorithm => HttpResponse::InternalServerError()
-                .json(ErrorInformation::new("Unsupported hash algorithm", "")),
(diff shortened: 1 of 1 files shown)


modules/fundamental/src/error.rs
```diff
@@ -51,32 +51,53 @@ impl ResponseError for Error {
...
-            Error::Data(msg) => HttpResponse::InternalServerError()
-                .json(ErrorInformation::new("Data-model corruption", msg)),
+
+            // The following match arms handle internal system errors that are not expected to occur.
+            // They are logged and a generic error response is returned to avoid leaking
+            // internal state to end users.
+            Self::Actix(err) => {
+                log::error!("System Actix: {}", err);
+                HttpResponse::InternalServerError()
+                    .json(ErrorInformation::new("Internal Server Error", ""))
```
jcrossley3, line 74: The json isn't really adding anything, so I think you can just...
```suggestion
                HttpResponse::InternalServerError().finish()
```

chirino: I do tend to like APIs that have consistent response bodies on all errors, including 500s.

modules/fundamental/src/error.rs
```diff
@@ -51,32 +51,53 @@ impl ResponseError for Error {
...
                 HttpResponse::BadRequest().json(ErrorInformation::new("Storage Key", err))
             }
-            Error::Data(msg) => HttpResponse::InternalServerError()
-                .json(ErrorInformation::new("Data-model corruption", msg)),
+
+            // The following match arms handle internal system errors that are not expected to occur.
+            // They are logged and a generic error response is returned to avoid leaking
+            // internal state to end users.
+            Self::Actix(err) => {
+                log::error!("System Actix: {}", err);
```
ctron, line 72: I don't really like the pattern of adding a custom log message. IIRC we are using `thiserror` and so the error message should already be handled by the `Display` trait implementation. So a simple `log::error!("{err}")` should be sufficient. And given the fact that all match patterns would then have the same implementation, I think we can simplify this to just:

```rust
err => {
  log::error!("{err}");
  HttpResponse::InternalServerError()
    .json(ErrorInformation::new("Internal Server Error", ""))
}
```

chirino: Sounds good.

ctron (approved):

PR #577, opened 2024-07-21
feat: add trustd subcommand to export openapi spec

chirino:
Usage:

```
trustd openapi export --file test.json
```

* defaults to json format
* exports as yaml if file ends with ‘yaml’ or ‘yml’

### DIFF
--- a/trustd/src/openapi.rs
+++ b/trustd/src/openapi.rs
@@ -1,0 +1,57 @@
+use std::fs;
+use std::path::PathBuf;
+use std::process::ExitCode;
+
+use anyhow::{anyhow, Error, Result};
+
+use trustify_server::openapi::openapi;
+
+#[derive(clap::Args, Debug)]
+pub struct Run {
+    #[command(subcommand)]
+    pub(crate) command: Command,
+}
+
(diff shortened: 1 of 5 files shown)


JimFuller-RedHat:
we might also save openapi.json to root dir and add a check for equiv in github CI action to ensure this openapi.json is always up to date.

trustd/src/openapi.rs
```diff
@@ -0,0 +1,57 @@
...
+#[derive(clap::Args, Debug)]
+pub struct Export {
+    /// The file the openapi spec should be exported to
+    #[arg(long, env)]
+    pub file: PathBuf,
+}
+
+impl Export {
+    pub async fn run(self) -> Result<ExitCode> {
+        let doc = match self.file.file_name() {
```
ctron, line 38: Works the same, just pointing it out. I guess you could also use `std::fs::PathBuf::extension` to get the extension and then do a match: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=30e4f2062ea0893781d7ad2845cd971a

ctron (approved):

PR #579, opened 2024-07-22
refactor: simplify tests by incorporating more TrustifyContext

jcrossley3:
Expanded CallService to return json and added a `caller` helper to return an impl of it, i.e. an App::new configured with the fundamental endpoints. Replaced the `with_upload` async closures with calls to `caller` and `ctx.ingest_document`. Added an explicit test for SBOM uploading in lieu of `with_upload`.

### DIFF
--- a/modules/fundamental/src/organization/endpoints/test.rs
+++ b/modules/fundamental/src/organization/endpoints/test.rs
@@ -10,18 +10,14 @@ use trustify_common::db::Transactional;
 use trustify_common::hashing::Digests;
 use trustify_common::model::Paginated;
 use trustify_module_ingestor::graph::advisory::AdvisoryInformation;
-use trustify_module_ingestor::graph::Graph;
 use trustify_test_context::TrustifyContext;
 
-#[test_context(TrustifyContext, skip_teardown)]
+#[test_context(TrustifyContext)]
 #[test(actix_web::test)]
-async fn all_organizations(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
-    let db = &ctx.db;
-    let graph = Graph::new(db.clone());
+async fn all_organizations(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
+    let app = caller(ctx).await?;
 
-    let app = caller(&ctx).await?;
-
-    graph
+    ctx.graph
         .ingest_advisory(
             "CAPT-1",
             ("source", "http://captpickles.com/"),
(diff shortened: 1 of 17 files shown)


modules/fundamental/src/vulnerability/endpoints/test.rs
```diff
@@ -106,13 +102,10 @@ async fn all_vulnerabilities(ctx: TrustifyContext) -> Result<(), anyhow::Error>
 #[test_context(TrustifyContext, skip_teardown)]
 #[test(actix_web::test)]
 async fn one_vulnerability(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
-    let db = ctx.db;
+    let db = &ctx.db;
```
ctron, line 105: Didn't the `TrustifyContext` even give you have `Graph` now?

ctron (approved):
There might be more that the `TrustyContext` could do now I guess. But it already is a good improvement!

modules/fundamental/src/vulnerability/endpoints/test.rs
bobmcwhirter: We can also remove `skip_teardown` to prevent unintentional partial-moves of `ctx.db` and friends. Forces folks to `clone()` if they need an owned.

bobmcwhirter: (because then the Context comes in via a &ref)

jcrossley3: Cool. I'll make another pass.

modules/fundamental/src/organization/endpoints/test.rs
```diff
@@ -61,7 +53,7 @@ async fn all_organizations(ctx: TrustifyContext) -> Result<(), anyhow::Error> {
 
     let request = TestRequest::get().uri(uri).to_request();
 
-    let response: Value = actix_web::test::call_and_read_body_json(&app, request).await;
+    let response: Value = app.call_and_read_body_json(request).await;
```
bobmcwhirter, line 56: I love this.

bobmcwhirter (approved):

modules/fundamental/src/organization/endpoints/test.rs
jcrossley3: Credit to @ctron for the idea.

PR #584, opened 2024-07-22
Improve the SQL version comparisons:

### LINKED ISSUE #466: GET /api/v1/package/{uuid} fails with some packages

bobmcwhirter:
- use bigger ints for jackholes who version stuff like 202410298675309.
- if a version cmp fails (because a sha:xxx "version" was given) don't fail, just false.

In theory fixes #466.

### DIFF
--- a/migration/src/m0000475_improve_version_comparison_fns/semver_cmp.sql
+++ b/migration/src/m0000475_improve_version_comparison_fns/semver_cmp.sql
@@ -1,0 +1,128 @@
+
+create or replace function semver_cmp(left_p text, right_p text)
+    returns integer
+as
+$$
+declare
+    left_parts text[];
+    right_parts text[];
+
+    left_major bigint;
+    left_minor bigint;
+    left_patch bigint;
+    left_pre text;
+    left_build text;
+
+    right_major bigint;
+    right_minor bigint;
+    right_patch bigint;
+    right_pre text;
+    right_build text;
+
+    left_numeric bool;
+    right_numeric bool;
+
+    cur integer;
+
+begin
+
+    raise notice 'semver_cmp % %', left_p, right_p;
+
+    left_parts = regexp_split_to_array(left_p, E'\\+');
+    left_build = left_parts[2];
+
+    left_parts = regexp_split_to_array(left_parts[1], E'-');
+    left_pre = left_parts[2];
+
+    left_parts = regexp_split_to_array(left_parts[1], E'\\.');
+    left_major = left_parts[1]::decimal;
+    left_minor = left_parts[2]::decimal;
+    left_patch = left_parts[3]::decimal;
+
+    right_parts = regexp_split_to_array(right_p, E'\\+');
+    right_build = right_parts[2];
+
+    right_parts = regexp_split_to_array(right_parts[1], E'-');
+    right_pre = right_parts[2];
+
+    right_parts = regexp_split_to_array(right_parts[1], E'\\.');
+    right_major = right_parts[1]::decimal;
+    right_minor = right_parts[2]::decimal;
+    right_patch = right_parts[3]::decimal;
+
+    if left_major > right_major then
+        return +1;
+    elsif left_major < right_major then
+        return -1;
+    end if;
+
+    if left_minor > right_minor then
+        return +1;
+    elsif left_minor < right_minor then
+        return -1;
+    end if;
+
+    if left_patch > right_patch then
+        return +1;
+    elsif left_patch < right_patch then
+        return -1;
+    end if;
+
+    if left_pre is null and right_pre is not null then
+        return +1;
+    elsif left_pre is not null and right_pre is null then
+        return -1;
+    elsif left_pre is not null and right_pre is not null then
+        left_parts = regexp_split_to_array(left_pre, E'\\.');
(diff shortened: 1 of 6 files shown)


migration/src/m0000475_improve_version_comparison_fns/semver_cmp.sql
```diff
@@ -0,0 +1,128 @@
...
+        return 0;
+    end if;
+
+    return null;
+exception
+    when others then
+        return null;
+end
+$$
+    language 'plpgsql';
```
JimFuller-RedHat, line 128: have u considered adding IMMUTABLE PARALLEL SAFE; here ?

migration/src/m0000475_improve_version_comparison_fns/semver_cmp.sql
```diff
@@ -0,0 +1,128 @@
...
+    right_build text;
+
+    left_numeric bool;
+    right_numeric bool;
+
+    cur integer;
+
+begin
+
+    raise notice 'semver_cmp % %', left_p, right_p;
```
JimFuller-RedHat, line 29: FWIW - **log_min_messages** configuration parameter to a higher level, such as WARNING or ERROR, to suppress NOTICE messages ... useful for production when we dont want to generate gigantic log files ;)

migration/src/m0000475_improve_version_comparison_fns/semver_cmp.sql
```diff
@@ -0,0 +1,128 @@
+
+create or replace function semver_cmp(left_p text, right_p text)
```
JimFuller-RedHat, line 2: it is unclear if this function covers off all of http://ftp.rpm.org/api/4.4.2.2/rpmvercmp_8c-source.html .... I did a version in sql here https://github.com/RedHatProductSecurity/component-registry/blob/main/corgi/core/migrations/0092_install_stored_proc.py ... if I have the time tmrw will cross reference between rpmvercmp and this

JimFuller-RedHat (commented):
I would test output of  this cmp with rpmvercmp ..

migration/src/m0000475_improve_version_comparison_fns/semver_cmp.sql
JimFuller-RedHat: and yes I know this comment is expanding scope ;)

JimFuller-RedHat (approved):

migration/src/m0000475_improve_version_comparison_fns/semver_cmp.sql
bobmcwhirter: Considering that I've no idea what that means... no. Should I?

migration/src/m0000475_improve_version_comparison_fns/semver_cmp.sql
bobmcwhirter: fwiw, I can never actually find the postgres logs. raise notice shows up in the rust log if `trace` at least.

migration/src/m0000475_improve_version_comparison_fns/semver_cmp.sql
bobmcwhirter: So far only semver cmp. Need to do *many more*.

bobmcwhirter: And I'll definitely steal your RPM version cmp.

migration/src/m0000475_improve_version_comparison_fns/semver_cmp.sql
JimFuller-RedHat: as this function does no mutation, and only works on 2 supplied inputs ... I would suggest it ... usually results in much faster function.

migration/src/m0000475_improve_version_comparison_fns/semver_cmp.sql
JimFuller-RedHat: ya in production, you might find this jams up logs unnec ... just mentioned

migration/src/m0000475_improve_version_comparison_fns/semver_cmp.sql
JimFuller-RedHat: +1 - we assumed rpm folks have encountered every variation and stole from them thusly ... I think my sql version is correct eg. ran it against their test cases (but did not embed in corgi)

PR #588, opened 2024-07-22
Add cargo xtask validate-openapi command that can be used to validate the generated openapi spec.

chirino:
Run with:

```
cargo xtask validate-openapi
```

### DIFF
--- a/xtask/src/openapi.rs
+++ b/xtask/src/openapi.rs
@@ -17,7 +17,9 @@ impl Validate {
         } else if command_exists("docker") {
             "docker"
         } else {
-            return Err(anyhow!("Neither podman nor docker is installed"));
+            return Err(anyhow!(
+                "This task requires podman or docker to be installed."
+            ));
         };
 
         let out_dir = env::temp_dir();
--- a/xtask/src/openapi.rs
+++ b/xtask/src/openapi.rs
@@ -1,0 +1,63 @@
+use std::path::Path;
+use std::process::Command;
+use std::{env, fs};
+
+use anyhow::anyhow;
+use clap::Parser;
+
+use trustify_server::openapi::openapi;
+
+#[derive(Debug, Parser)]
+pub struct Validate {}
+
+impl Validate {
+    pub fn run(self) -> anyhow::Result<()> {
+        let command = if command_exists("podman") {
+            "podman"
+        } else if command_exists("docker") {
+            "docker"
+        } else {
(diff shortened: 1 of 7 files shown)


chirino:
rebased.

xtask/src/openapi.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub fn run(self) -> anyhow::Result<()> {
+        let command = if command_exists("podman") {
+            "podman"
+        } else if command_exists("docker") {
+            "docker"
+        } else {
+            return Err(anyhow!("Neither podman nor docker is installed"));
+        };
+
+        let out_dir = env::temp_dir();
```
JimFuller-RedHat, line 23: Have you considered creating a separate Generate task to write out openapi.json,  and have Validate invoke that instead of 'baking it' in here.

xtask/src/openapi.rs
```diff
@@ -0,0 +1,63 @@
...
+pub struct Validate {}
+
+impl Validate {
+    pub fn run(self) -> anyhow::Result<()> {
+        let command = if command_exists("podman") {
+            "podman"
+        } else if command_exists("docker") {
+            "docker"
+        } else {
+            return Err(anyhow!("Neither podman nor docker is installed"));
```
JimFuller-RedHat, line 20: Alternately you could use more assertive language, for example "This task requires podman or docker to be installed.".

xtask/src/openapi.rs
```diff
@@ -0,0 +1,63 @@
...
+                "/src/openapi.yaml",
+            ])
+            .current_dir(out_dir.to_str().unwrap())
+            .status()
+            .map_err(|_| anyhow!("Failed to validate openapi.yaml"))?
+            .success()
+        {
+            Ok(())
+        } else {
+            Err(anyhow!("Failed to validate openapi.yaml"))
```
JimFuller-RedHat, line 53: I could not tell if this is going to emit specific validation errors to be able to diagnose the problem.

JimFuller-RedHat (approved):
LGTM though had a few questions/observations

xtask/src/openapi.rs
chirino: Actual validation errors are logged to console by the container.

xtask/src/openapi.rs
chirino: trustd has a sub command to do that.. I could add a cargo alias to trigger it easier.

Issue #563, opened 2024-07-18
I want to delete an SBOM (maybe other stuff too)


ctron:
I just uploaded and SBOM and want to get rid of it from the system.

* [x] Provide an API for deleting the document
* [x] https://github.com/trustification/trustify-ui/issues/375

JimFuller-RedHat:
belated observation: what is the rbac/acl mechanism for authorising a user to mutate ?

jcrossley3:
> belated observation: what is the rbac/acl mechanism for authorising a user to mutate ?

At present, we have [this](https://github.com/trustification/trustify/blob/acd702383529d7571d78bbea908866880623b848/common/auth/src/authorizer/mod.rs#L17-L48), used [like so](https://github.com/trustification/trustify/blob/acd702383529d7571d78bbea908866880623b848/modules/fundamental/src/sbom/endpoints/mod.rs#L219).

rh-jfuller:
I believe the spirit of this has been addressed

PR #590, opened 2024-07-23
feat: Add api to delete products

chirino:
A task of #563

### DIFF
--- a/modules/fundamental/src/error.rs
+++ b/modules/fundamental/src/error.rs
@@ -33,6 +33,8 @@ pub enum Error {
     Storage(anyhow::Error),
     #[error("Invalid data model {0}")]
     Data(String),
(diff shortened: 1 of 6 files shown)


dejanb (approved):
Looks good on it's own.

Products and product versions are created as a part of importing SBOMs (and in the future advisories). Does deleting a product should delete associated SBOMs as well?

chirino:
> Products and product versions are created as a part of importing SBOMs (and in the future advisories). Does deleting a product should delete associated SBOMs as well?

yeah that's worth a debate.  Someone might argue the reverse too, if you delete an SBOM should it delete the associated products?

dejanb:
Yup. No clear answers at this moment, so let's keep playing with it and things will start making sense

PR #591, opened 2024-07-23
Add RPM version comparison SQL fn.

bobmcwhirter:


JimFuller-RedHat (approved):
LGTM ... I am not a big fan of unit tests ... but maybe with these pg funcs we could add a test (can we do raw sql queries ?) near trustd/src/db.rs ?

like the following untested code ?
```
let rows: Vec<sea_orm::Row> = db.query("select * from rpmvercmp('1.8.3','1.3.5');").await?;

for row in rows {
assert_eq!(row, 1);
}
```

bobmcwhirter:
Yah we unit test semver cmp also. Will add.

PR #598, opened 2024-07-24
Optimize away some 1+N queries inside a loop to at least not be in a …

bobmcwhirter:
…loop.

### DIFF
--- a/modules/fundamental/src/advisory/model/mod.rs
+++ b/modules/fundamental/src/advisory/model/mod.rs
@@ -58,15 +58,19 @@ pub struct AdvisoryHead {
 impl AdvisoryHead {
     pub async fn from_advisory(
         entity: &advisory::Model,
-        issuer: Option<organization::Model>,
+        issuer: Option<Option<organization::Model>>,
         tx: &ConnectionOrTransaction<'_>,
     ) -> Result<Self, Error> {
-        let issuer = if let Some(issuer) = issuer {
-            Some(OrganizationSummary::from_entity(&issuer, tx).await?)
-        } else if let Some(issuer) = entity.find_related(organization::Entity).one(tx).await? {
-            Some(OrganizationSummary::from_entity(&issuer, tx).await?)
-        } else {
-            None
+        let issuer = match &issuer {
+            Some(Some(issuer)) => Some(OrganizationSummary::from_entity(issuer, tx).await?),
+            Some(None) => None,
+            None => {
+                if let Some(issuer) = entity.find_related(organization::Entity).one(tx).await? {
+                    Some(OrganizationSummary::from_entity(&issuer, tx).await?)
+                } else {
+                    None
+                }
+            }
         };
 
         Ok(Self {
--- a/modules/fundamental/src/advisory/model/summary.rs
+++ b/modules/fundamental/src/advisory/model/summary.rs
@@ -59,7 +59,7 @@ impl AdvisorySummary {
             let average_score = average_score.map(|score| Score::new(score).roundup());
 
             summaries.push(AdvisorySummary {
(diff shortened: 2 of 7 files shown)


modules/fundamental/src/advisory/model/mod.rs
```diff
@@ -58,15 +58,19 @@ pub struct AdvisoryHead {
 impl AdvisoryHead {
     pub async fn from_advisory(
         entity: &advisory::Model,
-        issuer: Option<organization::Model>,
+        issuer: Option<Option<organization::Model>>,
```
jcrossley3, line 61: This can't be right, can it? What behavior is it providing?

bobmcwhirter: It is in fact!

That method can take an organization from a previous query (which is `Option<org::Model>`) or can be deferred to let it do the lookup, if no caller has already queried it.

Therefore...

`None` -> I've not asked, please do it for me.
`Some(None)` -> I did the lookup, there isn't one, don't bother doing the query again.
`Some(Some(org_model))` -> I did the lookup, here it is, don't bother doing the query again.

JimFuller-RedHat (approved):
LGTM

modules/fundamental/src/advisory/model/mod.rs
```diff
@@ -58,15 +58,19 @@ pub struct AdvisoryHead {
...
     ) -> Result<Self, Error> {
-        let issuer = if let Some(issuer) = issuer {
-            Some(OrganizationSummary::from_entity(&issuer, tx).await?)
-        } else if let Some(issuer) = entity.find_related(organization::Entity).one(tx).await? {
-            Some(OrganizationSummary::from_entity(&issuer, tx).await?)
-        } else {
-            None
+        let issuer = match &issuer {
+            Some(Some(issuer)) => Some(OrganizationSummary::from_entity(issuer, tx).await?),
+            Some(None) => None,
```
bobmcwhirter, line 66: "I tried, but there's no org, be efficient."

modules/fundamental/src/advisory/model/mod.rs
```diff
@@ -58,15 +58,19 @@ pub struct AdvisoryHead {
...
+            None => {
```
bobmcwhirter, line 67: "I have no idea, please check for me."

modules/fundamental/src/advisory/model/summary.rs
```diff
@@ -59,7 +59,7 @@ impl AdvisorySummary {
             let average_score = average_score.map(|score| Score::new(score).roundup());
 
             summaries.push(AdvisorySummary {
-                head: AdvisoryHead::from_advisory(advisory, issuer, tx).await?,
+                head: AdvisoryHead::from_advisory(advisory, Some(issuer), tx).await?,
```
bobmcwhirter, line 62: Signals I've looked.

modules/fundamental/src/vulnerability/model/details/vulnerability_advisory.rs
```diff
@@ -70,29 +70,28 @@ impl VulnerabilityAdvisoryHead {
...
+                    let base = Cvss3Base::from((*e).clone());
                     base.score().value()
                 })
                 .reduce(|accum, e| accum + e);
 
-            let score = total_score.map(|score| Score::new(score / cvss3.len() as f64));
+            let score = total_score.map(|score| Score::new(score / advisory_cvss3s.len() as f64));
 
             heads.push(VulnerabilityAdvisoryHead {
                 head: AdvisoryHead::from_advisory(advisory, None, tx).await?,
```
bobmcwhirter, line 97: Signals I have *not* looked...

PR #599, opened 2024-07-24
Calculate average vuln severity and score in the DB, allow as a filte…

### LINKED ISSUE #411: Vulnerability GET /api/v1/vulnerability should support Filter by `average_severity`
Each vulnerability has the field `average_severity`. We should be able to filter by that field. There was very similar work done for the Advisories.

The current message I get when I try to filter by `average_severity` is: 

```
q=average_severity=low
```

{
    "error": "Query error",
    "message": "query syntax error: Invalid field name for filter: 'average_severity'"
}

bobmcwhirter:
…r/sort column.

Should fix #411 in theory.

### DIFF
--- a/modules/fundamental/src/vulnerability/service/mod.rs
+++ b/modules/fundamental/src/vulnerability/service/mod.rs
@@ -2,16 +2,22 @@ use crate::{
     vulnerability::model::{VulnerabilityDetails, VulnerabilitySummary},
     Error,
 };
-use sea_orm::{prelude::*, EntityTrait, QueryFilter};
+use sea_orm::{
+    prelude::*, EntityTrait, FromQueryResult, IntoIdentity, QueryFilter, QuerySelect, QueryTrait,
+};
+use sea_query::{ColumnRef, Func, IntoColumnRef, IntoIden, SimpleExpr};
+use trustify_common::db::limiter::LimiterAsModelTrait;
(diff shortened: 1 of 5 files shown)


jcrossley3 (approved):
Thanks for this! I'll add a test for it.

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -30,17 +36,99 @@ impl VulnerabilityService {
...
+            .column_as(
+                SimpleExpr::Column(ColumnRef::Column(
+                    "average_severity".into_identity().into_iden(),
+                ))
+                .cast_as("TEXT".into_identity()),
+                "average_severity",
+            )
+            .filtering_with(
+                search,
+                Columns::from_entity::<advisory::Entity>()
```
jcrossley3, line 90: @bobmcwhirter When I try a free text search, e.g. `q=CVE-2024-36974`, which should match the `identifier` field, I get an error from the db: missing FROM-clause entry for table "advisory". We're getting columns from `advisory::Entity` but we're querying `vulnerability::Entity`. I'm guessing we need a join of some type, but all my attempts have failed. Help?

bobmcwhirter: Ah, I think I just bad copy-pasta'd `advisory::Entity` and it should've been `vulnerability::Entiy` for the `Columns::from_entity`

Didja give that a whirl?

PR #601, opened 2024-07-25
feat: Add api to delete Vulnerabilities.

chirino:
A task of #563

### DIFF
--- a/modules/fundamental/src/vulnerability/endpoints/mod.rs
+++ b/modules/fundamental/src/vulnerability/endpoints/mod.rs
@@ -2,7 +2,8 @@
 mod test;
 
 use crate::vulnerability::service::VulnerabilityService;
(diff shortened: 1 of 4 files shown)


bobmcwhirter:
Does deleting a vulnerability result in advisories deleting also?

A vulnerability primarily is just an identifier, and should be upsert'd with at least it's `id` column if mentioned by any advisory.

I think it'd be startling to have deletion of a vulnerability to cause a cascade of advisory deletions. And if it doesn't, should violate a FK from advisory->vuln.  If we don't have that FK in place, we've got another problem.

chirino:
No advisories remain.  However, they will not list the related vulnerability that just got deleted. 
There are no FKs into vulnerability right now.

vulnerability_description does get cascaded deleted when the related advisory is deleted since it's got FK into advisories: 
image

bobmcwhirter (approved):

PR #602, opened 2024-07-25
Add some corner case tests for SPDX, and fix a discovered issue

ctron:
This fixes the two issues of https://github.com/trustification/trustify/discussions/555 but removing duplicated during the conversion and by renaming the enum variant to `All`, which better reflects what it currently does.

### DIFF
--- a/modules/fundamental/src/sbom/service/sbom.rs
+++ b/modules/fundamental/src/sbom/service/sbom.rs
@@ -1,9 +1,10 @@
 use super::SbomService;
-use crate::purl::model::summary::purl::PurlSummary;
-use crate::sbom::model::details::{SbomAdvisory, SbomDetails};
-use crate::sbom::model::SbomHead;
 use crate::{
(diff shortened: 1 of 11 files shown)


modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -460,7 +460,19 @@ impl SbomService {
             )
             .await?;
 
-        Ok(result.items.into_iter().map(|r| r.package).collect())
+        // TODO: this will break when adding pagination, as we effectively only process a single page
```
bobmcwhirter, line 463: can we leverage some `DISTINCT` on the query?

jcrossley3 (approved):

modules/fundamental/src/sbom/service/sbom.rs
ctron: Once we need this, we'd need to re-implement the query to return only packages, instead of full package-relationships. Currently, this re-uses the "fetch package relationships" function, stripping away the unnecessary information.

PR #604, opened 2024-07-25
fix: remove usage of FileSource and document how to use csaf walker to ingest local files

dejanb:
This PR tries to enable csaf importer to use file source and additionally to unify how sbom and csaf importers handle this case.

It should allow enable local file system importers like

```
  http POST localhost:8080/api/v1/importer/ds1-csaf "csaf[source]=file:///Users/dejanb/workspace/github.com/trustification/trustification/data/ds1/csaf/" "csaf[disabled]:=false" "csaf[period]=30s" "csaf[v3Signatures]:=true"

  http POST localhost:8080/api/v1/importer/ds1-sbom "sbom[source]=file:///Users/dejanb/workspace/github.com/trustification/trustification/data/ds1/sbom/" "sbom[disabled]:=false" "sbom[period]=30s" "sbom[v3Signatures]:=true"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Connect to PSQL:
 env PGPASSWORD=eggs psql -U postgres -d trustify -h localhost -p 5432
 ```
 
-If you don't have the `psql` command available, you can also use the `podman-compose` command: 
+If you don't have the `psql` command available, you can also use the `podman-compose` command:
 
 ```shell
 podman-compose -f etc/deploy/compose/compose.yaml exec postgres psql -U postgres -d trustify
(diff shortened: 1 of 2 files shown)


bobmcwhirter:
Can we chroot/limit that? I'd hate to have `file:///etc/passwd` end up shuffling something important to the S3 buckets once we have support for that.

dejanb:
Hm, not sure how to do this as we usually mount this to `/data` or something in the container. Also, it won't pick up random files, but expects the csaf compatible directory structure. 

@ctron any ideas on this topic?

bobmcwhirter:
It's also fine to assert `NOT_A_PROBLEM`. Just trying to be cognizant of attack vectors.

ctron:
I intentionally did not add `file` just because of all the problems that might cause. Like `/etc/passwd`, like not being that useful in a k8s setup.

For testing, one can easily spawn an http server like this: `python -m http.server 8000` and point the importer to that source.

So I believe that there are reasons for adding this. However, for me, there are more reasons to not add this.

modules/importer/src/server/csaf/mod.rs
```diff
@@ -38,13 +38,24 @@ impl super::Server {
         let fetcher = Fetcher::new(Default::default()).await?;
         let options = HttpOptions::new().since(last_success);
 
-        let source = match Url::parse(&importer.source) {
-            Ok(url) => HttpSource::new(url, fetcher, options),
+        let source: DispatchSource = match Url::parse(&importer.source) {
```
ctron, line 41: Should we really consider adding this, we should leverage the existing `new_source` function, which already does all the handling required:

https://github.com/ctron/csaf-walker/blob/f018d47cc5025ddf3f3a855f942cdf1f72726b3b/csaf/src/source/mod.rs#L41-L49

dejanb:
Make sense ... I think we can also use raw github as provider for testing sets

The only thing was that sbom importer supported files (kinda) and csaf didn't. I'll remove files support from sbom and create an example(s) on how to host and ingest custom sets

ctron:
> The only thing was that sbom importer supported files (kinda) and csaf didn't. I'll remove files support from sbom and create an example(s) on how to host and ingest custom sets

Ouch, it think it shouldn't either. Yes, that's a good idea!

ctron (approved):

PR #606, opened 2024-07-25
chore: tweak embedded database settings

JimFuller-RedHat:
Set the embedded_pg to more 'modern' settings:

From trustify we set

* db max conns = 300
* db min conns = 75

and in pg config
* **max_connections**=500

which lets the database 'breath'.

Additionally set **random_page_cost** to 1.1 because we all have SSD these days and the default setting of 4 is for _spinning rust_. There are a lot of pg configs that have dramatically conservative defaults which we can revisit but wanted to establish a baseline for the embedded_pg.

### DIFF
--- a/common/src/db/mod.rs
+++ b/common/src/db/mod.rs
@@ -131,7 +131,8 @@ impl Database {
         log::debug!("connect to {}", url);
 
         let mut opt = ConnectOptions::new(url);
-        opt.min_connections(16);
+        opt.max_connections(300);
+        opt.min_connections(75);
         opt.sqlx_logging_level(log::LevelFilter::Trace);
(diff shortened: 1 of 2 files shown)


bobmcwhirter (approved):

bobmcwhirter:
I don't love the map of strings for config, but I reckon that's outside of our control?

JimFuller-RedHat:
it is ... we could put a facade over this to make it easier to share/load pg configs eg. /etc/pg could contain sets of pg configs which we can dynamically load ... for this project I am guessing we want to be 'generic' but be able to easily test some other configs (for example to mirror our own stage/prod).

PR #600, opened 2024-07-24
feat: Add api to delete Advisories.

chirino:
A task of #563

PR #608, opened 2024-07-26
fix: sbom_node FK constraints

chirino:
Drop the node_id constrains since they modeled inheritance, not ownership, replace with sbom_id cascading constraints that do model ownership.

And then fix ingestion so that the sbom record is created before the sbom_node record, since the sbom record owns it.


chirino:
this PR has a migration numbered after the migration in https://github.com/trustification/trustify/pull/600/ so that one should likely go in first.

bobmcwhirter (approved):

PR #611, opened 2024-07-26
feat: GC Purls when sboms/advisories are deleted.

chirino:


ctron (commented):
Looks good from a technical perspective. I am not sure if that's the idea we have. I guess @bobmcwhirter can answer that.

bobmcwhirter (approved):
Looks good!

PR #614, opened 2024-07-26
Adding a criterion based benchmark for csaf ingestion.

chirino:
if you run `cargo bench` with this patch you can then 

```
open target/criterion/report/index.html
```

to view the benchmark report.  If you re-run the benchmark, it will show you perf difference compared to the previous run.

### DIFF
--- a/modules/fundamental/benches/bench.rs
+++ b/modules/fundamental/benches/bench.rs
@@ -1,40 +1,115 @@
-use criterion::{criterion_group, criterion_main};
+use std::time::Duration;
+
+use criterion::{criterion_group, criterion_main, Criterion};
 
 #[allow(clippy::expect_used, clippy::unwrap_used)]
 pub(crate) mod trustify_benches {
-    use criterion::Criterion;
+    use std::future::Future;
     use std::io;
+    use std::io::Error;
+    use std::ops::Add;
+    use std::sync::Arc;
+    use std::time::{Duration, Instant};
+
+    use bytes::Bytes;
+    use criterion::{black_box, Criterion};
+    use csaf::Csaf;
+    use futures_util::stream;
+    use futures_util::stream::Once;
+    use sea_orm::ConnectionTrait;
     use test_context::AsyncTestContext;
     use tokio::runtime::Runtime;
 
+    use trustify_common::db::Transactional;
     use trustify_entity::labels::Labels;
     use trustify_module_ingestor::service::Format;
-    use trustify_test_context::{document_stream, TrustifyContext};
+    use trustify_test_context::{document_bytes, TrustifyContext};
 
-    fn setup_runtime_and_ctx() -> (Runtime, TrustifyContext) {
+    pub fn ingestion(c: &mut Criterion) {
+        let (runtime, ctx) = setup_runtime_and_ctx();
+        c.bench_function("ingestion_csaf", |b| {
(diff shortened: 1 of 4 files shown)


chirino:
this might help out with #456

modules/fundamental/benches/bench.rs
```diff
@@ -0,0 +1,40 @@
...
+        let ctx = runtime.block_on(async { TrustifyContext::setup().await });
+        (runtime, ctx)
+    }
+
+    pub fn ingestion(c: &mut Criterion) {
+        let (runtime, ctx) = setup_runtime_and_ctx();
+        c.bench_function("ingestion_csaf", |b| {
+            b.to_async(&runtime).iter(|| async {
+                // todo: how do we reset the deb so we can re-import the doc?
+                let payload = document_stream("csaf/cve-2023-33201.json")
```
chirino, line 28: Anybody know how to quickly reset the DB, so that we can re-ingest.  Or maybe generate a new csaf doc each time?

chirino: @ctron ^^ what do you think?

ctron: I think that creating new CSAF documents might still lead to a different behavior. As the database grows. I would expect that later iterations run slower.

Resetting the DB on the other hand is definitely costly. Not sure if there's a way to take this out of the collected time.

Another idea might be to not only ingest a single document, but maybe 10, 50, … in a single run. And only then reset the DB. Then re-run that multiple times.

chirino: can you explain  the "creating new CSAF documents might still lead to a different behavior" issue more?

ctron: As far as I understood the idea, you wanted to write a test, which ingests a new document each time, but does not clear the database between runs.

That would lead to tables growing, which might have an impact on performance. So the first iteration would write into an empty database, while the 100th iteration would write into a database that already contains documents. I think such a benchmark should have the same condition (empty database) for each iteration.

Also, we do aggregate information. e.g. the purls get created only once. So the first iteration would check and create, while the other iterations would check and not create.

ctron: Right, that was the idea of the proposal to ingest multiple documents in a single iteration.

We could also do both. Have a test for a single (first) document. And then a test for adding 100 documents. Resetting the DB between those iterations.

chirino: Updated PR to do this.

modules/fundamental/benches/bench.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    async fn document_stream_generated_from(
+        path: &str,
+        rev: u64,
+    ) -> Once<impl Future<Output = Result<Bytes, Error>> + Sized> {
+        let payload = document_bytes(path).await.expect("load ok");
+        let mut doc: Csaf = serde_json::from_slice(payload.as_ref()).expect("parse ok");
+        doc.document.tracking.id = format!("{}-{}", doc.document.tracking.id, rev);
+        let data = serde_json::to_vec_pretty(&doc).expect("serialize ok");
```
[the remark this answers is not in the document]
chirino: implemented now.

ctron (approved):

PR #627, opened 2024-07-30
feat: S3 storage backend

### LINKED ISSUE #621: Implement S3 storage backend

jcrossley3:
Fixes #621

### DIFF
--- a/modules/storage/src/service/dispatch.rs
+++ b/modules/storage/src/service/dispatch.rs
@@ -2,7 +2,7 @@ use self::s3::S3Backend;
 
 use super::*;
 use bytes::Bytes;
-use futures::{Stream, TryStreamExt};
+use futures::{Stream, StreamExt, TryStreamExt};
 
 /// A common backend, dispatching to the ones we support.
 ///
@@ -45,12 +45,12 @@ impl StorageBackend for DispatchBackend {
             Self::Filesystem(backend) => backend
                 .retrieve(key)
                 .await
-                .map(|stream| stream.map(|stream| stream.map_err(anyhow::Error::from)))
+                .map(|stream| stream.map(|stream| stream.map_err(anyhow::Error::from).boxed()))
                 .map_err(anyhow::Error::from),
             Self::S3(backend) => backend
                 .retrieve(key)
                 .await
-                .map(|stream| stream.map(|stream| stream.map_err(anyhow::Error::from)))
+                .map(|stream| stream.map(|stream| stream.map_err(anyhow::Error::from).boxed()))
                 .map_err(anyhow::Error::from),
         }
     }
--- a/modules/storage/src/service/dispatch.rs
+++ b/modules/storage/src/service/dispatch.rs
@@ -1,3 +1,5 @@
+use self::s3::S3Backend;
+
 use super::*;
 use bytes::Bytes;
 use futures::{Stream, TryStreamExt};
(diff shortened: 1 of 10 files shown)


modules/storage/src/service/mod.rs
```diff
@@ -60,10 +62,15 @@ impl TryFrom<Vec<Id>> for StorageKey {
 
 #[derive(Clone, Debug)]
 pub struct StorageResult {
-    pub key: StorageKey,
     pub digests: Digests,
 }
 
+impl StorageResult {
+    pub fn key(&self) -> StorageKey {
+        StorageKey(self.digests.sha256.encode_hex())
```
jcrossley3, line 70: this is the convention I wanted to keep DRY, so no longer passing in a key, but only deriving it from the digests in a single place.

modules/storage/src/service/s3.rs
```diff
@@ -0,0 +1,84 @@
...
+impl S3Backend {}
+
+impl S3Backend {
+    pub fn new() -> Result<Self, S3Error> {
+        // TODO: not this
+        let bucket = Bucket::new(
+            "trustify-jcrossley",
+            Region::UsEast1,
+            Credentials::default().unwrap(),
+        )?;
```
jcrossley3, line 24: temporary hack until i add fields to `StorageConfig`

modules/storage/src/service/s3.rs
```diff
@@ -0,0 +1,84 @@
...
+
+        // Rename the object with a key calculated from its digests
+        self.bucket
+            .copy_object_internal(&path, &key)
+            .await
+            .map_err(StoreError::Backend)?;
+        self.bucket
+            .delete_object(&path)
+            .await
+            .map_err(StoreError::Backend)?;
```
jcrossley3, line 69: copying/deleting on S3 for now. we can do this on the filesystem if we think it'll be too expensive.

modules/storage/src/service/dispatch.rs
```diff
@@ -43,6 +47,11 @@ impl StorageBackend for DispatchBackend {
                 .await
                 .map(|stream| stream.map(|stream| stream.map_err(anyhow::Error::from)))
                 .map_err(anyhow::Error::from),
+            Self::S3(backend) => backend
+                .retrieve(key)
+                .await
+                .map(|stream| stream.map(|stream| stream.map_err(anyhow::Error::from)))
+                .map_err(anyhow::Error::from),
```
jcrossley3, line 54: this is where i could use some help/advice. rust compiler doesn't like seeing the different opaque types implementing `stream`. not sure of the best course here.

jcrossley3 (commented):
@ctron could use your help on the compiler error in `dispatch.rs`

modules/storage/src/service/dispatch.rs
bobmcwhirter: See subsequent commit I added for ya.

bobmcwhirter:
Maybe start here with TRUSTD_ env prefix?

modules/storage/src/service/mod.rs
ctron: Makes sense!

ctron (approved):
Looks good! And the changes to the key make perfect sense!

Not blocking the PR, but I would prefer a better way to handle the configuration switch between FS and S3. As far as I understand it, right now I could select FS (which also is the default) and provide S3 config options. And it would silently ignore the S3 config. I think there should be a away to conflict S3 options with the FS strategy (or vice versa).

jcrossley3:
> Maybe start here with TRUSTD_ env prefix?

Done

PR #629, opened 2024-07-31
better out of the box generic db conn settings

JimFuller-RedHat:
The default pg max conns is still 100 - there are plans to bump this in pg at some point though that day is not now, dialing back until then.


ctron (approved):

ctron:
I think it would be great to make this configurable.

JimFuller-RedHat:
Its worth mentioning that a many of pg default settings are 'ancient' in terms of modern hardware/software ... mainly this helps with backwards compat/usability on older platforms.

bobmcwhirter (approved):

ctron (approved):

PR #632, opened 2024-07-31
fix: make storage CLI options mutually exclusive

### LINKED ISSUE #631: Make storage CLI options mutually exclusive
Looks good! And the changes to the key make perfect sense!

Not blocking the PR, but I would prefer a better way to handle the configuration switch between FS and S3. As far as I understand it, right now I could select FS (which also is the default) and provide S3 config options. And it would silently ignore the S3 config. I think there should be a away to conflict S3 options with the FS strategy (or vice versa).

_Originally posted by @ctron in https://github.com/trustification/trustify/pull/627#pullrequestreview-2209399981_

jcrossley3:
Fixes #631

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7282,6 +7282,7 @@ dependencies = [
  "bytes",
  "futures",
  "hex",
+ "log",
  "rust-s3",
  "serde_json",
  "sha2",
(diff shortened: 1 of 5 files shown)


ctron (approved):
Looks great. And learned some new clap tricks! Thanks!

PR #638, opened 2024-08-05
Improve performance of fetching product summaries

dejanb:
By using load_many and load_one patterns we can now avoid querying in the loop

### DIFF
--- a/modules/fundamental/src/product/model/mod.rs
+++ b/modules/fundamental/src/product/model/mod.rs
@@ -54,7 +54,7 @@ impl ProductVersionHead {
     }
 
     pub async fn from_entities(
-        product_versions: &Vec<product_version::Model>,
+        product_versions: &[product_version::Model],
         tx: &ConnectionOrTransaction<'_>,
     ) -> Result<Vec<Self>, Error> {
         let mut heads = Vec::new();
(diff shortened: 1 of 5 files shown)


modules/fundamental/src/product/model/mod.rs
```diff
@@ -54,13 +54,13 @@ impl ProductVersionHead {
     }
 
     pub async fn from_entities(
-        product_versions: Vec<product_version::Model>,
+        product_versions: &Vec<product_version::Model>,
```
bobmcwhirter, line 57: I find clippy telling me to use `&[product_version::Model]` which you can get by `&`'ing a `Vec`.

modules/fundamental/src/product/model/summary.rs
```diff
@@ -46,10 +44,13 @@ impl ProductSummary {
...
         tx: &ConnectionOrTransaction<'_>,
     ) -> Result<Vec<Self>, Error> {
+        let versions = products.load_many(product_version::Entity, tx).await?;
+        let orgs = products.load_one(organization::Entity, tx).await?;
+
         let mut summaries = Vec::new();
 
-        for product in products {
-            summaries.push(ProductSummary::from_entity(product, tx).await?);
+        for (product, org, version) in izip!(products, orgs, versions) {
```
bobmcwhirter, line 52: Hrm, I like `izip!()` instead of all of my nested tuples.

Thanks!

bobmcwhirter (approved):

JimFuller-RedHat (approved):
nothing to add after @bobmcwhirter comments are addressed, LGTM

PR #639, opened 2024-08-05
ci: Continuously create image 

carlosthe19916:
Fix: https://github.com/trustification/trustify/issues/637

- Tested in my local fork here https://github.com/carlosthe19916/trustify/actions/runs/10250287369 . The tested container image was also generated correctly in my local fork https://github.com/carlosthe19916/trustify/pkgs/container/trustd
- Adding a workflow to continuously create container images of the server. It takes the code available in the branch and merges it with the latest available UI in the trustify-ui repository

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+target/
+.trustify/
--- a/.github/workflows/image-build.yaml
+++ b/.github/workflows/image-build.yaml
@@ -1,0 +1,26 @@
+name: Multiple Architecture Image Build
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - "main"
+      - "release-*"
+#    tags:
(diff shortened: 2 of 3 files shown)


ctron (changes requested):
I think it would be great to have a continuous stream of container images.

However, I strongly believe that we should re-use the existing build, and follow it's patterns. Instead of creating a completely new beast.

We already build binaries. And we already have a container build. We just need to orchestrate it in a way to be run with every push to main.

carlosthe19916:
I was trying to address 2 problems:
- Generation of images for each commit
- Stop relying on Cargo.lock for the trustify-ui dependency

I would avoid to touch the current gh actions as I wouldn't like to break anything. This PR was just an option. I think we can come up with proper solution in another PR. Closing this PR

PR #641, opened 2024-08-06
Add psql dump generator and schema generator

ctron:


Cargo.toml
```diff
@@ -146,6 +146,7 @@ trustify-test-context = { path = "test-context" }
 # This will fail the build on targets where we need vendoring for openssl. Using rustls instead works around this issue.
 postgresql_archive = { version = "0.14.0", default-features = false, features = ["theseus", "rustls-tls"] }
 postgresql_embedded = { version = "0.14.0", default-features = false, features = ["theseus", "rustls-tls"] }
+postgresql_commands = { version = "0.14.0", default-features = false, features = ["tokio"] }
```
JimFuller-RedHat, line 149: why have we dropped rusttls-tls and theseus ?

common/src/db/embedded.rs
```diff
@@ -0,0 +1,46 @@
...
+    .instrument(info_span!("start database"))
+    .await?;
+
+    let config = crate::config::Database {
+        username: "postgres".into(),
+        password: "trustify".into(),
+        host: "localhost".into(),
+        name: "test".into(),
+        port: postgresql.settings().port,
+        min_conn: 25,
```
JimFuller-RedHat, line 38: just a mention - min and max conn will be configurable soon, will make sure to refer to them here

Cargo.toml
ctron: for `postgresql_commands`? Those features aren't available on this crate.

xtask/src/log.rs
```diff
@@ -0,0 +1,23 @@
+use anyhow::Context;
+use tracing_subscriber::{
+    filter::LevelFilter, layer::SubscriberExt, util::SubscriberInitExt, EnvFilter,
+};
+
+pub fn init_log() -> anyhow::Result<()> {
```
JimFuller-RedHat, line 6: nice ;)

JimFuller-RedHat (approved):
apart from some minor observations, LGTM

xtask/src/dataset.rs
```diff
@@ -0,0 +1,165 @@
...
+    #[arg(short, long)]
+    input: Option<PathBuf>,
+}
+
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+pub struct Instructions {
+    import: Vec<ImporterConfiguration>,
+}
+
+impl GenerateDump {
```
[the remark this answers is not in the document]
ctron: The goal was to have a process that can generate a dump that you (or the CI) can then import/load again. Instead of doing that with a manual instance, having a CLI/xtask to perform all the sequential steps automatically.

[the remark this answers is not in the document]
JimFuller-RedHat: ok makes sense, thx!

PR #643, opened 2024-08-06
feat: improve product details to return basic sbom data

dejanb:
This PR adds basic sbom data (like published and labels) to the product version details, like

```
{
    "id": "urn:uuid:5a0ee1f0-b549-427c-ab8b-6cd0574ae22b",
    "name": "ADVANCED-VIRT-RHEL-8.2.1",
    "vendor": {
        "id": "0e56a55f-bb03-45cc-973b-27c0aea36201",
        "name": "Organization: Red Hat"
    },
    "versions": [
        {
            "id": "urn:uuid:11d6e984-dac2-45b6-ab3f-5a52c0b896b5",
            "sbom": {
                "labels": {
                    "file": "https://access.redhat.com/security/data/sbom/beta/spdx/rhel-av-8.2.1.z.json.bz2",
                    "importer": "redhat-sbom",
                    "source": "https://access.redhat.com/security/data/sbom/beta/",
                    "type": "spdx"
                },
                "published": "2024-05-28T10:46:38Z"
            },
            "sbom_id": "urn:uuid:019103a2-69e8-71f2-8e65-4a09320ac745",
            "version": "8.2.1.z"
        }
    ]
}
```

This should make product details page much more performant as it will reduct API call to fetch SBOM for every version (which can be problematic when there's a lot of product versions).

@carlosthe19916 I omitted SBOM name on purpose for now. It is in a separate table and I wanted to avoid querying it if it's not necessary. IMHO, we should just use actual version as the link to the SBOM. But we can discuss this in UX calls.

@bobmcwhirter On the similar topic, I find SBOM data and entities a bit weird. For example, name is in the separate table, but it is in the `SbomHead` which then needs to be joined every time. While for example `published` is in the same table, but only available in `SbomSummary`. Maybe we could add the name to `sbom` table and have a `SbomHead` contain all info from it? In this case this api could just return it for the every version.


bobmcwhirter:
Tagging @ctron because I just use the SBOM tables as delivered.

I _think_ a single SBOM can reference many things with names/versions via the `DESCRIBED_BY` relationship, so a single column is... wrong?

And conceivably my `SbomHead` could also be wrong.

bobmcwhirter:
But, I agree, if we can reduce the number of joins from an SBOM to _the root thing(s) being described_ that'd be great. I think @ctron looked as some sentinel value to define the root(s) of an SBOM?

dejanb:
Cool. BTW. this doesn't need to be fixed as a part of this PR. I just raised it to see if it's something worth looking into further

ctron:
> I _think_ a single SBOM can reference many things with names/versions via the `DESCRIBED_BY` relationship, so a single column is... wrong?

That's true. An SBOM references `0..*` packages as "describing".

dejanb:
That's all good, I'm just wondering if `name` can be "moved" from `sbom_node` to `sbom` table, so we can get all data about sbom from a single table.

No matter this, I'll refactor this PR a bit tomorrow to use (a bit modified) `SbomHead`.

ctron:
> That's all good, I'm just wondering if `name` can be "moved" from `sbom_node` to `sbom` table, so we can get all data about sbom from a single table.
> 
> No matter this, I'll refactor this PR a bit tomorrow to use (a bit modified) `SbomHead`.

Well, all nodes have names. So that would mean we would need to spread it out to all tables.

dejanb:
Yeah, I understand pattern much better now. I guess my only reservation is that sbom name as a basic metadata lives in a separate table. But that's fine for now.

I would go and merge this if no one objects. And try to improve a code a bit in a separate PR as it's in fact separate topic where we can discuss it further.

ctron (approved):

PR #644, opened 2024-08-06
Rewrap with Arc<T>s.

bobmcwhirter:


bobmcwhirter:
Apparently was wrong to remove the Arc<T>.

PR #656, opened 2024-08-07
Split off build parts for the CI

ctron:
This splits off part of the release build into re-usable components.


ctron:
That should work as a pre-req for creating "latest" containers

.github/workflows/build-binary.yaml
```diff
@@ -0,0 +1,216 @@
...
+
+      - name: Setup | Cache Cargo
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: "${{ matrix.target }}-release${{ matrix.edition }}-${{ hashFiles('**/Cargo.lock') }}"
```
jcrossley3, line 83:
```suggestion
        uses: Swatinem/rust-cache@v2
```

ctron: Doesn't that conflict with the other cache? Let me read into the use of that.

ctron: Looks like that requires the use of the `shared-key` argument, which I added.

bobmcwhirter (approved):

PR #657, opened 2024-08-07
chore - add indexes to assist api/v1/purl searching.

JimFuller-RedHat:
This speeds up SQL statements with 

```
WHERE ("base_purl"."type" ILIKE '%curl%') OR ("base_purl"."namespace" ILIKE '%curl%') OR ("base
 _purl"."name" ILIKE '%curl%')))
```

used in api/v1/purl?q=


migration/src/m0000530_base_purl_index.rs
```diff
@@ -0,0 +1,72 @@
...
+            )
+            .await?;
+
+        Ok(())
+    }
+}
+
+#[allow(clippy::enum_variant_names)]
+#[derive(DeriveIden)]
+pub enum Indexes {
```
bobmcwhirter, line 60: tip: we've been removing `pub` from these types of internal enum identifiers to avoid having any cross-migration references.

migration/src/m0000530_base_purl_index.rs
```diff
@@ -0,0 +1,72 @@
...
+    BasePurlTypeGinIdx,
+    BasePurlNamespaceGinIdx,
+    BasePurlNameGinIdx,
+}
+
+#[derive(DeriveIden)]
```
bobmcwhirter, line 66: ibid

migration/src/m0000530_base_purl_index.rs
```diff
@@ -0,0 +1,72 @@
...
+#[async_trait::async_trait]
+#[allow(deprecated)]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .get_connection()
+            .execute_unprepared(
+                r#"
+                CREATE EXTENSION IF NOT EXISTS pg_trgm;
+                CREATE INDEX BasePurlTypeGinIdx ON base_purl USING GIN ((type) gin_trgm_ops);
```
bobmcwhirter, line 15: I guess SeaORM doesn't provide enough API to do this type-safely instead of stringy?

Worth a PR upstream (separately; not blockingly)?

bobmcwhirter (dismissed):

migration/src/m0000530_base_purl_index.rs
JimFuller-RedHat: yes I found nothing stable for create extensions or referring to **gin_trgm_ops** ... will attempt PR upstream

ctron: I think there some options, not sure they lead to the required result:

* https://docs.rs/sea-query/latest/sea_query/index/struct.IndexCreateStatement.html#method.full_text
* https://docs.rs/sea-query/latest/sea_query/index/struct.IndexCreateStatement.html#method.index_type

JimFuller-RedHat: Yes, full_text() just creates a 'plain ole' GIN index ... I was not able to use or create a GIN index with **gin_trgm_ops** using anything sea_orm provides ... we also probably need to research if we need tsquery/full text search generically across trustify at some point.

PR #665, opened 2024-08-09
fix: refactor sbom dto objects to better represent data model

dejanb:
This is a first attempt to refactor SBOM DTOs to be more inline with the others. 

* `SbomHead` now maps all fields from `sboms` table
* `SbomSummary` is `SbomHead`+packages
* `SbomDetails` is `SbomSummary`+advisories
* All DTOs now implement `from_entity()` pattern

### DIFF
--- a/modules/fundamental/src/purl/service/test.rs
+++ b/modules/fundamental/src/purl/service/test.rs
@@ -671,7 +671,7 @@ async fn gc_purls(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
         assert_eq!(
             1,
             sbom_service
-                .delete_sbom(sbom.head.id, Transactional::None)
+                .delete_sbom(sbom.summary.head.id, Transactional::None)
                 .await?
         );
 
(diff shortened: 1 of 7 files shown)


ctron (approved):
Looks good to me. But I'd defer to @bobmcwhirter for making a final decision.

dejanb:
Thanks. Now on the second look, other entities are not actually including summaries into details, but using head and duplicate other fields. Perhaps we should do that here as well. Let's hear from @bobmcwhirter :)

bobmcwhirter:
So yep, `FooHead` should be flattened into `FooSummary` and `FooDetails`.

It's dependent on your particular use-case if `FooDetails` should include `FooSummary` or not.

In some cases, I do *not* include `FooSummary` directly in `FooDetails`, because the summary is simpler, maybe aggregating/averaging scores for vulns or such, while the `FooDetails` has all of the data points. Maybe it *should* have the summary variant, dunno.

Since we have very intertwingly DTOs, having the ability to get "just enough" for the context is useful.

e.g.

As list of Vulns (VulnSummary) where each has enough Advisory information (AdvisorySummary) to find it and drill in deeper.  While also possibly saying *how many SBOMs are affected* without providing _entirely too much information_ for a *list-of-vulns* page.

But, nothing is set in stone. Do whatever makes sense in your given context.  We can always shift.

dejanb:
@bobmcwhirter thanks for the context. So you're OK with merging this as is?

I think in the future summary and details can diverge, like for example summary having only count of packages (and advisories?) while details have all. But we can deal with that when needed.

PR #667, opened 2024-08-09
Add basic progress support, clean up a few things

ctron:
This moves out the `ImportRunner` to a more neutral location, as we use it for the `xtask` directly.

It then add progress support for the git based walkers (cve, and osv). Implementing progress support via the tracing crate.

CSAF and SBOM need some pre-req work, as currently the progress support there is tied to `indicatif`.


modules/importer/src/runner/common/walker/git.rs
```diff
@@ -58,10 +58,11 @@ impl Handler for () {
     }
 }
 
-pub struct GitWalker<H, T>
+pub struct GitWalker<H, T, P>
```
bobmcwhirter, line 61: question: should we default `P=()` in that parameter?

modules/importer/src/runner/progress.rs
```diff
@@ -0,0 +1,189 @@
...
+    fn start(&self, _work: usize) -> Self::Instance {}
+}
+
+impl ProgressInstance for () {
+    async fn increment(&mut self, _work: usize) {}
+
+    async fn finish(self) {}
+}
+
+pub trait ProgressInstance: Sized {
```
bobmcwhirter, line 25: Would it be worthwhile (or possible?) to know some total-expected-ticks?  Able to give us not just "things are progressing" but "you're 85% done" sort thing?

modules/importer/src/runner/progress.rs
```diff
@@ -0,0 +1,189 @@
...
+#[derive(Clone, Debug)]
+pub struct TracingProgress {
+    pub name: String,
+    pub period: Duration,
+}
+
+pub struct TracingProgressInstance {
+    name: String,
+
+    total: usize,
```
bobmcwhirter, line 56: Oh, I see total here.

bobmcwhirter (approved):

modules/importer/src/runner/progress.rs
ctron: That's what it should do today. For example:

```
2024-08-09T14:01:51.798535Z  INFO run_sync:walk: run: Import OSV: https://github.com/RConsortium/r-advisory-database: 10.00% (ETA: 0s) percent=10.0 base="/tmp/.tmpivFAHw/vulns"
```

ETA 0s, because it's rounded down to seconds, and it's below that in this case.

ctron: Ok, added `total` and `current` to the trace message too:

```
2024-08-09T14:30:52.929783Z  INFO run_sync:walk: run: Import OSV: https://github.com/RConsortium/r-advisory-database: 10.00% (ETA: 0s) current=1 total=10 percent=10.0 base="/tmp/.tmpTWGs40/vulns"
```

modules/importer/src/runner/common/walker/git.rs
ctron: Technically possible. But there's a `new` function, which has that and is expected to be used instead of manually creating the struct.

PR #680, opened 2024-08-13
Light up migrations and entities for:

bobmcwhirter:
* licenses, in general.
* sbom-centric license assertions
* add license info into purl DTOs.


entity/src/cpe_license_assertion.rs
```diff
@@ -0,0 +1,16 @@
...
+    pub id: Uuid,
+    pub license_id: Uuid,
+    pub cpe_id: Uuid,
+    pub sbom_id: Uuid,
+}
+
+#[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
+pub enum Relation {}
+
+impl ActiveModelBehavior for ActiveModel {}
```
dejanb, line 16: Looks good on first look. The only thing I noticed that relations for cpe assertion are missing. But that might be coming later

bobmcwhirter: Yah, currently I haven't tried any fetching for CPEs, which is where Relations come into play.

Do we have endpoints for CPEs currently? Directly or via Products?

dejanb: Not that I'm aware of. I hope to have something via products soon

dejanb (approved):

Issue #654, opened 2024-08-07
Improve memory consumption


ctron:
I noticed that, at least when importing, the server might consume huge amounts of memory (~3GiB). I know that some documents are just big. But we should keep an eye on this and try to improve the situation where possible.

JimFuller-RedHat:
observation: avoiding entire documents being loaded into memory for processing, usually imply streaming approach during ingestion though I have no clue how one does this in rust with serde.

ctron:
I fear a pattern we have for SBOMs is:

* load into memory
* decompress in memory
* deserialize in memory
* create database structures in memory
* execute statements

Holding on to results of every step, and only dropping all memory at the end.

jcrossley3:
> observation: avoiding entire documents being loaded into memory for processing, usually imply streaming approach during ingestion though I have no clue how one does this in rust with serde.

In [v1](https://github.com/trustification/trustification), we went to great pains to avoid loading an entire doc into memory, only to later realize that it was the only way we could accurately validate the document. In light of that experience, we avoided all that pain in v2, at the risk of memory exhaustion. We can probably strike a better balance, but we need to ensure that we can sufficiently validate a doc that doesn't fully reside in RAM. We should solve the validation problem first.

jcrossley3:
> Holding on to results of every step, and only dropping all memory at the end.

Dropping the byte array after deserializing might be the lowest hanging fruit?

helio-frota:
> Dropping the byte array after deserializing might be the lowest hanging fruit?

array -> array ? yes
Vec -> yes or I'm not understanding rust yet

bobmcwhirter:
Yep. Let's add some scope or explicit drops.

JimFuller-RedHat:
FWIW if validators also do not stream then that might be a problem addressed upstream (if we are validating with 3rd party parser).

jcrossley3:
Any thoughts on the best way to measure memory usage in rust programs? What's the simplest way I can determine what effect my changes make?

jcrossley3:
> Holding on to results of every step, and only dropping all memory at the end.

I need help finding an example of this. When I look at the loaders, they each deserialize a byte buffer, e.g.

```
        let csaf: Csaf =
            info_span!("parse document").in_scope(|| serde_json::from_reader(document))?;
```

But because `document` is moved into `from_reader`, I believe it's dropped once the `Csaf` struct is returned, right? Where is an example of us holding onto the bytes after deserialization?

[...]

PR #682, opened 2024-08-13
feat: replace SyncAdapter with SyncIoBridge to reduce RAM usage

jcrossley3:
Because serde deserialization is synchronous, and because we immediately read back the bytes we store when ingesting a doc (to ensure we can), we store them in RAM before asking serde to deserialize them.

So now we deserialize through the bridge, hopefully using less RAM in the process.

Helps with #654 though I'm sure there's more to do.

### DIFF
--- a/modules/ingestor/src/service/advisory/csaf/loader.rs
+++ b/modules/ingestor/src/service/advisory/csaf/loader.rs
@@ -14,9 +14,9 @@ use csaf::{
     Csaf,
 };
 use sbom_walker::report::ReportSink;
-use std::{io::Read, str::FromStr};
+use std::str::FromStr;
 use time::OffsetDateTime;
-use tracing::{info_span, instrument};
+use tracing::instrument;
 use trustify_common::{db::Transactional, hashing::Digests, id::Id};
 use trustify_cvss::cvss3::Cvss3Base;
 use trustify_entity::labels::Labels;
@@ -51,17 +51,14 @@ impl<'g> CsafLoader<'g> {
         Self { graph }
     }
 
-    #[instrument(skip(self, labels, document), err)]
-    pub async fn load<R: Read>(
+    #[instrument(skip(self, labels), err)]
+    pub async fn load(
         &self,
         labels: impl Into<Labels>,
-        document: R,
+        csaf: Csaf,
         digests: &Digests,
     ) -> Result<IngestResult, Error> {
         let warnings = Warnings::new();
-
-        let csaf: Csaf =
-            info_span!("parse document").in_scope(|| serde_json::from_reader(document))?;
 
         let tx = self.graph.transaction().await?;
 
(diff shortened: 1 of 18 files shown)


modules/ingestor/src/service/advisory/csaf/loader.rs
```diff
@@ -51,18 +51,15 @@ impl<'g> CsafLoader<'g> {
...
     }
 
-    #[instrument(skip(self, labels, document), err)]
-    pub async fn load<R: Read>(
+    #[instrument(skip(self, labels), err)]
+    pub async fn load(
         &self,
         labels: impl Into<Labels>,
-        document: R,
+        csaf: Csaf,
```
bobmcwhirter, line 58: I know WIP, but... would consuming the `Read` not also consume/drop the underlying buffer?  Or does `serde` maintain references into the original bytes? Maybe deserialize-owned?

test-context/src/lib.rs
```diff
@@ -27,6 +27,11 @@ pub struct TrustifyContext {
     postgresql: Option<PostgreSQL>,
 }
 
+use peak_alloc::PeakAlloc;
+
+#[global_allocator]
+static PEAK_ALLOC: PeakAlloc = PeakAlloc;
```
bobmcwhirter, line 33: This looks useful full-time/always, yeah?

jcrossley3: Maybe. I hadn't intended on leaving it in, but I'll reconsider.

modules/ingestor/src/service/advisory/csaf/loader.rs
jcrossley3: I think you're correct, but it's not the specific problem I'm attempting to solve. What I'm trying to avoid is the `SyncAdapter`. That's the thing loading the entire doc into memory. I'm hoping that using the `SyncIoBridge` instead will minimize memory usage at the cost of spawning a thread.

Ultimately, our troubles come down to serde not supporting async. And I'm trying to come up with a `Read` impl that doesn't blow out our RAM.

I'm open to other ideas, of course.

test-context/src/lib.rs
jcrossley3: Further, it'll be mostly useless unless you `--test-threads=1`

test-context/src/lib.rs
```diff
@@ -24,9 +25,13 @@ pub struct TrustifyContext {
     pub graph: Graph,
     pub storage: FileSystemBackend,
     pub ingestor: IngestorService,
+    pub mem_limit_mb: f32,
     postgresql: Option<PostgreSQL>,
 }
 
+#[global_allocator]
```
bobmcwhirter, line 32: question: this doesn't find its way into the final binary, yeah? Purely when testing?

bobmcwhirter (approved):

bobmcwhirter:
I believe you. No need for proof. Cocktail question really.

test-context/src/lib.rs
helio-frota: Confirmed :+1: 
https://github.com/helio-frota/mango?tab=readme-ov-file#mango

bobmcwhirter (approved):
LGTM! Glad you sorted the clone.

jcrossley3:
> LGTM! Glad you sorted the clone.

Thanks for your help on that!

PR #688, opened 2024-08-15
Begin loadering for ClearlyDefined "sbom" information about licenses.

bobmcwhirter:
Also, migrate from deprecated `serde_yaml` to `serde_yml`.


modules/ingestor/src/service/format.rs
```diff
@@ -170,7 +191,7 @@ fn masked<N: Mask>(mask: N, bytes: &[u8]) -> Result<Option<String>, Error> {
         .transpose()
 }
 
-async fn from_stream<S, T>(stream: S) -> Result<T, Error>
+async fn json_from_stream<S, T>(stream: S) -> Result<T, Error>
```
jcrossley3, line 194: This is a better name even without `yaml_from_stream` tbh

modules/ingestor/src/service/format.rs
```diff
@@ -50,34 +52,39 @@ impl<'g> Format {
...
-                let v: Value = from_stream(stream).await?;
+                let v: Value = json_from_stream(stream).await?;
                 let sbom = Bom::parse_json_value(v)
                     .map_err(|err| Error::UnsupportedFormat(format!("Failed to parse: {err}")))?;
 
                 loader.load(labels, sbom, digests).await
             }
+            Format::ClearlyDefined => {
+                let loader = ClearlyDefinedLoader::new(graph);
+                let v: serde_yml::Value = yaml_from_stream(stream).await?;
```
jcrossley3, line 85: Why not just create the Curation? Why parse it once here and then again inside the loader?
```suggestion
                let v: Curation = yaml_from_stream(stream).await?;
```

bobmcwhirter: Good call.

modules/ingestor/src/service/format.rs
```diff
@@ -120,40 +130,60 @@ impl<'g> Format {
...
     pub fn is_csaf(bytes: &[u8]) -> Result<bool, Error> {
-        Ok(masked(
+        match masked(
             key("document").and(key("csaf_version")).and(depth(2)),
             bytes,
-        )?
-        .is_some())
+        ) {
+            Ok(Some(_)) => Ok(true),
+            _ => Ok(false),
```
jcrossley3, line 138: The `_` prevents me from understanding why these changes to all the `is_*` fn's that call `masked` are required without any changes to `masked` itself. Have these always been wrong?

modules/ingestor/src/service/format.rs
```diff
@@ -184,6 +214,20 @@ where
...
+async fn yaml_from_stream<S, T>(stream: S) -> Result<T, Error>
+where
+    S: Stream<Item = Result<Bytes, anyhow::Error>> + Send + 'static,
+    T: serde::de::DeserializeOwned + Send + 'static,
+{
+    Ok(tokio::task::spawn_blocking(move || {
+        let stream = pin!(stream);
+        let stream = stream.map_err(|e| io::Error::new(io::ErrorKind::Other, format!("{e:?}")));
+        let reader = SyncIoBridge::new(StreamReader::new(stream));
+        info_span!("parse document").in_scope(|| serde_yml::from_reader(reader))
```
jcrossley3, line 226: I don't love that both `*_from_stream` fn's are identical except for this one line. We can refactor later, I suppose, but it'd be nice to remove all the redundancy if you can.

modules/ingestor/src/service/format.rs
bobmcwhirter: You'd prefer the explicit `Err(_) | Ok(None)` instead of the catch-all? I'm cool with that also.

modules/ingestor/src/service/format.rs
bobmcwhirter: De-duplicated.

modules/ingestor/src/service/format.rs
```diff
@@ -120,39 +130,59 @@ impl<'g> Format {
...
-        Ok(masked(
+        match masked(
             key("document").and(key("csaf_version")).and(depth(2)),
             bytes,
-        )?
-        .is_some())
+        ) {
+            Ok(Some(_)) => Ok(true),
+            Err(_) | Ok(None) => Ok(false),
+        }
```
jcrossley3, line 139: Don't hate me, please, but now that I understand why you're eating all the Errors potentially returned from the `is_*` fn's, I'm inclined to ask you to change all their return types from `Result<bool, Error>` to `bool`.

jcrossley3 (changes requested):

modules/ingestor/src/service/format.rs
bobmcwhirter: I did consider it, but there's places where ... it's valid JSON, it's valid CycloneDX (or SPDX) but... not a *supported* version. Which is a valid `Error` case.

```
            Ok(Some(x)) => Err(Error::UnsupportedFormat(format!(
                "CycloneDX version {x} is unsupported; try 1.3, 1.4, or 1.5"
            ))),
```

jcrossley3: Good point. I guess there's no harm allowing for the possibility.

bobmcwhirter: So while we catch all errors and turn them into false, some inherently non-error results can be turned into perfectly valid Errors.

jcrossley3 (approved):

PR #690, opened 2024-08-16
Reduce a bit more duplication on *_from_stream.

bobmcwhirter:


modules/ingestor/src/service/format.rs
```diff
@@ -230,14 +230,10 @@ where
...
-                .in_scope(|| serde_json::from_reader(reader))
-                .map_err(Error::Json),
-            StreamFlavor::Yaml => info_span!("parse document")
-                .in_scope(|| serde_yml::from_reader(reader))
-                .map_err(Error::Yaml),
-        }
+        info_span!("parse document").in_scope(|| match flavor {
+            StreamFlavor::Yaml => serde_yml::from_reader(reader).map_err(Error::Yaml),
+            StreamFlavor::Json => serde_json::from_reader(reader).map_err(Error::Json),
+        })
```
jcrossley3, line 236: I feel like the construction of the bridge should be a part of the parsing span, so will this work?

```rust
    tokio::task::spawn_blocking(move || {
        info_span!("parse document").in_scope(|| {
            let stream = pin!(stream);
            let stream = stream.map_err(|e| io::Error::new(io::ErrorKind::Other, format!("{e:?}")));
            let reader = SyncIoBridge::new(StreamReader::new(stream));
            match flavor {
                StreamFlavor::Json => serde_json::from_reader(reader).map_err(Error::Json),
                StreamFlavor::Yaml => serde_yml::from_reader(reader).map_err(Error::Yaml),
            }
        })
    })
    .await?
```

jcrossley3 (changes requested):

jcrossley3 (approved):

modules/ingestor/src/service/format.rs
jcrossley3: Nevermind. I changed my mind.

bobmcwhirter: Lemme try!

bobmcwhirter: Seems to.

PR #696, opened 2024-08-19
Stop deserializing and re-serializing OSV and CVE content during import.

bobmcwhirter:
Keep them all as a Vec<u8> until handed onwards to the ingestor to perform consistent and truthful hashing.


jcrossley3 (approved):
LGTM!

Issue #618, opened 2024-07-29
discovery of what it will take to implement 'find root package/component of a package/component'


JimFuller-RedHat:
Today we have packages (actually purls) ... we should experiment how we can represent dependency tree, that spans beyond any single package (for example, container > container > rpm > component > component)  and query to answer the question 'find root package/component'.

JimFuller-RedHat:
related to https://github.com/trustification/trustify/discussions/556

PR #709, opened 2024-08-26
implement /api/v1/analysis/root-component

JimFuller-RedHat:
This is an experimental attempt to address resolving a component's root-components (using petgraph) and partly addressing #618.

Points of interest:
* included draft arch decision at docs/adrs/00001-graph-analytics.md 
* added a postgres func  to resolve full pURL from a qualified_purl_id ... which we can rip out easier if we ever consolidate those tables
* added draft implementation of api/v1/analysis/root-component

There are lots of things to do eg. I wanted to get an end to end first, have not optimised and need to consolidate code but thought it better to get this out there earlier.


JimFuller-RedHat:
closing tihs to make cleaner PR

PR #710, opened 2024-08-26
Checkpoint importation of the f'ing XML CWE database.

bobmcwhirter:


bobmcwhirter:
Just trying to get CI kicked...

modules/ingestor/src/service/format.rs
```diff
@@ -92,19 +97,33 @@ impl<'g> Format {
...
+
+    pub fn is_format(&self, bytes: &[u8]) -> Result<bool, Error> {
+        match self {
+            Format::OSV => Self::is_osv(bytes),
+            Format::CSAF => Self::is_csaf(bytes),
+            Format::CVE => Self::is_cve(bytes),
+            Format::SPDX => Self::is_spdx(bytes),
+            Format::CycloneDX => Self::is_cyclonedx(bytes),
+            Format::ClearlyDefined => Self::is_clearly_defined(bytes),
+            Format::CweCatalog => Self::is_cwe_catalog(bytes),
```
ctron, line 115: Does that mean we are parsing the full catalog, just to "check" if its the type we are expecting it to be?

Maybe off topic, but I really think that this logic should be removed and the type should be provided by the user (uploader). And then just fail when it doesn't match the expected type. Instead of testing all combinations for all uploaded documents.

ctron (approved):
Looks good to me me. However, I really have concerns over the format detection. I am not sure this is worth the effort. As it affects all of the documents we upload. Times the number of formats we support.

modules/ingestor/src/service/format.rs
jcrossley3: > Does that mean we are parsing the full catalog, just to "check" if its the type we are expecting it to be?

No, but I share your concern...

> Maybe off topic,

Definitely 

> but I really think that this logic should be removed and the type should be provided by the user

Why?

modules/ingestor/src/service/format.rs
```diff
@@ -92,19 +97,33 @@ impl<'g> Format {
...
-                Err(Error::UnsupportedFormat(es)) => {
-                    Err(Error::UnsupportedFormat(format!("{ea}\n{es}")))
-                }
-                x => x,
-            },
-            x => x,
+        for format in Self::iter() {
+            if format.is_format(bytes)? {
+                return Ok(format);
+            }
```
jcrossley3, line 123: I don't love this. I don't trust the iteration order, and I think (in the spirit of @ctron 's concern above) that we should be intentional about the order of our format checks. The checks are fast (and when metrics reveal that to be false, easily parallelizable) but we should check for the ones we expect more frequently first. Hence, my divvying the logic up between advisories and SBOM's, as I expect to see far more of the former than the latter. Similarly we should check for the more common formats first.

jcrossley3 (changes requested):

modules/ingestor/src/service/format.rs
bobmcwhirter: I believe `strum` does assert that the `EnumIter` is in definition order, so we can put "expensive" ones last.

Talking to @ctron, I also agree, iff we know (e.g., from an importer) then pass directly, and reserve the "guess" case for "user is dragging/dropping a file into the browser".

modules/ingestor/src/service/format.rs
bobmcwhirter: fwiw, I could use something more streamy to detect the XML type, like your jsonpath stuff for detectoring correct JSON formats.

modules/ingestor/src/service/format.rs
jcrossley3: I disagree that rearranging the order of the variants and assuming that strum honors that order is more obvious than [the implementations of `advisory_from_bytes` and `sbom_from_bytes`](https://github.com/trustification/trustify/blob/6f0a558be9d4fe78c51eba550b6218b83e78d878/modules/ingestor/src/service/format.rs#L110-L136).

jcrossley3: > Talking to @ctron, I also agree, iff we know (e.g., from an importer) then pass directly, and reserve the "guess" case for "user is dragging/dropping a file into the browser".

Sure, adding an `Option<Format>` ingest argument makes sense.

bobmcwhirter: Either way, can we move that problem to a different subsequent PR?

jcrossley3: > Either way, can we move that problem to a different subsequent PR?

Yes

bobmcwhirter: Yah, I think keep the guessing for the human-centric usecase, but the CSAF Importer is pretty sure it's CSAF.  The OSV importer knows it's OSV, etc.

PR #712, opened 2024-08-26
feat: add progress support, backed by the database

ctron:
This extends the progress support to CSAF and SBOMs. It also persists the progress state in the database, and makes it available on the API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -387,7 +387,6 @@ dependencies = [
  "cfg-if",
  "getrandom",
  "once_cell",
- "serde",
  "version_check",
  "zerocopy",
 ]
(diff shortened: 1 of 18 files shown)


bobmcwhirter:
Looks good, just one question:

If it persists in the DB, and the system shuts down mid-ingestoring, and then restarts, do the tables in the DB represent stale data?

Or will it reset when the ingestor starts again?

I understand the DB is nice especially if we distribute importers to different processes than the API servers (or have multiple API servers in rotation).

ctron:
> Or will it reset when the ingestor starts again?

That's already happening. Clearing the progress data with it now.

PR #713, opened 2024-08-26
feat: speed up initial clones of git repos

jcrossley3:
Reduces clone time for CVE importer from ~30m down to ~1m

### DIFF
--- a/modules/importer/src/runner/common/walker/git.rs
+++ b/modules/importer/src/runner/common/walker/git.rs
@@ -203,7 +203,13 @@ where
             }
 
             let mut fo = Self::create_fetch_options();
-            fo.depth(2);
+            if self.continuation.0.is_none() {
+                if cfg!(test) {
+                    fo.depth(3);
+                } else {
+                    fo.depth(1);
+                }
+            }
             builder.fetch_options(fo).clone(&self.source, path)
         });
 
--- a/modules/importer/src/runner/common/walker/git.rs
+++ b/modules/importer/src/runner/common/walker/git.rs
@@ -202,7 +202,8 @@ where
                 builder.branch(branch);
             }
 
-            let fo = Self::create_fetch_options();
+            let mut fo = Self::create_fetch_options();
+            fo.depth(1);
             builder.fetch_options(fo).clone(&self.source, path)
         });
 
@@ -213,7 +214,7 @@ where
                 let repo = info_span!("open repository").in_scope(|| Repository::open(path))?;
 
                 info_span!("fetching updates").in_scope(|| {
-                    log::debug!("Fetching updates");
+                    log::info!("Fetching updates");
                     let mut remote = repo.find_remote("origin")?;
 
                     let mut fo = Self::create_fetch_options();
@@ -241,7 +242,7 @@ where
             }
         };
 
-        log::debug!("Repository cloned or updated");
+        log::info!("Repository cloned or updated");
 
         // discover files between "then" and now
 
--- a/modules/importer/src/runner/common/walker/git.rs
+++ b/modules/importer/src/runner/common/walker/git.rs
@@ -203,7 +203,7 @@ where
             }
 
             let mut fo = Self::create_fetch_options();
-            fo.depth(1);
+            fo.depth(2);
             builder.fetch_options(fo).clone(&self.source, path)
         });
 
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,5 @@
 [alias]
 xtask = "run --package xtask --"
+
+[env]
+CARGO_WORKSPACE_ROOT = { value = "", relative = true }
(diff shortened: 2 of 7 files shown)


modules/importer/src/runner/common/walker/git.rs
```diff
@@ -202,7 +202,8 @@ where
                 builder.branch(branch);
             }
 
-            let fo = Self::create_fetch_options();
+            let mut fo = Self::create_fetch_options();
+            fo.depth(1);
```
bobmcwhirter, line 206: This `depth` is depth of git history fetched, yeah? Not directory tree depth? If so, LGTM.

bobmcwhirter (approved):

modules/importer/src/runner/common/walker/git.rs
jcrossley3: I believe so, yes. I worry a bit that subsequent updates might somehow be affected by the initial shallow clone, but I couldn't find any evidence of that by bouncing the process. Seemed to pick up where it left off each time.

bobmcwhirter:
I still approve.

bobmcwhirter (approved):

modules/importer/src/runner/common/walker/git.rs
```diff
@@ -202,7 +202,8 @@ where
                 builder.branch(branch);
             }
 
-            let fo = Self::create_fetch_options();
+            let mut fo = Self::create_fetch_options();
+            fo.depth(2);
```
ctron, line 206: Ok, that might cause a problem later on in the code. In case of a continuation token (the git commit sha), having a limited depth will break that logic: https://github.com/trustification/trustify/blob/f66ef06d635a81f44c8e696b191b2fcc5144e6dd/modules/importer/src/runner/common/walker/git.rs#L249-L288

ctron (changes requested):
As I understand that would not allow having all git commits available. Which would break the logic.

ctron:
I did take a second look, because I noticed that the depth limit only affects the "initial" clone. And that would be fine.

However, when using an ephemeral store for the clone, it does matter. In that case, the original clone would be gone, and the next clone would require the full history. Only on the case of a persistent storage, the current logic would work.

An alternative would be that we only limit the depth when there is no existing clone **plus** we don't have a continuation token. If both cases are true, we know that we scan skip history.

jcrossley3:
> I did take a second look, because I noticed that the depth limit only affects the "initial" clone. And that would be fine.

Yeah, that's why I marked your initial question "resolved".

> However, when using an ephemeral store for the clone, it does matter. In that case, the original clone would be gone, and the next clone would require the full history. Only on the case of a persistent storage, the current logic would work.

I'm not sure I see the value of the ephemeral behavior when it takes ~30min to clone the repo.

> An alternative would be that we only limit the depth when there is no existing clone **plus** we don't have a continuation token. If both cases are true, we know that we scan skip history.

I probably don't understand the logic of the continuation token, but as long as it's "later" than the original shallow clone, we should be good, right?

ctron:
> > However, when using an ephemeral store for the clone, it does matter. In that case, the original clone would be gone, and the next clone would require the full history. Only on the case of a persistent storage, the current logic would work.
> 
> I'm not sure I see the value of the ephemeral behavior when it takes ~30min to clone the repo.
> 

Yes, that's true. So removing that would be an option too.

> > An alternative would be that we only limit the depth when there is no existing clone **plus** we don't have a continuation token. If both cases are true, we know that we scan skip history.
> 
> I probably don't understand the logic of the continuation token, but as long as it's "later" than the original shallow clone, we should be good, right?

In general for the importers, there is some string that the runner can hand back that it will receive the next time with the idea of: only check from there onward. For the git stuff, that's a git commit hash. It will evaluate only changed files between that token/commit and `HEAD`, and skip others.

jcrossley3:
> > I probably don't understand the logic of the continuation token, but as long as it's "later" than the original shallow clone, we should be good, right?
> 
> In general for the importers, there is some string that the runner can hand back that it will receive the next time with the idea of: only check from there onward. For the git stuff, that's a git commit hash. It will evaluate only changed files between that token/commit and `HEAD`, and skip others.

So the logic in the PR should be sound, right?

ctron:
> > > I probably don't understand the logic of the continuation token, but as long as it's "later" than the original shallow clone, we should be good, right?
> > 
> > 
> > In general for the importers, there is some string that the runner can hand back that it will receive the next time with the idea of: only check from there onward. For the git stuff, that's a git commit hash. It will evaluate only changed files between that token/commit and `HEAD`, and skip others.
> 
> So the logic in the PR should be sound, right?

Sorry, no. It would be if we either:

* Remove the ability to use an ephemeral working directory (and document that somehow)
* Check if the there's a continuation token present, and in this case fetch the full depth

ctron (approved):

PR #714, opened 2024-08-27
test: add csaf performance test

ctron:
This adds a hand full of CSAF performance test. The smaller ones look ok-ish. But the bigger they get, the slower they become. The "large" one (23 MiB uncompressed) takes ~40 seconds.


modules/ingestor/tests/performance/csaf.rs
```diff
@@ -0,0 +1,56 @@
+#![allow(clippy::expect_used)]
```
helio-frota, line 1: We can remove this line

ctron: Change pushed.

helio-frota (approved):

PR #717, opened 2024-08-27
fix performance regression, speeds things up ~10x

ctron:


modules/ingestor/src/service/advisory/csaf/loader.rs
```diff
@@ -98,51 +98,54 @@ impl<'g> CsafLoader<'g> {
...
-                        summary: None,
-                        description: None,
-                        discovery_date: vulnerability.discovery_date.and_then(|date| {
-                            OffsetDateTime::from_unix_timestamp(date.timestamp()).ok()
-                        }),
-                        release_date: vulnerability.release_date.and_then(|date| {
-                            OffsetDateTime::from_unix_timestamp(date.timestamp()).ok()
-                        }),
-                        cwes: vulnerability.cwe.as_ref().map(|cwe| vec![cwe.id.clone()]),
+        let Some(cve_id) = &vulnerability.cve else {
```
bobmcwhirter, line 101: I need to get `let/else` into my brain. Nice.

bobmcwhirter (approved):

modules/ingestor/src/service/advisory/csaf/loader.rs
ctron: It can safe you one level of indentation. I like that.

PR #718, opened 2024-08-27
fix: disable format detection for importers

### LINKED ISSUE #715: Remove auto-detection of file formats from critical paths.
I don't love this. I don't trust the iteration order, and I think (in the spirit of @ctron 's concern above) that we should be intentional about the order of our format checks. The checks are fast (and when metrics reveal that to be false, easily parallelizable) but we should check for the ones we expect more frequently first. Hence, my divvying the logic up between advisories and SBOM's, as I expect to see far more of the former than the latter. Similarly we should check for the more common formats first.

_Originally posted by @jcrossley3 in https://github.com/trustification/trustify/pull/710#discussion_r1732885794_

jcrossley3:
Fixes #715

We only auto-detect formats for the actix handlers providing our API.

### DIFF
--- a/modules/fundamental/benches/bench.rs
+++ b/modules/fundamental/benches/bench.rs
@@ -25,6 +25,7 @@ pub(crate) mod trustify_benches {
 
     use trustify_common::db::Transactional;
     use trustify_entity::labels::Labels;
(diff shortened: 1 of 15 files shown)


ctron (approved):
A nice compromise.

Issue #703, opened 2024-08-22
Info: pg-emb `1.15x+` requires rust `1.80.x`


helio-frota:
```console
Compiling postgresql_archive v0.15.0
error[E0658]: use of unstable library feature 'lazy_cell'
 --> /home/heliofrota/.cargo/registry/src/index.crates.io-6f17d22bba15001f/postgresql_archive-0.15.0/src/extractor/registry.rs:8:22
  |
8 | use std::sync::{Arc, LazyLock, Mutex, RwLock};
  |                      ^^^^^^^^
  |
  = note: see issue #109736 <https://github.com/rust-lang/rust/issues/109736> for more information

error[E0658]: use of unstable library feature 'lazy_cell'
  --> /home/heliofrota/.cargo/registry/src/index.crates.io-6f17d22bba15001f/postgresql_archive-0.15.0/src/extractor/registry.rs:10:18
   |
10 | static REGISTRY: LazyLock<Arc<Mutex<RepositoryRegistry>>> =
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #109736 <https://github.com/rust-lang/rust/issues/109736> for more information

error[E0658]: use of unstable library feature 'lazy_cell'
  --> /home/heliofrota/.cargo/registry/src/index.crates.io-6f17d22bba15001f/postgresql_archive-0.15.0/src/extractor/registry.rs:11:5
   |
11 |     LazyLock::new(|| Arc::new(Mutex::new(RepositoryRegistry::default())));
   |     ^^^^^^^^
   |
   = note: see issue #109736 <https://github.com/rust-lang/rust/issues/109736> for more information
```

ctron:
I don't think anything speaks against performing an upgrade to Rust 1.80.

helio-frota:
yeah sorry I messed with the title... updated :+1:

PR #719, opened 2024-08-28
build: update Rust to 1.80.1, update garage-door

ctron:
Pre-req of https://github.com/trustification/trustify/issues/703

This also updates garage-door in order to work around https://github.com/trustification/trustify-ui/issues/145

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2322,9 +2322,9 @@ dependencies = [
 
 [[package]]
 name = "garage-door"
-version = "0.1.0"
+version = "0.1.1"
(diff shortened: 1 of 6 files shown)


helio-frota (approved):
thanks for this :+1: 


I was trying to understand why we are having (not using yet) it... 

probably mock is weird for the amount of data we have
https://github.com/martelinho-de-ouro/moria-watcher/blob/39a1b1de5f9070490b213b016f74a19242b88879/service/tests/prepare.rs#L13-L47

PR #722, opened 2024-08-28
 implement /api/v1/analysis/root-component

JimFuller-RedHat:
This is an experimental attempt to address resolving a component's root-components (using petgraph) and partly addressing https://github.com/trustification/trustify/issues/618.

Points of interest:

    included draft arch decision at docs/adrs/00001-graph-analytics.md
    added a postgres func to resolve full pURL from a qualified_purl_id ... which we can rip out easier if we ever consolidate those tables
    added draft implementation of api/v1/analysis/root-component

There are lots of things to do eg. I wanted to get an end to end first, have not optimised and need to consolidate code but thought it better to get this out there earlier.


trustd/src/db.rs
```diff
@@ -74,6 +74,10 @@ impl Run {
                 "pg_stat_statements".to_string(),
             ),
             ("random_page_cost".to_string(), "1.1".to_string()),
+            (
+                " max_parallel_workers_per_gather".to_string(),
+                "4".to_string(),
```
JimFuller-RedHat, line 79: FWIW the default is 2

migration/src/m0000585_analysis_api_index.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[async_trait::async_trait]
+#[allow(deprecated)]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .get_connection()
+            .execute_unprepared(
+                r#"
+                CREATE EXTENSION IF NOT EXISTS pg_trgm;
```
bobmcwhirter, line 14: question: does this work with the embedded pgsql in PM mode?

migration/tests/getpurl.rs
```diff
@@ -0,0 +1,65 @@
+use migration::sea_orm::{EntityTrait, Statement};
+use migration::ConnectionTrait;
+use std::str::FromStr;
+use test_context::test_context;
+use test_log::test;
+use trustify_common::db::Database;
+use trustify_common::purl::Purl;
+use trustify_test_context::TrustifyContext;
+
+async fn getpurl(
```
bobmcwhirter, line 10: suggestion: `async fn get_purl(`

modules/fundamental/src/analysis/service/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+            ancestors,
+        }
+    }
+}
+
+pub struct AnalysisService {
+    db: Database,
+}
+
+#[allow(clippy::too_many_arguments)]
```
bobmcwhirter, line 54: I agree with clippy. That's a lot of stringly-typed arguments.

modules/fundamental/src/analysis/service/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+pub fn add_node(
+    g: &mut Graph<PackageNode, Relationship, petgraph::Directed>,
+    nodes: &mut HashMap<String, NodeIndex>,
+    component_purl: &String,
```
bobmcwhirter, line 58: suggestion: `&str` instead of `&String`.  Calling it with a reference to a `String` satisfies `&str` as does `"string literals"` generally used in tests.

Only ask for non-ref `String` if you want to take ownership, else `&str`.

modules/fundamental/src/analysis/service/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+    let sanitized_search_value = regex::escape(search_value);
+
+    let pattern = match Regex::new(&sanitized_search_value) {
+        Ok(pattern) => pattern,
+        Err(_) => return Vec::new(), // Return an empty vector if the regex pattern is invalid
+    };
+
+    nodes
+        .iter()
+        .filter(|(key, _)| pattern.is_match(key))
```
bobmcwhirter, line 93: HashMap might have `filter_map` like some other collections do. 

You can then do something akin to

```
nodes.iter()
   .filter_map(|key, value| 
        if pattern.is_match(key) { 
            Some(*value) 
        } else { 
            None 
        }
    )
    .collect()
```

modules/fundamental/src/analysis/service/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+        .map(|(_, value)| *value)
+        .collect()
+}
+
+pub fn exact_name_find(nodes: &HashMap<String, NodeIndex>, search_value: &str) -> Vec<NodeIndex> {
+    nodes
+        .iter()
+        .filter_map(|(key, &value)| {
```
bobmcwhirter, line 101: Hey, `filter_map`!

modules/fundamental/src/analysis/service/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+    graph: &petgraph::Graph<PackageNode, Relationship, petgraph::Directed>,
+    node: NodeIndex,
+    original_node: NodeIndex,
+) -> Vec<NodeIndex> {
+    // we need order
+    let mut ancestor_nodes: Vec<NodeIndex> = Vec::new();
+    if node != original_node {
+        ancestor_nodes.push(node);
+    }
+    fn dfs(
```
bobmcwhirter, line 123: I always forget about fn-local fns.

modules/fundamental/src/analysis/service/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+        original_node: NodeIndex,
+        ancestor_nodes: &mut Vec<NodeIndex>,
+    ) {
+        for neighbor in graph.neighbors_directed(node, Direction::Outgoing) {
+            if neighbor == original_node {
+                break; // no need to descend further past the original node in the tree
+            }
+            if !ancestor_nodes.contains(&neighbor) {
+                ancestor_nodes.push(neighbor);
+                dfs(graph, neighbor, original_node, ancestor_nodes);
```
bobmcwhirter, line 135: Are we concerned about recursion depth at all? Worth translating to a stack/queue-based algorithm for depth-first?

modules/fundamental/src/analysis/service/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+            .await?;
+
+        // keep track of nodes / indices for efficient searching
+        let mut nodes: HashMap<String, NodeIndex> = HashMap::new();
+
+        // load package relationships into graph
+        for row in relationship_results {
+            let sbom_published = row
+                .try_get("", "published")
+                .unwrap_or("NOVALUE".to_string()); // TODO: this is not right
```
bobmcwhirter, line 214: I see your comment.

One strategy is to do all the `try_get(...)` to get your collection of `Option`s, and later do something like

```
if let (Some(a), Some(b), Some(c)) = (a, b, c) {
    // you know you have them all, and can access them as `a`, `b` and `c` unwrapped
}
```

modules/fundamental/src/analysis/service/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+            );
+
+            g.add_edge(p1, p2, relationship);
+        }
+
+        let mut components: Vec<AdvisoryGraphSummary> = Vec::new();
+
+        for node_index in regex_purl_find(&nodes, query.q.as_str()) {
+            if let Some(find_match_package_node) = g.node_weight(node_index) {
+                let mut ancestor_nodes: Vec<NodeIndex> = ancestor_nodes(&g, node_index, node_index)
```
bobmcwhirter, line 272: I suspect you can elide the `Vec<NodeIndex>` type decl.

modules/fundamental/src/analysis/service/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+                    .into_iter()
+                    .collect();
+
+                ancestor_nodes.sort();
+
+                let mut root_components: Vec<PackageNode> = ancestor_nodes
```
bobmcwhirter, line 278: Likewise elision

modules/fundamental/src/analysis/service/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+                    .iter()
+                    .filter_map(|&ancestor_node| g.node_weight(ancestor_node).cloned())
```
bobmcwhirter, line 280: dunno if worthwhile or useful, but you can elide the `.cloned()` within the closure, and add a `.cloned()` after `.filter_map(...)`

modules/fundamental/src/analysis/service/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+            );
+
+            g.add_edge(p1, p2, relationship);
+        }
+
+        let mut components: Vec<AdvisoryGraphSummary> = Vec::new();
+
+        for node_index in exact_name_find(&nodes, component_name.as_str()) {
+            if let Some(find_match_package_node) = g.node_weight(node_index) {
+                let mut ancestor_nodes: Vec<NodeIndex> = ancestor_nodes(&g, node_index, node_index)
```
bobmcwhirter, line 413: elide type decl.

bobmcwhirter: Does clippy mention elision?

bobmcwhirter (commented):
This is looking good. Mostly just trying to be educational with my comments.

I'd like to see the sea_orm adjustments before we merge in, yeah?

migration/src/m0000585_analysis_api_index.rs
JimFuller-RedHat: yes

entity/src/relationship.rs
```diff
@@ -45,4 +46,12 @@ pub enum Relationship {
     DevToolOf,
     #[sea_orm(num_value = 13)]
     DescribedBy,
+    #[sea_orm(num_value = 14)]
+    PackageOf,
+}
+
+impl fmt::Display for Relationship {
```
ctron, line 53: I guess this works :) There's also a crate called `strum`, which gives you a set of `derive`s that can do the same for you.

modules/fundamental/src/analysis/endpoints/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+        (status = 200, description = "Retrieve component(s) root components by name or pURL.", body = PackageNode),
+    ),
+)]
+#[get("/v1/analysis/root-component/{key}")]
+pub async fn get_component_root_components(
+    service: web::Data<AnalysisService>,
+    key: web::Path<String>,
+    web::Query(paginated): web::Query<Paginated>,
+) -> actix_web::Result<impl Responder> {
+    if key.starts_with("pkg") {
```
ctron, line 61: A better check might be `pkg://`, as `pkg-config` should not match I guess.

modules/fundamental/src/analysis/service/mod.rs
```diff
@@ -0,0 +1,564 @@
...
+                product_name: product_name.to_string(),
+                product_version: product_version.to_string(),
+            });
+            nodes.insert(component_purl.to_string(), i);
+            i
+        }
+    }
+}
+
+pub fn regex_purl_find(nodes: &HashMap<String, NodeIndex>, search_value: &str) -> Vec<NodeIndex> {
```
ctron, line 83: Regexps can be quite costly sometimes. While the `Pattern` trait seems still to be nightly only, it might be worth having an enum like:

 ```rust
enum Search<'a> {
  Regexp(&'a str),
  Substring(&'a str),
}

impl <'a> Search<'a> {
  pub fn is_match(&self, other: &str) -> bool {}
}
```

File: modules/fundamental/src/analysis/service/mod.rs (hunk 9 above)
Comment (ctron) [reply modules/fundamental/src/analysis/service/mod.rs:141]: Yea, there should also be some checks/tests for loops in SBOMs. Which might be illegal, but also shouldn't crash the system.

File: modules/fundamental/src/analysis/service/mod.rs [hunk 18]
```diff
@@ -0,0 +1,564 @@
...
+        let mut nodes: HashMap<String, NodeIndex> = HashMap::new();
+
+        // load package relationships into graph
+        for row in relationship_results {
+            let sbom_published = row
+                .try_get("", "published")
+                .unwrap_or("NOVALUE".to_string()); // TODO: this is not right
+            let document_id = row
+                .try_get("", "document_id")
+                .unwrap_or("NOVALUE".to_string()); // TODO: this is not right
```
ctron, line 223: Independent of it's use here, it might be better to use: `.unwrap_or_else(||"NOVALUE".to_string())`.

The difference is that is will only allocate the string when it's really required. Where with `unwrap_or`, it would allocate that string passing it as an argument.

ctron (commented):
Just a few comments for now. In general it looks quite good and you already use a lot of more complex Rust constructs. Yay!

bobmcwhirter:
@JimFuller-RedHat if'n you don't mind, whenever you're Really Ready for a review, re-request and/or tag us here.

JimFuller-RedHat:
we use petgraph::Directed which should error if we encounter any circular dependencies.

modules/analysis/src/service.rs
```diff
@@ -0,0 +1,1024 @@
...
+        let connection = self.db.connection(&tx);
+        // retrieve all sboms in trustify
+        let distinct_sbom_ids: Vec<String> = sbom::Entity::find()
+            .select()
+            .order_by(sbom::Column::DocumentId, Order::Asc)
+            .order_by(sbom::Column::Published, Order::Desc)
+            .all(&connection)
+            .await?
+            .into_iter()
+            .map(|record| record.sbom_id.to_string()) // Assuming sbom_id is of type String
```
JimFuller-RedHat, line 290: comment: was unclear if we should use sbom_id::uuid or something like document_id::string

trustd/src/db.rs
```diff
@@ -74,6 +74,10 @@ impl Run {
                 "pg_stat_statements".to_string(),
             ),
             ("random_page_cost".to_string(), "1.1".to_string()),
+            (
+                " max_parallel_workers_per_gather".to_string(),
```
JimFuller-RedHat, line 78: comment: we can probably revert this but decided to keep it in for now.

modules/analysis/Cargo.toml
```diff
@@ -0,0 +1,65 @@
...
+
+actix-http = { workspace = true }
+actix-web = { workspace = true }
+anyhow = { workspace = true }
+async-graphql = { workspace = true, features = ["uuid", "time"] }
+cpe = { workspace = true }
+futures-util = { workspace = true }
+itertools = { workspace = true }
+log = { workspace = true }
+roxmltree = { workspace = true }
```
bobmcwhirter, line 22: I think @helio-frota has some cargo invocation to find "excess" dependencies?

Do you actually do XML parsing?

modules/analysis/src/error.rs
```diff
@@ -0,0 +1,67 @@
+use actix_http::StatusCode;
+use actix_web::body::BoxBody;
+use actix_web::{HttpResponse, ResponseError};
+use sea_orm::DbErr;
+use trustify_common::error::ErrorInformation;
+use trustify_common::id::IdError;
+use trustify_common::purl::PurlErr;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
bobmcwhirter, line 10: If this is a copy of the same errors from Fundamental, perhaps we can share?

If sharing seems ill-advised, do we need all of these variants? (IdKey for instance?)

bobmcwhirter: related: I think it'd be fine to have a dependency upon `fundamental`, particularly if we adjust the output of analysis to use fundamental things like `PurlHead` or whatnot, to throw in UUIDs.

modules/analysis/src/model.rs
```diff
@@ -0,0 +1,193 @@
...
+    pub purl: String,
+    pub name: String,
+    pub published: String,
+    pub document_id: String,
+    pub product_name: String,
+    pub product_version: String,
+    pub deps: Option<Vec<DepNode>>,
+}
+impl crate::model::AdvisoryDepSummary {
+    pub fn new(
```
bobmcwhirter, line 101: When a ctor has the same parameters as the struct fields, which are `pub`, I think there's more clarity in just constructing the struct directly, instead of via a function. Without inlay hints, it's hard to know which field is which position, using a positional arg fn.

modules/analysis/src/model.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub struct GraphMap {
+    map: HashMap<String, Graph<PackageNode, Relationship, petgraph::Directed>>,
+}
+
+// Static variables for the singleton instance
+static mut INSTANCE: Option<Arc<RwLock<GraphMap>>> = None;
```
bobmcwhirter, line 128: If we trade it around to Arc<RwLock<Option<GraphMap>> is that easier to avoid the static mut and it's related unsafe?

modules/analysis/src/model.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    // Retrieve a reference to a graph by its key (read access)
+    pub fn get(&self, key: &str) -> Option<&Graph<PackageNode, Relationship, petgraph::Directed>> {
+        self.map.get(key)
+    }
+
+    // Get the singleton instance of GraphMap
+    pub fn get_instance() -> Arc<RwLock<GraphMap>> {
+        INIT.call_once(|| unsafe {
```
bobmcwhirter, line 170: The unsafe gives me the heeby-jeebies.

modules/analysis/src/openapi.rs
```diff
@@ -0,0 +1,22 @@
...
+#[derive(OpenApi)]
+#[openapi(paths(), components(), tags())]
+pub struct ApiDoc;
+
+pub fn openapi() -> utoipa::openapi::OpenApi {
+    let mut doc = ApiDoc::openapi();
+
+    doc.merge(endpoints::ApiDoc::openapi());
+
+    if let Some(components) = doc.components.as_mut() {
```
bobmcwhirter, line 13: Is this required at this level, particularly as UUIDs are pretty spread out, maybe we do this higher-up where we merge all of the openapi's together?

modules/analysis/src/service.rs
```diff
@@ -0,0 +1,1024 @@
...
+pub struct AnalysisService {
+    db: Database,
+}
+
+pub fn dep_nodes(
+    graph: &petgraph::Graph<PackageNode, Relationship, petgraph::Directed>,
+    node: NodeIndex,
+) -> Vec<DepNode> {
+    // fast enough to directly instantiate DepNode in this function
+    let mut depnodes: Vec<DepNode> = Vec::new();
```
bobmcwhirter, line 36: type on the lvalue shouldn't be necessary here, yeah? Inferred since it's used as the `Vec<DepNode>` return.

modules/analysis/src/service.rs
```diff
@@ -0,0 +1,1024 @@
...
+    depnodes
+}
+
+pub fn ancestor_nodes(
+    graph: &petgraph::Graph<PackageNode, Relationship, petgraph::Directed>,
+    node: NodeIndex,
+    original_node: NodeIndex,
+) -> Vec<NodeIndex> {
+    // working directly with NodeIndex is a bit faster (we instantiate PackageNode later)
+    let mut ancestor_nodes: Vec<NodeIndex> = Vec::new();
```
bobmcwhirter, line 68: Elide

modules/analysis/src/service.rs
```diff
@@ -0,0 +1,1024 @@
...
+    }
+    dfs(graph, node, original_node, &mut ancestor_nodes);
+    ancestor_nodes
+}
+
+pub async fn get_relationships(
+    connection: &ConnectionOrTransaction<'_>,
+    distinct_sbom_id: &str,
+) -> Result<Vec<QueryResult>, DbErr> {
+    // TODO: We may convert this to sea_orm at some point though keeping this in its current form, to highlight
```
bobmcwhirter, line 96: Fair. The upside to converting to SeaORM at some point would be for refactoring and type-safety or whatnot, if we change the schema/entity models.

Non-blocking.

modules/analysis/src/service.rs
```diff
@@ -0,0 +1,1024 @@
...
+        let mut graph_write_gaurd = graph_map.write();
+
+        for distinct_sbom_id in distinct_sbom_ids {
+            if !graph_write_gaurd.contains_key(&distinct_sbom_id.to_string()) {
+                // lazy load graphs
+                let mut g: Graph<PackageNode, Relationship, petgraph::Directed> = Graph::new();
+
+                match get_relationships(connection, &distinct_sbom_id.to_string()).await {
+                    Ok(results) => {
+                        let mut nodes: HashMap<String, NodeIndex> = HashMap::new();
```
bobmcwhirter, line 160: Can we elide the lvalue type annotation?

modules/analysis/src/service.rs
```diff
@@ -0,0 +1,1024 @@
...
+                                    };
+                                    let new_node = PackageNode {
+                                        purl: left_purl_string.clone(),
+                                        name: left_purl.name.clone(),
+                                        published: sbom_published.clone(),
+                                        document_id: document_id.clone(),
+                                        product_name: product_name.clone(),
+                                        product_version: product_version.clone(),
+                                    };
+                                    let i = g.add_node(new_node);
```
bobmcwhirter, line 215: nit: characters for naming variables are free!

modules/analysis/src/service.rs
```diff
@@ -0,0 +1,1024 @@
...
+    ) -> Result<(), Error> {
+        let connection = self.db.connection(&tx);
+        load_graphs(&connection, &distinct_sbom_ids).await;
+
+        Ok(())
+    }
+    pub async fn load_all_graphs<TX: AsRef<Transactional>>(&self, tx: TX) -> Result<(), Error> {
+        let connection = self.db.connection(&tx);
+        // retrieve all sboms in trustify
+        let distinct_sbom_ids: Vec<String> = sbom::Entity::find()
```
bobmcwhirter, line 283: type elision.

modules/analysis/src/service.rs
```diff
@@ -0,0 +1,1024 @@
...
+    pub async fn retrieve_root_components<TX: AsRef<Transactional>>(
+        &self,
+        query: Query,
+        _paginated: Paginated,
+        tx: TX,
+    ) -> Result<PaginatedResults<AdvisoryGraphSummary>, Error> {
+        let connection = self.db.connection(&tx);
+
+        let search_sbom_node_name_subquery = sbom_node::Entity::find()
+            .filter(
```
bobmcwhirter, line 332: I think we can drop the `Condition::any().add(...)` wrapper.

```
.filter(Name.like(format!(...))
```

modules/analysis/src/service.rs
```diff
@@ -0,0 +1,1024 @@
...
+            .order_by(sbom::Column::Published, Order::Desc)
+            .all(&connection)
+            .await?
+            .into_iter()
+            .map(|record| record.sbom_id.to_string()) // Assuming sbom_id is of type String
+            .collect();
+
+        load_graphs(&connection, &distinct_sbom_ids).await;
+
+        let mut components: Vec<AdvisoryGraphSummary> = Vec::new();
```
bobmcwhirter, line 507: type elision

modules/analysis/src/service.rs
```diff
@@ -0,0 +1,1024 @@
...
+                                        .sorted_by_key(|node| node.published.to_string())
+                                        .collect(),
+                                ));
+                            }
+                        });
+                }
+            }
+        }
+
+        // TODO: limiting on Vec
```
bobmcwhirter, line 548: We should create a helper that can apply Paginated to Vec<T> appropriately.

modules/analysis/Cargo.toml
JimFuller-RedHat: not me ... maybe I picked it up at some point inadvertently ?

modules/analysis/src/model.rs
JimFuller-RedHat: oh that would prob be better ... I did it this way first and it worked and moved on ;) thx for the prod.

modules/analysis/src/model.rs
JimFuller-RedHat: me too - let me see if I can refactor that out

modules/analysis/Cargo.toml
helio-frota: I know a bit cargo bits :dark_sunglasses:

helio-frota: `RUSTFLAGS=-Wunused-crate-dependencies cargo build &> unused.txt`

helio-frota: some of them are not true (yeah I saw the un-true before)
https://gist.github.com/helio-frota/49ba11927a1f211a0f48b99f3e21bc24

modules/analysis/src/service.rs
bobmcwhirter: done in https://github.com/trustification/trustify/pull/769

modules/analysis/Cargo.toml
JimFuller-RedHat: thx @helio-frota very useful that!

modules/analysis/src/error.rs
JimFuller-RedHat: opted to amend and keep for the time being - it is unclear if modules/analysis could be used standalone someday ... hoping that is ok.

modules/analysis/src/model.rs
JimFuller-RedHat: I opted to use lazy_static

modules/analysis/src/model.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub product_version: String,
+    pub deps: Vec<DepNode>,
+}
+#[derive(Debug)]
+pub struct GraphMap {
+    map: HashMap<String, Graph<PackageNode, Relationship, petgraph::Directed>>,
+}
+
+// Static variables for the singleton instance
+lazy_static! {
```
ctron, line 86: IIRC we had a discussion about this. And I also wouldn't see this as a blocker, but I think the following pattern is the replacement for using `lazy_static`:

```rust
static G: OnceLock<Arc<Mutex<usize>>> = OnceLock::new();

fn g() -> &'static Arc<Mutex<usize>> {
    G.get_or_init(|| Arc::new(Mutex::new(0)))
}
```

ctron:
LGTM. A proposal to switch away from the `lazy_static` crate. But works either way. I'll leave it to @bobmcwhirter (or someone else) for a final +1. Consider this +1 from me. Great work!

bobmcwhirter (approved):

PR #728, opened 2024-08-28
Add CWE Catalog importer, daily interval.

bobmcwhirter:
Clean up some typos.
Regenerate OpenAPI.

### DIFF
--- a/modules/importer/src/runner/cwe/mod.rs
+++ b/modules/importer/src/runner/cwe/mod.rs
@@ -1,0 +1,59 @@
+mod walker;
+
+use crate::model::CweImporter;
+use crate::runner::{
+    context::RunContext,
+    cwe::walker::CweWalker,
+    report::{ReportBuilder, ScannerError},
+    RunOutput,
+};
+use std::sync::Arc;
+use tokio::sync::Mutex;
+use tracing::instrument;
+use trustify_module_ingestor::{graph::Graph, service::IngestorService};
+
+impl super::ImportRunner {
+    #[instrument(skip(self), ret)]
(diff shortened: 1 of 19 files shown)


bobmcwhirter:
since this importer processes exactly 1 fetched-over-http file (using Last-Modified header to determine continuationing), a lot of the nested Walker/Handler/Context stuff simplified out.

modules/importer/src/runner/cwe/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+        // no working-dir required
+
+        // one file, no progress to care about.
+
+        // run the walker
+
+        let walker = CweWalker::new(cwe_catalog.source.clone(), ingestor, report.clone())
+            .continuation(continuation);
+
+        match walker.run().await {
```
ctron, line 37: note to self: but I've seen the same pattern a bunch of times now. Might be worth pulling that out into a common place.

modules/importer/src/runner/cwe/walker.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /// Run the walker
+    #[instrument(skip(self), ret)]
+    pub async fn run(self) -> Result<LastModified, Error> {
+        let response = reqwest::get(&self.source).await?;
+
+        let last_modified = response
+            .headers()
+            .get("Last-Modified")
```
ctron, line 49: IIRC headers are case insensitive, and there are headers definitions from the `http` create. Does it make sense to use that?

ctron (approved):
Looks good! Just two comments, non-blocking.

Issue #725, opened 2024-08-28
embedded postgres issue in containers


ctron:
Starting a  most recent container in PM mode results in:

```
RUST_LOG is unset, using default: 'info,actix_web_prom=error'
2024-08-28T11:20:40.127981Z  WARN trustd::db: Setting up managed DB; not suitable for production use!    
Error: Command error: stdout=; stderr=/.trustify/postgres/16.3.0/bin/initdb: error while loading shared libraries: libgssapi_krb5.so.2: cannot open shared object file: No such file or directory
```

bobmcwhirter:
Container missing some Kerberos stuff? Why would we need Kerberos?

ctron:
I don't think this comes from our code, but the embedded psql instance.

helio-frota:
it works for me (both podman-compose and docker-compose) using this :

```yaml
version: "3"

services:
  db:
    image: postgres:16
    ports:
      - "5432:5432"
    environment:
      POSTGRES_DB: mydb
      POSTGRES_USER: myuser
      POSTGRES_PASSWORD: mypassword
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U myuser -d mydb"]
      interval: 10s
      timeout: 5s
      retries: 5

  trustify:
    image: ghcr.io/trustification/trustd:latest
    command:
      - 'api'
      - '--devmode'
    ports:
      - "8080:8080"
    environment:
      HTTP_SERVER_BIND_ADDR: 0.0.0.0
      TRUSTD_DB_USER: myuser
      TRUSTD_DB_PASSWORD: mypassword
      TRUSTD_DB_HOST: db
      TRUSTD_DB_PORT: 5432
      TRUSTD_DB_NAME: mydb
      AUTH_DISABLED: "true"
      RUST_LOG: trace
    depends_on:
      db:
        condition: service_healthy
```

helio-frota:
oops, 

not working...


```
[trustify] | 2024-08-28T20:09:27.863104Z TRACE setup:initialize:execute_command: mio::poll: deregistering event source from poller     program="initdb"
[trustify] | 2024-08-28T20:09:27.863123Z TRACE setup:initialize:execute_command: mio::poll: deregistering event source from poller     program="initdb"
[trustify] | 2024-08-28T20:09:27.863137Z TRACE setup:initialize:execute_command: mio::poll: deregistering event source from poller     program="initdb"
[...]

PR #729, opened 2024-08-29
build: add all required dependencies for embedded psql in the container

ctron:
See https://github.com/trustification/trustify/issues/725

This fixes the issue of the missing dependencies. But still psql doesn't seem to run because it thinks it's been run as root.


JimFuller-RedHat (approved):
LGTM

PR #738, opened 2024-08-30
chore: basic script based on ctron's step by step

helio-frota:
* https://github.com/trustification/trustify/issues/725#issuecomment-2320209288

### DIFF
--- a/.github/scripts/dev-build.sh
+++ b/.github/scripts/dev-build.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+# run from the root dir: ".github/scripts/dev-build.sh 0.1.0-alpha.14"
+# 13 as default in case we forget to pass the argument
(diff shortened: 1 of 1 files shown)


.github/scripts/dev-build.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# run from the root dir .github/scripts/dev-build.sh 13
+# 13 as default in case we forget to pass the argument
+VERSION=${1:-13}
```
ctron, line 5: I think it would be worth having the full version in here: `0.1.0-alpha.13`. So we don't to modify this script when we move on.

ctron (changes requested):
Looks good, but I think we should be more flexible on the version.

ctron (approved):

PR #741, opened 2024-08-30
feat: add depth to GitWalker for faster clones by limiting history

jcrossley3:
This started from a desire to add a `depth` field to all our git-based walkers: `CVE`, `OSV`, and `ClearlyDefined`. 

That led me to realize that those walkers were mostly just a collection of methods that delegated to their internal `GitWalker` field. And that the essential complexity distinguishing each walker was really in the handlers. 

So I thought things would be simpler by deleting all the walkers and using `GitWalker` directly.

This made it *much* simpler to add a `depth` field, getting rid of my hacky `if cfg!(test)` conditional.


ctron (commented):
I like that PR in general and I think it's a good improvement!

I am not a fan of dropping the wrappers. But I see the benefits. My concern is that this exposes features of the git walker, that aren't supported by the underlying data model/filesystem layout.

What I'd like to see changed, unless there are good reasons, is the default of `depth = 1`. A normal "git clone" would also not limit the depth, and so I think when it comes to expectations, `GitWalker` should behave the same.

jcrossley3:
> What I'd like to see changed, unless there are good reasons, is the default of `depth = 1`. A normal "git clone" would also not limit the depth, and so I think when it comes to expectations, `GitWalker` should behave the same.

Interesting. From where I sit, the only reason the `GitWalker` exists is to import a pile of docs stored in a git repo. That it clones anything at all is an implementation detail. There's no mention of cloning in its interface, and because an unlimited depth can take a ridiculously long time, I disagree that it's a reasonable default value.

I don't feel particularly strong about it, but since it didn't even occur to me to compare its behavior to a "normal git clone", I'm a little surprised you do.

ctron:
Well, that's just my expectation. Different people, different expectations. So maybe there should be a nice rustdoc section explaining it.

ctron (approved):

Issue #747, opened 2024-09-03
Implement PATCH for updating importer configurations


ctron:
Right now, a configuration can be replaced using `PUT`. However, it would be convenient to also use `PATCH` with some merge strategy, like JSON patch.

PR #749, opened 2024-09-03
feat: implement patch for updating importer configurations

ctron:
Closes: https://github.com/trustification/trustify/issues/747

Now I do know that the dependency `json-merge-patch = "0.0.1"` might trigger some discussion. However, it looks simple enough, well tested, and not too old. It just works, what else is required.

Also, the amount of code is pretty small. Does it make sense to re-write that from scratch? I think that would add more bugs. Can't we just copy over the code? Sure, the question is, why not try to properly consume that dependency first. It's so small, forking it will still be an option in the future.

It's released on crates.io, it seems to work, it has tests. I think we should give it a try. Unless anyone known a better implementation.


jcrossley3 (commented):
Why do we need this?

ctron:
> Why do we need this?

Because it makes life easier updating just a single field in the configuration.

bobmcwhirter (approved):

bobmcwhirter:
we do like PATCH

PR #753, opened 2024-09-04
feat: allow finding advisories with a specific ID, as claimed by the doc

ctron:
This also extracts a few of the query builder and processing code that seemed duplicate into dedicated, reusable functions.

### DIFF
--- a/modules/fundamental/src/advisory/endpoints/test.rs
+++ b/modules/fundamental/src/advisory/endpoints/test.rs
@@ -106,6 +106,89 @@ async fn all_advisories(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
 
 #[test_context(TrustifyContext)]
 #[test(actix_web::test)]
+async fn find_advisories_by_document_id(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
+    let app = caller(ctx).await?;
+
+    let advisory = ctx
+        .graph
+        .ingest_advisory(
+            "RHSA-1",
+            ("source", "https://redhat.com/"),
+            &Digests::digest("RHSA-1"),
+            AdvisoryInformation {
+                title: Some("RHSA-1".to_string()),
+                issuer: None,
+                published: Some(OffsetDateTime::now_utc()),
+                modified: None,
+                withdrawn: None,
+            },
+            (),
+        )
+        .await?;
+
+    let advisory_vuln = advisory
+        .link_to_vulnerability("CVE-123", None, Transactional::None)
+        .await?;
+    advisory_vuln
+        .ingest_cvss3_score(
+            Cvss3Base {
+                minor_version: 0,
+                av: AttackVector::Network,
+                ac: AttackComplexity::Low,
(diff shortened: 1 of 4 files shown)


modules/fundamental/src/advisory/endpoints/test.rs
```diff
@@ -104,6 +104,89 @@ async fn all_advisories(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
...
+                issuer: None,
+                published: Some(OffsetDateTime::now_utc()),
+                modified: None,
+                withdrawn: None,
+            },
+            (),
+        )
+        .await?;
+
+    let uri = "/api/v1/advisory/RHSA-1/by-document-id";
```
bobmcwhirter, line 165: since this returns a possibly-multiple response, it's really a subset of `/api/v1/advisory` content, filtered, maybe.

perhaps instead of `by-document-id` suffix, just `/api/v1/advisory?document_id=...` to make it clear it's still the collection of all advisories, winnowed down a bit?

jcrossley3: Or, if you can stomach the encoding, `/api/v1/advisory?q=identifier%3dRHSA-1`

ctron: > Why not use what we already have? `/api/v1/advisory?q=RHSA-1`

Because that would give you a different result. Searching all fields for that content, vs searching for specifically a single field.

ctron: > since this returns a possibly-multiple response, it's really a subset of `/api/v1/advisory` content, filtered, maybe.
> 
> perhaps instead of `by-document-id` suffix, just `/api/v1/advisory?document_id=...` to make it clear it's still the collection of all advisories, winnowed down a bit?

That would make that endpoint more complex. But if that's what we want, I can refactor it.

bobmcwhirter: That'd be my preference, since it really just feels like slicing a collection differently, just as how `q=` works.

jcrossley3: > > Or, if you can stomach the encoding, `/api/v1/advisory?q=identifier%3dRHSA-1`
> 
> So that would result in `identifier="RHSA-1"`? Or is that a `like`?

Yes. For a LIKE, you'd need a `~`, which I don't think needs to be encoded: `/api/v1/advisory?q=identifier~RHSA-1`

ctron: I wasn't aware that `q=identifier%3dRHSA-1` was a thing. However, the question is still is that an equals or sub-string match?

ctron: Sorry, that overlapped. I guess I shouldn't use GitHub as a chat :)

ctron:
It seems we can have the same with `q=identifier%3dRHSA-1`, which works for me.

PR #758, opened 2024-09-04
Add some one-liner API documentation for the UI.

bobmcwhirter:


bobmcwhirter:
Also add some more refined tags.

modules/fundamental/src/purl/endpoints/version.rs
```diff
@@ -18,6 +18,7 @@ use trustify_common::purl::Purl;
     ),
 )]
 #[get("/v1/purl/version/{key}")]
+/// Retrieve details of a versioned, non-qualified pURL
```
jcrossley3, line 21: I assume these comments show up on the openapi UI?

jcrossley3 (approved):

modules/fundamental/src/purl/endpoints/version.rs
bobmcwhirter: Yep!

PR #769, opened 2024-09-06
Add array-pagination fn to Paginated.

bobmcwhirter:


jcrossley3:
What problem is this solving?

bobmcwhirter:
Someone who has a vec of values not from a query but still want to support pagination.

JimFuller-RedHat (approved):
LGTM!

PR #776, opened 2024-09-09
chore: removes unused deps

helio-frota:
Based on `RUSTFLAGS=-Wunused-crate-dependencies cargo build`

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -32,8 +32,6 @@ sea-query = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 sqlx = { workspace = true }
-tempfile = { workspace = true }
-test-context = { workspace = true }
 thiserror = { workspace = true }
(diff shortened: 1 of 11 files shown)


common/Cargo.toml
```diff
@@ -27,13 +27,16 @@ regex = { workspace = true }
 reqwest = { workspace = true, features = ["native-tls"] }
 ring = { workspace = true }
 schemars = { workspace = true }
-sea-orm = { workspace = true, features = ["sea-query-binder", "sqlx-postgres", "runtime-tokio-rustls", "macros"] }
+sea-orm = { workspace = true, features = [
```
ctron, line 30: I'd like to keep the one-liners, as this makes sorting the deps alphabetically easier.

ctron (changes requested):
Thanks for cleaning up! :) … I'd just prefer to keep the one-liners.

ctron (approved):

Issue #777, opened 2024-09-09
Should we add cmp to Purl


JimFuller-RedHat:
```
pkg://oci/osc-operator-bundle@sha256:8f7815a6b7ac59c2ffe07cdfae5a1a2d6e97eb4f226216f68a3f9c8d0b101ce7?repository_url=registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle&tag=1.6.0-62

```
is equal to this
```
pkg://oci/osc-operator-bundle@sha256:8f7815a6b7ac59c2ffe07cdfae5a1a2d6e97eb4f226216f68a3f9c8d0b101ce7?tag=1.6.0-62&repository_url=registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle

```

param order is significant if one is only doing a string type comparison ;)

either we should be real careful to enforce canonical purl order (being mindful json can sometimes think up its own order) or have comparison operator that does the right thing when comparing Purl == Purl

JimFuller-RedHat:
adding from @jcrossley3 which might be another expression of the problem of comparison of pURLs 

CVE-2024-26308

said none of these purl's found an SBOM (but he couldn't tell me which SBOM should have:


```
'pkg://oci/logic-data-index-ephemeral-rhel8@sha256:1d4c65ddd65b54b387f21bdabf408d180bcc0d835fec714a2c06b643187279de',
 'pkg://oci/logic-data-index-ephemeral-rhel8@sha256:2223754df4f475fc7240df4d833c3ad3d757375ceb2dba359164bd6e8475d267',
 'pkg://oci/logic-data-index-ephemeral-rhel8@sha256:90938287390c5d53dd8311699daa4304444e0727fa1aed18e6b4712ef2da8ee4',
 'pkg://oci/logic-data-index-postgresql-rhel8@sha256:1050e0b388b09c494bcb2f9bc9d74eb1f12b1ef93218e3920434f7c09b22f9eb',
 'pkg://oci/logic-data-index-postgresql-rhel8@sha256:d6951064cd3ac48107a93d9d21df106157df0232645bb2d847589fda496b5c9a',
 'pkg://oci/logic-data-index-postgresql-rhel8@sha256:f188dc873609058aa3a4911526df0afc1f32c8b986c02646b403932750db5d19',
[...]

PR #780, opened 2024-09-10
Added test confirming purl comparison works #777

JimFuller-RedHat:


jcrossley3:
> test passes ;) so I think we can say 'it does'

I was assuming #777 was required because this test would NOT pass, so I expected the fix (cmp fn) to be included with the test that affirms it.

JimFuller-RedHat:
I was a little paranoid there might be more expressions of this problem ... which is why I raised the issue initially

JimFuller-RedHat:
ok sounds like ordering is not a problem, will raise a different issue if we find anything

PR #783, opened 2024-09-11
feat: allow importing complete datasets

ctron:
This adds an endpoint for ingesting an archive of documents. The idea is to have a a simpler version of ingesting "datasets", like the "dataset 1", that we had from Trustification.

This also adds a "dataset 3", which consists of "dataset 1", but adding the matching CVE files, so that all information in the UI is present.

The original idea was to ingest this using the existing "detect format" code path. However, it doesn't fix that pattern:

* Because there's actually no document that will be stored. So the result isn't an `IngestResult`, and doesn't have IDs or digests
* Because it actually returns multiple `IngestResult`, one for each file in the archive. That information should be reported back.

That's why this PR adds a new endpoint, for this specific use case. In the future, if we have more endpoints with a similar pattern, which make this more generic.


server/src/lib.rs
```diff
@@ -302,7 +302,7 @@ impl InitData {
 fn configure(
     svc: &mut web::ServiceConfig,
     db: db::Database,
-    storage: impl Into<DispatchBackend>,
+    storage: impl Into<DispatchBackend> + Clone,
```
JimFuller-RedHat, line 305: oh interesting, I did not know one could do this

JimFuller-RedHat (approved):
I did not find any tests, though I suspect this feature is mostly adhoc and helps data movement for testing purposes ... LGTM

modules/ingestor/src/service/mod.rs
```diff
@@ -166,15 +174,8 @@ impl IngestorService {
             .await
             .map_err(|err| Error::Storage(anyhow!("{err}")))?;
 
-        let stream = self
-            .storage
-            .retrieve(result.key())
-            .await
-            .map_err(Error::Storage)?
-            .ok_or_else(|| Error::Storage(anyhow!("file went missing during upload")))?;
-
```
jcrossley3, line 175: We were intentionally retrieving the doc after storing it to ensure that we could. If we don't find out now, we'll never know. Are we sure we want to eliminate this failsafe?

jcrossley3 (commented):
I wish you had broken out your ingestion changes in a separate commit. Or even a separate PR. Those are pretty significant changes to bury in a "datasets" feature PR, I think.

modules/ingestor/src/service/mod.rs
ctron: The reason for retrieving it was actually because the stream was consumed at that point. From my side, it was never intended as a check.

Since none of the internal APIs makes actual use of that stream, it didn't seem reasonable to keep that pattern. We anyway load the full document in memory. So why store megabytes of SBOMs, and then load them again? If the don't trust S3 when it says "ok, I stored it" I think we have a different problem.

PR #784, opened 2024-09-11
refactor: implement a cache for lookups

ctron:


ctron:
Doesn't seem to be worth the effort.

PR #787, opened 2024-09-11
Restore the shallow clone behavior of the git walkers

### LINKED ISSUE #775: Don't forget that the initial CVE walker's git clone takes ~30 minutes
The [CVE repo](https://github.com/CVEProject/cvelistV5) is over 3GB. Even a shallow clone is 2.7GB. 

The git cli takes almost 3 minutes to clone the full repo. The libgit2 rust crate takes almost 30 minutes. 
The git cli takes about 20s to create a shallow clone. The libgit2 rust crate takes about a minute.

Updates to the full clone are reasonably fast and it's efficient to determine which files need to be [re]ingested.

Updates to the shallow clone have [yet to be made to work](https://github.com/trustification/trustify/issues/764) successfully. 

Regardless, the "day 0" experience for new users is sucky. There is no visual indication whatsoever (in the UI or the server log) that the app is doing anything... for almost 30 minutes.

Consider this issue a placeholder for improving the "day 0" experience for enabling importers monitoring large git repos.

Possible alternatives:

- Figure out how to update a shallow clone with our existing `libgit2` crate.
- Shell out to the `git` CLI, which doesn't use `libgit2`, possibly enabling the deepening of an initial shallow clone in order to properly determine the changed files.
- Create a new shallow clone for every import cycle and calculate the digest of every file (~3GB worth of them) and query the db to determine whether they need to be ingested.

jcrossley3:
Resolves #775

I finally figured out how to update the shallow clone, setting the fetch depth to `i32::MAX` to signal infinite fetch depth, per https://git-scm.com/docs/shallow

I had wrongly assumed depth=0 signaled infinite depth -- it does for an initial clone, but not an update of a shallow one.

Essentially, this improves the "day 0" experience -- the first walk of the shallow clone takes no more than a few minutes. But the annoying ~30 minute fetch is deferred to the first update of the shallow clone, effectively turning it into a deep clone, after which subsequent fetches are fast.

I think that's a workable compromise until libgit2's perf issue is addressed.

### DIFF
--- a/modules/importer/src/runner/common/walker/git.rs
+++ b/modules/importer/src/runner/common/walker/git.rs
@@ -75,6 +75,9 @@ impl<H> GitWalker<H, (), ()>
 where
     H: Handler,
 {
+    /// Create a new GitWalker for a given repo and handler. By
+    /// default, a "shallow clone" (depth=1) of the repo will be
+    /// walked.
     pub fn new(source: impl Into<String>, handler: H) -> Self {
         Self {
             source: source.into(),
@@ -84,7 +87,7 @@ where
             working_dir: (),
             handler,
             progress: (),
-            depth: 0,
+            depth: 1, // shallow clone, by default
         }
     }
 }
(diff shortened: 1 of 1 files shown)


ctron (approved):
I think that's a good compromise.

PR #793, opened 2024-09-12
Doc table

bobmcwhirter:
Add a `source_document` table, with columns sha256/384/512.
Remove digest columns from `sbom` and `advisory`.
Add a _nullable_ `source_document_id` to both `sbom` and `advisory`.
Update queries to join appropriately to the new `source_document` table.

**GOAL**

To refactor common elements regarding documents from different tables into a shared table.
Will enable easier scrobbling for sigstore signatures (as an importer) regardless of document type.
And will enable future authoring/originating of advisories etc from Trustify in which there is _no_ backing doc.


modules/fundamental/src/advisory/model/details/mod.rs
```diff
@@ -40,23 +42,29 @@ impl AdvisoryDetails {
...
+                &advisory.advisory,
+                Memo::Provided(advisory.issuer.clone()),
+                tx,
+            )
+            .await?,
+            source_document: if let Some(doc) = &advisory.source_document {
+                Some(SourceDocument::from_entity(doc, tx).await?)
+            } else {
+                None
+            },
```
jcrossley3, line 64: This is a breaking api change, right? I can't decide whether the `source_document` element is important enough to clients to have it broken out.

jcrossley3 (commented):
As we start to have some consumers of this app, e.g. AI interns, we need to ensure that we signal BREAKING API CHANGES somehow in the release notes. I think we have a convention for doing that, but I'm not sure what it is.

ctron:
I'd like to hear @JimFuller-RedHat's thoughts on the DB design.

I think the PR is ok. I just struggle with the naming "source document id". Simply, because I have the feeling this adds to the confusion around documents and IDs.

For me a "source document ID" would be the ID provided by the source document (namespace, serialNumber, tracking id, …). 

I don't really have a better name though. But I still think we should find one.

modules/fundamental/src/advisory/model/details/mod.rs
bobmcwhirter: Adjusted to flatten into the owning DTO.

jcrossley3 (approved):

Issue #792, opened 2024-09-12
openapi: uuid types don't look like uuids. 


chirino:
The quarkus openapi client generator does not seem to like the data type of the ProductSummary.id field:

```
Caused by: com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `java.util.UUID` from String "urn:uuid:ed54fc78-d4db-48d8-82e8-22c1f38323cc": UUID has to be represented by standard 36-char representation
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 17] (through reference chain: org.github.trustify.model.PaginatedProductSummary["items"]->java.util.ArrayList[0]->org.github.trustify.model.ProductSummary["id"])
```

The spec declare the `id` field as follows:
```
        id:
          type: string
          format: uuid
```

PR #794, opened 2024-09-12
fixes: openapi: uuid types don't look like uuids #792

chirino:
Closes https://github.com/trustification/trustify/issues/792


ctron:
They don't look like UUIDs?

ctron:
Ah, just found the related issue. Makes sense.

ctron (approved):

PR #799, opened 2024-09-13
feat: support compressed upload

ctron:
Also, drop unnecessary error enum and Sync/Send traits.

### DIFF
--- a/modules/fundamental/src/advisory/endpoints/mod.rs
+++ b/modules/fundamental/src/advisory/endpoints/mod.rs
@@ -2,13 +2,18 @@ mod label;
 #[cfg(test)]
 mod test;
 
-use crate::purl::service::PurlService;
-use crate::Error::Internal;
-use crate::{advisory::service::AdvisoryService, Error};
-use actix_web::{delete, get, post, web, HttpResponse, Responder};
+use crate::{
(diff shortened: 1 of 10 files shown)


modules/fundamental/src/advisory/endpoints/mod.rs
```diff
@@ -166,8 +171,10 @@ struct UploadParams {
 pub async fn upload(
     service: web::Data<IngestorService>,
     web::Query(UploadParams { issuer, labels }): web::Query<UploadParams>,
+    content_type: Option<web::Header<header::ContentType>>,
     bytes: web::Bytes,
 ) -> Result<impl Responder, Error> {
+    let bytes = decompress_async(bytes, content_type.map(|ct| ct.0)).await??;
```
bobmcwhirter, line 177: Just to clarify, this means the ingestor never sees the compressed file, yeah? So hashes are for the underlying document after decompression? If so, looks great.

bobmcwhirter (approved):

modules/fundamental/src/advisory/endpoints/mod.rs
ctron: Correct.

PR #813, opened 2024-09-17
chore: reusing sample_advosiry ingestion function

helio-frota:


modules/fundamental/src/advisory/service/test.rs
```diff
@@ -12,27 +12,34 @@ use trustify_cvss::cvss3::{
...
 };
+
 use trustify_test_context::TrustifyContext;
 
-async fn ingest_and_link_advisory(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
-    let advisory = ctx
-        .graph
+async fn ingest_sample_advisory(
+    ctx: &TrustifyContext,
+    title: String,
```
jcrossley3, line 22: I think if you take a `&str` instead, you can remove all the `.to_string()` calls from the callers. You may have to call `.to_string()` within this function, depending on what `AdvisoryInformation` expects, but better one than many.
```suggestion
    title: &str,
```

jcrossley3 (changes requested):
I like the refactoring. Just one minor nit.

modules/fundamental/src/advisory/service/test.rs
helio-frota: thanks for the review... but now I'm entering in a weirdo lifetimes thing that I should learn more about : )

jcrossley3 (approved):

modules/fundamental/src/advisory/service/test.rs
```diff
@@ -12,27 +12,34 @@ use trustify_cvss::cvss3::{
...
 };
+
 use trustify_test_context::TrustifyContext;
 
-async fn ingest_and_link_advisory(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
-    let advisory = ctx
-        .graph
+async fn ingest_sample_advisory<'a>(
+    ctx: &'a TrustifyContext,
+    title: &'a str,
```
jcrossley3, line 22: BTW, not a big deal, but you don't technically need the lifetime on `title`. Because `AdvisoryContext` requires a lifetime, and we're using `ctx` to return one, we're telling the compiler that `ctx` must live at least as long as the value we're returning... if that makes any sense.

PR #815, opened 2024-09-17
fix: prevent uploading compressed files which might exhaust the memory

ctron:


JimFuller-RedHat (approved):
I am not so familiar with the test context stuff, otherwise LGTM

PR #818, opened 2024-09-17
Add tests to see if maven/semver version_matches is right or not.

bobmcwhirter:
Seems to be right.

### DIFF
--- a/migration/src/m0000610_improve_version_cmp_fns/mavenver_cmp.sql
+++ b/migration/src/m0000610_improve_version_cmp_fns/mavenver_cmp.sql
@@ -14,12 +14,16 @@ declare
     left_qualifier text;
     left_build bigint;
 
+    left_cardinality integer;
+
     right_major bigint;
     right_minor bigint;
     right_revision bigint;
     right_qualifier_or_build text;
     right_qualifier text;
     right_build bigint;
+
+    right_cardinality integer;
 
     left_numeric bool;
     right_numeric bool;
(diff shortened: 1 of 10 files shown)


jcrossley3 (commented):
Is `migration/` the right place for these tests to live?

migration/src/m0000610_improve_version_cmp_fns/mavenver_cmp.sql
```diff
@@ -0,0 +1,110 @@
+
+create or replace function mavenver_cmp(left_p text, right_p text)
```
jcrossley3, line 2: I guess we have to do this logic in the db? Too much potentially returned to post-process? All the SQL makes my eyes bleed on to the vomit I just threw up. :eyes: :fork_and_knife: :vomiting_face:

bobmcwhirter: Yep, we want to do it in the DB, so that we can also support suitable pagination.

Else, we have to bring them *all* back, and scrobble around on the rust side winnowing things out of the version ranges.

bobmcwhirter: Add just a touch of Sriracha, and the bloodvomit will be lovely.

bobmcwhirter: I was once in a band named "bloodvomit"

modules/graphql/src/test.rs
```diff
@@ -315,6 +315,10 @@ async fn get_cves_by_sbom(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
     assert_eq!(cves[1]["vulnerabilityId"], "CVE-2024-29025");
     assert_eq!(cves[1]["status"], "not_affected");
 
+    /*
+    TODO: Gilles!
+    These statuses are *not* reflected in the CSAF regarding the quarkus 3.2.11 SBOM, which is the `sbom_id` from results[1] above.
+
```
jcrossley3, line 321: Create an issue for this?

jcrossley3 (approved):
Thanks for the quick turnaround!

modules/graphql/src/test.rs
helio-frota: sounds good

bobmcwhirter:
Thanks for merging! Your report, your privilege.

PR #825, opened 2024-09-18
Fix phantom bug about pURL advisory matches.

bobmcwhirter:


jcrossley3 (approved):
This does seem to fix the issue I was seeing locally. My advisories array is empty in the response for `/api/v1/purl` `key=pkg://maven/org.apache.commons/commons-compress@1.26.0.redhat-00001`

Issue #811, opened 2024-09-16
Add API endpoint to interface with a Trustify tooled AI assistant.


chirino:
The endpoints should provide a chat interface with an AI assistant that uses the Trustify database to respond to user questions.

chirino:
This basically means figure out a good API that a web UI can use to do similar things to what was prototyped in this branch:
https://github.com/chirino/trustify/tree/ai

chirino:
@carlosthe19916 hey do you want to help design the API?

carlosthe19916:
@chirino I've been looking at some applications that use the "chat" way of interacting with AI. And almost all of them (including ChatGPT) use just a single REST endpoint with event streaming for sending the required text from the backend to the client.

Something like POST `/conversation`.

I am not sure how much of a complexity we might have but even for Chat GPT a single endpoint seemed to be enough. For starters, we you mentioned in the internal chat, having a blocking REST endpoints that gives back the whole response text would be enough. But as I mentioned I still need more exposure to this kind of APIs for better understanding.

chirino:
Ok let's start simple and evolve as we go.

jcrossley3:
Assuming fixed by #830

PR #830, opened 2024-09-19
Add AI endpoint under: `/api/v1/ai/completions` to query the DB for information.

chirino:
Add a `cargo xtask ai` cli interface to the AI service 

It uses that langchanin-rust apis interface with an LLM of your choice.

To run it against an OpenAI compatible endpoint, configure and export the following env variables:

* OPENAI_API_KEY
* OPENAI_API_BASE
* OPENAI_MODEL

To run against OpenAI, just set the 'OPENAI_API_KEY' env var.

Use ollama.com to run it all locally:

* install https://ollama.com/
* `ollama pull llama3.1:70b`
* `OPENAI_API_KEY=ollama OPENAI_API_BASE=http://localhost:11434/v1 OPENAI_MODEL=llama3.1:70b cargo xtask ai`

### DIFF
--- a/modules/fundamental/Cargo.toml
+++ b/modules/fundamental/Cargo.toml
@@ -17,9 +17,12 @@ actix-http = { workspace = true }
 actix-web = { workspace = true }
 anyhow = { workspace = true }
 async-graphql = { workspace = true, features = ["uuid", "time"] }
+async-trait = { workspace = true }
+base64 = { workspace = true }
 cpe = { workspace = true }
 futures-util = { workspace = true }
 itertools = { workspace = true }
+langchain-rust = { workspace = true }
 log = { workspace = true }
 sea-orm = { workspace = true }
 sea-query = { workspace = true }
--- a/modules/fundamental/src/ai/endpoints/test.rs
+++ b/modules/fundamental/src/ai/endpoints/test.rs
@@ -1,0 +1,40 @@
+use crate::ai::model::ChatState;
+use crate::ai::service::test::ingest_fixtures;
+use crate::ai::service::AiService;
+use crate::test::caller;
+use actix_http::StatusCode;
+use actix_web::test::TestRequest;
+use test_context::test_context;
+use test_log::test;
+use trustify_test_context::call::CallService;
+use trustify_test_context::TrustifyContext;
+
+#[test_context(TrustifyContext)]
+#[test(actix_web::test)]
+async fn configure(ctx: &TrustifyContext) -> anyhow::Result<()> {
+    let service = AiService::new(ctx.db.clone());
+    if !service.enabled() {
+        return Ok(()); // skip test
+    }
+
+    ingest_fixtures(ctx).await?;
+
+    let app = caller(ctx).await?;
+    let mut req = ChatState::new();
+    req.add_human_message("What is the latest version of Trusted Profile Analyzer?".into());
+
+    let request = TestRequest::post()
+        .uri("/api/v1/ai/completions")
+        .set_json(req)
+        .to_request();
+
+    let response = app.call_service(request).await;
+    log::debug!("Code: {}", response.status());
+    assert_eq!(response.status(), StatusCode::OK);
+
+    let result: ChatState = actix_web::test::read_body_json(response).await;
+    log::info!("result: {:?}", result);
+    assert!(result.messages.last().unwrap().content.contains("37.17.9"));
+
+    Ok(())
+}
--- a/modules/fundamental/src/ai/service/mod.rs
+++ b/modules/fundamental/src/ai/service/mod.rs
@@ -1,0 +1,227 @@
+pub mod tools;
+
+use crate::Error;
+
+use trustify_common::db::{Database, Transactional};
+
+use crate::ai::model::{ChatMessage, ChatState, LLMInfo, MessageType};
+
+use crate::ai::service::tools::{CVEInfo, ProductInfo, ToolLogger};
+use crate::product::service::ProductService;
+use crate::vulnerability::service::VulnerabilityService;
+
+use base64::engine::general_purpose::STANDARD;
+use base64::engine::Engine as _;
+use langchain_rust::chain::options::ChainCallOptions;
+use langchain_rust::chain::Chain;
+use langchain_rust::language_models::options::CallOptions;
+use langchain_rust::tools::OpenAIConfig;
+use langchain_rust::{
+    agent::{AgentExecutor, OpenAiToolAgentBuilder},
+    llm::openai::OpenAI,
+    memory::SimpleMemory,
+    prompt_args,
+    tools::Tool,
+};
+
+use std::env;
+
+use langchain_rust::schemas::{BaseMemory, Message};
+use std::fmt::Write;
+use std::sync::Arc;
+
+pub const PREFIX: &str = include_str!("prefix.txt");
+
+pub struct AiService {
+    llm: Option<OpenAI<OpenAIConfig>>,
+    llm_info: Option<LLMInfo>,
+    pub tools: Vec<Arc<dyn Tool>>,
+}
+
+impl AiService {
+    /// Creates a new instance of the AI service.  It can be run against any OpenAI compatible
+    /// API endpoint.  The service is disabled if the OPENAI_API_KEY environment variable is not set.
+    /// You can configure the following environment variables to run against different OpenAI compatible:
+    ///
+    /// * OPENAI_API_KEY
+    /// * OPENAI_API_BASE (default: https://api.openai.com/v1)
+    /// * OPENAI_MODEL (default: gpt-4o)
+    ///
+    /// ## Running Against OpenAI:
+    /// OpenAI tends to provide cutting edge proprietary models, but they are not open source.
+    ///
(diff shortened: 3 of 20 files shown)


chirino:
Implementation of #811

carlosthe19916:
Interesting PR indeed! The new endpoint definition should more than enough.

When i first tried to use the endpoint http://localhost:8080/api/v1/ai/completions i got:

```
Json deserialize error: unknown variant `string`, expected one of `system`, `ai`, `human`, `tool` at line 6 column 30
```
So although the `message_type` could be an enum the backend response is very helpful so I can understand what I am doing wrong and then I fixed my mistake.  

I also tried to play with the whole thing following the commands in this PR but unfortunately I do not have the resources to run in locally (nothing to fix on this PR just sharing I need an upgrade of laptop with more resources I guess):

```
ingesting documents...
DONE!
Trustify Assistant〉hello                                                                                                                                                                                                                                                                                         >>
Error: Agent error: Error in agent planning: Chain error: LLM error: OpenAI error: api_error: model requires more system memory (40.8 GiB) than is available (20.2 GiB)
```

chirino:
I updated the API to expose the valid enum values.

Yeah the llama3.1:70b needs lots of memory.  If you don't have a big GPU, i'd recommend running against [OpenAI](https://platform.openai.com/) or [groq.com](https://groq.com/)

modules/fundamental/Cargo.toml
```diff
@@ -17,9 +17,12 @@ actix-http = { workspace = true }
 actix-web = { workspace = true }
 anyhow = { workspace = true }
 async-graphql = { workspace = true, features = ["uuid", "time"] }
+async-trait = { workspace = true }
```
bobmcwhirter, line 20: I _think_ that latest Rust version we're using means we don't need `async-trait` crate any more.

modules/fundamental/src/ai/service/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+    tools::Tool,
+};
+
+use std::env;
+
+use langchain_rust::schemas::{BaseMemory, Message};
+use std::fmt::Write;
+use std::sync::Arc;
+
+pub const PREFIX: &str = r#"
```
bobmcwhirter, line 33: I'd prefer, personally (non-blocking), a `prompt.txt` and `include_str!()` it.

modules/fundamental/src/ai/service/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+
+Always use the available tools to provide up to date information.
+
+Do not talk about your knowledge cutoff.
+"#;
+
+pub struct AiService {
+    llm: Option<OpenAI<OpenAIConfig>>,
+    llm_info: Option<LLMInfo>,
+    // advisory_service: AdvisoryService,
```
bobmcwhirter, line 56: Will these be returning?

modules/fundamental/src/ai/service/test.rs
```diff
@@ -0,0 +1,59 @@
+use crate::ai::model::ChatState;
+use crate::ai::service::AiService;
+use test_context::test_context;
+use test_log::test;
+use trustify_common::db::Transactional;
+use trustify_common::hashing::Digests;
+use trustify_module_ingestor::graph::product::ProductInformation;
+use trustify_test_context::TrustifyContext;
+
+pub async fn ingest_fixutures(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
```
bobmcwhirter, line 10: spelling: "fixtures"

xtask/src/ai.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    // export EXTERNAL_TEST_DB to use an external database, otherwise
+    // a temporary database will be created for testing.
+    let ctx = TrustifyContext::setup().await;
+
+    // if we are not using an external database, ingest some documents so that
+    // we use the data in them to respond to questions.
+    if env::var("EXTERNAL_TEST_DB").is_err() {
+        println!("{}", LightGray.paint("ingesting documents..."));
+        _ = ctx
```
bobmcwhirter, line 32: I think the `?` counts as using the result enough you don't have to assign to `_`

modules/fundamental/Cargo.toml
chirino: When I remove it and it's use I get this compiler error: 

```
error[E0195]: lifetime parameters or bounds on method `call` do not match the trait declaration
  --> modules/fundamental/src/ai/service/tools.rs:26:18
   |
26 |     async fn call(&self, input: &str) -> Result<String, Box<dyn Error>> {
   |                  ^ lifetimes do not match method in trait
```

modules/fundamental/src/ai/service/mod.rs
chirino: Nah don't think so.. will drop.

xtask/src/ai.rs
chirino: Your right... simplified now.

modules/fundamental/Cargo.toml
bobmcwhirter: I'll gander Monday.

xtask/src/ai.rs
```diff
@@ -0,0 +1,109 @@
...
+                        println!("\nChat history cleared...");
+                        continue;
+                    }
+                    _ => {}
+                }
+
+                chat_state.add_human_message(buffer.clone());
+                let pos = chat_state.messages.len();
+
+                let new_state = service.completions(&chat_state, ()).await?;
```
jcrossley3, line 100: What's the xtask for? Is it just another way of accessing the same service accessible through `/api/v1/ai/completions`?

jcrossley3 (commented):
I tried running it locally, but it doesn't seem practical without a GPU. I never got a response back from either `/api/v1/ai/completions` through the swagger-ui or the xtask, but the ollama process was hogging all my CPU.

Can you provide more detailed instructions for running it against a remote LLM?

chirino:
> Can you provide more detailed instructions for running it against a remote LLM?

## Using OpenAI:
Likely the best model you can test against but it's not an open-source model, and costs money.

1. generate an API key at: https://platform.openai.com/settings/profile?tab=api-keys 
2. 
        export OPENAI_API_KEY=xxxx

## Using Groq:
You can test against open source models and has a free their.

1. generate an API key at: https://console.groq.com/keys
2.    
        export OPENAI_API_KEY=xxxx
        export OPENAI_API_BASE=https://api.groq.com/openai/v1
        export OPENAI_MODEL=llama3-groq-70b-8192-tool-use-preview

xtask/src/ai.rs
chirino: Right.  a CLI version of it.

jcrossley3:
> > Can you provide more detailed instructions for running it against a remote LLM?
> 
> ## Using OpenAI:
> Likely the best model you can test against but it's not an open-source model, and costs money.

<snip>

Sorry, I meant as a part of this PR, maybe in the rustdoc for `AiService::new`?

modules/fundamental/src/ai/endpoints/test.rs
```diff
@@ -0,0 +1,40 @@
...
+        .set_json(req)
+        .to_request();
+
+    let response = app.call_service(request).await;
+    log::debug!("Code: {}", response.status());
+    assert_eq!(response.status(), StatusCode::OK);
+
+    let result: ChatState = actix_web::test::read_body_json(response).await;
+    log::info!("result: {:?}", result);
+    assert!(result.messages.last().unwrap().content.contains("37.17.9"));
```
jcrossley3, line 37: At the risk of exposing my AI ignorance, how does this work? Where is the magic that integrates the LLM with the contents of our DB?

chirino: A list of tools can be added to the AI agent.  The first call to the LLM contains the user question and a list of tools that are available.  The LLM will responds with a tool request if it wants use a tool, the agent then invokes the tool and adds it's results to the chat context. and calls the LLM again and so on.

the tool which is providing that DB info for that test is implemented at:
https://github.com/trustification/trustify/pull/830/files#diff-6b73c5edc3e0873b62eaf3ca01c547ec2b770f03663025856e3e3ae13bf7bc77R54

jcrossley3: Ah, so the magic is in our implementation of `langchain_rust::tools::Tool`? Makes sense, thanks.

modules/fundamental/src/ai/service/mod.rs
```diff
@@ -39,6 +39,48 @@ pub struct AiService {
...
+    ///
+    /// 1. install https://ollama.com/
+    /// 2. run `ollama pull llama3.1:70b`
+    /// 3. export the following env variables:
+    /// ```bash
+    /// export OPENAI_API_KEY=ollama
+    /// export OPENAI_API_BASE=http://localhost:11434/v1
+    /// export OPENAI_MODEL=llama3.1:70b
+    /// ```
+    ///
```
jcrossley3, line 83: Perfect, thanks!

jcrossley3 (approved):

chirino:
rebasing and squashing.

Issue #829, opened 2024-09-19
get 500 https://trustify.dentrassi.de/packages/10ff899d-3f9e-5d8b-8449-1290b8b050fb


JimFuller-RedHat:
curious as to what the logs say on this

ctron:
Yes, that might have been the case. I ran `dnf update` and rebooted at least twice.

JimFuller-RedHat:
dupe of #824

```
1234567891011121314151617181920212223242526272829303132333435362024-09-19 12:58:52.727 UTC [164] FATAL:  terminating background worker "parallel worker" due to administrator command
2024-09-19 12:58:52.727 UTC [165] FATAL:  terminating background worker "parallel worker" due to administrator command
2024-09-19 12:58:52.730 UTC [1] LOG:  background worker "parallel worker" (PID 165) exited with exit code 1
2024-09-19 12:58:52.730 UTC [1] LOG:  background worker "parallel worker" (PID 164) exited with exit code 1
2024-09-19 12:58:56.974 UTC [107] ERROR:  cannot start subtransactions during a parallel operation
2024-09-19 12:58:56.974 UTC [107] CONTEXT:  PL/pgSQL function semver_cmp(text,text) line 27 during statement block entry
        PL/pgSQL function semver_version_matches(text,version_range) line 7 at assignment
        PL/pgSQL function version_matches(text,version_range) line 4 at RETURN
2024-09-19 12:58:56.974 UTC [107] STATEMENT:  SELECT DISTINCT ON ("purl_status"."id") "purl_status"."id", "purl_status"."advisory_id", "purl_status"."vulnerability_id", "purl_status"."status_id", "purl_status"."base_purl_id", "purl_status"."version_range_id", "purl_status"."context_cpe_id", "version_range"."id", "version_range"."version_scheme_id", "version_range"."low_version", "version_range"."low_inclusive", "version_range"."high_version", "version_range"."high_inclusive" FROM "purl_status" LEFT JOIN "version_range" ON "purl_status"."version_range_id" = "version_range"."id" LEFT JOIN "base_purl" ON "purl_status"."base_purl_id" = "base_purl"."id" WHERE "purl_status"."base_purl_id" = $1 AND version_matches($2, "version_range".*)
2024-09-19 12:58:56.975 UTC [168] ERROR:  could not map dynamic shared memory segment
2024-09-19 12:58:56.976 UTC [169] ERROR:  could not map dynamic shared memory segment
[...]

Issue #824, opened 2024-09-18
Query Error: error returned from database: cannot start subtransactions during a parallel operation


JimFuller-RedHat:
```
2024-09-17T15:31:14.737291Z ERROR trustify_module_fundamental::error: Query Error: error returned from database: cannot start subtransactions during a parallel operation    
2024-09-17T15:31:14.737420Z ERROR trustify_module_fundamental::error: Query Error: error returned from database: cannot start subtransactions during a parallel operation    
2024-09-17T15:31:14.737420Z ERROR trustify_module_fundamental::error: Query Error: error returned from database: cannot start subtransactions during a parallel operation    
2024-09-17T15:31:14.738355Z  INFO actix_web::middleware::logger: ::1 "GET /api/v1/vulnerability/CVE-2024-25710 HTTP/1.1" 500 37 "http://localhost:8080/packages/02ffbf32-3526-52d4-84eb-900ba1741e3c" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.0.0 Safari/537.36" 0.043097    
2024-09-17T15:31:14.739111Z  INFO actix_web::middleware::logger: ::1 "GET /api/v1/vulnerability/CVE-2024-26308 HTTP/1.1" 500 37 "http://localhost:8080/packages/02ffbf32-3526-52d4-84eb-900ba1741e3c" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.0.0 Safari/537.36" 0.043820    
2024-09-17T15:31:14.739564Z  INFO actix_web::middleware::logger: ::1 "GET /api/v1/vulnerability/CVE-2023-42503 HTTP/1.1" 500 37 "http://localhost:8080/packages/02ffbf32-3526-52d4-84eb-900ba1741e3c" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.0.0 Safari/537.36" 0.044289    
2024-09-17T15:31:18.120901Z ERROR trustify_module_fundamental::error: Query Error: error returned from database: cannot start subtransactions during a parallel operation    
[...]

PR #831, opened 2024-09-20
mark version_matches pg funcs as unsafe.

JimFuller-RedHat:
mitigates #824 #829 - once purl refactor beds in we may reverse this decision.


ctron (approved):
I've no idea about the functions themselves, but how the get applied looks good to me.

JimFuller-RedHat:
for clarity - none of the function bodies were amended ... this just marks existing functions as not safe for parallel db processing.

PR #832, opened 2024-09-20
Try to come up with a consistent idea about logging/tracing

ctron:


docs/design/log_tracing.md
```diff
@@ -0,0 +1,112 @@
...
+
+A quick summary:
+
+* Don't panic, don't println (see below for exceptions).
+* Log messages and tracing should be relevant to operating personell and developers, not end users. Ops will either run
+  with a level of `warn` or `info` when sending to a tracing solution (like Jaeger).
+* Errors and warnings, which are relevant to the end user, must be sent back using the same channel (API call, UI, …)
+  not the logging system
+* Be reasonable with the amount of data being logged (might be transmitted to a tracing system, like Jaeger)
+* Instrumentation (`#[instrument]` and friends)
```
JimFuller-RedHat, line 15: minor nit: maybe add link to tracing cargo

common/src/decompress.rs
```diff
@@ -27,6 +28,7 @@ pub enum Error {
 /// **NOTE:** Depending on the size of the payload, this method might take some time. In an async
 /// context, it might be necessary to run this as a blocking function, or use [`decompress_async`]
 /// instead.
+#[instrument(skip(bytes), fields(bytes_len=bytes.len()), err(level=tracing::Level::INFO))]
```
JimFuller-RedHat, line 31: you explained this in the design doc but wanted to explicitly ask - the `err(level=tracing::Level::INFO)` is saying log errors from this func to info ... though this is not stopping halt of the application itself or wrapping up error mode behaviour.. right ? I could probably ask this question better!

docs/design/log_tracing.md
```diff
@@ -0,0 +1,112 @@
...
+```rust
+#[instrument(err(level=tracing::Level::INFO))]
+fn example() -> Result<LargeStruct, Error> {}
+```
+
+## println and panic
+
+An application should not panic during normal runtime. Also, all output should use logging or tracing.
+
+There is one exception to this rule (aside from tests): the start of the application. During the logging/tracing system
```
JimFuller-RedHat, line 108: what about migrations (I see usage of .panic there)

modules/importer/src/server/mod.rs
```diff
@@ -55,7 +55,7 @@ struct Server {
 }
 
 impl Server {
-    #[instrument(skip_all, err)]
+    #[instrument(skip_all, ret)]
```
JimFuller-RedHat, line 58: (more for my education) can you expand a little why err is removed here ?

modules/ingestor/src/service/dataset/mod.rs
```diff
@@ -36,7 +36,7 @@ impl<'g> DatasetLoader<'g> {
         }
     }
 
-    #[instrument(skip(self, buffer), ret)]
+    #[instrument(skip(self, buffer), err(level=tracing::Level::INFO))]
```
JimFuller-RedHat, line 39: related to other comment (though in reverse) eg. are we not interested in return values as well ?

JimFuller-RedHat (approved):
I like very much ... LGTM though interested in others thoughts.

common/src/decompress.rs
ctron: Correct, the only affect of this that the error gets logged as "info", rather than "error". Nothing else changes. The idea behind it being, that at this point, the error does not affect the application itself. So that doesn't warrant an "error log".

ctron: Not sure if there's something to add to the doc.

docs/design/log_tracing.md
ctron: Good point. I expanded that section a bit.

ctron:
@jcrossley3 @bobmcwhirter @mrizzi @chirino Any thoughts, ideas, feedback, opinion on this?

docs/design/log_tracing.md
```diff
@@ -0,0 +1,137 @@
...
+
+A few ideas around logging and tracing.
+
+## TL;DR
+
+A quick summary:
+
+* Don't panic, don't println (see below for exceptions).
+* [Log messages](https://crates.io/crates/log) and [tracing](https://crates.io/crates/tracing) should be relevant to
+  operating personell and developers, not end users. Ops will either run
```
jcrossley3, line 11:
```suggestion
  operating personnel and developers, not end users. Ops will either run
```

docs/design/log_tracing.md
```diff
@@ -0,0 +1,137 @@
...
+
+Logging and tracing is intended for either developers or operators (humans) of the system. Not for end users.
+
+This means that errors or warnings relevant to the end user should be propagated through the API and UI to the user,
+not the console or logging/tracing systems like opentelemetry.
+
+## Levels
+
+* `error` – Something went seriously wrong, the application can no longer work properly and will most likely exit next.
+* `warning` – The application encountered a serious issue. However, it can continue to run and future operations are likely to succeed of the cause has been resolved.
```
jcrossley3, line 32:
```suggestion
* `warning` – The application encountered a serious issue. However, it can continue to run and future operations are likely to succeed if the cause has been resolved.
```

docs/design/log_tracing.md
```diff
@@ -0,0 +1,137 @@
...
+* `info` – Something noteworthy to an operator in the case where more information is required. This might include
+  important configuration settings, but must exclude secrets and credentials (which should be obfuscated).
+* `debug` – Something noteworthy to a developer in the case where more information is required.
+* `tracing` – Most detailed information possible.
+
+A default setting for production workload would be `warning`. An alternate setting when operation issues are being
```
jcrossley3, line 38:
```suggestion
A default setting for production workload would be `warn`. An alternate setting when operation issues are being
```

docs/design/log_tracing.md
```diff
@@ -0,0 +1,137 @@
...
+Function calls can be wrapped with the `#[instrument]` attribute, which allows tracing calls to functions, including
+their arguments and return values.
+
+Adding instrumentation like this adds some overhead. So it should be applied with care. However, on the other side, it
+can be used to profile the application, also in production environments. This can be a considerable help, when things
+go wrong in production.
+
+### Arguments
+
+Arguments get serialized in a string format which can either be printed other sent to a tracing system. However, this
```
jcrossley3, line 55:
```suggestion
Arguments get serialized in a string format which can either be printed or sent to a tracing system. However, this
```

docs/design/log_tracing.md
```diff
@@ -0,0 +1,137 @@
...
+configuration, it might be ok to use println. And if the application cannot be started up, failing with a panic might
+be ok too.
+
+The same is true for simple command line applications, which are expected to perform a task, and then finish. Either
+successful, or with an error (returned as exit code). For example, the migrations. They can run and fail, even with
+a panic.
+
+### Rationale
+
+The main reason for all of this is: panics cannot be treated as graceful as `Result::Err`. Panics might abort the
```
jcrossley3, line 119:
```suggestion
The main reason for all of this is: panics cannot be treated as gracefully as `Result::Err`. Panics might abort the
```

jcrossley3 (approved):
Nice write-up! Found a few typos.

helio-frota:
I also found that a nice doc

helio-frota (approved):

PR #845, opened 2024-09-23
Provide a meaningful way for updating/deleting CSAF advisories

### LINKED ISSUE #856: Document identifier for CSAF advisories
We have an `id` field, and an `identifier` field. The former is a unique, hash based, id of the document. If a bit of the serialized document changes, that ID must change.

The latter is an identifier provided by the document content itself. For CSAF, that is currently the `document.tracking.id`. However, that is only unique in the context of the `document.tracking.namespace`. So I guess we should add that, when ingesting the document.

ctron:
That PR is more of a discussion than a PR (as of now).

The goal was to find out, how the system behaves when updating a CSAF advisory. It looks like any change creates a new advisory. Despite the combination of namespace and ID being unique.

This means, that changing an affected product to fixed, will leave two advisories. One saying the product is affected, one saying the product is not affected. Both have the same "ID", but also ignore the namespace.

I think this is something we need to discuss. As right now, it's seems a bit weird.

The new tests are either failing, or passing (incorrectly), but there are some TODOs, explaining the expectations.

## ToDo

* [x] Update when deleting an advisory
* [x] Add test for deleting an advisory

Closes  #856

### DIFF
--- a/migration/src/m0000650_alter_advisory_tracking/update_deprecated_advisory.sql
+++ b/migration/src/m0000650_alter_advisory_tracking/update_deprecated_advisory.sql
@@ -2,15 +2,12 @@ CREATE OR REPLACE FUNCTION update_deprecated_advisory(identifier_input TEXT DEFA
     RETURNS VOID AS
 $$
 BEGIN
-    WITH MostRecent AS (SELECT DISTINCT ON (identifier) id
-                        FROM advisory
-                        WHERE identifier = COALESCE(identifier_input, identifier)
-                        ORDER BY identifier, modified DESC)
     UPDATE advisory
-    SET deprecated = CASE
-                         WHEN id IN (SELECT id FROM MostRecent) THEN FALSE
-                         ELSE TRUE
-        END
+    SET deprecated = (id != (SELECT id
+                             FROM advisory a
+                             WHERE a.identifier = COALESCE(identifier_input, advisory.identifier)
+                             ORDER BY a.modified DESC
+                             LIMIT 1))
     WHERE identifier = COALESCE(identifier_input, identifier);
 END;
 $$ LANGUAGE plpgsql;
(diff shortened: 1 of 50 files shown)


ctron:
@JimFuller-RedHat @jcrossley3 @bobmcwhirter @mrizzi I'd like your thoughts on this. I think it's one of those hot potatoes we try to avoid for a while now. Let's focus on advisories first.

ctron:
After a chat with @bobmcwhirter we'll try an approach of "deprecating" old documents, based on the CSAF tracking/version information.

bobmcwhirter:
And iffff that works, figure out similar approach for other advisory types that may lack version numbers.

but if they come from a git repo, we can at least know "more recent" from git commit shas.

ctron:
I implemented the idea, adding the versions in our database, but sticking to the "modifed" timestamp for performing the evaluation. The state right now is first draft if this, which seems to work fine, but needs some fine-tuning.

ctron:
@bobmcwhirter @JimFuller-RedHat if you find some time, maybe take a second look. Tests seem to pass.

migration/src/m0000650_alter_advisory_tracking/update_deprecated_advisory.sql
```diff
@@ -0,0 +1,16 @@
...
+                        WHERE identifier = COALESCE(identifier_input, identifier)
+                        ORDER BY identifier, modified DESC)
+    UPDATE advisory
+    SET deprecated = CASE
+                         WHEN id IN (SELECT id FROM MostRecent) THEN FALSE
+                         ELSE TRUE
+        END
+    WHERE identifier = COALESCE(identifier_input, identifier);
+END;
+$$ LANGUAGE plpgsql;
```
JimFuller-RedHat, line 16: there maybe a more efficient func eg ... is distinct really needed ? dont we want to just return 1 thing from this ? using **!=** operator instead of **IN** probably avoids temporary result set (though I would have to check with explain analyze ... kinda hard with functions). 

```
CREATE OR REPLACE FUNCTION update_deprecated_advisory(identifier_input TEXT DEFAULT NULL)
    RETURNS VOID AS
$$
BEGIN
    UPDATE advisory
    SET deprecated = (id != (SELECT id
                             FROM advisory a
                             WHERE a.identifier = COALESCE(identifier_input, advisory.identifier)
                             ORDER BY a.modified DESC
                             LIMIT 1))
    WHERE identifier = COALESCE(identifier_input, identifier);
END;
$$ LANGUAGE plpgsql;
```

JimFuller-RedHat (approved):
@ctron walked me through the code, LGTM

PR #846, opened 2024-09-23
Also generate the openapi client in trustify-ui when the openapi spec changes.

chirino:


.github/workflows/openapi.yaml
```diff
@@ -24,9 +29,14 @@ jobs:
         run: |
           rm ./trustify-ui/client/openapi/trustd.yaml
           cp ./trustify/openapi.yaml ./trustify-ui/client/openapi/trustd.yaml
-          
           cd ./trustify-ui
           git diff
+
+          if ! git diff --quiet; then
+            npm install
```
ctron, line 36: Wondering if this shouldn't be `ci` instead of `i`, honoring the lock file?

.github/workflows/openapi.yaml
```diff
@@ -24,9 +29,14 @@ jobs:
         run: |
           rm ./trustify-ui/client/openapi/trustd.yaml
           cp ./trustify/openapi.yaml ./trustify-ui/client/openapi/trustd.yaml
-          
           cd ./trustify-ui
           git diff
+
+          if ! git diff --quiet; then
```
carlosthe19916, line 35: This current GH action will only be triggered when a change to the file openapi.yaml is done, therefore this verification doesn't hurt but seems redundant. The `if` condition will always be `true`

https://github.com/trustification/trustify/blob/bc8a88defd2666bd6e6551be29ed215627ffc365/.github/workflows/openapi.yaml#L3-L8

chirino: Ah.. I had not noticed that.. I'll simplify.

.github/workflows/openapi.yaml
chirino: I'm not sure I understand.  Could you rephrase?

carlosthe19916: I think the suggestion is to use the command `npm ci` rather than `npm install` (`npm i` is the shorter command version).  If we copy and paste the command from the README file of the trustify-ui repository the command would be `npm clean-install --ignore-scripts` or `npm ci --ignore-scripts`

ctron: Yup, thanks for the explanation!

chirino: ahhh... I had never used `clean-install` before.  Seems like it's what you would want to do for a release build.  But this is not a release build.    It's doing what a developer would do on his machine to submit a PR request.  If that means updating lock files, I think that's ok.  The lock file update will be included in the PR review.

ctron: :shrug: ok, then :grin:

ctron (approved):

PR #848, opened 2024-09-24
fix: handle NONE and NOASSERTION as relationship targets

### LINKED ISSUE #552: Handle `NONE` and `NOASSERTION` relationships in SPDX
Haven't seen them so far. However, technically it is possible to have a relationship with the target `NONE` or `NOASSERTION`. `NONE` seems simple, the latter I have no idea what to do with.

https://spdx.github.io/spdx-spec/v2.3/relationships-between-SPDX-elements/#1111-description

> In cases where there are "known unknowns", the use of the keyword `NOASSERTION` can be used on the right hand side of a relationship to indicate that the author is not asserting whether there are other SPDX elements (package/file/snippet) that are connected by relationships or not. That is, there could be some, but the author is not asserting one way or another.
>
> Similarly, the use of the keyword `NONE` can be used to indicate that an SPDX element (package/file/snippet) has no other elements connected by some relationship to it.

ctron:
Closes #552

### DIFF
--- a/modules/ingestor/src/graph/sbom/common/relationship.rs
+++ b/modules/ingestor/src/graph/sbom/common/relationship.rs
@@ -30,7 +30,29 @@ impl RelationshipCreator {
         }
     }
 
+    /// Record a relationship.
+    ///
+    /// To store those relationships, it is required to call [`Self::create`].
+    ///
+    /// It is possible to record invalid relationship targets, which might fail the actual creation
+    /// process later on. It is possible to validate relationships using [`Self::validate`].
     pub fn relate(&mut self, left: String, rel: Relationship, right: String) {
+        // The idea of `NOASSERTION` is to state that there is a relationship, but the element it
+        // relates to is unknown.
+        //
+        // The idea of `NONE` is to state there it is known that there is no relationship to that
+        // element.
+        //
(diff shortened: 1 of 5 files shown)


modules/ingestor/src/graph/sbom/common/relationship.rs
```diff
@@ -30,7 +30,27 @@ impl RelationshipCreator {
...
+        // The idea of `NOASSERTION` is to state that there is a relationship, but the element it
+        // relates to is unknown.
+        //
+        // The idea of `NONE` is to state there it is known that there is no relationship to that
+        // element.
+        //
+        // At the moment, both those pieces of information don't add value to our system and
+        // only cause complexity when storing. So we simply drop it.
+
+        if let ("NONE" | "NOASSERTION", _) | (_, "NONE" | "NOASSERTION") = (&*left, &*right) {
```
JimFuller-RedHat, line 49: might be worth putting a TODO: here as a flag to some poor future programmer ;) otherwise this looks good.

ctron: Added a line.

JimFuller-RedHat (approved):
LGTM

PR #863, opened 2024-09-30
fix: avoid workflow failures in forks

jcrossley3:
There are some workflows we want to run in downstream forks, and others we don't.


ctron (approved):

ctron:
I had to revert part of it, as it was failing: https://github.com/trustification/trustify/actions/runs/11119332590

PR #867, opened 2024-10-01
Ensure CVEs & OSV can be updated/deleted too

ctron:
Builds on top of https://github.com/trustification/trustify/pull/845


ctron:
@bobmcwhirter @JimFuller-RedHat I rebased it, should be a lot smaller and clearer now.

JimFuller-RedHat (approved):
looks good to me!

Issue #837, opened 2024-09-20
poorly performing query when hitting api/v1/purl/{uuid}


JimFuller-RedHat:
At scale, hitting api/v1/purl/{uuid} blows up

```
1232024-09-20T14:11:27.271481Z  INFO actix_web::middleware::logger: ::1 "GET /api/v1/purl/8543edb3-05b9-5b64-a438-c7af4c58cd65 HTTP/1.1" 200 1782 "http://localhost:8080/packages/8543edb3-05b9-5b64-a438-c7af4c58cd65" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.0.0 Safari/537.36" 0.082499    
[...]

PR #880, opened 2024-10-02
Add id indexes to speed up purl resolution.

JimFuller-RedHat:
This mitigates perf problem seen in #837 by adding simple indexes on **id** for base_purl, versioned_purl and qualified_purl tables. 

Probably the last of the 'obvious' indexes.

### DIFF
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -82,6 +82,7 @@ mod m0000630_create_product_version_range;
 mod m0000631_alter_product_cpe_key;
 mod m0000640_create_product_status;
 mod m0000650_alter_advisory_tracking;
+mod m0000660_purl_id_indexes;
 
 pub struct Migrator;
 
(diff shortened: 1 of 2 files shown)


JimFuller-RedHat:
adding @ctron for review so he sees I am using 660 migration ;)

migration/src/lib.rs
```diff
@@ -168,6 +170,7 @@ impl MigratorTrait for Migrator {
             Box::new(m0000630_create_product_version_range::Migration),
             Box::new(m0000631_alter_product_cpe_key::Migration),
             Box::new(m0000640_create_product_status::Migration),
+            Box::new(m0000660_purl_id_indexes::Migration),
```
ctron, line 173: Thanks for leaving the gap. :heart:

ctron (approved):

PR #882, opened 2024-10-02
Wire up the full ClearlyDefined definition importer from their blobst…

bobmcwhirter:
…ore.


jcrossley3 (commented):
I don't understand why we need both a `ClearlyDefinedLoader` taking a `Value` and a `ClearlyDefinedCurationLoader` taking a `Curation`.

I'm not familiar enough with the walker interfaces, so I'll defer to @ctron on that.

bobmcwhirter:
> I don't understand why we need both a `ClearlyDefinedLoader` taking a `Value` and a `ClearlyDefinedCurationLoader` taking a `Curation`.

The `Curation` YAML was fairly small/easy to make a `serde` impl for.

The `Definition` JSON is more complex, and since we're just plucking a few small bits out of it, figured JsonPath was sufficient against `Value`.

tl;dr: lazy

modules/importer/src/runner/clearly_defined/walker.rs
```diff
@@ -0,0 +1,208 @@
...
+        }
+    }
+
+    pub fn continuation(mut self, continuation: ClearlyDefinedItemContinuation) -> Self {
+        self.continuation = continuation;
+        self
+    }
+
+    pub async fn run(mut self) -> Result<ClearlyDefinedItemContinuation, Error> {
+        let changes = reqwest::get(self.changes_index_url()).await?;
```
ctron, line 47: This creates a new connection pool for each request. I think we should create and keep a client.

Same for the other calls.

modules/ingestor/src/service/sbom/clearly_defined.rs
```diff
@@ -14,50 +22,165 @@ impl<'g> ClearlyDefinedLoader<'g> {
...
+                document_id: previously_found.document_id,
+                warnings: vec![],
+            });
+        }
 
-        let sbom = self
-            .graph
-            .ingest_sbom(labels, digests, &curation.document_id(), &curation, &tx)
-            .await?;
+        println!("{:#?}", item);
```
ctron, line 44: No `println` please.

ctron (changes requested):
A few small things.

ctron (approved):

PR #884, opened 2024-10-03
chore: update to otel 0.24

### LINKED ISSUE #876: info: not able to update otel deps yet after actix-web-opentelemetry updates
This project updated to otel 0.24 and that means we could update all the other otel-deps
https://github.com/OutThereLabs/actix-web-opentelemetry/releases

I did some initial attempts and got blocked by compilation problem with https://github.com/trustification/trustify/blob/471a47220d4ed30a015e70b33536619a4ab5c8c8/common/infrastructure/src/tracing.rs#L140-L141

( I hope to be missing something  )

Related with this comment https://github.com/trustification/trustify/pull/731

helio-frota:
Closes #876


common/infrastructure/src/tracing.rs
```diff
@@ -112,33 +112,32 @@ fn init_otlp(name: &str) {
...
 
+    println!("Using Jaeger tracing.");
+    println!("{:#?}", provider);
+
     let formatting_layer = tracing_subscriber::fmt::Layer::default();
 
     if let Err(e) = tracing_subscriber::Registry::default()
         .with(tracing_subscriber::EnvFilter::from_default_env())
-        .with(tracing_opentelemetry::layer().with_tracer(tracer))
+        .with(tracing_opentelemetry::layer().with_tracer(provider.tracer("")))
```
ctron, line 140: I am wondering if putting in an empty string here is the right thing to do?!

ctron: From the rust docs:

> The `name` should be the application name or the name of the library providing instrumentation.

So I think it makes sense to actually pass in the `name` argument of the function.

ctron (approved):

helio-frota:
rebased

PR #891, opened 2024-10-04
chore(ai): split out the tools into their own files

chirino:
Also collocate the test for each tool in the same file. This should make it easier to find and work on each tool individually


jcrossley3 (approved):
I dig the refactor, thanks!

PR #892, opened 2024-10-04
fix: test failure due to .DS_Store files being present

chirino:
We now ignore .DS_Store entries.


helio-frota (approved):

ctron (approved):
I honestly don't want to add such code.

Issue #870, opened 2024-10-01
Allow dealing with versionless PURLs


ctron:
Ingesting OSV documents creates "versionless PURLs" (aka base PURLs). It also adds a entries in the `version_range` table.

However, I don't see any way to get back the originally ingested OSV document through such PURLs. My expectation would be, that there exists a PURL (versioned or not), which I can use to query the system, receiving the originally ingested OSV documents in return.

PR #897, opened 2024-10-07
feat: create versioned purls from OSV

ctron:
Also see: https://github.com/trustification/trustify/issues/870

### DIFF
--- a/modules/ingestor/src/service/advisory/osv/loader.rs
+++ b/modules/ingestor/src/service/advisory/osv/loader.rs
@@ -4,6 +4,7 @@ use crate::{
             advisory_vulnerability::{Version, VersionInfo, VersionSpec},
             AdvisoryInformation, AdvisoryVulnerabilityInformation,
         },
+        purl::creator::PurlCreator,
         Graph,
     },
     model::IngestResult,
@@ -70,6 +71,8 @@ impl<'g> OsvLoader<'g> {
                 .await?;
         }
 
+        let mut purl_creator = PurlCreator::new();
+
         for cve_id in cve_ids {
             self.graph.ingest_vulnerability(&cve_id, (), &tx).await?;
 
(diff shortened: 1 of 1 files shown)


modules/ingestor/src/service/advisory/osv/loader.rs
```diff
@@ -104,97 +107,85 @@ impl<'g> OsvLoader<'g> {
...
+                        }
+
+                        if let (_, Some(fixed)) = &parsed_range {
+                            advisory_vuln
+                                .ingest_package_status(
+                                    None,
+                                    &purl,
+                                    "fixed",
+                                    VersionInfo {
+                                        // TODO detect better version scheme
```
JimFuller-RedHat, line 174: can we expand on this or raise an issue with what you are thinking ?

modules/ingestor/src/service/advisory/osv/loader.rs
```diff
@@ -104,97 +107,85 @@ impl<'g> OsvLoader<'g> {
...
+
+                let Some(package) = &affected.package else {
+                    continue;
+                };
 
-                    purls.extend(translate::to_purl(package).map(Purl::from));
+                // extract PURLs
+
+                let mut purls = vec![];
+                purls.extend(translate::to_purl(package).map(Purl::from));
```
JimFuller-RedHat, line 119: is this some kind of hyper specific translate to purl ?

modules/ingestor/src/service/advisory/osv/loader.rs
```diff
@@ -104,97 +107,85 @@ impl<'g> OsvLoader<'g> {
             }
 
             for affected in &osv.affected {
-                if let Some(package) = &affected.package {
-                    let mut purls = vec![];
+                // we only process it when we have a package
```
JimFuller-RedHat, line 110: might be worth a log::trace!

JimFuller-RedHat (commented):
I do not think existing test has any assertions for checking versions for this specific fix might be worth adding, otherwise minor comments - LGTM

modules/ingestor/src/service/advisory/osv/loader.rs
ctron: Yup.

PR #901, opened 2024-10-07
feat: create versioned purls from OSV

ctron:
Re-open for #897


JimFuller-RedHat (approved):
LGTM though we might consider an assert to agitate this in existing test.

ctron:
> LGTM though we might consider an assert to agitate this in existing test.

I'll create a follow up PR.

PR #906, opened 2024-10-08
test: add test for updating CWEs

ctron:


modules/fundamental/tests/weakness/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+        .list_weaknesses(
+            Default::default(),
+            Paginated {
+                offset: 0,
+                limit: 10,
+            },
+        )
+        .await?;
+
+    assert_eq!(964, all.total);
```
helio-frota, line 51: nit to replace the magic number with a const here and on line 83

`const TOTAL_ITEMS_FOUND:u64 = 964;`

ctron: Done.

helio-frota (approved):

PR #907, opened 2024-10-08
Refactor and enhance version comparison

ctron:


bobmcwhirter (commented):
Looking good.

migration/src/m0000670_version_cmp/generic_version_matches.sql
```diff
@@ -0,0 +1,28 @@
...
+            if version_p = range_p.high_version  then
+                return true;
+            end if;
+        end if;
+    end if;
+
+    return false;
+
+end
+$$
```
JimFuller-RedHat, line 26: you could rewrite to be a bit more compact:
 
```
create or replace function generic_version_matches(version_p text, range_p version_range)
    returns bool
as
$$
begin
    return (
        (range_p.low_version is not null and range_p.low_inclusive and version_p = range_p.low_version)
        or (range_p.high_version is not null and range_p.high_inclusive and version_p = range_p.high_version)
    );
end
$$ language plpgsql immutable;
```
untested,  I doubt this would result in such a speed up that it is worth reducing readability

modules/ingestor/src/service/advisory/osv/prefix.rs
```diff
@@ -0,0 +1,45 @@
...
+    }
+}
+
+struct PrefixMapping {
+    prefix: String,
+    name: String,
+}
+
+pub fn get_well_known_prefixes() -> &'static PrefixMatcher {
+    WELL_KNOWN_PREFIXES.get_or_init(|| {
```
JimFuller-RedHat, line 33: I wonder (one for the future) if WELL_KNOWN_PREFIXES needs an env var

modules/ingestor/src/service/advisory/osv/prefix.rs
```diff
@@ -0,0 +1,45 @@
...
+        self.prefixes.push(PrefixMapping {
+            prefix: prefix.into(),
+            name: name.into(),
+        })
+    }
+
+    pub fn detect(&self, input: &str) -> Option<String> {
+        self.prefixes
+            .iter()
+            .find(|each| input.starts_with(&each.prefix))
```
JimFuller-RedHat, line 22: have you considered using a find_map here ?

JimFuller-RedHat (approved):
apart from some minor observations, LGTM

PR #908, opened 2024-10-08
fix: xtask generate-dump

helio-frota:


helio-frota:
```
2024-10-08T22:55:29.490776Z  INFO Done ingesting
2024-10-08T22:55:29.801560Z  INFO Dumped to: dump.sql
```

```
➜  trustify git:(fix-dump) tail dump.sql
--

ALTER TABLE ONLY public.version_range
    ADD CONSTRAINT version_range_version_scheme_id_fkey FOREIGN KEY (version_scheme_id) REFERENCES public.version_scheme(id);


--
-- PostgreSQL database dump complete
--

➜  trustify git:(fix-dump)
```

xtask/src/dataset.rs
```diff
@@ -110,21 +110,17 @@ impl GenerateDump {
...
             .dbname(db.name())
             .file(&self.output)
-            .build_tokio()
-            .output()
-            .await?;
+            .build();
+        let (stdout, stderr) = pg_dump.execute()?;
 
-        if !output.status.success() {
+        if !stderr.is_empty() {
```
ctron, line 120: This feels a bit awkward. What is the contract of the `execute` function? Will it fail in case of running `pg_dump` fails? If that's the case, then we have three outcomes now (instead of two from before):

* pg_dump failed
* `execute` failed
* success

The two failures being handled in a different way.

If the `exeute `call does not handle the error of `pg_dump`, then we're assuming that output on `stderr` means that the application failed. Which isn't necessarily the case.

helio-frota: > This feels a bit awkward

I agree, I cherry-picked that from tests https://github.com/theseus-rs/postgresql-embedded/blob/344b47846cb9dbc2040982d542195fb03ef36f30/postgresql_embedded/tests/dump_command.rs#L36

* Execute returns a [Result] tuple with 2 strings
  * https://github.com/theseus-rs/postgresql-embedded/blob/main/postgresql_commands/src/traits.rs#L131

* The impl handles the success and returns the tuple
  * https://github.com/theseus-rs/postgresql-embedded/blob/main/postgresql_commands/src/traits.rs#L213-L217

ctron: I think it is ok to keep the `execute` call in general. The question to me would be: how to deal with the outcome?

If the call fails, I'd like to understand "why". The stdout/stderr output did help me in the past. So I'd still like to see that in the case of an error.

If the command success, telling the user "failed to run" seems wrong. As the `execute` call did succeed.

I think one way to deal with this could be to:

* ensure we see something on the console in case the command files (including stdout/stderr)
* if the command succeeds, simply dump stdout/stderr to the logging system (not failing the function itself)

helio-frota: sorry, I'm not sure if I understood 

> If the call fails, I'd like to understand "why"

I've changed the database name to `abc` and it fails with a proper message, although we can see `stdout` is empty (according to the function signature / contract / impl).

```
2024-10-09T11:36:39.871578Z  INFO Done ingesting
Error: Command error: stdout=; stderr=pg_dump: error: connection to server at "localhost" (::1), port 42535 failed: FATAL:  database "abc" does not exist
```

> if the command succeeds, simply dump stdout/stderr to the logging system (not failing the function itself)

that means the `else` (`stderr.is_empty()`), keeping the current code logging the dump-file-name and returning `Ok`

```rust
        } else {
            log::info!("Dumped to: {}", self.output.display());
            Ok(())
        }
    }
```

Although both stdout and stderr are `empty` when command succeeds :shrug: 

```rust
    log::info!("{stdout}");
    log::info!("{stderr}");
    log::info!("Dumped to: {}", self.output.display());
```

```
2024-10-09T11:54:53.841753Z  INFO Done ingesting
2024-10-09T11:54:53.900340Z  INFO
2024-10-09T11:54:53.900375Z  INFO
2024-10-09T11:54:53.900383Z  INFO Dumped to: dump.sql
```

ctron: I think that's ok. IIRC the info level isn't shown by default. If it is, then maybe lower that logging statement do debug. Prefix with `stdout:` and `stderr:` and it should be good.

helio-frota: debug level for the `success` also showing nothing
```
2024-10-09T12:35:47.131645Z DEBUG stdout:
2024-10-09T12:35:47.131656Z DEBUG stderr:
```

ctron: True, but that's ok. If there's nothing, we show nothing. But hide it as debug information.

helio-frota: ok so updated the pr adding 2 extra lines of debug in case success :+1:

ctron: Cool. But can't we just remove the `!stderr.is_empty()` branch? If the command had failed, that would be handled by the `Err` response of the `execute` call.

helio-frota: cool it works fine, let's remove 

```
2024-10-09T14:31:29.703513Z DEBUG Result: 1
stdout:
stderr: pg_dump: error: connection to server at "localhost" (::1), port 42981 failed: FATAL:  database "aaa" does not exist

Error: Command error: stdout=; stderr=pg_dump: error: connection to server at "localhost" (::1), port 42981 failed: FATAL:  database "aaa" does not exist
```

helio-frota: pr updated

ctron (approved):

Issue #804, opened 2024-09-16
V2 CI/CD pipeline and deployment to Trustify Staging


PhilipCattanach:
Create a new Trustify Staging environment on infrastructure paid for by Red Hat. 
Essentially a V2 equivalent of https://staging.trustification.dev/
Create the Trustify build pipeline so that every commit to the main branch triggers a new build and executes the integration tests. On a nightly basis the latest build must be automatically deployed to Trustify Staging.

PhilipCattanach:
@jcrossley3 - Both Jens and Dejan help manage the costs of cloud based infrastructure. I think they were both involved in the set-up of the https://staging.trustification.dev/ and https://trustification.dev/ environments. Speaking to those guys would be where to start.

jcrossley3:
@ctron @dejanb you've been summoned. help?

jcrossley3:
Correct me if I'm wrong, but this issue seems to be a request to migrate https://trustify.dentrassi.de/ to https://staging.trustification.dev/ and since @ctron set the former up, I would be entirely relying on his expertise to close this issue.

ctron:
I think it would be something new. `https://trustify.dentrassi.de/` is more like a personal instance, not using any OCP stuff.

So we'd need to come up with new Helm charts, or an operator, or some Ansible scripts to get there.

jcrossley3:
Well, I'm personally clueless wrt all of those. :smile:

carlosthe19916:
My two cents:
Whatever is the strategy used for deploying Trustify (as described in this gh issue) it would be great not to create an "independent way of deploying things just for this use case". E.g. for the old Trustification it seems we had a whole independent way of deploying it through https://github.com/trustification/trustification.rocks but as the time goes we stopped maintaining it in favor of the pipelines defined by Conflux and the official Helm charts we distribute as RHTPA.

So ideally , we should reuse whatever strategy that work on the long term (so we guarantee that we will always maintain it) without creating too much of an extra work.

bobmcwhirter:
We have budget for environments on AWS.

I don't know that we necessarily need (at this point) OpenShift, since "deploy on RHEL" is also a goal.

I think order of operations is probably:

1. Get it building in Konflux.
2. Get each builds deployed ... somewhere easy and simple
3. No idea
4. Enjoy some cake
5. Figure out deployment into something less easy and simple, probably involving k8s.

jcrossley3:
> 1. Get it building in Konflux.

[...]

PR #912, opened 2024-10-09
feat: CI pipeline to build, package, test, and deploy each merged PR

jcrossley3:
Fixes: #804

Requires an `OPENSHIFT_TOKEN` secret that enables privileged access to an OCP instance, e.g. https://api.cluster.trustification.rocks:6443

The following commands were used to generate the token on the OCP host:

```
oc create serviceaccount ${ACCT_NAME}
oc create clusterrolebinding ${ACCT_NAME} --clusterrole=cluster-admin --serviceaccount=default:${ACCT_NAME}
oc create token ${ACCT_NAME} --duration=$((365*24))h
```
Upon successful completion of the `latest-release` workflow, the app will be available here:
https://trustify-latest-staging.apps.cluster.trustification.rocks/

### DIFF
--- a/.github/workflows/latest-release.yaml
+++ b/.github/workflows/latest-release.yaml
@@ -1,4 +1,4 @@
-name: container-latest
+name: latest-release
 
 # Build "latest" container, run e2e tests against it, and deploy it to
 # https://trustify-latest-staging.apps.cluster.trustification.rocks/
@@ -9,7 +9,7 @@ on:
       - main
   workflow_dispatch:
 
-concurrency: container-latest
+concurrency: latest-release
 
 env:
   IMAGE_NAME: trustd
--- a/.github/workflows/latest-release.yaml
+++ b/.github/workflows/latest-release.yaml
@@ -77,6 +77,7 @@ jobs:
       run_ui_tests: true
 
   deploy:
+    if: ${{ github.repository == 'trustification/trustify' }}
     runs-on: ubuntu-22.04
     needs:
       - publish
--- a/.github/workflows/container-latest.yaml
+++ b/.github/workflows/container-latest.yaml
@@ -1,35 +1,41 @@
 name: container-latest
 
-# Build a "latest" container for every commit
+# Build "latest" container, run e2e tests against it, and deploy it to
+# https://trustify-latest-staging.apps.cluster.trustification.rocks/
 
 on:
   push:
     branches:
       - main
+  workflow_dispatch:
 
 concurrency: container-latest
 
+env:
+  IMAGE_NAME: trustd
+  IMAGE_TAG: latest
+  OPENSHIFT_NAMESPACE: trustify-latest
+  OPENSHIFT_SERVER: https://api.cluster.trustification.rocks:6443
+  APP_NAME: staging
+
 jobs:
   build:
-    if: ${{ github.repository == 'trustification/trustify' }}
     uses: ./.github/workflows/build-binary.yaml
     with:
       version: latest
 
   publish:
-    if: ${{ github.repository == 'trustification/trustify' }}
     needs: [ build ]
     runs-on: ubuntu-22.04
+
+    outputs:
+      image: ${{ steps.push.outputs.registry-path }}
 
     permissions:
       contents: write
       packages: write
       id-token: write
       attestations: write
-
-    env:
-      IMAGE_NAME: trustd
-      IMAGE_TAG: latest
 
     steps:
 
@@ -53,6 +59,7 @@ jobs:
       # Push to ghcr.io
 
       - name: Push to ghcr.io
+        id: push
         uses: redhat-actions/push-to-registry@v2
         with:
           image: ${{ env.IMAGE_NAME }}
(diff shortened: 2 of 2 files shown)


jcrossley3:
You can view the results of the workflow on my personal branch: https://github.com/jcrossley3/trustify/actions/runs/11264784849

helio-frota (approved):

.github/workflows/latest-release.yaml
```diff
@@ -0,0 +1,101 @@
...
+    with:
+      server_image: ${{ needs.publish.outputs.image }}
+      run_api_tests: true
+      run_ui_tests: true
+
+  deploy:
+    runs-on: ubuntu-22.04
+    needs:
+      - publish
+      - e2e-test
```
carlosthe19916, line 83: @jcrossley3 I arrived late to the party as this PR has been merged already.

1. I wanted to point our attention to the fact that we will deploy the code only if the e2e tests passed. I think we should deploy it regardless of the e2e pass or not. Actually, I think the e2e test might or not might fail. Just yesterday you faced an e2e error due to the fact that the the UI of the backend had an outdated version. But I do not hold a strong opinion so I'll leave that to you.
2. I thought the idea was to deploy the downstream version of Trustify not the upstream one. That might be just me not getting the requirement correctly though. In any case this is a good milestone!
3. I created this issue https://github.com/trustification/trustify-operator/issues/34 in regards to the fact that you made the trustify-operator scripts to delete the namespace. I think that is a dangerous thing to be set in the trustify-operator scripts and the action of deletion should be done by the user itself (a conscious decision). So Just like you are using `redhat-actions/oc-login@v1` to login in OCP you should delete the Namepace in a command defined in this workflow.
  - oc login (`uses: redhat-actions/oc-login@v1`)
  - oc delete project MY_PROJECT
  - Install the oprator `uses: trustification/trustify-operator/.github/actions/install-trustify@main`

jcrossley3: > 1. I wanted to point our attention to the fact that we will deploy the code only if the e2e tests passed. I think we should deploy it regardless of the e2e pass or not. Actually, I think the e2e test might or  might not fail. Just yesterday you faced an e2e error due to the fact that the the UI of the backend had an outdated version. But I do not hold a strong opinion so I'll leave that to you.

Interesting. I think of a CI pipeline as sequential (build, package, test, deploy, etc), failing immediately if any step fails. I like knowing that the thing deployed has passed all our tests, at least the ones known at the time. In the case of the failed ui tests yesterday, we'd only know about that after merging a PR in this repo, which isn't great, but at least we'd know about it eventually, in an automated way.

But if there's a majority opinion either way, I'm happy to abide.

> 2. I thought the idea was to deploy the downstream version of Trustify not the upstream one. That might be just me not getting the requirement correctly though. In any case this is a good milestone!

Easy enough to do -- now that I know how -- but right now both upstream and downstream are identical, except for the konflux config, which isn't buying us anything yet.

Personally, I like for the deployment of the latest thing merged to come out of the repo that merged it.

> 3. I created this issue [Do not delete the NS and let the user do it by himself trustify-operator#34](https://github.com/trustification/trustify-operator/issues/34) in regards to the fact that you made the trustify-operator scripts to delete the namespace. I think that is a dangerous thing to be set in the trustify-operator scripts and the action of deletion should be done by the user itself (a conscious decision). So Just like you are using `redhat-actions/oc-login@v1` to login in OCP you should delete the Namepace in a command defined in this workflow.

Yeah, I commented on that issue. Deleting the namespace was a hack that gave me a crude, idempotent behavior. And I didn't think it would affect anything you were currently using that script for, because you're always using it in a "virgin" minikube. With a long-lived OCP instance, we don't have that luxury.

> * oc login (`uses: redhat-actions/oc-login@v1`)
> * oc delete project MY_PROJECT
> * Install the oprator `uses: trustification/trustify-operator/.github/actions/install-trustify@main`

To be clear, I don't want to delete the namespace. I want the operator to install a new version of trustify into that namespace. This would significantly reduce the time the app is down. This is the way operators should work, imo, but I need help refactoring that script to achieve that behavior.

carlosthe19916: > To be clear, I don't want to delete the namespace. I want the operator to install a new version of trustify into that namespace. This would significantly reduce the time the app is down. This is the way operators should work, imo, but I need help refactoring that script to achieve that behavior.

Agree. Deleting the NS also means deleting the whole date that was before. I don't want to block your PR, it has been approved, feel free to merge it. But it will be worth investigating how to achieve what you mentioned. Not sure I have too much time to support you in that journey though, but I'll try to find some free time to also suggest some solutions.

carlosthe19916: wait, I'll mention what I mentioned somewhere in other thread:

This is how Operators work.
- Step 1: The operator is released
- Step 2: you install your operator in OCP 
- Step 3: the team releases a new version of the backend (which included the UI) every night.
- Step 4: each night the a new version of the operator is realeased
- Step 5: OLM detects there is a new operator released and automatically updates the container images used in previous release and updates the application to the latest one.

Important to note that the Operator is installed only once in STEP 2.

This is how a real application receives updates when they install an operator. E.g. Today my company installs the Keycloak Operator in my OCP instance and in 1 month the Keycloak teams releases a new version of Keycloak (also the operator). ME as a company should not move a finger to receive updates from Keycloak as my operator, through OLM, should update the Keycloak version as soon as there is a new one.

carlosthe19916: That magic happens when you install an operator and enable the AUTOMATIC upgrades monitor:

ocp-operator-install-details1

That way, OLM does the magic and you don't need to actually deploy the whole thing after each commit.

But for that to happen a requirement would be:

-  The deployment is done nightly and not after each PR
- Nightly a new version is released under a K8s channel `development` and it takes the latest version of container


That is just an idea, feel free to discard it :)

jcrossley3: That all sounds good to me, and resonates with what I recall from working with operators in the early days (pre bundles) back when they rarely worked. :wink: 

> * The deployment is done nightly and not after each PR

Why? What does the frequency matter?

PR #913, opened 2024-10-10
Add report to source_document meta

### LINKED ISSUE #896: SBOM quality reporting
Valid SBOMs can have a range of issues; missing pURLs, missing or invalid licenses, or just 'garbage' data. Ingesting such SBOMs can often lead to surprising results which we should minimally at least be able to report on.

Existing [reporting](https://github.com/trustification/trustify/blob/d7ac13374432a964b4d6140b8cbe5cfc506f223f/modules/ingestor/src/graph/sbom/spdx.rs#L306) should be improved to be able to 'score' an SBOM's quality and make explicit so end users can assess risk properly.

Towards this goal - developing a small, focused [utility](https://github.com/JimFuller-RedHat/sbomsleuth) that we can consider adding as a dep to trustify.

JimFuller-RedHat:
Generate [sbomsleuth](https://github.com/JimFuller-RedHat/sbomsleuth) report when ingesting sbom which fixes #896

We will need to discuss where this report data is exposed .. possibly on **api/v1/sbom** endpoints ?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -399,6 +399,7 @@ dependencies = [
  "cfg-if",
  "getrandom",
  "once_cell",
+ "serde",
  "version_check",
  "zerocopy",
 ]
@@ -1183,6 +1184,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bytecount"
+version = "0.6.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5ce89b21cab1437276d2650d57e971f9d548a2d9037cc231abdc0562b97498ce"
+
+[[package]]
 name = "byteorder"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


ctron:
I think the idea is good. But what I don't like is that we now have two "reports". One from sbomsleuth, the other existing report. And that we seem to have two different ways dealing with them.

I'd prefer if we could merge the sbomsleuth report and the `warnings` in a single thing. Which we collect in the importer run report, report back on the API, and store in the source document table.

JimFuller-RedHat:
sure one thing better then two ;)

The importer run report does not currently work on any sboms uploaded via rest api (or maybe I missed it?) ... I was hoping to get some direction on unifying on that axis as well.

Also - do we want to 'bake in' heuristics on reports in trustify or push that to some calling client to decide (and maintain) ? 

Sounds like we need to discuss further.

ctron:
image

ctron:
:rofl: Let's have a chat.

PR #914, opened 2024-10-10
feat: populate cpe_key in organization and product tables

dejanb:


modules/fundamental/src/product/endpoints/test.rs
```diff
@@ -30,6 +31,7 @@ async fn all_products(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
             "AMQ Broker",
             ProductInformation {
                 vendor: Some("Red Hat".to_string()),
+                cpe: None,
```
JimFuller-RedHat, line 34: maybe worthwhile to populate ?

modules/ingestor/src/graph/product/mod.rs
```diff
@@ -120,11 +120,12 @@ impl<'g> ProductContext<'g> {
 #[derive(Clone, Default, Debug)]
 pub struct ProductInformation {
     pub vendor: Option<String>,
+    pub cpe: Option<Cpe>,
 }
 
 impl ProductInformation {
     pub fn has_data(&self) -> bool {
-        self.vendor.is_some()
+        self.vendor.is_some() || self.cpe.is_some()
```
JimFuller-RedHat, line 128: I think idiomatic Rust might look like ` self.vendor.or(self.cpe).is_some()`

entity/src/product.rs
```diff
@@ -9,6 +9,7 @@ pub struct Model {
     pub id: Uuid,
     pub name: String,
     pub vendor_id: Option<Uuid>,
+    pub cpe_key: Option<String>,
```
JimFuller-RedHat, line 12: have you considered using **cpe** crate (we already have in Cargo.toml)  to also validate cpe with `Cpe::parse(cpe_str)?;` somewhere ?

modules/ingestor/src/graph/sbom/spdx.rs
```diff
@@ -157,7 +159,9 @@ impl SbomContext {
                     "cpe22Type" => match Cpe::from_str(&r.reference_locator) {
                         Ok(cpe) => {
                             refs.push(PackageReference::Cpe(cpe.uuid()));
-                            cpes.add(cpe);
+                            cpes.add(cpe.clone());
+                            // assign only the first cpe for now
```
JimFuller-RedHat, line 163: Can you expand a little bit on rationale ... is there a TODO here ?

JimFuller-RedHat (approved):
apart from failing CI, made some observations LGTM

modules/ingestor/src/graph/product/mod.rs
dejanb: This pattern is used for all similar code. Let's change them all in a separate PR?

modules/fundamental/src/product/endpoints/test.rs
dejanb: Actually cpe_key is not propagated to the endpoint at the moment. It used only internally for correlation. Maybe we should wait with introducing it into the api until we finish with todo mentioned below?

dejanb: It's being tested in 'service' btw

modules/ingestor/src/graph/sbom/spdx.rs
dejanb: Add some more context. Thanks

JimFuller-RedHat (approved):
LGTM

PR #919, opened 2024-10-12
chore(ai): endpoint improvements.

chirino:
* Move more of the resource info out of the system messages/prefix into the Tools descriptions
* Disable the product-info tool for now, it’s better to use the sbom-info for similar information.
* Bump async-openai version
* switch to new langchain-rust patch branch
* Add more info to tool response.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -694,8 +694,9 @@ dependencies = [
 
 [[package]]
 name = "async-openai"
-version = "0.24.0"
-source = "git+https://github.com/chirino/async-openai?branch=optional-fields#7b9fe4b17c9f1985a296a313adf9e641a5812163"
+version = "0.24.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c6db3286b4f52b6556ac5208fb575d035eca61a2bf40d7e75d1db2733ffc599f"
 dependencies = [
  "async-convert",
  "backoff",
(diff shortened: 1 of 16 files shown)


ctron (approved):
I think this is good in general. I noticed the use of some Red Hat specific/preferred terms (RHEL, quarkus). Maybe there's a better way dealing with this.

helio-frota:
@chirino that is sorta nit and not required :+1: ...

as you already wrote tests for this feature could you please double check with 
`cargo clippy -- -W clippy::redundant_clone` ? not required but, if you can a void a clone, that means you avoided a memory allocation ... as my noob perspective my only safe/backup/shield thing is the test when I call `cargo clippy -- -W clippy::redundant_clone` nowadays :+1:

chirino:
> @chirino that is sorta nit and not required 👍 ...
> 
> as you already wrote tests for this feature could you please double check with `cargo clippy -- -W clippy::redundant_clone` ? not required but, if you can a void a clone, that means you avoided a memory allocation ... as my noob perspective my only safe/backup/shield thing is the test when I call `cargo clippy -- -W clippy::redundant_clone` nowadays 👍

Then we should add that to `cargo xtask precommit` and ci bits so that it consistently gets enforced.

chirino:
> I think this is good in general. I noticed the use of some Red Hat specific/preferred terms (RHEL, quarkus). Maybe there's a better way dealing with this.

Could do.  Just need different example product names.

PR #920, opened 2024-10-14
Enhance query filters to handle Uuid and enable analysis graph to use filters.

JimFuller-RedHat:
Enhances query.rs to be able to perform comparisons with **Uuid** when supplied via inline query params.

Enabling the analysis graph (api/v1/analysis) to filter on **sbom_id**, **node_id** and exact **name**.
```
api/v1/analysis/root-component?q=name=A&node_id=SPDXRef-A&sbom_id=urn:uuid:019289ff-232c-7780-b0d5-22b6e5dc4be5
```
where we retain **contains** style search with:
```
api/v1/analysis/root-component?q=libcurl
```
to filter exact match use:
```
api/v1/analysis/root-component?q=name=libcurl
```

and similar for dep endpoints eg.
```
api/v1/analysis/dep?q=name=A&node_id=SPDXRef-A&sbom_id=urn:uuid:019289ff-232c-7780-b0d5-22b6e5dc4be5
api/v1/analysis/dep?q=libcurl
api/v1/analysis/dep?q=name=libcurl
```

**Note** - In real usage - the above inline _query name=values&name=value_ should always be escaped.

Most of the 'moving parts' of this feature is due to querying both an ORM and non ORM entity at the same time (sbom_node and analysis graph) ... kept most of the code changes within analysis graph.

### DIFF
--- a/common/src/db/query.rs
+++ b/common/src/db/query.rs
@@ -810,16 +810,16 @@ mod tests {
             r#"(("advisory"."location" ILIKE '%foo%') OR ("advisory"."title" ILIKE '%foo%')) AND "advisory"."location" = 'bar'"#
         );
         assert_eq!(
-            where_clause(r"m\&m's&location=f\&oo&id=13")?,
-            r#"(("advisory"."location" ILIKE E'%m&m\'s%') OR ("advisory"."title" ILIKE E'%m&m\'s%')) AND "advisory"."location" = 'f&oo' AND "advisory"."id" = 13"#
+            where_clause(r"m\&m's&location=f\&oo&id=0e840505-e29b-41d4-a716-665544004400")?,
+            r#"(("advisory"."location" ILIKE E'%m&m\'s%') OR ("advisory"."title" ILIKE E'%m&m\'s%')) AND "advisory"."location" = 'f&oo' AND "advisory"."id" = '0e840505-e29b-41d4-a716-665544004400'"#
         );
         assert_eq!(
             where_clause("a|b|c")?,
             r#"("advisory"."location" ILIKE '%a%') OR ("advisory"."title" ILIKE '%a%') OR ("advisory"."location" ILIKE '%b%') OR ("advisory"."title" ILIKE '%b%') OR ("advisory"."location" ILIKE '%c%') OR ("advisory"."title" ILIKE '%c%')"#
         );
         assert_eq!(
-            where_clause("a|b&id=1")?,
-            r#"(("advisory"."location" ILIKE '%a%') OR ("advisory"."title" ILIKE '%a%') OR ("advisory"."location" ILIKE '%b%') OR ("advisory"."title" ILIKE '%b%')) AND "advisory"."id" = 1"#
+            where_clause("a|b&id=0e840505-e29b-41d4-a716-665544004400")?,
+            r#"(("advisory"."location" ILIKE '%a%') OR ("advisory"."title" ILIKE '%a%') OR ("advisory"."location" ILIKE '%b%') OR ("advisory"."title" ILIKE '%b%')) AND "advisory"."id" = '0e840505-e29b-41d4-a716-665544004400'"#
         );
         assert_eq!(
             where_clause("a&b")?,
@@ -924,7 +924,11 @@ mod tests {
                 )
                 .unwrap()
                 .build(sea_orm::DatabaseBackend::Postgres)
-                .to_string()[45..]
+                .to_string()
+                .split("WHERE")
+                .last()
+                .expect("problem splitting string")
+                .trim()
                 .to_string();
             assert_eq!(stmt, expected);
         };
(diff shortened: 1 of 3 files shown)


ctron (commented):
LGTM, but I'll defer to @jcrossley3

common/src/db/query.rs
```diff
@@ -473,6 +474,7 @@ impl Arg {
             return Ok(Arg::Null);
         }
         Ok(match ct {
+            ColumnType::Uuid => Arg::Value(Value::from(s.parse::<Uuid>().map_err(err)?)),
```
jcrossley3, line 477: Ideally, we'd have a test for this in this file. Can you change the type for `id` from `i32` to `Uuid` in the dummy advisory model at the bottom of this file, and re-run the tests? I expect you'll need to add some quotes around some of the expected where clauses.

If that's too big of a hassle, I'm happy to do it after this PR merges.

modules/analysis/src/service.rs
```diff
@@ -273,6 +274,23 @@ pub async fn load_graphs(
...
+                    value = value.strip_prefix("urn:uuid:").unwrap_or(value);
+                }
+                query_map.insert(key.to_owned(), value.to_owned());
+            }
+        }
+    } else {
+        query_map.insert("re_name".to_owned(), query.q.clone());
+    }
+    query_map
+}
```
jcrossley3, line 292: Now I understand why you were asking how to get the parsed value out of `Query`. If we find ourselves wanting this in other places, we can refactor to hang this fn off `Query`.

Minor style nit: I don't love the "mutable hashmap plus for loop" pattern here. Smells like old java. :stuck_out_tongue_winking_eye:  Next time, consider calling `reduce` or `fold` on the `split`.

jcrossley3 (commented):
Glad you figured it out!

common/src/db/query.rs
JimFuller-RedHat: will add!

modules/analysis/src/service.rs
JimFuller-RedHat: ah ya, I made this loop explicit as I could not figure out why things were not working -  then grokked the need to strip urn:uuid: for the graph query bit ... will rewrite, thx!!

common/src/db/query.rs
```diff
@@ -810,16 +810,16 @@ mod tests {
...
         );
         assert_eq!(
             where_clause("a|b|c")?,
             r#"("advisory"."location" ILIKE '%a%') OR ("advisory"."title" ILIKE '%a%') OR ("advisory"."location" ILIKE '%b%') OR ("advisory"."title" ILIKE '%b%') OR ("advisory"."location" ILIKE '%c%') OR ("advisory"."title" ILIKE '%c%')"#
         );
         assert_eq!(
-            where_clause("a|b&id=1")?,
-            r#"(("advisory"."location" ILIKE '%a%') OR ("advisory"."title" ILIKE '%a%') OR ("advisory"."location" ILIKE '%b%') OR ("advisory"."title" ILIKE '%b%')) AND "advisory"."id" = 1"#
+            where_clause("a|b&id=0e840505-e29b-41d4-a716-665544004400")?,
+            r#"(("advisory"."location" ILIKE '%a%') OR ("advisory"."title" ILIKE '%a%') OR ("advisory"."location" ILIKE '%b%') OR ("advisory"."title" ILIKE '%b%')) AND "advisory"."id" = '0e840505-e29b-41d4-a716-665544004400'"#
```
jcrossley3, line 822: Thanks!

modules/analysis/src/service.rs
```diff
@@ -273,6 +274,23 @@ pub async fn load_graphs(
...
+                pair.split_once('=').map(|(key, value)| {
+                    let value = value.strip_prefix("urn:uuid:").unwrap_or(value);
+                    (key.to_owned(), value.to_owned())
+                })
+            })
+            .collect()
+    } else {
+        HashMap::from([("re_name".to_owned(), query.q.clone())])
+    }
+}
```
jcrossley3, line 292: I'm not sure this is more clear. :smile:

modules/analysis/src/service.rs
```diff
@@ -371,7 +390,23 @@ impl AnalysisService {
...
+                                        .map_or(true, |sbom_id| node.sbom_id.eq(sbom_id));
+                                    let matches_node_id = graph_query_map
+                                        .get("node_id")
+                                        .map_or(true, |node_id| node.node_id.eq(node_id));
+                                    let matches_name =
+                                        graph_query_map.get("name").map_or(true, |name| {
+                                            !name.is_empty() && node.name.eq(name)
+                                        });
+                                    matches_sbom_id && matches_node_id && matches_name
+                                }
```
jcrossley3, line 409: This lump of code is repeated below, which is crying out for a refactor into a single, common fn. If it were me, I'd probably rename/modify `convert_query_to_hashmap` to take both a `Query` and a `Node` and then return a `bool`, and call it here. 

I'd prolly use regex's instead of the temporary hashmap, too, so feel free to ignore me. :smile:

jcrossley3 (commented):
Did you want to fix the `graph_read_gaurd` typo while you're in there?

modules/analysis/src/service.rs
JimFuller-RedHat: this area may evolve depending on @dejanb usage ... so will leave it for now.

jcrossley3 (approved):

PR #930, opened 2024-10-18
Introduce a Constraint struct to bridge Query user input and Conditions

jcrossley3:
This refactoring clarifies the logic in `Filter::parse`, hopefully, and opens the door to facilitating access to the parsed components of the `Query`, useful in the analysis module


modules/analysis/src/service.rs
```diff
@@ -630,27 +600,13 @@ impl AnalysisService {
...
-                                    let matches_name =
-                                        graph_query_map.get("name").map_or(true, |name| {
-                                            !name.is_empty() && node.name.eq(name)
-                                        });
-                                    matches_sbom_id && matches_node_id && matches_name
-                                }
-                            } else {
-                                false // Return false if the node does not exist
-                            }
+                            graph.node_weight(i).is_some_and(|node| {
```
JimFuller-RedHat, line 603: ohh ... **is_some_and** ;)

modules/analysis/src/service.rs
```diff
@@ -274,23 +277,6 @@ pub async fn load_graphs(
     }
 }
 
-fn convert_query_to_hashmap(query: &Query) -> HashMap<String, String> {
```
JimFuller-RedHat, line 277: glad to see this go ;)

common/src/db/query.rs
```diff
@@ -422,35 +531,53 @@ impl From<Filter> for ConditionExpression {
...
+struct Sort {
+    field: ColumnRef,
+    order: Order,
+}
+
+impl Sort {
+    fn parse(s: &str, columns: &Columns) -> Result<Self, Error> {
+        let (field, order) = match s.split(':').collect::<Vec<_>>()[..] {
+            [f] => (f, String::from("asc")),
+            [f, dir] => (f, dir.to_lowercase()),
```
JimFuller-RedHat, line 546: to address my own ignorance - what is **dir** in this context ?

JimFuller-RedHat (approved):
LGTM

common/src/db/query.rs
jcrossley3: direction. sort fields can have an optional suffix of either `:asc` or `:desc`. default is `asc` if omitted.

JimFuller-RedHat: ug, yes of course ...

PR #937, opened 2024-10-23
enhance testing around cyclonedx ingestion

JimFuller-RedHat:


common/src/db/query.rs
```diff
@@ -1132,20 +1162,37 @@ mod tests {
 
     #[test(tokio::test)]
     async fn apply_to_context() -> Result<(), anyhow::Error> {
+        use time::format_description::well_known::Rfc2822;
```
helio-frota, line 1165: I'm confused, should we use Rfc3339 instead ?

JimFuller-RedHat: for different reasons - I'm confused as well as my commit/PR did not contain this - I believe this was in @jcrossley3 previous PR ... not sure what git/github is doing here as I had updated my branch before raising this PR.

As for your specific question you will have to follow up with @jcrossley3

jcrossley3: Any ol' format will do. I just needed a hardcoded date. It's only used for creating a test fixture.

PR #940, opened 2024-10-23
enhance testing around cyclonedx ingestion

JimFuller-RedHat:


modules/analysis/src/service.rs
```diff
@@ -134,7 +134,7 @@ pub async fn get_relationships(
         LEFT JOIN
             sbom_package_purl_ref t2 ON sbom.sbom_id = t2.sbom_id AND t2.node_id = package_relates_to_package.right_node_id
         WHERE
-            package_relates_to_package.relationship IN (0, 8, 14)
+            package_relates_to_package.relationship IN (0, 1, 8, 14)
```
jcrossley3, line 137: What does the added 1 mean?

jcrossley3 (approved):

modules/analysis/src/service.rs
JimFuller-RedHat: matches relationship enum

PR #941, opened 2024-10-23
refactor: break up Query behavior for better clarity

jcrossley3:
Reorganize the tests as well, though many use the same dummy `Advisory` model in the parent `tests` module.


JimFuller-RedHat (approved):
nicely done ... boring observation (that perhaps others have already mentioned in the past):  It is always tricky to know how far to go hand crafting parser(s) when something anchored (for example with EBNF) would be comprehensive/contained ... **query.rs** does a reasonable job at balancing off pros/cons  and even better if this finds its way into **sea_orm** in some far future day ... otherwise LGTM.

helio-frota (approved):
more elegant code that also helps when using mem and cpu profiling tools

jcrossley3:
> nicely done ... boring observation (that perhaps others have already mentioned in the past): It is always tricky to know how far to go hand crafting parser(s) when something anchored (for example with EBNF) would be comprehensive/contained ... **query.rs** does a reasonable job at balancing off pros/cons and even better if this finds its way into **sea_orm** in some far future day ... otherwise LGTM.

Thanks... I think! :smile: 

I've always claimed I'd be open to using a parser (i've looked at chumsky) for our query syntax when introducing it makes the code more clear. As it stands, I think the regex suffices for clarity, but we'll see how our query needs evolve.

JimFuller-RedHat:
to be clear I meant very nicely done! no sideways snide with the EBNF mention intended !

PR #942, opened 2024-10-24
chore: Enhance testing around cyclonedx ingestion

JimFuller-RedHat:
We need to ensure (when similarly encoded) that ingestion of cyclonedx matches spdx ingestion. 

We also should take care to avoid any race conditions if we encounter sbom that encode circular dependencies.

### DIFF
--- a/modules/analysis/src/service.rs
+++ b/modules/analysis/src/service.rs
@@ -17,6 +17,7 @@ use crate::model::{
     AnalysisStatus, AncNode, AncestorSummary, DepNode, DepSummary, GraphMap, PackageNode,
 };
 use crate::Error;
+use petgraph::algo::is_cyclic_directed;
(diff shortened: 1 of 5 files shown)


modules/analysis/src/service.rs
```diff
@@ -134,7 +134,7 @@ pub async fn get_relationships(
         LEFT JOIN
             sbom_package_purl_ref t2 ON sbom.sbom_id = t2.sbom_id AND t2.node_id = package_relates_to_package.right_node_id
         WHERE
-            package_relates_to_package.relationship IN (0, 8, 14)
+            package_relates_to_package.relationship IN (0, 1, 8, 14)
```
JimFuller-RedHat, line 137: 1 = DependencyOf relationship which is how we encode ingesting cyclonedx **dependsOn**

bobmcwhirter (approved):

Issue #925, opened 2024-10-16
Implement storage compression for S3 too


ctron:
We do have storage compression for the plain filesystem. In a real-life example, using zstd reduced the consumed storage from ~100GiG to 6GiB.

Unless there's a better way with S3, I think we should support the same mechanism from S3 too.

PR #944, opened 2024-10-25
feat: enable compression of docs stored on S3

jcrossley3:
Fixes: #925

### DIFF
--- a/modules/storage/src/service/compression/mod.rs
+++ b/modules/storage/src/service/compression/mod.rs
@@ -2,12 +2,13 @@ use async_compression::tokio::bufread;
 use async_compression::tokio::write::ZstdEncoder;
 use std::fmt::{Display, Formatter};
 use std::pin::Pin;
-use std::str::FromStr;
 use std::task::{Context, Poll};
 use tokio::io;
 use tokio::io::{AsyncBufRead, AsyncRead, AsyncWrite, BufReader, ReadBuf};
 
-#[derive(Copy, Clone, Eq, PartialEq, Default, Debug, strum::EnumIter, clap::ValueEnum)]
+#[derive(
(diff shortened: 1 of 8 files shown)


modules/storage/src/service/compression/mod.rs
```diff
@@ -22,6 +23,17 @@ impl Display for Compression {
     }
 }
 
+impl FromStr for Compression {
```
ctron, line 26: I think we already use `strum`, which should be able to do the same.

modules/storage/src/service/s3.rs
```diff
@@ -47,11 +60,18 @@ impl StorageBackend for S3Backend {
...
+            .compress(file.reader().await.map_err(errhdlr)?)
+            .await;
         let result = file.result();
 
         self.bucket
-            .put_object_stream(&mut source, result.key().to_string())
+            .put_object_stream_with_content_type(
+                &mut source,
+                result.key().to_string(),
+                "application/json",
```
ctron, line 73: Does that require all documents to be JSON? IIRC, that's not the case for some OSV documents, which are YAML instead.

ctron (commented):
Looks good. Two minor things to sort out.

modules/storage/src/service/compression/mod.rs
jcrossley3: Updated. Got a little carried away handling the potential `ParseError` though.

modules/storage/src/service/s3.rs
jcrossley3: Good call. I realized we don't even use the content-type, so the default will do. Besides, I don't think we want to ripple a content-type argument through our existing storage api.

ctron (approved):

PR #947, opened 2024-10-30
Upgrade dependencies, most important to utoipa 5, dropping a bunch of workarounds

ctron:
This updates a few dependencies. Most notably to utoipa 5, which now supports nesting of specs, and generic type arguments. Allowing us to drop some workarounds we had in place. It also auto-collects schemas now.


ctron:
@chirino @carlosthe19916 not sure this has an impact in the UI side of things.

carlosthe19916:
I can see the openapi.yaml will change a bit. I don't see too many changes though (it is something I can manage). I think it should not be a blocker for the UI. So those changes should be fine

common/auth/src/swagger_ui.rs
```diff
@@ -85,10 +86,10 @@ impl SwaggerUiOidc {
...
 
             oauth2.extensions = Some({
-                let mut map = HashMap::new();
-                map.insert("x-client-id".into(), self.client_id.clone().into());
-                map.insert("x-default-scopes".into(), "openid".into());
-                map
+                Extensions::builder()
+                    .add("x-client-id", self.client_id.clone())
+                    .add("x-default-scopes", "openid")
+                    .build()
```
jcrossley3, line 92: clean!

modules/fundamental/src/advisory/endpoints/mod.rs
```diff
@@ -18,21 +20,27 @@ use trustify_common::{
...
+    upload_limit: usize,
+) {
+    let advisory_service = AdvisoryService::new(db.clone());
+    let purl_service = PurlService::new(db);
 
     config.service(
-        web::scope(CONTEXT_PATH)
+        utoipa_actix_web::scope(CONTEXT_PATH)
             .app_data(web::Data::new(advisory_service))
+            .app_data(web::Data::new(purl_service))
```
jcrossley3, line 43: I see that this is used in the `delete` fn, but if it wasn't there before, how did it work?

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -26,32 +29,45 @@ use trustify_common::{
...
-        .service(label::update);
+pub const CONTEXT_PATH: &str = "/v1/sbom";
+
+pub fn configure(
+    config: &mut utoipa_actix_web::service_config::ServiceConfig,
+    db: Database,
+    upload_limit: usize,
+) {
+    let sbom_service = SbomService::new(db.clone());
+    let purl_service = PurlService::new(db);
```
jcrossley3, line 51: same question

modules/analysis/src/endpoints.rs
```diff
@@ -28,35 +33,25 @@ pub fn configure(config: &mut web::ServiceConfig, db: Database) {
...
-        crate::model::AncNode,
-        crate::model::DepSummary,
-        crate::model::DepNode,
-    )),
     tags()
 )]
 pub struct ApiDoc;
 
 #[utoipa::path(
-    context_path = "/api",
```
jcrossley3, line 44: I like that this is gone.

modules/analysis/src/endpoints.rs
```diff
@@ -28,35 +33,25 @@ pub fn configure(config: &mut web::ServiceConfig, db: Database) {
...
     tag = "analysis",
     operation_id = "status",
     responses(
         (status = 200, description = "Analysis status.", body = AnalysisStatus),
     ),
 )]
-#[get("/v1/analysis/status")]
+#[get("/status")]
```
jcrossley3, line 47: Personally, I like seeing the components of the path most relevant to the fn right next to the fn. In this case, I'd prefer `#[get("/analysis/status")]`, I think. I could live with the `/v1` in there, too, but we could defer that until we have a `/v2`.

My only reason is that it makes it easy to construct the URL for curl, etc rather than having to look up the `CONTEXT_PATH` const, so it's a minor nit if you feel strongly otherwise.

ctron:
@jcrossley3 I'll work through the comments. In the meantime, I pushed some more changes. Which gets rid of the manually collected openapi spec altogether and uses the actually registered handlers to construct it.

modules/fundamental/src/advisory/endpoints/mod.rs
ctron: It worked because we didn't scope the data. Now the instance is added to the scope only.

modules/fundamental/src/sbom/endpoints/mod.rs
ctron: Same answer. It wasn't scoped before.

modules/analysis/src/endpoints.rs
ctron: It made live a bit easier before the most recent change (of dropping the manual collection altogether). Now that we no longer need to hand-craft the OpenAPI struct, I can try to revert this. Though I am not sure how this affect the spec generation. But I'll find out.

ctron: Ok, I reverted that part. I think it should work just fine now.

jcrossley3 (approved):

PR #948, opened 2024-10-30
feat: csaf correlation - correlate vulnerability to purls and sboms

dejanb:
This PR should be able to correlate VEX advisories from CSAF files that don't have product identifier helpers (purls) to correct sboms.

To test it, run a local instance of trustify

Ingest one of the [datasets](https://github.com/trustification/trustify/blob/main/etc/datasets/README.md) (ds1 and ds3 should both work)

`http POST localhost:8080/api/v1/dataset @etc/datasets/ds3.zip`

And inspect results for some of the documents, e.g. CVE-2023-0044 and quarkus-sbom

`http GET localhost:8080/api/v1/vulnerability/CVE-2023-0044 | jq .advisories`

(find uuid of quarkus-sbom and the provide it below)

`http GET http://localhost:8080/api/v1/sbom/urn%3Auuid%3A01931b9e-8c5f-7681-8343-c8265a1257f1/advisory
`
This PR doesn't link found components to actual SBOM purls. That will be done in a separate effort.

### DIFF
--- a/entity/src/product_status.rs
+++ b/entity/src/product_status.rs
@@ -8,7 +8,7 @@ pub struct Model {
     pub advisory_id: Uuid,
     pub vulnerability_id: String,
     pub status_id: Uuid,
-    pub component: Option<String>,
+    pub package: Option<String>,
     pub product_version_range_id: Uuid,
     pub context_cpe_id: Option<Uuid>,
 }
--- a/entity/src/product_status.rs
+++ b/entity/src/product_status.rs
@@ -8,7 +8,7 @@ pub struct Model {
     pub advisory_id: Uuid,
     pub vulnerability_id: String,
     pub status_id: Uuid,
-    pub base_purl_id: Option<Uuid>,
+    pub component: Option<String>,
     pub product_version_range_id: Uuid,
     pub context_cpe_id: Option<Uuid>,
 }
@@ -20,12 +20,6 @@ pub enum Relation {
         to = "super::product_version_range::Column::Id"
     )]
     ProductVersionRange,
-
-    #[sea_orm(belongs_to = "super::base_purl::Entity",
-        from = "Column::BasePurlId"
-        to = "super::base_purl::Column::Id"
-    )]
-    BasePurl,
 
     #[sea_orm(belongs_to = "super::vulnerability::Entity",
         from = "Column::VulnerabilityId"
@@ -61,12 +55,6 @@ pub enum Relation {
 impl Related<super::product_version_range::Entity> for Entity {
     fn to() -> RelationDef {
         Relation::ProductVersionRange.def()
-    }
-}
-
-impl Related<super::base_purl::Entity> for Entity {
-    fn to() -> RelationDef {
-        Relation::BasePurl.def()
     }
 }
 
--- a/migration/src/m0000641_update_product_status.rs
+++ b/migration/src/m0000641_update_product_status.rs
@@ -11,16 +11,42 @@ impl MigrationTrait for Migration {
             .alter_table(
                 Table::alter()
                     .table(ProductStatus::Table)
-                    .drop_column(ProductStatus::BasePurlId)
+                    .add_column(ColumnDef::new(ProductStatus::Component).string())
                     .to_owned(),
             )
+            .await?;
+
+        // Transform data: populate `component` from `base_purl`
+        let update_sql = r#"
+            UPDATE product_status
+            SET component =
+                (SELECT namespace || '/' || name FROM base_purl WHERE base_purl.id = product_status.base_purl_id)
+            WHERE base_purl_id IS NOT NULL;
+        "#;
+        manager
+            .get_connection()
+            .execute_unprepared(update_sql)
+            .await?;
+
+        // Delete original `purl_status` entries now that data has been migrated
+        let delete_statuses_sql = r#"
+            DELETE FROM purl_status
+            WHERE base_purl_id IN (
+                SELECT id FROM base_purl
+                WHERE id IN (SELECT DISTINCT base_purl_id FROM product_status WHERE base_purl_id IS NOT NULL)
+                AND TYPE = 'generic'
+            )
+        "#;
+        manager
+            .get_connection()
+            .execute_unprepared(delete_statuses_sql)
             .await?;
 
         manager
             .alter_table(
                 Table::alter()
                     .table(ProductStatus::Table)
-                    .add_column(ColumnDef::new(ProductStatus::Component).string())
+                    .drop_column(ProductStatus::BasePurlId)
                     .to_owned(),
             )
             .await?;
@@ -33,15 +59,6 @@ impl MigrationTrait for Migration {
             .alter_table(
                 Table::alter()
                     .table(ProductStatus::Table)
-                    .drop_column(ProductStatus::Component)
-                    .to_owned(),
-            )
-            .await?;
-
-        manager
-            .alter_table(
-                Table::alter()
-                    .table(ProductStatus::Table)
                     .add_column(ColumnDef::new(ProductStatus::BasePurlId).uuid())
                     .add_foreign_key(
                         TableForeignKey::new()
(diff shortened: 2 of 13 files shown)


dejanb:
First stab at finding sboms and components affected by advisory when purls are not available.

@JimFuller-RedHat I'll try next to use graph analysis to try to search for purl when appropriate information is available

entity/src/product_status.rs
```diff
@@ -8,7 +8,7 @@ pub struct Model {
     pub advisory_id: Uuid,
     pub vulnerability_id: String,
     pub status_id: Uuid,
-    pub base_purl_id: Option<Uuid>,
+    pub component: Option<String>,
```
jcrossley3, line 11: I like the name much better

migration/src/m0000641_update_product_status.rs
```diff
@@ -0,0 +1,72 @@
...
+                    .drop_column(ProductStatus::BasePurlId)
+                    .to_owned(),
+            )
+            .await?;
+
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(ProductStatus::Table)
+                    .add_column(ColumnDef::new(ProductStatus::Component).string())
```
jcrossley3, line 23: Can we migrate existing data as well? Or is there no correlation between BasePurl's and Components?

modules/fundamental/src/sbom/model/details.rs
```diff
@@ -74,6 +80,35 @@ impl SbomDetails {
...
+            .join(JoinType::Join, product_status::Relation::Advisory.def())
+            .distinct_on([
+                (product_status::Entity, product_status::Column::ContextCpeId),
+                (product_status::Entity, product_status::Column::StatusId),
+                (product_status::Entity, product_status::Column::Component),
+                (
+                    product_status::Entity,
+                    product_status::Column::VulnerabilityId,
+                ),
+            ])
```
jcrossley3, line 101: All these joins feel like a smell to me, but I got no better suggestion.

jcrossley3 (commented):
I don't feel qualified to review some of the logic -- it's a big PR.

I'm a tad concerned about terminology. That the UX refers to "packages" and the backend to "purls" makes me nervous about introducing "component". Ideally, we could agree on a single term.

carlosthe19916:
I can see data for vulnerabilities that is great!!!!!!!!!!

image


The only thing I noticed is that if I hit http://localhost:8080/api/v1/sbom/urn:uuid:019320e4-4b70-7a90-a81b-8003340dc7bf/advisory then I can get packages without purl but only with name e.g.

```shell
curl http://localhost:8080/api/v1/sbom/urn:uuid:019320e4-4b70-7a90-a81b-8003340dc7bf/advisory | jq
```

```json
[
    {
        "uuid": "urn:uuid:e07bc26b-0445-454c-b621-5ba663529bd7",
        "identifier": "https://www.redhat.com/#CVE-2023-1664",
        "issuer": {
            "id": "91267f8b-91c9-4a50-99e5-73bd1140fc3e",
            "name": "Red Hat Product Security",
            "cpe_key": null,
            "website": null
        },
        "published": "2023-03-27T00:00:00Z",
        "modified": "2023-11-14T17:39:21Z",
        "withdrawn": null,
        "title": "keycloak: Untrusted Certificate Validation",
        "labels": {
            "type": "csaf",
            "datasetFile": "csaf/2023/cve-2023-1664.json"
        },
        "status": [
            {
                "vulnerability_id": "CVE-2023-1664",
                "status": "affected",
                "context": {
                    "cpe": "cpe:/a:redhat:quarkus:2:*:*:*"
                },
                "packages": [
                    {
                        "id": "",
                        "name": "org.keycloak/keycloak-core",
                        "version": null,
                        "purl": [],
                        "cpe": []
                    }
                ]
            }
        ]
    },
```

I cut the previous JSON to only see the relevant data.
Having a package only with name and without a purl intended? Currently, at least on the UI, there must be a purl for it to be rendered otherwise it won't be rendered.

dejanb:
> I cut the previous JSON to only see the relevant data. Having a package only with name and without a purl intended? Currently, at least on the UI, there must be a purl for it to be rendered otherwise it won't be rendered.

@carlosthe19916  I wanted to mentioned this earlier today, but forgot :) 

Find the purl based on the component name is the next step and I filled the separate issue for it #994. But in general we can run into the case where we can't find it and I think we should display component name only in that case.

modules/fundamental/src/sbom/model/details.rs
dejanb: If you look existing search for purls, it's the similar join-armargedon :)

https://github.com/trustification/trustify/blob/main/modules/fundamental/src/sbom/model/details.rs#L42

There's A LOT of room for improvement, but we need to start somewhere.

migration/src/m0000641_update_product_status.rs
dejanb: I was thinking about empty database and reimport documents. Let me see if I can come up with a way to avoid it.

entity/src/product_status.rs
JimFuller-RedHat: I prefer component as packages are overloaded - though good as long as they are the same wherever (package or component).

migration/src/m0000641_update_product_status.rs
```diff
@@ -0,0 +1,130 @@
...
+            .execute_unprepared(update_sql)
+            .await?;
+
+        // Delete original `purl_status` entries now that data has been migrated
+        let delete_statuses_sql = r#"
+            DELETE FROM purl_status
+            WHERE base_purl_id IN (
+                SELECT id FROM base_purl
+                WHERE id IN (SELECT DISTINCT base_purl_id FROM product_status WHERE base_purl_id IS NOT NULL)
+                AND TYPE = 'generic'
```
JimFuller-RedHat, line 37: for my own education ... why the TYPE='generic' filter is relevant ?

modules/fundamental/src/vulnerability/model/details/vulnerability_advisory.rs
```diff
@@ -476,6 +650,46 @@ impl VulnerabilitySbomStatus {
...
+                            Some(cpe) => {
+                                // This is where the matching logic for multiple advisory entries lives
+                                // At the moment we try to use the biggest range we have.
+                                // For example: if we have entries for 2.0, 2.7 and 2.13 we will choose the last one
+                                if let Some(version) = &product.cpe.version {
+                                    let version_semver = lenient_semver::parse(version);
+                                    let current_version = &cpe.version().clone().to_string();
+                                    let current_semver = lenient_semver::parse(current_version);
+
+                                    if match (&version_semver, &current_semver) {
```
JimFuller-RedHat, line 678: Have you considered using existing version related functions (e.g. psql funcs) so we do not introduce a 2nd code path (using **lenient_semver**) - I believe all those functions can be used directly in rust/sea_orm (or maybe even better performing in the query itself.

modules/fundamental/src/vulnerability/service/test.rs
```diff
@@ -207,11 +207,35 @@ async fn commons_compress(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
...
 async fn product_statuses(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
-    let service = VulnerabilityService::new(ctx.db.clone());
+    let vuln_service = VulnerabilityService::new(ctx.db.clone());
+    let sbom_service = SbomService::new(ctx.db.clone());
 
-    ctx.ingest_documents(["csaf/cve-2023-0044.json"]).await?;
+    let ingest_results = ctx
+        .ingest_documents([
+            "csaf/cve-2023-0044.json",
+            "quarkus/v2/quarkus-bom-2.13.8.Final-redhat-00004.json",
```
JimFuller-RedHat, line 216: Considering how much variation there is with sboms ... might be worthwhile to test a few different ones (quarkus is pretty opinionated).

migration/src/m0000641_update_product_status.rs
dejanb: This is how previous logic stored those "fake" purls, so basically `io.quarkus/quarkus-vertx-http` was stored as `generic://io.quarkus/quarkus-vertx-http`. So we need to clean those up

modules/fundamental/src/vulnerability/model/details/vulnerability_advisory.rs
dejanb: I did, but I was pushing what I was able to make work at this point with pure sql. I would leave that for the future improvement, I would love to have all this incorporated in the original query

modules/fundamental/src/vulnerability/service/test.rs
dejanb: Yes, that's where #986 should come to play

entity/src/product_status.rs
dejanb: I changed them to package. I like the term component better as well, but let's discuss this on a whole code base level once we are ready to change that

migration/src/m0000641_update_product_status.rs
dejanb: I added data migration logic as well

modules/fundamental/src/vulnerability/service/test.rs
jcrossley3: If it's at all possible to test the variations in a unit test, i'd prefer that to #986

migration/src/m0000641_update_product_status.rs
```diff
@@ -0,0 +1,130 @@
...
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(ProductStatus::Table)
+                    .add_column(ColumnDef::new(ProductStatus::Package).string())
+                    .to_owned(),
+            )
+            .await?;
+
+        // Transform data: populate `package` from `base_purl`
```
jcrossley3, line 19: Thanks for going to the effort -- seems a tad gnarly. I confirmed the migration of my local data worked fine.

jcrossley3 (approved):

modules/fundamental/src/vulnerability/model/details/vulnerability_advisory.rs
JimFuller-RedHat: sounds good -  might be worth inserting a TODO here.

modules/fundamental/src/vulnerability/service/test.rs
dejanb: Let's definitely add more tests here as we go along and find the cases that we want to cover while improving the matching logic

modules/fundamental/src/vulnerability/model/details/vulnerability_advisory.rs
dejanb: I'll add it, but it definitely won't be forgotten :)

migration/src/m0000641_update_product_status.rs
dejanb: It is :/ Thanks for checking it

PR #965, opened 2024-11-04
Refactor sea_orm-dependent logic out of query and into filtering

jcrossley3:
Also added a Display impl for Columns to assist in debugging.

### DIFF
--- a/common/src/db/query/filtering.rs
+++ b/common/src/db/query/filtering.rs
@@ -1,6 +1,6 @@
 // SeaORM impls
 
-use super::{Columns, Error, IntoColumns, Query};
+use super::{Columns, Error, IntoColumns, Query, Sort};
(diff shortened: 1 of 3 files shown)


common/src/db/query/filtering.rs
```diff
@@ -16,7 +16,22 @@ pub trait Filtering<T: EntityTrait> {
     where
         Self: Sized + QueryFilter + QueryOrder,
     {
-        search.query(self, &context.columns())
+        let Query { ref q, ref sort } = search;
```
helio-frota, line 19: interesting... 
https://doc.rust-lang.org/std/keyword.ref.html

`Using the ref keyword, the value is only borrowed, not moved, making it available for use after the match statement:`

I already saw clippy asking to avoid match statements containing only one arm, there is a direct correlation with if and else then.

helio-frota (approved):

Issue #976, opened 2024-11-06
CVEs Related Advisories missing props for v1 parity


kahboom:
In v1 CVE detail page you have the following properties visible for its related advisories, which are not available now:

- `Revision`
- `Vulnerabilities`
- No `aggregate severity` showing up in column value despite `severity` existing? This might just be a "not enough data" issue on my part

See also [this](https://docs.google.com/spreadsheets/d/1mhBd3-UYvSANIXwRPDrCdGplN7cq5rQPJhIywo5mcDY/edit?gid=1088076328#gid=1088076328) spreadsheet for a complete list of properties missing in the current version of the UI.

![Screenshot 2024-11-06 at 2 04 54 PM copy](https://github.com/user-attachments/assets/31672464-40f7-46cb-a77f-21a71cbaa223)

ctron:
I think that "Aggregated severity" is already available:

![image](https://github.com/user-attachments/assets/42c0bac5-be99-4baf-9fba-80709e4f6757)

It's just the wrong column header. Because that's a severity, not a score.

ctron:
Taking a look at the "revision", I think it's just the advisory's date. So you should be able to use `modified`.

PR #978, opened 2024-11-06
fix: add missing "number_of_vulnerabilities" field

ctron:
Closes https://github.com/trustification/trustify/issues/976

cc @kahboom

### DIFF
--- a/entity/src/advisory_vulnerability.rs
+++ b/entity/src/advisory_vulnerability.rs
@@ -1,5 +1,5 @@
 use crate::{advisory, cvss3, purl_status, vulnerability};
-use sea_orm::entity::prelude::*;
+use sea_orm::{entity::prelude::*, LinkDef};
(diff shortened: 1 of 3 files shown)


JimFuller-RedHat (approved):
this is a bit of a 'lift' just to get number of vulns ... that being said I could not figure out a more elegant way to get there as you have here ;) ... there maybe opportunities for enhancing performance though this LGTM and we can get it into branch for further testing.

ctron:
> this is a bit of a 'lift' just to get number of vulns ... that being said I could not figure out a more elegant way to get there as you have here ;) ... there maybe opportunities for enhancing performance though this LGTM and we can get it into branch for further testing.

I agree there should be more elegant way, instead of performing this in the for loop. Then again, there are bunch of similar patterns following the next lines in that for loop. Just adding to that :D

PR #981, opened 2024-11-07
chore: enforce permissions

ctron:


bobmcwhirter:
LGTM
narrator: he didn't look

ctron:
> LGTM narrator: he didn't look

At least you commented. I appreciate that. 😆

JimFuller-RedHat (approved):
The pedant in me feels the need to point out that this is a coarse grained perms on REST API eg. that is likely not to easily evolve past  securing CRUD on endpoints (ex. securing a single vulnerability, advisory, etc). From an infra pov - we could alternately have nginx or api gateway enforce such things. All that being said - LGTM as designed (just making explicit).

ctron:
> The pedant in me feels the need to point out that this is a coarse grained perms on REST API eg. that is likely not to easily evolve past securing CRUD on endpoints (ex. securing a single vulnerability, advisory, etc). From an infra pov - we could alternately have nginx or api gateway enforce such things. All that being said - LGTM as designed (just making explicit).

I agree, it's the best we can do today. I think, moving forward, we need much finer grained permissions. And ACL/RBAC/…

PR #995, opened 2024-11-12
chore: removes some redundant clone()

helio-frota:


jcrossley3 (approved):
Surprised clippy didn't find these

helio-frota:
yeah, for this we need `cargo clippy -- -W clippy::redundant_clone` more specifically

PR #996, opened 2024-11-13
Add version and handle circular deps in analysis graphs

JimFuller-RedHat:
This PR contains a few interconnected changes to analysis graph:

* add **version** to node
* enable filtering with **version**
* ensure we handle circular dependencies (ex. _etc/test-data/spdx-ubi-examples/v2.json_)
*  heavy refactoring of graph routines

Additionally we should no longer extract name from the component **pURL** - as often a component will  not have a pURL ! - so we retrieve from **sbom_node**. 

**Note**:  We can further refactor in retrieve_ fns but left that till whenever we contemplate doing a MRU style forest of graphs.

### DIFF
--- a/modules/analysis/src/service.rs
+++ b/modules/analysis/src/service.rs
@@ -3,7 +3,7 @@ use sea_orm::{
     prelude::ConnectionTrait, ColumnTrait, DbErr, EntityOrSelect, EntityTrait, QueryFilter,
     QueryOrder, QueryResult, QuerySelect, QueryTrait, Statement,
 };
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use tracing::instrument;
 use trustify_common::{
     db::{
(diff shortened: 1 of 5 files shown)


modules/analysis/src/service.rs
```diff
@@ -723,7 +814,6 @@ impl AnalysisService {
                             .filter(|&i| {
                                 if let Some(node) = graph.node_weight(i) {
                                     node.name.eq(&component_name.to_string())
-                                // Use eq for exact match
                                 } else {
                                     false // Return false if the node does not exist
                                 }
```
jcrossley3, line 819: I see this pattern in a few places in this file. Can you convert to something like this, please?
```
.filter(|&i| graph.node_weight(i).is_some_and(|node| node.name.eq(&component_name)))
```

JimFuller-RedHat: thx (great idea!)- note this PR is still in draft (and undergoing change) though feel free to drip feed any thoughts.

ctron (approved):

PR #998, opened 2024-11-13
chore: add 'ci-ai' job

mrizzi:
Adding a CI workflow dedicated to the AI code to be executed only when and if changes are done to the AI module, i.e. within `modules/fundamental/src/ai/`, to avoid slowing down the "main" [`ci`](https://github.com/trustification/trustify/blob/main/.github/workflows/ci.yaml) job with unnecessary AI tests.

Obviously I'm happy to remove the path filter if these tests are considered worth being executed for every PR, independently from the changes in the PR itself.

### DIFF
--- a/.github/workflows/ci-ai.yaml
+++ b/.github/workflows/ci-ai.yaml
@@ -1,0 +1,51 @@
+name: ci-ai
+
+on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


chirino (approved):
LGTM

PR #1005, opened 2024-11-14
feat: provide scores in sbom details response

### LINKED ISSUE #952: Dashboard Data - Last 10 SBOMs Barchart
The image below is how V1 looks like in regards of the Dashboard. In the left center part there is BarChart that renders the last 10 SBOMs ingested in the system. Each bar renders the number of vulnerabilities that each SBOM has

image-2024-10-23-11-43-39-294

Here are some things I'd like to have (we can separate each task on its own issue if that works too):

- How to get the last 10 SBOMs  "Ingested"? Currently I can sort SBOMs by `published` but as far as I understand that will sort by the date in  which the SBOM was published which does not necessarily match the date in which it has been ingested into the system.
  - So I can not use `/api/v1/sbom?sort=published:asc` I need something else.
- How to get Vulnerabilities for all 10 SBOMs?
  - Here is the problem I have: If the UI wants to determine vulnerabilities for those 10 SBOMs then I would need to follow these steps:
  - First: Get the last SBOMs with something like `/api/v1/sbom?limit=10&offset=0&sort=published:asc`
  - Second: For each SBOM fetch the advisories available using `/api/v1/sbom/{id}/advisory`. So this endpoint is called 10 times. This endpoint returns a list of advisories associated to each SBOM; each advisory will have a list of Vulnerability IDs (it can be from 0 to hundreds)
  - Third. For each Vulnerability identified in the previous step I need to fetch the vulnerability itself to know its severify and appropriately group them and make sure it is able to be rendered in the BarChart. If we identified hundred of vulnerabilities in the previous step then we will call `/api/v1/vulnerability/{id}` hundred times.

The final math of calls for `/api/v1/vulnerability/{id}` will be = `10SBOM multiplied by #Advisories multiplied by #Vulnerabilities`. This can easily explode from hundreds to thousands. This is not going to scale.

dejanb:
Fixes #952

### DIFF
--- a/cvss/src/cvss3/score.rs
+++ b/cvss/src/cvss3/score.rs
@@ -1,5 +1,7 @@
 use crate::cvss3::severity::Severity;
 
+use super::Cvss3Base;
+
 /// CVSS V3.1 scores.
 ///
 /// Formula described in CVSS v3.1 Specification: Section 5:
@@ -64,5 +66,23 @@ impl From<Score> for f64 {
 impl From<Score> for Severity {
     fn from(score: Score) -> Severity {
         score.severity()
+    }
+}
+
+impl FromIterator<Cvss3Base> for Score {
+    fn from_iter<I: IntoIterator<Item = Cvss3Base>>(iter: I) -> Self {
+        let mut count: usize = 0;
+        let mut sum = 0.0;
+
+        for v in iter {
+            sum += v.score().value();
+            count += 1;
+        }
+
+        if count > 0 {
+            Self::new(sum / (count as f64))
+        } else {
+            Self::default()
+        }
     }
 }
--- a/cvss/src/cvss3/score.rs
+++ b/cvss/src/cvss3/score.rs
@@ -37,16 +37,12 @@ impl Score {
 
     /// Convert the numeric score into a `Severity`
     pub fn severity(self) -> Severity {
-        if self.0 < 0.1 {
-            Severity::None
-        } else if self.0 < 4.0 {
-            Severity::Low
-        } else if self.0 < 7.0 {
-            Severity::Medium
-        } else if self.0 < 9.0 {
-            Severity::High
-        } else {
-            Severity::Critical
+        match self.0 {
+            x if x < 0.1 => Severity::None,
+            x if x < 4.0 => Severity::Low,
+            x if x < 7.0 => Severity::Medium,
+            x if x < 9.0 => Severity::High,
+            _ => Severity::Critical,
         }
     }
 }
@@ -73,12 +69,10 @@ impl FromIterator<Cvss3Base> for Score {
     fn from_iter<I: IntoIterator<Item = Cvss3Base>>(iter: I) -> Self {
         let mut count: usize = 0;
         let mut sum = 0.0;
-
         for v in iter {
             sum += v.score().value();
             count += 1;
         }
-
         if count > 0 {
             Self::new(sum / (count as f64))
         } else {
--- a/entity/src/cvss3.rs
+++ b/entity/src/cvss3.rs
@@ -28,8 +28,8 @@ pub struct Model {
     pub severity: Severity,
 }
 
-impl From<Model> for cvss3::Cvss3Base {
-    fn from(value: Model) -> Self {
+impl From<&Model> for cvss3::Cvss3Base {
+    fn from(value: &Model) -> Self {
         Self {
             minor_version: value.minor_version as u8,
             av: value.av.into(),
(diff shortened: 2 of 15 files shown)


carlosthe19916 (changes requested):
@dejanb thanks for putting this PR in place.

Generally speaking this is exactly what I need so I would be happy to have this PR merged but here are some things I've just notice that we might need to address before:

- I've uploaded the files from D3 https://github.com/trustification/trustify/tree/main/etc/datasets/ds3
- Then try to hit the endpoint `http://localhost:8080/api/v1/sbom/{id}/advisory` and the response is huge. The one I got has 40Mb of size which I think is too much.
  - I guess the size of the response could be explained because the response includes `status.vulnerability.advisories` where `advisories` is the field that weights the most. From the top of my mind I would suggest removing `status.vulnerability.advisories` so the response is not that huge but I am not sure if that is fine with you.

dejanb:
Yeah, as mentioned in the issue, I think we should actually return [VulnerabilitySummary](https://github.com/trustification/trustify/blob/main/modules/fundamental/src/vulnerability/model/summary.rs#L17) here, but that needs a bit more changes to fetch for a single vulnerability. It should be faster as well as it will not reach database to fetch unnecessary data.

jcrossley3:
I took a swipe at it, using `VulnerabilityHead` instead of `VulnerabilityDetails` and adding a `severity` field. 

Can you give it a shot @carlosthe19916 and see if we're any closer to what you need? Hoping to add some unit tests tomorrow.

carlosthe19916:
@jcrossley3 Thank you very much!!!!!
This reduced the number of call drastically and I was able to generate the Dashboard using my WIP draft PR https://github.com/trustification/trustify-ui/pull/232

image

- From the pure point of view "It Works". I think YESSS it works your changes gives me back what I need. We can call it DONE. Thanks.
- From the point of long term API:
  - If I hit `/api/v1/sbom/{id}/advisory` then the response is similar to:

```json
[
    {        
        "status": [
            {
                "vulnerability": {
                    "normative": true,
                    "identifier": "CVE-2023-20860",
                    "title": null,
                    "description": null,
                    "reserved": null,
                    "published": null,
                    "modified": null,
                    "withdrawn": null,
                    "discovered": null,
                    "released": null,
                    "cwes": []
                },
                "severity": "high",
                "status": "not_affected",                
            },
```

You can see that the field `severity` is outside of the `vulnerability` object. However, if I hit `/api/v1/vulnerability/{id}` then the severity is in a field named `average_severity` and it is inside the object itself (a vulnerability).

```json
{
  "title": "string", 
  "average_severity": null,
// other fields
}
```

I assume `average` and `average_severity` are the same in this case.

In the long term, the APIs always grow, and having a strategy for having consistency and avoiding fields to be spread in different places and with different names would be necessary for the future. 

For the time being, I am happy to accept this PR as it is because it will allow me to have the Dashboard being implemented! Thank you!

jcrossley3:
> @jcrossley3 Thank you very much!!!!! This reduced the number of call drastically and I was able to generate the Dashboard using my WIP draft PR [trustification/trustify-ui#232](https://github.com/trustification/trustify-ui/pull/232)

Good to hear! :smile: 

> You can see that the field `severity` is outside of the `vulnerability` object. However, if I hit `/api/v1/vulnerability/{id}` then the severity is in a field named `average_severity` and it is inside the object itself (a vulnerability).

I can fix that. I'll flatten all the fields and rename `severity` to `average_severity`.

> For the time being, I am happy to accept this PR as it is because it will allow me to have the Dashboard being implemented! Thank you!

Part of the reason it took me so long to grok this part of the code is its lack of unit tests, so I'd like to fill some in there and clean up the code a bit. Hoping to have that in today, if that's not too long for you to wait for the merge.

carlosthe19916:
I can wait surely. Thanks @jcrossley3 !

jcrossley3:
@carlosthe19916 I refactored things a bit and flattened the structure per your request -- lemme know if it needs further tweaking

@dejanb I would've requested a review from you but since you originated the PR, I can't :shrug:

cvss/src/cvss3/score.rs
```diff
@@ -35,16 +37,12 @@ impl Score {
...
-            Severity::None
-        } else if self.0 < 4.0 {
-            Severity::Low
-        } else if self.0 < 7.0 {
-            Severity::Medium
-        } else if self.0 < 9.0 {
-            Severity::High
-        } else {
-            Severity::Critical
+        match self.0 {
```
ctron, line 40: :heart:

ctron (approved):
Love all the improvements!

dejanb (commented):
It looks good to me. The UI is currently broken as sbom/vulnerability page still expects the old response. Maybe we should wait with merging until UI catches up? @carlosthe19916

carlosthe19916:
@jcrossley3 Thanks, I've tested it and it works as expected. Whenever you feel it is the time let's merge this PR.

carlosthe19916 (approved):

Issue #999, opened 2024-11-13
Encountering deadlocks when uploading advisories


JimFuller-RedHat:
on latest master - after uploading (via ux) etc/test-data/csaf/*  I see the following errors

```
2024-11-13T15:39:16.521222Z ERROR ThreadId(52) trustify_module_ingestor::service::advisory::csaf::loader: error=Query Error: error returned from database: deadlock detected
2024-11-13T15:39:16.523532Z ERROR ThreadId(52) trustify_module_ingestor::service: error=Query Error: error returned from database: deadlock detected
```

accompanied by some slow warnings:

```
2024-11-13T15:42:43.699888Z  WARN ThreadId(55) sqlx::query: slow statement: execution time exceeded alert threshold summary="UPDATE \"organization\" SET \"cpe_key\" …" db.statement="\n\nUPDATE\n  \"organization\"\nSET\n  \"cpe_key\" = $1,\n  \"website\" = $2\nWHERE\n  \"organization\".\"id\" = $3\nRETURNING\n  \"id\",\n  \"name\",\n  \"cpe_key\",\n  \"website\"\n" rows_affected=0 rows_returned=1 elapsed=12.21970387s elapsed_secs=12.21970387 slow_threshold=1s
```

helio-frota:
good catch I saw the same https://gist.github.com/helio-frota/85d7fa081f195a05caeaf59442811cd8

I thought this [clone](https://github.com/trustification/trustify/pull/995/files#diff-a17b5f36604df0d4a36a89bb00935dbf1064cb9771bf28f90efe815e4625e974L19) was hiding a bug after I removed it. But I added it back and saw [2 deadlocks](https://gist.github.com/helio-frota/94b775236750fe28f6b5494238e45be1)

helio-frota:
I found the point when the message changes:

* Query Error: error returned from database: duplicate key value violates unique constraint "package_pkey"
commit 7f57a812714e6210eccd4f23fa9416be95addc3f

* Query Error: error returned from database: deadlock detected
commit 47be91437abf51f0c6a38e1b1c69df1fa5f1f497

helio-frota:
> I found the point when/where the message changes:

This statement ^ is wrong :+1: 
It also depends on input.

the test pass with:

```rust
let files = vec![
    "csaf/cve-2023-0044.json",
    "csaf/rhsa-2023_5835.json",
    "csaf/rhsa-2024_2776.json",
    "csaf/CVE-2023-20862.json",
[...]

PR #1009, opened 2024-11-15
test: this will break the ci same as #999

helio-frota:


helio-frota:
I'll close this in the 'spirit of cleanup'

PR #1014, opened 2024-11-18
feat: find purls for affected packages

dejanb:


JimFuller-RedHat:
are we going to close this ?

PR #1016, opened 2024-11-18
fix: s3 storage config options and encoding case insensitivity

jcrossley3:
I confirmed that both compressed and uncompressed files rendered correctly when downloaded via the trustify app using [this bucket](https://us-east-1.console.aws.amazon.com/s3/buckets/trustify-jcrossley?region=us-east-1)

### DIFF
--- a/modules/storage/src/service/compression/mod.rs
+++ b/modules/storage/src/service/compression/mod.rs
@@ -11,7 +11,9 @@ use tokio::io::{AsyncBufRead, AsyncRead, AsyncWrite, BufReader, ReadBuf};
 )]
 pub enum Compression {
     #[default]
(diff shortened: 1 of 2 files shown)


modules/storage/src/service/compression/mod.rs
```diff
@@ -11,7 +11,9 @@ use tokio::io::{AsyncBufRead, AsyncRead, AsyncWrite, BufReader, ReadBuf};
 )]
 pub enum Compression {
     #[default]
+    #[strum(ascii_case_insensitive)]
```
ctron, line 14: Nice!

ctron (approved):

PR #1019, opened 2024-11-19
ci: cache the Ollama install script and downloaded models.

chirino:


ctron (approved):

ctron:
Looks like the CI fails with that. Not sure why.

chirino:
Looks like you can't actually cache the models 😢 

If we can't do that I don't think it's worth adding the caching.

PR #1020, opened 2024-11-19
chore: makes some function calls more readable

helio-frota:
* By explicitly saying that those are not using transaction.

Thanks to @jcrossley3 for this

### DIFF
--- a/modules/ingestor/src/graph/organization.rs
+++ b/modules/ingestor/src/graph/organization.rs
@@ -78,7 +78,10 @@ impl Graph {
         let name = name.into();
         let information = information.into();
 
-        if let Some(found) = self.get_organization_by_name(&name, &tx).await? {
+        if let Some(found) = self
+            .get_organization_by_name(&name, Transactional::None)
+            .await?
+        {
             if information.has_data() {
                 let mut entity = organization::ActiveModel::from(found.organization);
                 entity.website = Set(information.website);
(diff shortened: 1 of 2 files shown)


helio-frota:
> More readable is debatable.

yeah to say that -- this thing is Transactional::None as a db `select`

> Are we absolutely sure we're not changing behavior? 

the tests say - no

modules/ingestor/src/graph/organization.rs
```diff
@@ -78,7 +78,10 @@ impl Graph {
         let name = name.into();
         let information = information.into();
 
-        if let Some(found) = self.get_organization_by_name(&name, &tx).await? {
+        if let Some(found) = self
+            .get_organization_by_name(&name, Transactional::None)
```
ctron, line 82: That isn't about readability, but actually changing logic, isn't it. There's a difference between re-using the existing transaction (or none) vs explicitly not using a transaction.

ctron (changes requested):

modules/ingestor/src/graph/organization.rs
helio-frota: > vs explicitly not using a transaction

yeah so when this is explicit then is more easy to read and understand 

> In computer programming, [readability](https://en.wikipedia.org/wiki/Readability) refers to the ease with which a human reader can comprehend the purpose,

[link](https://en.wikipedia.org/wiki/Computer_programming#:~:text=the%20aforementioned%20attributes.-,Readability%20of%20source%20code,usability%20and%20most%20importantly%20maintainability.)

ctron: Yea, but it's changing the behavior of the application. And I don't think that's the correction action here.

helio-frota: I'm ok closing the PR, also this is not solving the problem with the upload :+1:

Issue #975, opened 2024-11-06
CVEs Related Products missing properties for v1 parity


kahboom:
In v1 CVE detail page you have the following properties visible for its related products, which are not available now:

- `Version`
- `Dependencies`
- `Supplier`
- `Created on` - unless it's the same as `published`?

See also [this](https://docs.google.com/spreadsheets/d/1mhBd3-UYvSANIXwRPDrCdGplN7cq5rQPJhIywo5mcDY/edit?gid=1088076328#gid=1088076328) spreadsheet for a complete list of properties missing in the current version of the UI.

![Screenshot 2024-11-06 at 2 04 48 PM copy](https://github.com/user-attachments/assets/41ffd4ed-7ba4-4598-9d1c-6d5fd23222ab)

ctron:
I think the following fields are already available:

* Version
* Supplier
* Created on

![image](https://github.com/user-attachments/assets/a72da201-1c30-47b6-89c2-57391d2e944a)

kahboom:
>I think the following fields are already available

thanks @ctron , not sure how I missed that 😅

ctron:
> > I think the following fields are already available
> 
> thanks @ctron , not sure how I missed that 😅

No worries, I am working on the other fields.

PR #1021, opened 2024-11-20
fix: move number_of_packages to SbomHead

### LINKED ISSUE #1006: Package count is missing from nested sboms returned from /api/v1/vulnerability/{id}
The UI has the page CVE/Vulnerability Page details where we render all the SBOMs that somehow have a relation with a Vulnerability. See the image below:

image

## Current UI process

-  Fetch the vulnerability `/api/v1/vulnerability/{id}` that returns a response similar to:

```json
{
  "advisories": [
    {
      "identifier": "Advisory1",
      "sboms": [
        {
          "id": "Sbom1",
          "status": ["affected"]
          // other fields
        },
        {
          "id": "Sbom2",
          "status": ["not_affected", "fixed"]
          // other fields
        }
      ]
    },
    {
      "identifier": "Advisory2",
      "sboms": [
        {
          "id": "Sbom2",
          "status": ["not_affected"]
          // other fields
        },
        {
          "id": "Sbom3",
          "status": ["not_affected"]
          // other fields
        }
      ]
    }
  ]
}
```

- The UI flattens the field `advisories.sboms`:

```json
[
  {
    "id": "Sbom1",
    "status": ["affected"]
  },
  {
    "id": "Sbom2",
    "status": ["not_affected", "fixed"]
  },
  {
    "id": "Sbom2",
    "status": ["not_affected"]
  },
  {
    "id": "Sbom3",
    "status": ["not_affected"]
  }
]
```

- We can see that the `status` field is an array so we flatten that array too. The logic is to create an element of the array for each `status` . In our example since there was an element of the array with `"status": ["not_affected", "fixed"]` then that is decoupled in 2 elements of the array. The final result is:

```json
[
  {
    "id": "Sbom1",
    "status": "affected"
  },
  {
    "id": "Sbom2",
    "status": "not_affected"
  },
  {
    "id": "Sbom2",
    "status": "fixed"
  },
  {
    "id": "Sbom2",
    "status": "not_affected"
  },
  {
    "id": "Sbom3",
    "status": "not_affected"
  }
]
```

- Then we remove duplicates:
```json
[
  {
    "id": "Sbom1",
    "status": "affected"
  },
  {
    "id": "Sbom2",
    "status": "not_affected"
  },
  {
    "id": "Sbom2",
    "status": "fixed"
  },
  {
    "id": "Sbom3",
    "status": "not_affected"
  }
]
```

- For the previous array we need the field `number_of_packages` for each sbom. Therefore we call `/api/v1/sbom/{id}` for each element of the previous array. Then enrich the previous array to end up having something like:

```json
[
  {
    "id": "Sbom1",
    "status": "affected",
    "number_of_packages": 123
  },
  {
    "id": "Sbom2",
    "status": "not_affected",
    "number_of_packages": 123
  },
  {
    "id": "Sbom2",
    "status": "fixed",
    "number_of_packages": 123
  },
  {
    "id": "Sbom3",
    "status": "not_affected",
    "number_of_packages": 123
  }
]
```

Finally the array is ready to be rendered in the UI. This approach suffer the same problems described at https://github.com/trustification/trustify/issues/952

-  Too many calls to the backend
- The client implementing its own algorithm to deal with multiple advisories and its complexity

## What I think we could evaluate changing

- Naming is very important. The response of `/api/v1/vulnerability/{id}` contains the field `advisories.sboms` which implies, from the naming, that what that field contains are SBOMS. The problem is that `advisories.sboms` does not match what `/api/v1/sbom/{id}` has as only `/api/v1/sbom/{id}` has the field `number_of_packages` but `advisories.sboms` doesn't.
  - Generally speaking it is important to define what an SBOM is and what are their fields so whenever any part of the API mentions `sbom` then the client is 100% sure what he expects there. 
  - If we have the field `advisories.sboms`  then I should get there exactly what I see in `/api/v1/sbom/{id}`
- The response from `/api/v1/vulnerability/{id}` contains a field `advisories.sboms.status` which is an array and not a single field. That technically means that an a Single advisory can say that an SBOM has the status `affected` and `not_affected` at the same time. E.g.

```json
{
  "advisories": [
    {
      "identifier": "Advisory1",
      "sboms": [
        {
          "id": "Sbom1",
          "status": ["affected", "not_affected"]
          // other fields
        },
      ]
    }
  ]
}
```

This structure forces the client to come up with some kind of algorithm to parse the data and solve conflicts between multiple `advisory.sboms.status` values which might contradict each other.

jcrossley3:
Fixes #1006 by making the sbom #packages available in `/api/v1/vulnerability/{id}` as well as `/api/v1/sbom/{id}`

### DIFF
--- a/modules/fundamental/src/ai/endpoints/test.rs
+++ b/modules/fundamental/src/ai/endpoints/test.rs
@@ -66,8 +66,7 @@ async fn flags(ctx: &TrustifyContext) -> anyhow::Result<()> {
         json!({
             "completions": service.completions_enabled(),
         }),
-        "result:\n{}",
-        serde_json::to_string_pretty(&result)?
+        "result:\n{result:#?}"
     );
 
     Ok(())
@@ -88,12 +87,7 @@ async fn tools(ctx: &TrustifyContext) -> anyhow::Result<()> {
 
     let expected: serde_json::Value =
         serde_json::from_str(include_str!("expected_tools_result.json"))?;
-    assert_eq!(
-        result,
-        expected,
-        "result:\n{}",
-        serde_json::to_string_pretty(&result)?
-    );
+    assert_eq!(result, expected, "result:\n{result:#?}");
 
     Ok(())
 }
--- a/modules/fundamental/src/purl/endpoints/test.rs
+++ b/modules/fundamental/src/purl/endpoints/test.rs
@@ -328,7 +328,7 @@ async fn package_with_status(ctx: &TrustifyContext) -> Result<(), anyhow::Error>
     let request = TestRequest::get().uri(&uri).to_request();
     let response: Value = app.call_and_read_body_json(request).await;
 
-    log::debug!("{}", serde_json::to_string_pretty(&response)?);
+    log::debug!("{response:#?}");
 
     Ok(())
 }
--- a/modules/fundamental/src/sbom/endpoints/test.rs
+++ b/modules/fundamental/src/sbom/endpoints/test.rs
@@ -45,7 +45,7 @@ async fn get_sbom(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
     let uri = format!("/api/v1/sbom/{id}");
     let req = TestRequest::get().uri(&uri).to_request();
     let sbom: Value = app.call_and_read_body_json(req).await;
-    log::debug!("{}", serde_json::to_string_pretty(&sbom)?);
+    log::debug!("{sbom:#?}");
 
     // assert expected fields
     assert_eq!(sbom["id"], id);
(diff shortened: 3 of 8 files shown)


ctron:
I think was added by a request from the UI: https://github.com/trustification/trustify/issues/975 So is moving this breaking the other case? Or did I get something wrong?

jcrossley3:
> I think was added by a request from the UI: #975 So is moving this breaking the other case? Or did I get something wrong?

I don't think you went quite far enough with #975, which is why #1006 was created. The #packages is required not only from `/api/v1/sbom/{id}` but also in the nested sboms returned by `/api/v1/vulnerability/{id}`. This PR should add the latter without affecting the former, but I'll add some tests to confirm.

ctron:
> > I think was added by a request from the UI: #975 So is moving this breaking the other case? Or did I get something wrong?
> 
> I don't think you went quite far enough with #975, which is why #1006 was created. The #packages is required not only from `/api/v1/sbom/{id}` but also in the nested sboms returned by `/api/v1/vulnerability/{id}`. This PR should add the latter without affecting the former, but I'll add some tests to confirm.

Ah ok. Got it.

ctron (approved):

PR #1028, opened 2024-11-20
feat: enable custom trust anchors for swagger oidc

ctron:
Also, fix a small docs issue.


ctron:
This is required for CRC, deployments with a custom trust anchor, used by the Helm charts.

Issue #860, opened 2024-09-30
Correlate purls to advisories using product statues


dejanb:
With #853 we started ingesting product statuses for csaf statements that don't include real purls, but just package names. We need to be able to find advisory statements for these purls by searching for product statuses by package names in sboms that contain them. This data should show up at `http://localhost:3000/packages/{}`

PR #1033, opened 2024-11-21
feat: correlate vulnerabilties for purls using product statuses

dejanb:
This PR addresses #860

### DIFF
--- a/modules/fundamental/src/purl/model/details/purl.rs
+++ b/modules/fundamental/src/purl/model/details/purl.rs
@@ -5,9 +5,10 @@ use crate::{
     vulnerability::model::VulnerabilityHead,
     Error,
 };
+use ::cpe::uri::OwnedUri;
 use sea_orm::{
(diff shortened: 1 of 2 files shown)


modules/fundamental/src/purl/model/details/purl.rs
```diff
@@ -275,3 +395,33 @@ impl FromQueryResultMultiModel for LicenseCatcher {
             .try_model_columns(license::Entity)
     }
 }
+
+#[derive(Debug)]
+#[allow(dead_code)] //TODO sbom field is not used at the moment, but we will probably need it for graph search
```
ctron, line 400: Does this comment still apply? It says "field".

dejanb: It doesn't. It was copied over from another query catcher. Fixed now, thanks for spotting it.

ctron (commented):
:+1:  LGTM. But I'd want @JimFuller-RedHat to have a look too, especially at the DB related stuff.

JimFuller-RedHat (approved):
I think this is fine to get into master - it will need some optimisation but I will do that as part of final perf pass, LGTM

PR #1034, opened 2024-11-21
Able to upload 290+ files adding this lockType

helio-frota:
Related to #999


helio-frota:
I'll close this issue because I'm not so happy to have a explicit lock when we already found other strange behavior
https://github.com/trustification/trustify/issues/1036

Issue #1038, opened 2024-11-22
Don't check if bucket exists, requires "list all" permission


ctron:
Currently we do this during startup:

https://github.com/trustification/trustify/blob/9ee0d138ddd7b43c8450ab24a795a4bc7998cc2d/modules/storage/src/service/s3.rs#L38

The problem with this check is, it performs a "list all" operation:

```rust
    #[maybe_async::maybe_async]
    pub async fn exists(&self) -> Result<bool, S3Error> {
        let credentials = self.credentials().await?;

        let response = Self::list_buckets(self.region.clone(), credentials).await?;

        Ok(response
            .bucket_names()
            .collect::<std::collections::HashSet<String>>()
            .contains(&self.name))
    }
```

This requires that we give the user accessing the backup the ability to discover all buckets of an AWS account. Or it will fail:

```
Error: Got HTTP 403 with content '<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>User: arn:aws:iam::469459763996:user/storage-jreimann is not authorized to perform: s3:ListAllMyBuckets because no identity-based policy allows the s3:ListAllMyBuckets action</Message><RequestId>1Q9K3Y0K8M4MXYCY</RequestId><HostId>AabcZXTDj6CmDy/Z7CkcZzEU/h0D4LIA3K6P9CNy9SynPDt2U17PXtnBeikjI7MrjTPdcGJG+WJQfGEKq3PBMxJ9SJRiNwbmfKJGAq5Sw+8=</HostId></Error>'
```

I am also not really happy about the hidden panic, through the `assert` macro. I think we should keep this to debug code.

PR #1039, opened 2024-11-22
fix: don't check if the bucket exists.

ctron:
Performing an "exists" check requires the s3:ListAllMyBuckets permission. Which feels excessive for such a simple check.

Closes: https://github.com/trustification/trustify/issues/1038


JimFuller-RedHat (approved):
LGTM

PR #1041, opened 2024-11-22
Add retrieve_all_sbom_roots_by_name() func

JimFuller-RedHat:
Add **retrieve_all_sbom_roots_by_name** function that searches for a component(s) by name in a specific sbom, then returns that components root component.

Can be used thusly:
```
        let roots = service
            .retrieve_all_sbom_roots_by_name(sbom_id, component_name, ())
            .await?;
```

Intended for use in #1014

**Note** - most of graph is based on string values ... this was an explicit design decision I do with graph stuff up front, its likely at some point I will revert these values back to some normal types at some point (ex. sbom_id would be stored in graph as a Uuid).

### DIFF
--- a/modules/analysis/src/service.rs
+++ b/modules/analysis/src/service.rs
@@ -710,6 +710,40 @@ impl AnalysisService {
         Ok(paginated.paginate_array(&components))
     }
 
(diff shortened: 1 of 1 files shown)


dejanb (approved):
PR on it's own look good. Let me see if it can be used for the use case.

Issue #1043, opened 2024-11-22
`/api/v1/sbom/{id}/advisory` contains packages without id.


carlosthe19916:
Steps to reproduce:
- Star the backend with the latest version 
- Upload all data from D1 https://github.com/trustification/trustify/tree/main/etc/datasets/ds1
- get the quarkus-sbom advisories using `/api/v1/sbom/urn%3Auuid%3A01935352-6dc7-7582-9241-b2577f404d3f/advisory`
 The response will be similar to 

```json
[
    {
        "uuid": "urn:uuid:e635c662-fc76-4c07-87b3-908226d95fd1",
        "identifier": "https://www.redhat.com/#CVE-2023-2798",
        "issuer": {
            "id": "a80a05f2-7a3c-48a3-9a17-5252f9f1c17e",
            "name": "Red Hat Product Security",
            "cpe_key": null,
            "website": null
        },
        "published": "2023-05-25T00:00:00Z",
        "modified": "2023-08-10T12:02:50Z",
        "withdrawn": null,
        "title": "Stack overflow crash causes Denial of Service (DoS)",
        "labels": {
            "type": "csaf"
        },
        "status": [
            {
                "normative": true,
                "identifier": "CVE-2023-2798",
                "title": null,
                "description": null,
                "reserved": null,
                "published": null,
                "modified": null,
                "withdrawn": null,
                "discovered": null,
                "released": null,
                "cwes": [],
                "average_severity": "high",
                "status": "not_affected",
                "context": {
                    "cpe": "cpe:/a:redhat:quarkus:2:*:*:*"
                },
                "packages": [
                    {
                        "id": "",
                        "name": "htmlUnit",
                        "version": null,
                        "purl": [],
                        "cpe": []
                    }
                ]
            }
        ]
    },
```

> I cut the response for better understanding

If you see the field `status.packages` then we have a package without `id` and only `name`.

The problem is that I am using this current data for the image below:

![image](https://github.com/user-attachments/assets/2fa0cb09-d35a-45cd-b1e0-60bad5d1224f)

- If you see the `affected dependencies` column you will see a count of packages affected by the Vulnerability.
- After clicking on a row|number at the `affected dependencies` column then there is supposed to be a table with all packages. Each package is link-able to the Page `Package Details`, which means that for each package it should be possible to go to the Package Details page. For navigating to the UI Package Details page we need an ID for each package.

jcrossley3:
[...]

Issue #994, opened 2024-11-12
Find purl based on component name and sbom


dejanb:
#948 is able to make vulnerability to sbom correlations (and back) based only on the component names mentioned in advisories. The next step is to use graph analysis to try and find purls for these component names in SBOMs and include them into the response

dejanb:
This is still pedning based on #1014 completion.

PR #1048, opened 2024-11-25
feat: find purls for affected packages

dejanb:
This is db implementation of feature needed for #994.

I would go with it to unlock work on UI (#1043) and see if and how to replace same functionality with graph in the future.

I'm keeping #1014 draft PR as a placeholder for this future work

### DIFF
--- a/modules/fundamental/src/sbom/model/details.rs
+++ b/modules/fundamental/src/sbom/model/details.rs
@@ -9,14 +9,14 @@ use crate::{
     vulnerability::model::VulnerabilityHead,
     Error,
 };
-use cpe::uri::OwnedUri;
+use cpe::{cpe::Cpe, uri::OwnedUri};
 use sea_orm::{
-    ConnectionTrait, DbErr, EntityTrait, FromQueryResult, JoinType, ModelTrait, QueryFilter,
+    ConnectionTrait, DbErr, EntityTrait, FromQueryResult, Iden, JoinType, ModelTrait, QueryFilter,
     QueryOrder, QueryResult, QuerySelect, RelationTrait, Select,
 };
 use sea_query::{Asterisk, Expr, Func, SimpleExpr};
 use serde::{Deserialize, Serialize};
-use std::collections::{hash_map::Entry, HashMap};
+use std::collections::HashMap;
 use trustify_common::{
     cpe::CpeCompare,
     db::{
@@ -48,7 +48,7 @@ impl SbomDetails {
         service: &SbomService,
         tx: &C,
     ) -> Result<Option<SbomDetails>, Error> {
-        let relevant_advisory_info = sbom
+        let mut relevant_advisory_info = sbom
             .find_related(sbom_package::Entity)
             .join(JoinType::Join, sbom_package::Relation::Node.def())
             .join(JoinType::LeftJoin, sbom_package::Relation::Purl.def())
@@ -83,7 +83,7 @@ impl SbomDetails {
             .all(tx)
             .await?;
 
-        let product_advisory_info = sbom
+        let mut product_advisory_info = sbom
             .find_related(product_version::Entity)
             .join(JoinType::LeftJoin, product_version::Relation::Product.def())
             .join(JoinType::LeftJoin, product::Relation::Cpe.def())
@@ -97,6 +97,16 @@ impl SbomDetails {
                 JoinType::Join,
                 product_status::Relation::Vulnerability.def(),
             )
+            // Joins for purl-related tables
+            .join(JoinType::Join, sbom::Relation::Node.def())
+            .join(JoinType::Join, sbom_node::Relation::Package.def())
+            .join(JoinType::Join, sbom_package::Relation::Purl.def())
+            .join(JoinType::Join, sbom_package_purl_ref::Relation::Purl.def())
+            .join(
+                JoinType::Join,
+                qualified_purl::Relation::VersionedPurl.def(),
+            )
+            .join(JoinType::Join, versioned_purl::Relation::BasePurl.def())
             .distinct_on([
                 (product_status::Entity, product_status::Column::ContextCpeId),
                 (product_status::Entity, product_status::Column::StatusId),
@@ -109,12 +119,37 @@ impl SbomDetails {
             .order_by_asc(product_status::Column::ContextCpeId)
             .order_by_asc(product_status::Column::StatusId)
             .order_by_asc(product_status::Column::Package)
-            .order_by_asc(product_status::Column::VulnerabilityId);
-
-        let product_advisory_statuses = product_advisory_info
-            .try_into_multi_model::<ProductStatusCatcher>()?
+            .order_by_asc(product_status::Column::VulnerabilityId)
+            // Filter for product_status.package
+            .filter(
+                Expr::col((product_status::Entity, product_status::Column::Package))
+                    .is_null()
+                    .or(Expr::col((product_status::Entity, product_status::Column::Package)).eq(""))
+                    .or(SimpleExpr::Binary(
+                        Box::new(
+                            Expr::col((product_status::Entity, product_status::Column::Package))
+                                .into(),
+                        ),
+                        sea_query::BinOper::Like,
+                        Box::new(SimpleExpr::FunctionCall(
+                            Func::cust(CustomFunc::Concat).args([
+                                Expr::col((base_purl::Entity, base_purl::Column::Namespace)).into(),
+                                Expr::val("/").into(),
+                                Expr::col((base_purl::Entity, base_purl::Column::Name)).into(),
+                            ]),
+                        )),
+                    ))
+                    .or(
+                        Expr::col((product_status::Entity, product_status::Column::Package))
+                            .eq(Expr::col((base_purl::Entity, base_purl::Column::Name))),
+                    ),
+            )
+            .select_only()
+            .try_into_multi_model::<QueryCatcher>()?
             .all(tx)
             .await?;
+
+        relevant_advisory_info.append(&mut product_advisory_info);
 
         let summary = SbomSummary::from_entity((sbom, node), service, tx).await?;
 
@@ -122,9 +157,8 @@ impl SbomDetails {
             Some(summary) => Some(SbomDetails {
                 summary: summary.clone(),
                 advisories: SbomAdvisory::from_models(
-                    &summary.clone().described_by,
+                    &summary.described_by,
                     &relevant_advisory_info,
-                    &product_advisory_statuses,
                     tx,
                 )
                 .await?,
@@ -132,6 +166,12 @@ impl SbomDetails {
             None => None,
         })
     }
+}
+
+#[derive(Iden)]
+enum CustomFunc {
+    #[iden = "CONCAT"]
+    Concat,
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, ToSchema)]
@@ -145,7 +185,6 @@ impl SbomAdvisory {
     pub async fn from_models<C: ConnectionTrait>(
         described_by: &[SbomPackage],
         statuses: &[QueryCatcher],
-        product_statuses: &[ProductStatusCatcher],
         tx: &C,
     ) -> Result<Vec<Self>, Error> {
         let mut advisories = HashMap::new();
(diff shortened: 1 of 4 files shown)


carlosthe19916 (commented):
@dejanb Thanks for the PR. I've just tested it.

- I imported all data from https://github.com/trustification/trustify/tree/main/etc/datasets/ds1
- If I see the Quarkus SBOM Using the UI I see the following result

Screenshot From 2024-11-25 16-29-01

- You see that that expanded are of the new row contains a full PURL therefore I can link the purl to the Package Details page in the UI
- The second row does not have a purl and only name. I guess this PR will help identifying some Purls but not all of them right? there will still be some package that won't contains IDs?

Also I tested with all the other SBOMS from https://github.com/trustification/trustify/tree/main/etc/datasets/ds1 and ALL sboms, except the `quarkus-sbom` are not able to generate the IDs. It looks like this PR enhanced things but only for the quarkus-sbom and not for the rest of them. But that might just be that the other SBOMs do not have enough data to achieve and make the magic?

ctron:
We should be careful with terminology here and how we use data. "packages" (aka components) only make sense in the context of an SBOM. The same is true for "names of packages". PURLs (as well as CPEs) are different. Packages may have a PURL or CPE, which has a meaning outside of the SBOM.

At Red Hat, we may choose to name packages using names that make sense externally, but that's just our decision.

modules/fundamental/src/sbom/model/details.rs
```diff
@@ -255,11 +259,31 @@ impl SbomAdvisory {
...
+                let name = if parts.len() == 1 { parts[0] } else { parts[1] };
+
+                let query = Query {
+                    q: format!("name={}", name),
+                    sort: Default::default(),
                 };
-                packages.push(package);
+
+                let mut sbom_packages = sbom_service
+                    .fetch_sbom_packages(summary.head.id, query, Default::default(), ())
```
ctron, line 274: I think this should pass on the transaction.

modules/fundamental/src/sbom/model/details.rs
```diff
@@ -255,11 +259,31 @@ impl SbomAdvisory {
...
+                    .await?;
+                if !sbom_packages.items.is_empty() {
```
ctron, line 276: Can we swap this?

```
if is_emty() {
  // add basic info
} else {
  // append all
}
```

modules/fundamental/src/sbom/model/details.rs
```diff
@@ -255,11 +259,31 @@ impl SbomAdvisory {
 
             let mut packages = vec![];
             if let Some(package) = &product.product_status.package {
-                let package = SbomPackage {
-                    name: package.to_string(),
-                    ..Default::default()
+                // We need to enable search on namespace/name pattern as well
+                // Also, from the brief look at it, the search is case sensitive
+                // which doesn't cover all use cases
+                let parts = package.split('/').collect::<Vec<_>>();
```
ctron, line 265: To my understanding the "package" here is the CSAF (advisory) internal package name. Taking that and looking up through a PURL "by name" with a sub-string match.

I think this is highly Red Hat specific and error prone.

And, as we seem to this this in a for loop, this might have a performance impact that we should check.

ctron (changes requested):

modules/fundamental/src/sbom/model/details.rs
dejanb: I didn't find any examples where we are getting `Transactional` from `ConnectionOrTransaction`. I tried quickly implementing something like

```
impl<'db> From<ConnectionOrTransaction<'db>> for Transactional {
    fn from(conn_or_tx: ConnectionOrTransaction<'db>) -> Self {
        match conn_or_tx {
            ConnectionOrTransaction::Connection(_) => Transactional::None,
            ConnectionOrTransaction::Transaction(tx) => Transactional::Some((*tx).clone()),
        }
    }
}
```
But failed miserably as `DbTransaction` is not cloneable and trying to make some changes to `Transactional` cascades all over the place.
If you have some example on how to do this easily, please let me know. Otherwise, I would deal with it in a separate PR as it is potentially affecting a lot of code.

dejanb:
@ctron This is the logic from v1 that made it possible to find these purls. It is the "best effort" kind of thing so it shouldn't affect others (non-RH advisories).
As for the performance, that's where in-mem graph search instead of db can come to play if needed (and/or is ready)

@carlosthe19916 so the large number of purls we don't find is consistent with v1, but in that place we don't show the vulnerability. Maybe it's the same logic we should apply here as well. We need to think more about what's more useful: showing that sbom is vulnerable although we can't find the component inside or not showing it to be affected in that case.

Anyways, I will come back to this next week. Please feel free to comment, change anything around this in the meantime.

modules/fundamental/src/sbom/model/details.rs
ctron: Ok, I just peeked into this. `ConnectionOrTransaction` carries exactly that. Where the other function accepts an indication if a new connection should be created, or an existing transaction be re-used. Those two concepts conflict with each other.

Normally, the more "outer" service functions accept the `Transactional`, getting a new database from the connection pool is required. While the more inner ones accept a an existing "connection or transaction". Especially when we're talking about the "from model" functions, which don't have a connection pool themselves.

Honestly, it feels a bit like we're mixing things here. My concern also is that this will be a performance issue.

I guess the only feasible option is to have `fetch_sbom_packages`  accept a `ConnectionTrait` instead of an `AsRef<Transactional>`. Maybe create a `fetch_sbom_packages_with` and have it accept the `ConnectionTrait`, keeping the existing one with the `AsRef<Transactional>`.

dejanb: Sounds good, thanks. Yeah, I think the new function is needed here anyways as the current one doesn't do all functionalities needed. And we need to provide a way to have in-mem graph implementation as well.

carlosthe19916:
@dejanb sorry for my late reply.

I uploaded the files from D3 and then using the current UI go to the details of an SBOM page. and I see the following message in the log of the backend:

```shell
2024-12-04T12:04:34.055378Z  WARN ThreadId(49) sqlx::query: slow statement: execution time exceeded alert threshold summary="SELECT DISTINCT ON (\"product_status\".\"context_cpe_id\", …" db.statement="\n\nSELECT\n  DISTINCT ON (\n    \"product_status\".\"context_cpe_id\",\n    \"product_status\".\"status_id\",\n    \"product_status\".\"package\",\n    \"product_status\".\"vulnerability_id\"\n  ) \"advisory\".\"id\" AS \"advisory$id\",\n  \"advisory\".\"identifier\" AS \"advisory$identifier\",\n  \"advisory\".\"version\" AS \"advisory$version\",\n  \"advisory\".\"document_id\" AS \"advisory$document_id\",\n  \"advisory\".\"deprecated\" AS \"advisory$deprecated\",\n  \"advisory\".\"issuer_id\" AS \"advisory$issuer_id\",\n  \"advisory\".\"published\" AS \"advisory$published\",\n  \"advisory\".\"modified\" AS \"advisory$modified\",\n  \"advisory\".\"withdrawn\" AS \"advisory$withdrawn\",\n  \"advisory\".\"title\" AS \"advisory$title\",\n  \"advisory\".\"labels\" AS \"advisory$labels\",\n  \"advisory\".\"source_document_id\" AS \"advisory$source_document_id\",\n  \"vulnerability\".\"id\" AS \"vulnerability$id\",\n  \"vulnerability\".\"title\" AS \"vulnerability$title\",\n  \"vulnerability\".\"reserved\" AS \"vulnerability$reserved\",\n  \"vulnerability\".\"published\" AS \"vulnerability$published\",\n  \"vulnerability\".\"modified\" AS \"vulnerability$modified\",\n  \"vulnerability\".\"withdrawn\" AS \"vulnerability$withdrawn\",\n  \"vulnerability\".\"cwes\" AS \"vulnerability$cwes\",\n  \"base_purl\".\"id\" AS \"base_purl$id\",\n  \"base_purl\".\"type\" AS \"base_purl$type\",\n  \"base_purl\".\"namespace\" AS \"base_purl$namespace\",\n  \"base_purl\".\"name\" AS \"base_purl$name\",\n  \"versioned_purl\".\"id\" AS \"versioned_purl$id\",\n  \"versioned_purl\".\"base_purl_id\" AS \"versioned_purl$base_purl_id\",\n  \"versioned_purl\".\"version\" AS \"versioned_purl$version\",\n  \"qualified_purl\".\"id\" AS \"qualified_purl$id\",\n  \"qualified_purl\".\"versioned_purl_id\" AS \"qualified_purl$versioned_purl_id\",\n  \"qualified_purl\".\"qualifiers\" AS \"qualified_purl$qualifiers\",\n  \"qualified_purl\".\"purl\" AS \"qualified_purl$purl\",\n  \"sbom_package\".\"sbom_id\" AS \"sbom_package$sbom_id\",\n  \"sbom_package\".\"node_id\" AS \"sbom_package$node_id\",\n  \"sbom_package\".\"version\" AS \"sbom_package$version\",\n  \"sbom_node\".\"sbom_id\" AS \"sbom_node$sbom_id\",\n  \"sbom_node\".\"node_id\" AS \"sbom_node$node_id\",\n  \"sbom_node\".\"name\" AS \"sbom_node$name\",\n  \"status\".\"id\" AS \"status$id\",\n  \"status\".\"slug\" AS \"status$slug\",\n  \"status\".\"name\" AS \"status$name\",\n  \"status\".\"description\" AS \"status$description\",\n  \"cpe\".\"id\" AS \"cpe$id\",\n  \"cpe\".\"part\" AS \"cpe$part\",\n  \"cpe\".\"vendor\" AS \"cpe$vendor\",\n  \"cpe\".\"product\" AS \"cpe$product\",\n  \"cpe\".\"version\" AS \"cpe$version\",\n  \"cpe\".\"update\" AS \"cpe$update\",\n  \"cpe\".\"edition\" AS \"cpe$edition\",\n  \"cpe\".\"language\" AS \"cpe$language\"\nFROM\n  \"product_version\"\n  INNER JOIN \"sbom\" ON \"sbom\".\"sbom_id\" = \"product_version\".\"sbom_id\"\n  LEFT JOIN \"product\" ON \"product_version\".\"product_id\" = \"product\".\"id\"\n  LEFT JOIN \"cpe\" ON \"product\".\"cpe_key\" = \"cpe\".\"product\"\n  JOIN \"product_status\" ON \"cpe\".\"id\" = \"product_status\".\"context_cpe_id\"\n  JOIN \"status\" ON \"product_status\".\"status_id\" = \"status\".\"id\"\n  JOIN \"advisory\" ON \"product_status\".\"advisory_id\" = \"advisory\".\"id\"\n  JOIN \"vulnerability\" ON \"product_status\".\"vulnerability_id\" = \"vulnerability\".\"id\"\n  JOIN \"sbom_node\" ON \"sbom\".\"sbom_id\" = \"sbom_node\".\"sbom_id\"\n  JOIN \"sbom_package\" ON \"sbom_node\".\"sbom_id\" = \"sbom_package\".\"sbom_id\"\n  AND \"sbom_node\".\"node_id\" = \"sbom_package\".\"node_id\"\n  JOIN \"sbom_package_purl_ref\" ON \"sbom_package\".\"sbom_id\" = \"sbom_package_purl_ref\".\"sbom_id\"\n  AND \"sbom_package\".\"node_id\" = \"sbom_package_purl_ref\".\"node_id\"\n  JOIN \"qualified_purl\" ON \"sbom_package_purl_ref\".\"qualified_purl_id\" = \"qualified_purl\".\"id\"\n  JOIN \"versioned_purl\" ON \"qualified_purl\".\"versioned_purl_id\" = \"versioned_purl\".\"id\"\n  JOIN \"base_purl\" ON \"versioned_purl\".\"base_purl_id\" = \"base_purl\".\"id\"\nWHERE\n  \"sbom\".\"sbom_id\" = $1\n  AND (\n    \"product_status\".\"package\" IS NULL\n    OR \"product_status\".\"package\" = $2\n    OR \"product_status\".\"package\" LIKE CONCAT(\"base_purl\".\"namespace\", $3, \"base_purl\".\"name\")\n    OR \"product_status\".\"package\" = \"base_purl\".\"name\"\n  )\nORDER BY\n  \"product_status\".\"context_cpe_id\" ASC,\n  \"product_status\".\"status_id\" ASC,\n  \"product_status\".\"package\" ASC,\n  \"product_status\".\"vulnerability_id\" ASC\n" rows_affected=22 rows_returned=22 elapsed=1.469311664s elapsed_secs=1.4693116640000001 slow_threshold=1s
```

So there are mentions to `query: slow statement: execution time `

I am just reporting it, I am not saying it is wrong. Just wanted you to be aware of it in any case :)

- On the other hand, I don't know why. When I use this PR branch (backend) and then import the D1 or D3 files then I cannot see vulnerabilities in the UI (in the SBOM Details page - VUlnerabilities tab) for sboms different from "qurakus-sbom". I mean:
  - using this PR, the only sbom that has vulnerabilities with the status "affected" is the "quarkus-sbom"
  - using the main branch, all sboms have at least one vulnerability with status "affected".

I don't know if that is thing to consider or nor. I just wanted you point that out.

To me, as the person on the client side, I am happy with the result. I can approve this if my review is required :) . I can not judge the code :)

dejanb:
@carlosthe19916 Thanks Carlos. Those both are valid and expected things.

The query is a bit slowish, but I can think we can further improve it.
As for the results, this is now in line with v1 parity. Currently in the main we show vulnerabilities for packages we can't find in the sbom. I think the new behaviour is correct.

carlosthe19916 (approved):
LGTM! From the client side point of view. I'll let the others decide the rest.

JimFuller-RedHat (approved):
yes the query will need some optimisation, but for main LGTM.

ctron (approved):

PR #1057, opened 2024-11-28
test: adds the concurrent test to mimic the upload behavior

helio-frota:
run with `cargo test advisory::endpoints::test::upload_default_csaf_format_multiple_concurrent -- --nocapture`

Related to #999

### DIFF
--- a/modules/fundamental/src/advisory/endpoints/test.rs
+++ b/modules/fundamental/src/advisory/endpoints/test.rs
@@ -8,7 +8,8 @@ use hex::ToHex;
 use jsonpath_rust::JsonPathQuery;
 use serde_json::{json, Value};
 use sha2::{Digest, Sha256};
-use test_context::test_context;
+use std::rc::Rc;
+use test_context::{futures, test_context};
 use test_log::test;
 use time::OffsetDateTime;
 use trustify_common::{db::Transactional, hashing::Digests, id::Id, model::PaginatedResults};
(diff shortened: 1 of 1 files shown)


modules/ingestor/src/graph/organization.rs
```diff
@@ -83,8 +83,14 @@ impl Graph {
                 let mut entity = organization::ActiveModel::from(found.organization);
                 entity.website = Set(information.website);
                 entity.cpe_key = Set(information.cpe_key);
-                let model = entity.update(&self.connection(&tx)).await?;
-                Ok(OrganizationContext::new(found.graph, model))
+                let model = match entity.update(&self.connection(&tx)).await {
+                    Ok(model) => Ok(model),
+                    Err(e) => {
+                        eprintln!("Failed to update organization: {:?}", e);
```
ctron, line 89: We have some guidelines on how to deal with this: https://github.com/trustification/trustify/blob/main/docs/design/log_tracing.md

ctron (changes requested):

modules/ingestor/src/graph/organization.rs
helio-frota: :facepalm: thanks, and tbh I forgot to remove `eprintln!`

I'll update based on this statement:
`Log messages and tracing should be relevant to operating personell and developers, not end users.` with log::error! macro thanks :+1:

modules/ingestor/src/graph/organization.rs
```diff
@@ -83,8 +83,14 @@ impl Graph {
                 let mut entity = organization::ActiveModel::from(found.organization);
                 entity.website = Set(information.website);
                 entity.cpe_key = Set(information.cpe_key);
-                let model = entity.update(&self.connection(&tx)).await?;
-                Ok(OrganizationContext::new(found.graph, model))
+                let model = match entity.update(&self.connection(&tx)).await {
+                    Ok(model) => Ok(model),
+                    Err(e) => {
+                        log::error!("Failed to update organization: {e}");
```
ctron, line 89: Unless this terminates or seriously affects the the application, this should be reduced to info or debug.

helio-frota:
I'll revert the `modules/ingestor/src/graph/organization.rs` code and since the test is already ignored we can safe-merge-it

PR #1060, opened 2024-11-29
ci: disable musl builds for now

ctron:
The reason for this, that we enable sqlx's TLS mode. Which requires, openssl, even for the build-dependency. In combination with musl, this creates some major problems, as vendoring doesn't seem to work.

For now, we just disable musl.


JimFuller-RedHat (approved):
LGTM

PR #1062, opened 2024-11-29
chore: using log macro

helio-frota:
* See https://github.com/trustification/trustify/blob/main/docs/design/log_tracing.md#tldr

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,7 +42,7 @@ jobs:
       - name: Check
         run: cargo check
       - name: Clippy
-        run: cargo clippy --all-targets --all-features -- -D warnings -D clippy::unwrap_used -D clippy::expect_used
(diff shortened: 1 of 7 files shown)


ctron (changes requested):
I am not sure this is the right approach. In general, I like the idea of having a clippy lint. However, converting all `(e)print(ln)` macros is not correct.

https://github.com/trustification/trustify/blob/main/docs/design/log_tracing.md#println-and-panic:

> An application should not panic during normal runtime. Also, all output should use logging or tracing.
> 
> There are exceptions to this rule (aside from tests): the start of the application. During the logging/tracing system configuration, it might be ok to use println. And if the application cannot be started up, failing with a panic might be ok too.
> 
> The same is true for simple command line applications, which are expected to perform a task, and then finish. Either successful, or with an error (returned as exit code). For example, the migrations. They can run and fail, even with a panic.

There are cases where those make sense, or are actually necessary.

helio-frota:
yeah I agree I"ll close the PR thanks

PR #1063, opened 2024-11-29
chore: allow running a dedicated importer instance

ctron:
The idea is to allow setups with scaled up API pods, but keeping the importer process scaled to 1.

Closes: https://github.com/trustification/trustify/issues/1059


server/src/profile/importer.rs
```diff
@@ -0,0 +1,146 @@
...
+};
+use trustify_module_ui::{endpoints::UiResources, UI};
+use utoipa::openapi::{Info, License};
+use utoipa::OpenApi;
+use utoipa_rapidoc::RapiDoc;
+use utoipa_redoc::{Redoc, Servable};
+
+/// Run the API server
+#[derive(clap::Args, Debug)]
+pub struct Run {
```
jcrossley3, line 50: To me, it's confusing to have both this module *and* the server's `with_importer` argument. I would think one or the other would suffice.

My vote would be for keeping just the `with_importer` feature of the server and delete this module. To run a dedicated importer, we'd simply set the `with_importer` to true on one of the deployed `api` servers.

trustd/src/main.rs
```diff
@@ -60,16 +64,28 @@ async fn pm_mode() -> anyhow::Result<ExitCode> {
         "api",
         #[cfg(feature = "garage-door")]
         "--embedded-oidc",
+        "--sample-data",
+        "--db-port",
+        &postgres.settings().port.to_string(),
+    ]);
+
+    let importer = Trustd::parse_from([
```
jcrossley3, line 72: Wouldn't it be simpler to add `--with-importer` to the `api` var and omit the `importer` var?

jcrossley3 (changes requested):

ctron:
The idea was to split up the "all-in-one" thing into two processes/instances: one running the importer, one running the API server.

I think, right now, we're overloading the `api` sub-command. We have the `db` sub-commands, but enable that functionality with the `--devmode` on the `api` sub-command. Running the importers on the `api` sub-command also feels wrong. Then we also have the "no subcommand" mode, which does run the DB stuff, as well as the API server.

I think it's ok to have an "all in one" command. But adding more complexity to the `api` sub-command feels wrong. Then again, forcing people today to run two processes, if they only want to use the `api` sub-command felt like leading to frustration.

That was the reasoning behind having both the dedicated command, as well as the "with" flag.

If we want to approach this from a perspective of making this clean and focused, I think we should drop the "with" flag on the API server, and run this together with the API server when running in "pm mode".

jcrossley3:
> If we want to approach this from a perspective of making this clean and focused, I think we should drop the "with" flag on the API server, and run this together with the API server when running in "pm mode".

+1

ctron:
@jcrossley3 I updated the PR, I hope that's what you had in mind.

jcrossley3 (approved):

PR #1065, opened 2024-12-02
fix: always create "described by" relationships

ctron:
In cases where the metadata component doesn't have an explicit `bom-ref`, we now create an artificial one, so that the relationship gets created.

CycloneDX doesn't need such a reference for the relationship, but we do. So instead of ignoring the information, we create an artificial value. Just like primary IDs in the database.

Also, improve some test coverage.

### DIFF
--- a/modules/ingestor/src/graph/sbom/cyclonedx.rs
+++ b/modules/ingestor/src/graph/sbom/cyclonedx.rs
@@ -91,7 +91,7 @@ impl SbomContext {
     #[instrument(skip(tx, sbom), ret)]
     pub async fn ingest_cyclonedx<TX: AsRef<Transactional>>(
         &self,
-        sbom: Bom,
+        mut sbom: Bom,
         tx: TX,
     ) -> Result<(), anyhow::Error> {
         let db = &self.graph.db.connection(&tx);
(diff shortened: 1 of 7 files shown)


modules/ingestor/src/graph/sbom/cyclonedx.rs
```diff
@@ -91,16 +91,23 @@ impl SbomContext {
...
 
-        if let Some(metadata) = &sbom.metadata {
-            if let Some(component) = &metadata.component {
+        // extract "describes"
+
+        if let Some(metadata) = &mut sbom.metadata {
+            if let Some(component) = &mut metadata.component {
+                let bom_ref = component
+                    .bom_ref
+                    .get_or_insert_with(|| Uuid::new_v4().to_string())
```
JimFuller-RedHat, line 108: we generate a uuid if no bom_ref exists ?

modules/ingestor/src/graph/sbom/cyclonedx.rs
```diff
@@ -246,6 +266,7 @@ impl<'a> Creator<'a> {
                 .bom_ref
                 .as_ref()
                 .cloned()
+                // TODO: we might re-consider this
```
JimFuller-RedHat, line 269: elsewhere we generate uuid - here we seem to be ok using the name ?

JimFuller-RedHat (approved):
a few comments - otherwise LGTM

modules/ingestor/src/graph/sbom/cyclonedx.rs
ctron: Yea, "todo". Didn't change the logic in this PR. Not happy about it. I'll take another look.

modules/ingestor/src/graph/sbom/cyclonedx.rs
ctron: Yes, we need to link it in the database. As the database uses references, where the cyclonedx uses a "contains" model to describe this information.

modules/ingestor/src/graph/sbom/cyclonedx.rs
ctron: Follow up PR: https://github.com/trustification/trustify/pull/1067

PR #1066, opened 2024-12-02
chore: add a document id for advisories

ctron:
Fix an issue for the UI, where the document ID was reported as `https://www.redhat.com/#CVE-2023-0044` for CSAF documents. However, in v1 the ID was `CVE-2023-0044`.

This is due to the fact that we have a list of IDs for each advisory: an internal DB one (`id`), one that uniquely identifies the docment (`identifier`), and now one which carries the ID as reported by the document (`document_id`). The latter one should be show in the UI.

@carlosthe19916 This also means that at the UI needs to pick up the new field, once this PR is merged.


JimFuller-RedHat (approved):
LGTM

PR #1069, opened 2024-12-03
refactor: drop Transactional and ConnectionOrTransaction

ctron:
Directly use the `ConnectionTrait` from sea_orm.

### DIFF
--- a/modules/analysis/src/endpoints.rs
+++ b/modules/analysis/src/endpoints.rs
@@ -13,10 +13,11 @@ use trustify_auth::{
 use trustify_common::{db::query::Query, db::Database, model::Paginated, purl::Purl};
 
 pub fn configure(config: &mut utoipa_actix_web::service_config::ServiceConfig, db: Database) {
-    let analysis = AnalysisService::new(db);
+    let analysis = AnalysisService::new();
 
     config
         .app_data(web::Data::new(analysis))
+        .app_data(web::Data::new(db))
         .service(search_component_root_components)
         .service(get_component_root_components)
         .service(analysis_status)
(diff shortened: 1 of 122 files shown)


modules/analysis/src/endpoints.rs
```diff
@@ -34,12 +35,13 @@ pub fn configure(config: &mut utoipa_actix_web::service_config::ServiceConfig, d
...
 pub async fn analysis_status(
     service: web::Data<AnalysisService>,
+    db: web::Data<Database>,
     user: UserInformation,
     authorizer: web::Data<Authorizer>,
     _: Require<ReadSbom>,
 ) -> actix_web::Result<impl Responder> {
     authorizer.require(&user, Permission::ReadSbom)?;
-    Ok(HttpResponse::Ok().json(service.status(()).await?))
+    Ok(HttpResponse::Ok().json(service.status(&**db).await?))
```
jcrossley3, line 44: I love concise code, but is this a tad too magic? I can't know at a glance what `&**db` is doing. Are we taking a reference of the result of calling first the `Deref` impl of `web::Data` and then that of `Database`? I'm not a rust noob, and I'm not sure. :smile: 

Is something like `service.status(db.get_ref().connection())` too wordy or less idiomatic?

ctron: > Are we taking a reference of the result of calling first the `Deref` impl of `web::Data` and then that of `Database`?

Btw, close: The first `Deref` takes the `Arc` from `web::Data`, the second takes `T` (`Database`) from the `Arc`.

ctron: Replaced this with `db.as_ref()`.

JimFuller-RedHat (approved):
took me a little while to get through, LGTM!

helio-frota (approved):
it is hard to see upload deadlock #999  using this refactor https://gist.github.com/helio-frota/931dc5cf93a850996b210dcf7f0003d8

jcrossley3 (approved):

PR #1070, opened 2024-12-03
chore: Enable persistent DB during tests

helio-frota:
This commits adds a new env-var `DB_PATH` for us to specify the postgresql path.

`DB_PATH=/home/heliofrota/Desktop/testdb cargo test advisory::endpoints::test::upload_default_csaf_format_multiple`

With this we can take a look at the logs when the test finish

`cat testdb/data/start.log`

### DIFF
--- a/common/src/db/embedded.rs
+++ b/common/src/db/embedded.rs
@@ -4,31 +4,40 @@ use postgresql_embedded::{PostgreSQL, Settings, VersionReq};
 use std::path::Path;
 use tracing::{info_span, Instrument};
 
-/// Create common default settings for the embedded database
-fn default_settings() -> anyhow::Result<Settings> {
+/// Create common default settings for the embedded database.
+/// Use the bool `temp` parameter to indicate if the pg_data
+/// directory is temporary or not.
+fn default_settings(temp: bool) -> anyhow::Result<Settings> {
     let version = VersionReq::parse("=16.3.0").context("valid psql version")?;
     Ok(Settings {
         version,
         username: "postgres".to_string(),
         password: "trustify".to_string(),
-        temporary: true,
+        temporary: temp,
         ..Default::default()
     })
 }
 
 /// Create a new, embedded database instance
-pub async fn create() -> anyhow::Result<(Database, PostgreSQL)> {
-    create_for(default_settings()?).await
+/// Use the bool `temporary` parameter to indicate if the pg_data
+/// directory is temporary or not.
+pub async fn create(temporary: bool) -> anyhow::Result<(Database, PostgreSQL)> {
+    create_for(default_settings(temporary)?).await
 }
 
 /// Create a new, embedded database instance in a specific directory
-pub async fn create_in(base: impl AsRef<Path>) -> anyhow::Result<(Database, PostgreSQL)> {
+/// Use the bool `temporary` parameter to indicate if the pg_data
+/// directory is temporary or not.
+pub async fn create_in(
+    base: impl AsRef<Path>,
+    temporary: bool,
+) -> anyhow::Result<(Database, PostgreSQL)> {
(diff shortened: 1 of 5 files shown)


common/src/db/embedded.rs
```diff
@@ -4,31 +4,40 @@ use postgresql_embedded::{PostgreSQL, Settings, VersionReq};
 use std::path::Path;
 use tracing::{info_span, Instrument};
 
-/// Create common default settings for the embedded database
-fn default_settings() -> anyhow::Result<Settings> {
+/// Create common default settings for the embedded database.
+/// Use the bool `temp` parameter to indicate if the pg_data
+/// directory is temporary or not.
+fn default_settings(temp: bool) -> anyhow::Result<Settings> {
```
jcrossley3, line 10: Passing a param to this function kind of betrays its name. :wink:

common/src/db/embedded.rs
```diff
@@ -4,31 +4,40 @@ use postgresql_embedded::{PostgreSQL, Settings, VersionReq};
...
+/// directory is temporary or not.
+pub async fn create(temporary: bool) -> anyhow::Result<(Database, PostgreSQL)> {
+    create_for(default_settings(temporary)?).await
 }
 
 /// Create a new, embedded database instance in a specific directory
-pub async fn create_in(base: impl AsRef<Path>) -> anyhow::Result<(Database, PostgreSQL)> {
+/// Use the bool `temporary` parameter to indicate if the pg_data
+/// directory is temporary or not.
+pub async fn create_in(
```
jcrossley3, line 31: I don't think adding this function buys us any more than we already have with the `EXTERNAL_TEST_DB` feature.

docs/env-vars.md
```diff
@@ -50,6 +50,7 @@
 
 | Environment Variable | Description                    | Default Value |
 |----------------------|--------------------------------|---------------|
+| `DB_PATH`       | Specifies a location for postgres data directory making it persistent after the test |           |
```
jcrossley3, line 53: Don't we have enough env vars already? :smile:

test-context/src/lib.rs
```diff
@@ -126,11 +126,25 @@ impl AsyncTestContext for TrustifyContext {
...
+                .expect("Create an embedded database");
+
+            TrustifyContext::new(db, postgresql).await
+        } else {
+            let (db, postgresql) = db::embedded::create(true)
+                .await
+                .expect("Create an embedded database");
+
+            TrustifyContext::new(db, postgresql).await
+        }
```
jcrossley3, line 147: Again, I don't think this code justifies its existence.

jcrossley3 (commented):
Can't we already achieve the goal of this PR with the `EXTERNAL_TEST_DB` env var? You could set `DB_PORT` to that of your "PM mode" instance and all the logs would be available to you in `.trustify/data/start.log`.

common/src/db/embedded.rs
helio-frota: LOL that's right :smile:

helio-frota: > Can't we already achieve the goal of this PR with the EXTERNAL_TEST_DB env var? You could set DB_PORT to that of your "PM mode" instance and all the logs would be available to you in .trustify/data/start.log.

it works with `cargo test` without running the app ? :thinking:

test-context/src/lib.rs
helio-frota: > Again, I don't think this code justifies its existence.

ok I'm +1 here...

a) because of upload issue I would like to see the DB logs after run `cargo test`
b) I sent this PR because of upload issue...
c) I saw the logs and it was exactly the same (the error message) as the app

I'll close the PR :+1:

PR #1071, opened 2024-12-04
chore: update nu-ansi-term

helio-frota:
This is used on AI part of the code only:

```
➜  trustify git:(chores) ✗ rg "nu_ansi_term"
xtask/src/ai.rs
3:use nu_ansi_term::Color::Blue;
```

See: https://github.com/nushell/nu-ansi-term/releases

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9675,7 +9675,7 @@ dependencies = [
  "clap",
  "langchain-rust",
  "log",
- "nu-ansi-term 0.46.0",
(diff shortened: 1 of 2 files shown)


helio-frota:
wait my neovim is formatting cargo.toml :thinking:

ctron (changes requested):
Please keep the other lines as they are. Especially the single line dependencies, as that way it's easy to sort them alphabetically.

PR #1073, opened 2024-12-05
fix: enable filtering/sorting on qualified purl json fields

### LINKED ISSUE #888: Package List - V1 Parity - SORTING
image

V1 parity work is being done. I'd like us to agree on which columns the UI will be able to sort and how. If possible be able to sort by all columns the table has. Otherwise, make conscious decisions about why not.

V1 does not have a single column Sort. We didn't do Sorting in v1, I believe due to a design that limited us to do such thing but those limitations are solved now I think right?.

- **Sort by Name**: Which is the REST query to achieve it?
- **Sort by Version**: Which is the REST query to achieve it?
- **Sort by Type**: Which is the REST query to achieve it?
- **Sort by Vulnerabilities**: this column will render the total the count of vulnerabilities affecting the package. This might need a deeper discussion (we there are multiple advisories for the same package) but a brainstorm would be welcome.
- **Sort by SBOMs**: this column will render the total count of SBOMs that contain the package. I believe we can not do that at the moment, then let's decide whether or not we are able to do it.

### LINKED ISSUE #951: Package List - Filtering
This is how it looks V1 Package List page:

Screenshot From 2024-10-30 16-25-07

- In v2 UI I am using the endpoint `/api/v1/purl` for populating the table in the UI.

There are 2 main filters I need to implement in V2.

- Type Filter: 
  - I am currently using `q=type=maven|rpm|npm|oci` and it works! So I guess there is nothing to do on that.
- Architecture Filter:
  - I was not able to make it work with `q=arch=x86_64|aarch64|ppc64le|s390x|noarch` so would you advice me how to filter by Architecture?

### LINKED ISSUE #972: Curious unused code snippet relating to querying jsonb fields
@JimFuller-RedHat I'm thinking this is related to something I'm working on, #888, and I'm curious why it's not in a comment. :smile: 
 
https://github.com/trustification/trustify/blob/68d08bc52e6e65ac8d352ad524904df7a20d2cf8/modules/fundamental/src/purl/service/mod.rs#L336-L343

jcrossley3:
Fixes #888
Fixes #951
Fixes #972

This introduces some potential naming conflicts, specifically that `type` is a valid PURL _qualifier_ as well as an attribute of the PURL itself, therefore we currently alias `ty` to `type`, which effectively prohibits using the `type` _qualifier_ in queries; only the type of the PURL itself may be used.

We need to decide which qualifiers we want to support.

### DIFF
--- a/common/auth/src/authorizer/mod.rs
+++ b/common/auth/src/authorizer/mod.rs
@@ -32,7 +32,7 @@ impl Authorizer {
         permission: Permission,
     ) -> Result<(), AuthorizationError> {
         if self.config.is_none() {
-            log::warn!("Authorization disabled, all permissions granted");
+            log::debug!("Authorization disabled, all permissions granted");
             return Ok(());
         }
 
--- a/common/src/db/query/columns.rs
+++ b/common/src/db/query/columns.rs
@@ -137,8 +137,17 @@ impl Columns {
         self
     }
 
-    pub fn iter(&self) -> impl Iterator<Item = &(ColumnRef, ColumnDef)> {
-        self.columns.iter()
+    /// Return the columns that are string-ish
+    pub(crate) fn strings(&self) -> impl Iterator<Item = Expr> + '_ {
+        self.columns
+            .iter()
+            .filter_map(|(col_ref, col_def)| match col_def.get_column_type() {
+                ColumnType::String(_) | ColumnType::Text => Some(Expr::col(col_ref.clone())),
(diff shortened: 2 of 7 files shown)


JimFuller-RedHat (approved):
very nice, LGTM

ctron (approved):

carlosthe19916 (approved):
Tested it with UI and it works! Thank you!

PR #1076, opened 2024-12-05
CycloneDX 1.6 PoC

ctron:
Here's a PoC, trying to leverage a different parsing library for the CycloneDX SBOMs. It supports 1.6 and has a simpler approach to generating the structs from the schema. Thus I'd expect less delays with future versions.

Just to be clear, this doesn't extract any of the new 1.6 information. It just ensures we can parse such files.


JimFuller-RedHat:
Not that I think we should overload this PR to implement 'all the things'...

**tldr on cdx 1.6 support for prodsec**: 

* Set relationships to represent SRPM->binaryRPM, image index -> arch-specific image, image contains component, and others). 
* Components with identities represented by multiple identifiers (aka multiple purls and CPEs for one component).
* Evidence, pedigree and provenance object support introspection (eg. able to retrieve/query on these values).

ctron:
I think we should discuss if we want to merge this, or not. I know it doesn't add any new data extraction, only allows to actually parse CDX 1.6.

ctron:
The one thing I am not sure about is, it it is good enough to only parse using the 1.6 structures. All our tests (which are pre 1.6) pass. So it looks ok.

JimFuller-RedHat (approved):
I think this is a important step - we take earlier rather then later - LGTM

jcrossley3:
> The one thing I am not sure about is, it it is good enough to only parse using the 1.6 structures. All our tests (which are pre 1.6) pass. So it looks ok.

I guess we have some gaps in our testing because [this](https://github.com/trustification/trustify/blob/cd26af6207b0f9101c84d73023c0572ce53d22b2/modules/ingestor/src/service/format.rs#L192-L200) doesn't seem correct in light of this PR.

ctron:
> > The one thing I am not sure about is, it it is good enough to only parse using the 1.6 structures. All our tests (which are pre 1.6) pass. So it looks ok.
> 
> I guess we have some gaps in our testing because [this](https://github.com/trustification/trustify/blob/cd26af6207b0f9101c84d73023c0572ce53d22b2/modules/ingestor/src/service/format.rs#L192-L200) doesn't seem correct in light of this PR.

Indeed.

PR #1078, opened 2024-12-06
test: added a test case for TC-1817

ctron:


mrizzi (commented):
@ctron I just need some context to be sure I'm not missing anything on the trustify side of this issue.

In trustification, [TC-1817](https://issues.redhat.com/browse/TC-1817) is caused by having used the `id` user-provided value as the file name into the S3-compatible storage.

In trustify, AFAIU, when storing into the S3-compatible storage, it's the `sha256` value to be used as file name:
https://github.com/trustification/trustify/blob/60c6097304e8531d57581a86e3afb64f6dd97625/modules/storage/src/service/s3.rs#L67

So basically trustify looks decoupled from all user input when storing files, is this right?

ctron:
> So basically trustify looks decoupled from all user input when storing files, is this right?

Yes, that's right.

Still, there's a `name` field in the SBOM, which in this case has the value `registry.k8s.io/coredns/coredns:v1.9.3`. I think adding this test anyway should keep us a bit safer for the future.

mrizzi (approved):
@ctron LGTM, thanks 👍

Issue #1077, opened 2024-12-05
Sort SBOMs and Advisories by "ingestion time"


carlosthe19916:
This issue is a follow up of https://github.com/trustification/trustify/issues/952

- The use case is the Dashboard where we render the last 10 SBOMs ingested and on the right side we render the date in which the last advisory was ingested.
  - Currently what I am doing is to fetch `GET /api/v1/sbom/`, sort it by `published` and in this way I have the 10 SBOMs the dashboard will render. But what we need it to be able to sort SBOMs by "when the date in which the file was imported into the system". Same for Advisories

@jcrossley3 tagging you here as you implemented nicely the Dashboard use case, and you are familiar with the use case.

jcrossley3:
@carlosthe19916 you should now be able to call `/api/v1/sbom?sort=ingested%3Adesc&limit=10` to get the 10 most recently ingested SBOM's.

I didn't realize you needed it for advisories, too, though. PR forthcoming.

PR #1079, opened 2024-12-06
feat: add ingestion time to source documents

ctron:
This adds an ingestion timestamp.

@jcrossley3 not sure this makes it available for search too.

Also see: https://github.com/trustification/trustify/issues/1077


jcrossley3 (changes requested):

jcrossley3:
> @jcrossley3 not sure this makes it available for search too.

Should be easy enough to add a test for it, no?

ctron:
Renamed the field.

ctron:
Added a test case.

jcrossley3:
> Added a test case.

But not one that affirms what you weren't sure about. C'mon, you can do it!

ctron:
> > Added a test case.
> 
> But not one that affirms what you weren't sure about. C'mon, you can do it!

I am sorry, I don't understand.

jcrossley3 (approved):

PR #1081, opened 2024-12-06
chore: adding row lock with behavior

helio-frota:
To avoid organization deadlock

### DIFF
--- a/modules/ingestor/src/graph/organization.rs
+++ b/modules/ingestor/src/graph/organization.rs
@@ -1,4 +1,7 @@
-use sea_orm::{ActiveModelTrait, ColumnTrait, ConnectionTrait, EntityTrait, QueryFilter, Set};
+use sea_orm::{
+    ActiveModelTrait, ColumnTrait, ConnectionTrait, EntityTrait, QueryFilter, QuerySelect, Set,
+};
+use sea_query::{LockBehavior, LockType};
 use std::fmt::Debug;
 use tracing::instrument;
 use trustify_entity::organization;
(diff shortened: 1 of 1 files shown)


modules/ingestor/src/graph/organization.rs
```diff
@@ -62,6 +65,7 @@ impl Graph {
     ) -> Result<Option<OrganizationContext>, Error> {
         Ok(organization::Entity::find()
             .filter(organization::Column::Name.eq(name.into()))
+            .lock_with_behavior(LockType::Update, LockBehavior::SkipLocked)
```
ctron, line 68: We are getting an organization, we do we need to lock that row?

Also, why would we want to skip that row in a case it's locked?

ctron (changes requested):
Not sure I understand what's the motivation.

modules/ingestor/src/graph/organization.rs
helio-frota: from the [docs](https://www.postgresql.org/docs/current/sql-select.html#:~:text=With%20SKIP%20LOCKED%20%2C%20any%20selected,accessing%20a%20queue-like%20table.):
`With SKIP LOCKED, any selected rows that cannot be immediately locked are skipped. Skipping locked rows provides an inconsistent view of the data, so this is not suitable for general purpose work, but can be used to avoid lock contention with multiple consumers accessing a queue-like table.`

my main motivation for this is `but can be used to avoid lock contention with multiple consumers`
to solve this [update problem](https://gist.github.com/helio-frota/079780ca7e8d0d194ea687af7ec3e5c7#file-current_main-txt-L11 ) 

I sent the PR mostly to share the app behavior with the team, I'm not fan of adding db locks

ctron: So the result would be that anyone getting the organization would lock it (until the transaction expires) and that would make other, concurrent request skip that record?

helio-frota: What I understand of this is that it blocks the select row and ignores the already blocked, by other concurrent calls.
I think is the same what you wrote.
Again I'm not happy with this also I'm using a small set of files for the upload test via ui.

helio-frota: yeah not a problem :+1:  the PR is more for me to have second thought

Issue #953, opened 2024-10-30
How to render data for V1 parity


carlosthe19916:
cc: @mid998 @PhilipCattanach 

- It has been decided that for having parity we will only ingest OSV + CSAF data; every other source of data will be disabled and will only be available after "V2"
- Currently the models can tell us that certain vulnerabilities can have multiple statuses like `fixed | note_affected | known_not_affected | affected`. My understanding is that we will only take into account "affected" statuses as in V1 we only show vulnerabilities that affect a package or SBOM.

I'd like to know exactly which steps I need to follow to see Relationships in the System: SBOM - Vulnerabilities / Package - Vulnerabilities, etc.

I expect the answer to be something like: "enable X,Y,Z importers". But I'd like the steps officially documented and give zero chance to misinterpretation.

The reason is that it is getting difficult to get "affected" packages with only the OSV + CSAF importers. Of course there is a high chance that I am doing something wrong but having an official statement about it will help not just me but anyone who tries Trustify.

ctron:
If we're not going to ingest CVE data, we might lack a bunch of information that we do have in V1.

ctron:
> What would be the recommended set of sources/data to import in order to have the same results as in V1?

Same as v1: SBOMs from Red Hat, Advisories from Red Hat, CVEs from https://github.com/CVEProject/cvelistV5

carlosthe19916:
Closing this PR. As the right set of importers will be defined by the Helm chart

PR #1084, opened 2024-12-06
feat: reduce default importers to only those used in V1

jcrossley3:
This will facilitate the transition of V1 users to V2.

Relates to #953


ctron (changes requested):
As this is never exposed to a v1 user, I am against merging this. To my understanding also @mid998 is relying on this sometimes.

jcrossley3:
I just think the essence of the feature is better conveyed through a minimum of default examples. Thirteen importers is overwhelming to a new user unfamiliar with the feature. Worse, the only sample SBOM importer doesn't even appear on the first page!

IMO the 10 importers I removed belong in the docs as examples accompanied by supporting text descriptions of each field.

ctron:
I agree that 10 importers might not be the best approach. On the other side, ripping out the existing other times leaves us without any example of using OSV, CWE, ClearlyDefined, … In order to compensate, we should improve docs to enable users to get there. Or, just stick to the existing importers, on which PM relies today.

PR #1086, opened 2024-12-09
fix: report 500 in case of storage errors, align with other errors

ctron:
Also: add instrumentation to `store` and align DB passwords.


helio-frota (commented):
nit: update readme.md changing eggs password

jcrossley3 (approved):

Issue #1089, opened 2024-12-09
`m0000720_alter_sbom_fix_null_array` fails with data


ctron:
```
trustify-migrate-1  | 2024-12-09T16:45:53.030996Z  INFO ThreadId(01) migrate: sea_orm_migration::migrator: Applying migration 'm0000720_alter_sbom_fix_null_array'
trustify-migrate-1  | 2024-12-09T16:45:53.031011Z DEBUG ThreadId(01) migrate: sea_orm::database::transaction: ALTER TABLE "sbom" ALTER COLUMN "data_licenses" TYPE text[], ALTER COLUMN "data_licenses" SET NOT NULL, ALTER COLUMN "data_licenses" SET DEFAULT NULL
trustify-migrate-1  | 2024-12-09T16:45:53.031956Z ERROR ThreadId(01) migrate: trustify_common::db: error=Execution Error: error returned from database: column "data_licenses" of relation "sbom" contains null values: error returned from database: column "data_licenses" of relation "sbom" contains null values: error returned from database: column "data_licenses" of relation "sbom" contains null values: column "data_licenses" of relation "sbom" contains null values
trustify-migrate-1  | Error: Execution Error: error returned from database: column "data_licenses" of relation "sbom" contains null values
trustify-migrate-1  | Caused by:
trustify-migrate-1  |   error returned from database: column "data_licenses" of relation "sbom" contains null values
trustify-migrate-1  |   error returned from database: column "data_licenses" of relation "sbom" contains null values
trustify-migrate-1  |   column "data_licenses" of relation "sbom" contains null values
```

PR #1091, opened 2024-12-10
fix: actually set a default value, instead of `null`

ctron:
Closes: https://github.com/trustification/trustify/issues/1089


JimFuller-RedHat (approved):
LGTM

PR #1097, opened 2024-12-12
chore: update rust

helio-frota:


ctron (changes requested):
I am ok with this, but the CI needs to be fixed. Looks like new clippy lints.

ctron (approved):

PR #1098, opened 2024-12-13
Reimplament vulnerability endpoint logic

dejanb:
The idea is to change the API to return list of purl statuses for every sbom that is returned.

The new statuses should be under advisory->sboms->purl_statuses

I intentionally left statuses for now in order not to break current UI. We can change that later on, once UI catches up

@carlosthe19916 Please give it a try and let me know if this is something you can work with

I'll go document and clean code some more next


carlosthe19916 (approved):
@dejanb Thanks for the quick PR. I tested it and it looks great. I can use that for the UI!!

> I intentionally left statuses for now in order not to break current UI. We can change that later on, once UI catches up

Thank, but after testing this PR I realized the amount of changes I will need to do in the UI is minimal. So we can either remove the current `statuses` field or put the current `purl_statuses` into that field. Whatever makes more sense for you!

Issue #1093, opened 2024-12-11
Got "Visitor error" when I canceled the importer 


helio-frota:
![2024-12-11_07-54](https://github.com/user-attachments/assets/35fa29cf-1a0f-4ca7-b097-8418a20d7f23)

maybe nice to have `Error:` instead of `Visitor error: ` ?

helio-frota:
I found a place when this is happening

```diff
-            .map_err(|err| ScannerError::Normal {
-                err: err.into(),
+            .map_err(|_err| ScannerError::Normal {
+                err: anyhow!("aaa"),
```

![2024-12-13_08-41](https://github.com/user-attachments/assets/52244f21-b251-4090-bac3-1d7f6928caff)


I'm trying to understand how to make the change in proper way

PR #1100, opened 2024-12-13
chore: Better error message for end users

helio-frota:
Related to #1093


ctron (changes requested):
Having an appropriate error message is good. But I am not a fan of post-processing messages.

There's a way this message gets created, and we are in full control of this. So we should ensure the message it good from the start.

helio-frota:
thanks for the review!
I agree :+1:

PR #1114, opened 2024-12-19
fix: use predictable ids for product related entities and additionally speed up csaf ingestion

dejanb:
This PR handles two things. 

First it removes select in loop that checks for preexisting purl statuses. Instead I added a bit of logic to return early from ingesting already ingested csaf document, which I believe is the case this was protecting from.

Additionally, now product related entities use predictable ids that was missing in yesterdays batch insert PR. There are still some workarounds in this area that will be tackled next, when the sam work is done for ingesting SBOMs.

This all result in further speedup of csaf ingestion, where `cve-2023-44487` goes locally from 50 to 20 sec.

The whole dataset 3 now ingests for 30 sec

```
time http POST localhost:8080/api/v1/dataset @etc/datasets/ds3.zip
Executed in   35.23 secs      fish           external
   usr time  152.97 millis    0.09 millis  152.89 millis
   sys time   66.68 millis    1.29 millis   65.40 millis   
```  

I'm still testing the importer, but it looks like 2024 RH CSAF should be imported in 1.5h, by the current pace.

### DIFF
--- a/entity/src/version_scheme.rs
+++ b/entity/src/version_scheme.rs
@@ -1,6 +1,19 @@
 use sea_orm::entity::prelude::*;
+use serde::{Deserialize, Serialize};
 
-#[derive(Copy, Clone, Eq, Hash, Debug, PartialEq, EnumIter, DeriveActiveEnum, strum::Display)]
+#[derive(
+    Copy,
+    Clone,
+    Eq,
+    Hash,
+    Debug,
+    PartialEq,
+    EnumIter,
+    DeriveActiveEnum,
+    strum::Display,
+    Serialize,
+    Deserialize,
+)]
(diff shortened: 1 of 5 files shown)


etc/deploy/compose/compose.yaml
```diff
@@ -2,7 +2,7 @@ services:
   postgres:
     image: docker.io/library/postgres:16
     ports:
-      - "5432:5432"
+      - "5433:5432"
```
helio-frota, line 5: :question:

dejanb: Fixed. Sorry about that, local change as I run other (guac) db at 5432

helio-frota (approved):

Issue #1099, opened 2024-12-13
/api/v1/purl/{id} does not contain `advisories.status.vulnerability.severity` field


carlosthe19916:
Fetching data from `/api/v1/purl/{id}` returns a response like the JSON below (I removed field and leave only the main ones):

```json
{
  "uuid": "269e3024-cbc2-5c63-9bb8-da09f3a5a067",
  "purl": "pkg:rpm/redhat/libnghttp2@1.33.0-5.el8_8?arch=x86_64",
  "advisories": [
    {
      "document_id": "CVE-2023-44487",
      "status": [
        {
          "vulnerability": {
            "normative": true,
            "identifier": "CVE-2023-44487",
            "title": "The HTTP/2 protocol allows a denial of service (server resource consumption) because request cancellation can reset many streams quickly, as exploited in the wild in August through October 2023.",
            "description": "The HTTP/2 protocol allows a denial of service (server resource consumption) because request cancellation can reset many streams quickly, as exploited in the wild in August through October 2023.",
            "reserved": "2023-09-29T00:00:00Z",
            "published": "2023-10-10T00:00:00Z",
            "modified": "2024-08-19T07:48:04.546Z",
            "withdrawn": null,
            "discovered": null,
            "released": null,
            "cwes": []
          },
          "status": "fixed"
        }
      ]
    }
  ],
  "licenses": []
}
```

The `advisories.status.vulnerability` object does not have "severity".

The reason to have "severity" is that there is the UI table below:

![Screenshot From 2024-12-13 15-06-22](https://github.com/user-attachments/assets/51c19b80-d78d-4f1d-a9d8-db9819e17910)

For each of the vulnerabilities found, there is the need to render the Severity of it. Currently the UI is fetching the severity for each Vulnerability calling to `/api/v1/vulnerability/{id}`. So, if the table has 10 elements then `/api/v1/vulnerability/{id}` will be called 10 times too.

If we were able to include the severity then we could significantly reduce the number of calls done to the backend.

@jcrossley3 tagging you here as you did previous work similar to this one to reduce the number of calls from the UI to the backend.

PR #1128, opened 2025-01-08
feat: add average_severity to PurlStatus

jcrossley3:
Fixes: #1099

### DIFF
--- a/modules/fundamental/src/purl/endpoints/test.rs
+++ b/modules/fundamental/src/purl/endpoints/test.rs
@@ -339,6 +339,10 @@ async fn package_with_status(ctx: &TrustifyContext) -> Result<(), anyhow::Error>
     let response: Value = app.call_and_read_body_json(request).await;
 
     assert_eq!(uuid, Uuid::parse_str(response["uuid"].as_str().unwrap())?);
+    assert_eq!(
+        "critical",
+        response["advisories"][0]["status"][0]["average_severity"]
+    );
 
     Ok(())
 }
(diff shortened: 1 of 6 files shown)


carlosthe19916 (approved):
Thanks for the enhancement. It works as expected.

Only one thing I noticed is that the "average_severity" is outside the "vulnerability" object. while in other endpoints the "average_severity|severity" is part of the "vulnerability" object itself (next to the "identifier" field). But the current response is more than enough for me :) 

```json
{
    "advisories": [
            {                
                "status": [
                    {
                        "vulnerability": {
                            "identifier": "CVE-2023-28867"                            
                        },
                        "average_severity": "high",
                        "status": "not_affected"
                    }
                ]
            }
    ]
}
```

jcrossley3:
I agree it's weird, but it was the least invasive solution, because that `vulnerability` object is re-used in other places that don't return severity. Because there is a cost to calculating severity, I didn't want to add it in all those other places. 

One trivial alternative is to "flatten" the structure, removing the `vulnerability` name, like so:
```
{
    "advisories": [
            {                
                "status": [
                    {
                        "identifier": "CVE-2023-28867"            
                        ...                
                        "average_severity": "high",
                        "status": "not_affected"
                    }
                ]
            }
    ]
}
```
Would that be better?

carlosthe19916:
Let's leave it as it is (I approved this PR already as it is) :)

Both options works fine technically (the current code and what you suggested). My comment was more on aesthetics

jcrossley3:
Ok, but I'm now thinking that it might be ok to add severity/score to that object everywhere it's used. It seems that, after this change, all but one place return the severity/score with the object, so I think we can safely just add those fields to it.

I'll do that in a separate PR, though.

PR #1132, opened 2025-01-10
feat[ai]: Add conversation endpoints.

chirino:
LLM internal state messages is simplified to be a single field either on ChatState or in the Conversation table. We now keep timestamps for individual ChatMessages.

Also drop the db connection arg from the completions method as the DB is Not needed for those calls.

You can test out the completions API using the UI at: https://github.com/chirino/ai-assistant-vite

just `npm i && npm run dev` and use the browser link outputted.

### DIFF
--- a/modules/fundamental/src/ai/endpoints/mod.rs
+++ b/modules/fundamental/src/ai/endpoints/mod.rs
@@ -2,15 +2,27 @@
 mod test;
 
 use crate::{
-    ai::model::{AiFlags, AiTool, ChatState},
-    ai::service::AiService,
+    ai::{
+        model::{AiFlags, AiTool, ChatState, Conversation, ConversationSummary},
+        service::AiService,
+    },
     Error,
 };
-use actix_http::header;
-use actix_web::{get, post, web, HttpResponse, Responder};
+use actix_web::{
+    delete, get,
+    http::header::{self, ETag, EntityTag, IfMatch},
+    post, put, web, HttpResponse, Responder,
+};
+
+use crate::ai::model::ChatMessage;
 use itertools::Itertools;
+use time::OffsetDateTime;
+use trustify_auth::authenticator::user::UserDetails;
 use trustify_auth::{authorizer::Require, Ai};
+use trustify_common::db::query::Query;
 use trustify_common::db::Database;
+use trustify_common::model::{Paginated, PaginatedResults};
+use uuid::Uuid;
 
 pub fn configure(config: &mut utoipa_actix_web::service_config::ServiceConfig, db: Database) {
     let service = AiService::new(db.clone());
@@ -19,7 +31,12 @@ pub fn configure(config: &mut utoipa_actix_web::service_config::ServiceConfig, d
         .service(completions)
         .service(flags)
         .service(tools)
-        .service(tool_call);
+        .service(tool_call)
+        .service(create_conversation)
+        .service(update_conversation)
+        .service(list_conversations)
+        .service(get_conversation)
+        .service(delete_conversation);
 }
 
 #[utoipa::path(
@@ -35,11 +52,10 @@ pub fn configure(config: &mut utoipa_actix_web::service_config::ServiceConfig, d
 #[post("/v2/ai/completions")]
 pub async fn completions(
     service: web::Data<AiService>,
-    db: web::Data<Database>,
     request: web::Json<ChatState>,
     _: Require<Ai>,
 ) -> actix_web::Result<impl Responder> {
-    let response = service.completions(&request, db.as_ref()).await?;
+    let response = service.completions(&request).await?;
     Ok(HttpResponse::Ok().json(response))
 }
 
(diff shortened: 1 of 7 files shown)


modules/fundamental/src/ai/endpoints/mod.rs
```diff
@@ -1,16 +1,22 @@
 #[cfg(test)]
 mod test;
 
+use crate::ai::model::{Conversation, ConversationSummary};
```
ctron, line 4: I would be great if you could group the `use` statements.

modules/fundamental/src/ai/endpoints/mod.rs
```diff
@@ -123,3 +133,236 @@ pub async fn tool_call(
...
+#[utoipa::path(
+    tag = "ai",
+    operation_id = "getConversation",
+    params(
+        ("id", Path, description = "Opaque ID of the conversation")
+    ),
+    responses(
+        (status = 200, description = "The resulting conversation", body = Conversation),
+        (status = 400, description = "The request was invalid"),
+        (status = 404, description = "The AI service is not enabled or the conversation was not found")
```
ctron, line 269: It looks to me like there's an inconsistency in generating this. When the conversion is not found, it returns "ok, but emtpy" if the conversion is found, but doesn't belong to the user, it returns "not found".

modules/fundamental/src/ai/service/mod.rs
```diff
@@ -205,156 +259,164 @@ impl AiService {
...
+        let internal_state = InternalState {
+            messages: history,
+            timestamps,
         };
-        Ok(model.insert(connection).await?)
+        Ok(internal_state)
     }
 
-    pub async fn update_conversation<C: ConnectionTrait>(
+    pub async fn upsert_conversation<C: ConnectionTrait>(
```
ctron, line 338: In the implementation is says that the update can take a while. How will that affect concurrent calls this this function. Which to my understanding can happen when multiple calls via the API happen.

ctron (changes requested):
Maybe a few things to understand or fix.

modules/fundamental/src/ai/endpoints/mod.rs
ctron: Can you add that then?

modules/fundamental/src/ai/endpoints/mod.rs
ctron: Ok, but in that case the part of "or the conversation was not found" is not true, is it?

modules/fundamental/src/ai/endpoints/mod.rs
chirino: I personally don't mind it being ungrouped, my IDE automatically adds them that way.  I also don't know how to get the format checker to enforce that.

modules/fundamental/src/ai/endpoints/mod.rs
chirino: well yes.  It actually means that the conversation belongs to another user, but I'd rather lie than tell them that.

modules/fundamental/src/ai/service/mod.rs
chirino: Could do.

modules/fundamental/src/ai/endpoints/mod.rs
chirino: Looks like https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#imports_granularity is what we would want.  but it's only in unstable right now.

modules/fundamental/src/ai/service/mod.rs
chirino: but given seq is part of the data model response object, seems odd to post it via headers.

ctron: The pattern of `If-Match` and `ETag` seems to be a standard thing.

modules/fundamental/src/ai/endpoints/mod.rs
ctron: But you're leaking that information anyway. Every time you get back a 404, it means "not allowed", or "other other".

modules/fundamental/src/ai/endpoints/mod.rs
chirino: If we can't automatically fix it for them /w cargo fmt, we should not be nitpicking about it in PR reviews.  This is all just stylistic right?

modules/fundamental/src/ai/service/mod.rs
chirino: I think that tend to be a thing when the etag is not part of the document.  Example: hash of a file etc.

chirino: But I guess it can't hurt to use the etag system. It should help with caching too.

ctron: Why invent something new, if there's an existing pattern?!

modules/fundamental/src/ai/endpoints/mod.rs
```diff
@@ -2,15 +2,27 @@
...
+use actix_web::{
+    delete, get,
+    http::header::{self, ETag, EntityTag, IfMatch},
+    post, put, web, HttpResponse, Responder,
+};
+
+use crate::ai::model::ChatMessage;
 use itertools::Itertools;
+use time::OffsetDateTime;
+use trustify_auth::authenticator::user::UserDetails;
```
ctron, line 20: I'd still prefer to have this nested. Or at least, consistent.

ctron (approved):

chirino:
will squash and rebase..

Issue #1112, opened 2024-12-18
uptick the api version to `2`


ctron:
Currently the REST APIs are mounted under `/api/v1`. However, we consider trustify "v2". So we should mount all REST APIs under `/api/v2` instead, just to be clear on that, and also to (maybe) allow compatibility APIs.

As that has a direct impact on the UI, maybe the first step would be to mount those APIs twice: once under `v1` once under `v2`.

carlosthe19916:
> As that has a direct impact on the UI, maybe the first step would be to mount those APIs twice: once under v1 once under v2.

For the UI I don't think it is needed. The backend can switch straight to `v2`. The backend will generate a new openapi.yaml file and the UI can quickly catch up with it.

ctron:
> What problem does this solve? I wouldn't expect an `/api/v2` endpoint to be served at a particular host:port without `/api/v1` also being mounted there. To me, trustification and trustify are two completely different upstream projects with completely different API's. Conflating the two _might_ be a downstream concern but I can't imagine anyone would want to deploy both of the upstream API's in a single downstream product, would they?
> 
> Maybe I'm not fully understanding the motivation for this?

See: https://github.com/trustification/trustify/pull/1133#issuecomment-2587202704

carlosthe19916:
> @carlosthe19916,
> 
> > The backend can switch straight to `v2`.
> 
> You meant the frontend can switch to use v2 directly ?

I meant that if the backend changes the REST paths then it will be reflected in the openapi.yaml the trustify repo has therefore triggering this action https://github.com/trustification/trustify/blob/main/.github/workflows/openapi.yaml

PR #1133, opened 2025-01-13
chore: change API version to v2

ctron:
BREAKING-CHANGE: This changes the prefix of the API from /api/v1 to /api/v2 as this is the successor API of trustification (v1).

Closes: https://github.com/trustification/trustify/issues/1112

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ zipped archive of SBOMs and/or Advisories like so:
 ```shell
 cd etc/datasets
 make
-http POST localhost:8080/api/v1/dataset @ds1.zip
+http POST localhost:8080/api/v2/dataset @ds1.zip
 ```
 
 #### Upload
@@ -44,8 +44,8 @@ There is an "Upload" menu option in the GUI: http://localhost:8080/upload
 You can also interact with the API directly in a shell:
 
 ```shell
-cat some-sbom.json | http POST localhost:8080/api/v1/sbom
-cat some-advisory.json | http POST localhost:8080/api/v1/advisory
+cat some-sbom.json | http POST localhost:8080/api/v2/sbom
+cat some-advisory.json | http POST localhost:8080/api/v2/advisory
 ```
 
 #### Importers
@@ -140,7 +140,7 @@ If you haven't already, [get started!](https://www.rust-lang.org/learn/get-start
 
 #### If test failures on OSX
 
-Potentially our concurrent Postgres installations during testing can
+Potentially, our concurrent Postgres installations during testing can
 exhaust shared-memory.  Adjusting shared-memory on OSX is not
 straight-forward.  Use [this
 guide](https://unix.stackexchange.com/questions/689295/values-from-sysctl-a-dont-match-etc-sysctl-conf-even-after-restart).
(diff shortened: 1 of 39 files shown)


jcrossley3 (changes requested):
This seems like an arbitrary, disruptive change that offers no real value. What problem are we solving?

ctron:
> This seems like an arbitrary, disruptive change that offers no real value. What problem are we solving?

We are solving the problem that people might consider `/api/v1` of RHTPA 1 compatible to `/api/v1` of RHTPA 2. Which it is not.

jcrossley3:
> > This seems like an arbitrary, disruptive change that offers no real value. What problem are we solving?
> 
> We are solving the problem that people might consider `/api/v1` of RHTPA 1 compatible to `/api/v1` of RHTPA 2. Which it is not.

What's RHTPA? :stuck_out_tongue_winking_eye: 

(apologies for being snarky -- I still cling to the idea this work may attract upstream contributors, and changing upstream to accommodate a downstream product makes me queasy)

ctron:
> > > This seems like an arbitrary, disruptive change that offers no real value. What problem are we solving?
> > 
> > 
> > We are solving the problem that people might consider `/api/v1` of RHTPA 1 compatible to `/api/v1` of RHTPA 2. Which it is not.
> 
> What's RHTPA? 😜
> 
> (apologies for being snarky -- I still cling to the idea this work may attract upstream contributors, and changing upstream to accommodate a downstream product makes me queasy)

Right, but even upstream, we do consider trustify the successor of trustification. So that would make sense here too.

gildub (commented):
It works fine meanwhile we might want to bump the version number of Trustify since this is a breaking change from 0.1.1 to 0.2.0 ?

ctron:
> It works fine meanwhile we might want to bump the version number of Trustify since this is a breaking change from 0.1.1 to 0.2.0 ?

Ideally yes. But then again, we never promised anything so far: https://github.com/trustification/trustify/blob/main/RELEASE.md#what-we-promise

gildub (approved):
LGTM

JimFuller-RedHat:
bumping the rest api version is a good idea for RHTPA v2 - a few questions :

1) Are we explicitly saying we will do not support multiple rest api versions from a same instance ? 
2) Could we not define version in one place and refer to that everywhere ... instead of bake in the string everywhere ... eg. if we have to make an exception we could bake into the string (tests are fine to do this).
3) What kind of churn for the ux does this cause ?

ctron:
> bumping the rest api version is a good idea for RHTPA v2 - a few questions :
> 
>     1. Are we explicitly saying we will do not support multiple rest api versions from a same instance ?

I don't think we are. But we are (currently) not supporting any `v1` API.

> 
>     2. Could we not define version in one place and refer to that everywhere ... instead of bake in the string everywhere ... eg. if we have to make an exception we could bake into the string (tests are fine to do this).

Might be tricky, as the string is part of a macro. And IIRC, string concatenation in macros is still tricky. We are however also using the macros to generate the openapi spec. So I'd hate so loose those.

> 
>     3. What kind of churn for the ux does this cause ?

I don't think it should any problems. The UI should simply use the newly generated client. All others have to replace `v1` with `v2`. Which we might need to do in some examples that we currently have.

JimFuller-RedHat (approved):
LGTM

PR #1134, opened 2025-01-13
refactor: dry up redundant test app creation logic

jcrossley3:
We have 3 places that are essentially doing the same thing. That thing is more correctly done in a single place and test context seemed the most appropriate.


JimFuller-RedHat (approved):
LGTM

ctron (approved):

Issue #1131, opened 2025-01-10
Denote src to binary rpm relationship from CYCLONEDX


jcrossley3:
This is a copy of https://issues.redhat.com/browse/TC-2050

When a cdx v1.6 SBOM is encoded with provides dependency denoting a relationship between a src rpm component and binary rpm component, ingesting into trustify should establish GENERATED_FROM trustify relationship.

```js
"components": [ 

     {packageA:srpm}, {packageB:binary_rpm}, ... ],

 "dependencies": [ 

     { "ref": packageA.bom_ref, 

        "provides": [packageB.bom_ref, ] } ]
```

pURL endpoint should expose GENERATED_FROM relationship

HTTP GET /api/v1/purl

HTTP GET /api/v1/purl?q={....}

```js
{

  "uuid": "0f525b2a-74ea-5f84-9555-55250c358861",

  "purl":  "pkg:rpm/redhat/openssl@3.0.7-18.el9_2?arch=x86_64",

  ...

  "generated_from": {[

    	"uuid": "fa88f614-d555-5088-bf47-0399fb182d35",

    	"purl": "pkg:rpm/redhat/openssl@3.0.7-18.el9_2?arch=src"

   }, ... ],

...

}
```

GENERATED_FROM relationship will be included on other  analysis graph endpoints.

HTTP GET /api/v1/analysis/root-component?q={...}

HTTP GET /api/v1/analysis/root-component/{purl|name}

HTTP GET /api/v1/analysis/dep?q=

HTTP GET /api/v1/analysis/dep/{purl|name}

```js
{
  	"sbom_id": "0193d5a4-a15c-76a0-8263-dc7334891dc8",
  	"node_id": "SPDXRef-e32409d9-5f5f-416c-96e5-e8d7b53957f3",
  	"purl": "pkg:rpm/redhat/openssl@3.0.7-18.el9_2?arch=x86_64",
        "generated_from":[
        {
      	    "purl": "pkg:rpm/redhat/openssl@3.0.7-18.el9_2?arch=src",
    	},
        ...],
...
}
```

jcrossley3:
Sample file here: https://gitlab.cee.redhat.com/brew/rh-koji-plugins/-/merge_requests/65/diffs#diff-content-ad4b10bba0f9157d82a2fcf54cd475e466430d48

Relevant section: 
```js
  "dependencies": [
    {
[...]

PR #1135, opened 2025-01-13
feat: create GeneratedFrom relationship from CycloneDX dependencies

jcrossley3:
Relates to #1131

I'm not entirely sure my right and left is correct, because I don't fully grok the expected output of `analysis/root-component` and `analysis/dep`.

### DIFF
--- a/modules/ingestor/src/graph/sbom/cyclonedx.rs
+++ b/modules/ingestor/src/graph/sbom/cyclonedx.rs
@@ -192,6 +192,17 @@ impl SbomContext {
             for right in left.depends_on.iter().flatten() {
                 creator.relate(right.clone(), Relationship::DependencyOf, left.ref_.clone());
             }
+
+            // https://github.com/trustification/trustify/issues/1131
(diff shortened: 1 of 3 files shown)


modules/ingestor/src/graph/sbom/cyclonedx.rs
```diff
@@ -192,6 +192,17 @@ impl SbomContext {
             for right in left.depends_on.iter().flatten() {
                 creator.relate(right.clone(), Relationship::DependencyOf, left.ref_.clone());
             }
+
+            // https://github.com/trustification/trustify/issues/1131
+            // Do we need to qualify this so that only "arch=src" refs
+            // get the GeneratedFrom relationship?
```
JimFuller-RedHat, line 198: good question - originally I thought we should think of 'red hat' mode as only applying to ingestion rules eg. we do not have to bake anything into rest api ... or add any additional metadata because we know a src vs binary package already.

JimFuller-RedHat (approved):
perfect, LGTM!

ctron (approved):

PR #1138, opened 2025-01-14
chore: update tokio

helio-frota:
* Past week I did a basic investigation with new tokio version and importers
  * We are using a tokio version from ~1 year ago
  * I need to run the importers with hotspot to see if we have differences with the tokio versions
    * result: nothing major changed


ctron:
I think it's ok to upgrade. But it's not that bad. We require tokio `1.30`, but in fact we are using `1.42`, as the `.lock` file shows.

`Cargo.toml` lists what is required, `Cargo.lock` what is being used. As we do check in the lock file, we need to refresh it. But that is done during the release process.

ctron (approved):

helio-frota:
you right https://github.com/trustification/trustify/blob/main/Cargo.lock#L7987 1.42
I didn't checked that :+1: 
thanks for the extras

Issue #1018, opened 2024-11-19
Terribly slow performance of vulnerabilties page


ctron:
Opening up https://trustify.dentrassi.de/vulnerabilities/CVE-1999-0001, it spins forever. Postgres goes crazy on the server side.

I think this is something which needs to be fixed. Not sure what the exact API endpoint is.

JimFuller-RedHat:
not very clear reviewing https://www.trustification.io/trustify-load-test-runs/ when this started ...

ctron:
those tests need to be fixed. I think one (or more) of the migrations are terribly slow too. Causing the timeout.

PR #1139, opened 2025-01-14
chore: perf improvements

JimFuller-RedHat:
Most visibly (and easy to adhoc test) this resolves list of vulns faster on http://localhost:8080/sboms listing and  http://localhost:8080/vulnerabilities.

Also various speed improvements in various places.

Addresses both #1018 and  #1140.

### DIFF
--- a/modules/fundamental/src/sbom/model/details.rs
+++ b/modules/fundamental/src/sbom/model/details.rs
@@ -47,6 +47,7 @@ impl SbomDetails {
         (sbom, node): (sbom::Model, Option<sbom_node::Model>),
         service: &SbomService,
         tx: &C,
+        statuses: Vec<String>,
     ) -> Result<Option<SbomDetails>, Error> {
         let mut relevant_advisory_info = sbom
             .find_related(sbom_package::Entity)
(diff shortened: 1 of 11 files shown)


JimFuller-RedHat:
@dejanb there is one failing test - because we need to make a decision if we parameterise this for status (currently baked in as 'affected' ) ... the current test needs to change but want to discuss with you first.

ctron (approved):

dejanb:
@JimFuller-RedHat I think we should propagate the filter all the way to the API, so that we can fetch only certain status like we need for this use case but allow folks to fetch all statuses if they need to.

modules/fundamental/src/sbom/model/details.rs
```diff
@@ -149,7 +151,7 @@ impl SbomDetails {
             JOIN "version_range" ON "product_version_range"."version_range_id" = "version_range"."id" AND version_matches("product_version"."version", "version_range".*)
 
             -- now find matching purls in these statuses
-            JOIN base_purl ON "product_status"."package" LIKE CONCAT("base_purl"."namespace", '/', "base_purl"."name") OR "product_status"."package" = "base_purl"."name"
+            JOIN base_purl ON product_status.package = base_purl.name OR product_status.package LIKE CONCAT(base_purl.namespace, '/%')
```
dejanb, line 154: @JimFuller-RedHat I believe this line should be

`JOIN base_purl ON product_status.package = base_purl.name OR product_status.package LIKE CONCAT(base_purl.namespace, '/', "base_purl"."name")`

Like this it returns too many false positives.

dejanb (approved):

Issue #1137, opened 2025-01-14
Denote cpe/pURL aliases on a single component from CYCLONEDX


ctron:
This is a copy of https://issues.redhat.com/browse/TC-2046

---

 When a cdx v1.6 SBOM is encoded with evidence/identity component metadata it denotes an additional CPE or pURL alias.

CPE:

```
 "components": [ 
      { "bom-ref": packageA.bom_ref 
        packageA...
        "cpe": main_cpe 
        "evidence": {
             "identity": [ 
                 {"field": "cpe",
                  "concludedValue": cpe_alias}, 
... ] }, ... ]
```

PURL:

```
 "components": [ 
   { "bom-ref": packageA.bom_ref packageA...
     "purl": main_purl.
     "evidence": { 
         "identity": [ 
            {"field": "purl", 
             "concludedValue": purl_alias}
              , ... ] 
      }, ... ]
```

**HTTP GET** /api/v1/cpe?q={....}
**HTTP GET** /api/v1/cpe/{cpe-id|alias}

```
 {
  "uuid": "0f525b2a-74ea-5f84-9555-55250c358861",
  "cpe":  "...",
  ...
  "alias": {[
        "uuid": "fa88f614-d555-5088-bf47-0399fb182d35",
        "cpe": "cpe_alias"
   }, ... ],
...
}
```

**HTTP GET** /api/v1/purl?q={....}
**HTTP GET** /api/v1/purl/{purl|alias|name}

```
 {
  "uuid": "0f525b2a-74ea-5f84-9555-55250c358861",
  "purl":  "pkg:oci/kernel-module-management-rhel9-operator@sha256%3Ad8...98?arch=amd64",
  ...
  "alias": {[
        "uuid": "fa88f614-d555-5088-bf47-0399fb182d35",
        "purl": "purl_alias"
   }, ... ],
...
}
```

ctron:
This should work the same way for SPDX SBOMs.

jcrossley3:
> This should work the same way for SPDX SBOMs.

Which I believe is the subject of https://issues.redhat.com/browse/TC-2047, correct?

ctron:
Here's a proposal for a change. If that's accepted, we need to update the JIRA for that too.

[...]

PR #1149, opened 2025-01-16
Allow searching by CPE and provide CPE information

ctron:
* [x] Handle the case of multiple CPEs and PURLs
* [x] Refactor away from a global, singleton instance
* [x] Fix an issue with nodes which aren't packages
* [x] Allow searching for CPEs with the analysis graph
* [x]  Allow querying the database for SBOMs by CPE
* [x] Add more unit tests

Closes https://github.com/trustification/trustify/issues/1137


common/Cargo.toml
```diff
@@ -23,6 +23,7 @@ log = { workspace = true }
 native-tls = { workspace = true }
 packageurl = { workspace = true }
 pem = { workspace = true }
+percent-encoding = { workspace = true }
```
JimFuller-RedHat, line 26: nice !

modules/analysis/README.md
```diff
@@ -0,0 +1,7 @@
+# Analysis Graph
```
JimFuller-RedHat, line 1: ya I should do another pass on this

server/src/profile/api.rs
```diff
@@ -47,9 +50,6 @@ use utoipa::{
 use utoipa_rapidoc::RapiDoc;
 use utoipa_redoc::{Redoc, Servable};
 
-#[cfg(feature = "garage-door")]
```
JimFuller-RedHat, line 50: curious - why does this go ?

JimFuller-RedHat (approved):
this is a big update - some of it concerned with much better code layout and I appreciate you taking the time/effort to clean up my crappy code ;) ... otherwise everything makes sense LGTM ... I would rather get this in so I can test at scale to make sure (before we do a new release).

ctron:
@chirino I think this PR might also partially close https://github.com/trustification/trustify/issues/1151, as it ingests the pedigree information.

JimFuller-RedHat (approved):
makes the code clearer, better perforrming and glad to see CPE surfaced up ... oh and thx for fixing global graph ... I still do not know what my rationale was but glad to see it back in the service LGTM

modules/fundamental/src/test/common.rs
```diff
@@ -1,3 +1,4 @@
+use trustify_module_analysis::service::AnalysisService;
```
jcrossley3, line 1: I've had other reasons for wanting to move analysis beneath fundamental, fwiw

PR #1156, opened 2025-01-17
Allow more relationships in analysis graph

JimFuller-RedHat:
We need to allow VariantOf and AncestorOf relationships in analysis graph.


chirino (approved):
lgtm

PR #1157, opened 2025-01-17
OSV loader: set properly VersionScheme in version_range

mrizzi:
Enhanced OSV loader to manage the (range)`type` and `ecosystem` properly when dealing with the `version_range.version_scheme_id` value.
In this way the GHSA advisories in `ds3` correlate with the `quarkus-bom` and in this regard a test has been added.


ctron (approved):

dejanb (approved):
Works great

PR #1159, opened 2025-01-17
fix: /api/v2/purl/{purl} now honors qualifiers, or lack thereof

### LINKED ISSUE #1158: bug: /api/v2/purl/{purl} doesn't always return the correct purl
It seems to ignore the qualifiers, e.g. if I request a purl qualified by `arch=src`, I may get one qualified by `arch=x86_64` because it's just returning the first one (of possibly many) it finds.

jcrossley3:
Fixes #1158


ctron (approved):
Looks good to me. Though I am not sure what the intention of the original implementation was.

PR #1161, opened 2025-01-20
refactor: re-use some code, add endpoint for single component

ctron:


JimFuller-RedHat (approved):
LGTM - I think we might want to consider immediate relationships on the nodes as well

PR #1164, opened 2025-01-20
License export

bxf12315:


ctron:
Please move the markdown file to `https://github.com/trustification/trustify/tree/main/docs/adrs`, name it accordingly, and only have the markdown file in there.

The idea of the ADR/RFC is to come up with an agreement before doing any code changes. So that we don't waste time and agree on the final solution in a wider team.

bxf12315:
> can we close this PR ... it does not make much sense to raise a lot of code before ADR has been discussed, reviewed and agreed upon.

Ok

JimFuller-RedHat:
thx we can re-open once ADR hsa gotten approved (just want to avoid confusion)

PR #1165, opened 2025-01-20
Add unit test affirming TC-2051

jcrossley3:
See https://issues.redhat.com/browse/TC-2051

This doesn't include the request to show relationships in the response of `/api/v2/purl` as there's a pending request to remove that requirement.


JimFuller-RedHat (approved):
LGTM

PR #1166, opened 2025-01-20
Add unit test affirming TC-2053

jcrossley3:
See https://issues.redhat.com/browse/TC-2053

This doesn't include relationships in the response of /api/v2/purl as there's a pending request to remove that requirement.


JimFuller-RedHat (approved):
great work, LGTM

Issue #1151, opened 2025-01-16
Denote upstream relationship from CycloneDX


chirino:
### Story:

As devsecops, I want to track upstream components.

When a cdx v1.6 SBOM is encoded with pedigree/ancestors objects denoting a relationship with an upstream component, ingesting into trustify should establish ANCESTOR_OF trustify relationship.

```
"components": [ 
 { upstream_component, 
   packageA... 
     "pedigree": 
         { "ancestors": [{upstream_component}]} 
}]
```

### Acceptance Criteria:

- View this information via api/v1/purl, api/v1/cpe, api/v1/analysis/root-component and api/v1/analysis/dep endpoints
- Existing support for sorting and querying extended
-  Expected Trustify internal relationship:
             upstream.package ANCESTOR_OF packageA 
-  Multiple and/or nested pedigree/ancestor relationships parsed from cdx MUST result in equivalent number of internal relationships
- Given an array of pedigree/ancestor components, the last component is the one true upstream - in the case that last item has nested ancestor we continue to apply the rules until we find the last item (of the last array) - every other component would be considered midstream.

### Search/Query:

pURL endpoint should expose this relationship

HTTP GET `/api/v2/purl`
HTTP GET `/api/v2/purl?q={....}`

```
{
  "uuid": "0f525b2a-74ea-5f84-9555-55250c358861",
  "purl": "pkg:rpm/redhat/basesystem@11-13.el9?arch=noarch",
 ...
  "relationship":{
     "ancestor_of":[{
         "uuid": "fa88f614-d555-5088-bf47-0399fb182d35",      
         "purl": "pkg:generic/basesystem"}, {} ... ],
   },
...
}
```

ANCESTOR_OF relationship should be included on other analysis graph endpoints.

HTTP GET `/api/v2/analysis/root-component?q={...}`
HTTP GET `/api/v2/analysis/root-component/{purl|name}`
_(ex. /api/v2/analysis/root-component/pkg%3Arpm/redhat/basesystem@11-13.el9%3Farch%3Dnoarch)_

```
{
   "sbom_id": "01947003-905b-70d1-bf49-76707fe75401",
   "node_id": "SPDXRef-013c6f62-83b1-43da-8f86-ad40a7397033",
   "purl": "pkg:rpm/redhat/basesystem@11-13.el9?arch=noarch",
   "name": "basesystem",
   ...
   "ancestors": [
        {
          "sbom_id": "01947003-905b-70d1-bf49-76707fe75401",
[...]

PR #1168, opened 2025-01-20
feat: create AncestorOf relationship from cdx pedigree/ancestors

jcrossley3:
Fixes #1151

PR #1169, opened 2025-01-21
Add unit tests for issue trustification#1151: Denote upstream relationship from CycloneDX

chirino:


chirino:
This resolves #1151

jcrossley3:
I guess you didn't see #1168 , huh?

One of us is wrong -- probably me -- because as I understood the issue, I needed to swap the left and right sides of the relationship in the ingestor. I do show the relationships in both `root-component` and `deps` but I could be passing the wrong purl's to each. :shrug: 

Also, that purl endpoint isn't correct -- it's only working because the `bom-ref`'s are actual purl's in the test-data, but I don't think we can rely on that, so @ctron suggested we not expose the relationships through that endpoint at all.

We can discuss further tomorrow.

chirino:
@jcrossley3 rebased off your changes, only thing this PR adds now is an additional unit test for the purl endpoint.

jcrossley3:
> @jcrossley3 rebased off your changes, only thing this PR adds now is an additional unit test for the purl endpoint.

I'm in the process of reverting the purl endpoint changes, per @JimFuller-RedHat, so no relationships will be returned from it. I think you can safely delete this PR. :smile:

PR #1176, opened 2025-01-21
fix: sbom license migration to update nulls

dejanb:
This is a fix for https://github.com/trustification/trustify-load-test-runs/issues/5

The fix is to use `USING COALESCE` to update NULL values before set the constraint. As I couldn't find an easy way to do that in sea-orm, I converted the migration to raw sql


JimFuller-RedHat (approved):
LGTM - though curious that any other existing tests around license do not bork on this change (or said tests are not changed in this PR) ... perhaps this indicates we need a few more tests ?

Issue #1140, opened 2025-01-14
Denote relationship between Product SBOM and other SBOM from SPDX


chirino:
This is a linked to https://issues.redhat.com/browse/TC-2049

### Story

As develops, I want to be able to ingest spdx SBOM that relates back to their Product SBOM (when defined in a separate spdx SBOM).

This relationship should be traverseable via the REST-API/UX.

When spdx v2.3 component SBOM is encoded with a relationship, ingesting into trustify should establish a PACKAGE_OF trustify relationship.

```json
{
  "spdxElementId": "SPDXRef-openssl-3.0.7-18.el9-2",
  "relationshipType": "PACKAGE_OF",
  "relatedSpdxElement": "SPDXRef-RHEL-9.2-EUS"
}
```
 
Alternatively, when spdx v2.3 product SBOM is encoded with an externalDocumentRefs, ingesting into trustify should establish a PACKAGE_OF trustify relationship.

```json
{
  "SPDXID" : "example-product-sbom",
  "spdxVersion": "SPDX-2.3",
  "dataLicense": "CC0-1.0",
  "created": "2023-10-01T12:00:00Z",
  "externalDocumentRefs": [
    {
      "referenceDocumentIdentifier": "SPDXRef-other-sbom",
      "referenceCategory": "PACKAGE",
      "referenceType": "SPDX",
      "referenceLocator": "https://example.com/path/to/other-spdx-document.spdx"
    }
  ],
  ...
}
```

### Acceptance Criteria:

- View this information on api/v1/sbom, api/v1/analysis/root-component and api/v1/analysis/dep endpoints
- List components for specific product (by CPE)
- Existing support for sorting and querying extended
- Expected Trustify internal relationship: pkg:oci/ose-cluster PACKAGE_OF cpe:/a:redhat:openshift

### Search/Query

HTTP GET `/api/v1/sbom`
HTTP GET `/api/v1/sbom?q={...}`

```json
{
  "id": "urn:uuid:01939af7-697e-7242-99d5-cbf11729fa79",
  "document_id": "https://www.redhat.com/comp.json",
  "published": "2006-08-14T02:34:56Z",
  "name": "kernel-module-management-operator-container-1.1.2-25",
  ...,

  "package_of": [
    {
      "id": "SPDXRef-image-index",
      "name": "kernel-module-management-product",
      "version": "1.1.2-25",
      "purl": [
        {
          "uuid": "2771e255-2ba5-5577-ab2c-3c43073f8841",
[...]

PR #14, opened 2024-03-06
Remove kubectl commands as it is early to integrate the UI to remote k8s-backend

carlosthe19916:
Fixes https://github.com/trustification/trustify/issues/12

PR #1179, opened 2025-01-21
Add support for spdx "relationshipType": "PACKAGE_OF"

chirino:
Part of issue #1140

The relationships table did not have records for #14 DescribedBy and #15 PackageOf

### DIFF
--- a/migration/src/m0000840_add_relationship_14_15.rs
+++ b/migration/src/m0000840_add_relationship_14_15.rs
@@ -1,0 +1,45 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let data = [(14, "DescribedBy"), (15, "PackageOf")];
+
+        for (id, description) in data {
+            let insert = Query::insert()
+                .into_table(Relationship::Table)
+                .columns([Relationship::Id, Relationship::Description])
+                .values_panic([id.into(), description.into()])
+                .to_owned();
+
+            manager.exec_stmt(insert).await?;
+        }
+
+        Ok(())
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let data = [(14, "DescribedBy"), (15, "PackageOf")];
+
+        for (id, _) in data {
+            let insert = Query::delete()
+                .from_table(Relationship::Table)
+                .and_where(Expr::col(Relationship::Id).lt(id))
+                .to_owned();
+
+            manager.exec_stmt(insert).await?;
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(DeriveIden)]
+pub enum Relationship {
+    Table,
+    Id,
+    Description,
+}
--- a/migration/src/m0000840_add_relationship_14_15.rs
+++ b/migration/src/m0000840_add_relationship_14_15.rs
@@ -2,13 +2,11 @@ use sea_orm_migration::prelude::*;
 
 #[derive(DeriveMigrationName)]
 pub struct Migration;
-
+const DATA: [(i32, &str); 2] = [(14, "DescribedBy"), (15, "PackageOf")];
 #[async_trait::async_trait]
 impl MigrationTrait for Migration {
     async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
-        let data = [(14, "DescribedBy"), (15, "PackageOf")];
-
-        for (id, description) in data {
+        for (id, description) in DATA {
             let insert = Query::insert()
                 .into_table(Relationship::Table)
                 .columns([Relationship::Id, Relationship::Description])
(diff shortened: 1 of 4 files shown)


JimFuller-RedHat:
I think we have support in that if an spdx defines this relationship it will appear in analysis graph - if not then that is a bug

chirino:
PR has a test.  Fix is included.

modules/ingestor/src/graph/sbom/spdx.rs
```diff
@@ -251,31 +251,32 @@ impl<'spdx> TryFrom<(&'spdx str, &'spdx RelationshipType, &'spdx str)> for SpdxR
...
+            RelationshipType::Describes => Ok((right, Relationship::DescribedBy, left)),
             RelationshipType::DevDependencyOf => Ok((left, Relationship::DevDependencyOf, right)),
+            RelationshipType::DevToolOf => Ok((left, Relationship::DevToolOf, right)),
+            RelationshipType::ExampleOf => Ok((left, Relationship::ExampleOf, right)),
+            RelationshipType::GeneratedFrom => Ok((left, Relationship::GeneratedFrom, right)),
+            RelationshipType::Generates => Ok((right, Relationship::GeneratedFrom, left)),
             RelationshipType::OptionalDependencyOf => {
                 Ok((left, Relationship::OptionalDependencyOf, right))
             }
+            RelationshipType::PackageOf => Ok((right, Relationship::PackageOf, left)),
```
chirino, line 271: This is the real change.  The rest is just me sorting the list of cases so that it's easier to spot missing ones in the future.

migration/src/m0000840_add_relationship_14_15.rs
```diff
@@ -0,0 +1,45 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
```
jcrossley3, line 5: nit: make the data a global constant so you don't have to repeat it in 2 places.
```suggestion
const DATA: [(i32, &str); 2] = [(14, "DescribedBy"), (15, "PackageOf")];
```

migration/src/m0000840_add_relationship_14_15.rs
```diff
@@ -0,0 +1,45 @@
...
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let data = [(14, "DescribedBy"), (15, "PackageOf")];
+
+        for (id, description) in data {
```
jcrossley3, line 11:
```suggestion
        for (id, description) in DATA {
```

migration/src/m0000840_add_relationship_14_15.rs
```diff
@@ -0,0 +1,45 @@
...
+            manager.exec_stmt(insert).await?;
+        }
+
+        Ok(())
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let data = [(14, "DescribedBy"), (15, "PackageOf")];
+
+        for (id, _) in data {
```
jcrossley3, line 27:
```suggestion
        for (id, _) in DATA {
```

modules/analysis/src/endpoints/test.rs
```diff
@@ -613,3 +613,45 @@ async fn issue_tc_2054(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
 
     Ok(())
 }
+
+#[test_context(TrustifyContext)]
+#[test(actix_web::test)]
+async fn issue_spdx_package_of(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
```
jcrossley3, line 619: another nit, inspired by you, actually :wink: 
```suggestion
async fn spdx_package_of(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
```

modules/analysis/src/endpoints/test.rs
```diff
@@ -613,3 +613,45 @@ async fn issue_tc_2054(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
...
+        .as_array()
+        .into_iter()
+        .flatten()
+        .filter(|m| {
+            m == &&json!({
+                "sbom_id": sbom["sbom_id"],
+                "node_id": "SPDXRef-83c9faa0-ca85-4e48-9165-707b2f9a324b",
+                "relationship": "PackageOf",
+                "purl": [],
+                "cpe": m["cpe"], // long list assume it's correct
```
jcrossley3, line 646: I'm good if you are, but worth filtering on something to target a specific element to make sure it's there?

jcrossley3 (approved):
lgtm. act on my nits at your discretion.

modules/analysis/src/endpoints/test.rs
chirino: Isn't this doing filtering?

jcrossley3: I was referring to the comment. Is there anything about the "long list" that you might possibly filter to find something in it that verifies the test? Or is enough just to assume it's correct if it's "long"? It may be overkill, but the comment gives me pause.

chirino: oh I see, yeah I think including that would just add a bunch of noise to the test.  I think matching on the other object properties is plenty good enough.

PR #1182, opened 2025-01-22
fix: report CPE parsing as "bad request"

ctron:
Right now, it is being reported as an internal server error.


helio-frota (approved):

JimFuller-RedHat (approved):
LGTM

PR #1186, opened 2025-01-22
Add tests for spdx "relationshipType": "PACKAGE_OF"

chirino:
Verify the relationship type shows up in 
`/api/v2/analysis/root-component` API calls.

Part of issue #1140

### DIFF
--- a/modules/analysis/src/endpoints/test.rs
+++ b/modules/analysis/src/endpoints/test.rs
@@ -654,7 +654,7 @@ async fn spdx_package_of(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
     let uri = format!("/api/v2/analysis/dep/{}", urlencoding::encode(purl));
     let request: Request = TestRequest::get().uri(&uri).to_request();
     let response: Value = app.call_and_read_body_json(request).await;
-    log::debug!("{response:#?}");
+    log::debug!("{}", serde_json::to_string_pretty(&response)?);
 
     let sbom = &response["items"][0];
     let matches: Vec<_> = sbom["deps"]
@@ -671,6 +671,34 @@ async fn spdx_package_of(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
                 "name": "SATELLITE-6.15-RHEL-8",
                 "version": "6.15",
                 "deps": [],
+            })
+        })
+        .collect();
+
+    assert_eq!(1, matches.len());
+
+    let uri = format!(
+        "/api/v2/analysis/root-component?q={}",
+        urlencoding::encode("SATELLITE-6.15-RHEL-8")
+    );
+    let request: Request = TestRequest::get().uri(&uri).to_request();
+    let response: Value = app.call_and_read_body_json(request).await;
+    log::info!("{}", serde_json::to_string_pretty(&response)?);
+
+    let sbom = &response["items"][0];
+    let matches: Vec<_> = sbom["ancestors"]
+        .as_array()
+        .into_iter()
+        .flatten()
+        .filter(|m| {
+            m == &&json!({
+              "sbom_id":  sbom["sbom_id"],
+              "node_id": m["node_id"],
+              "relationship": "PackageOf",
+              "purl": m["purl"], // long list assume it's correct
+              "cpe": m["cpe"], // long list assume it's correct
+              "name": "rubygem-google-cloud-compute",
+              "version": "0.5.0-1.el8sat"
             })
         })
         .collect();
--- a/modules/analysis/src/endpoints/test.rs
+++ b/modules/analysis/src/endpoints/test.rs
@@ -6,7 +6,7 @@ use serde_json::{json, Value};
 use std::collections::HashMap;
 use test_context::test_context;
 use test_log::test;
-use trustify_test_context::{call::CallService, TrustifyContext};
+use trustify_test_context::{call::CallService, subset::ContainsSubset, TrustifyContext};
 
 #[test_context(TrustifyContext)]
 #[test(actix_web::test)]
@@ -662,16 +662,11 @@ async fn spdx_package_of(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
         .into_iter()
         .flatten()
         .filter(|m| {
-            m == &&json!({
-                "sbom_id": sbom["sbom_id"],
-                "node_id": "SPDXRef-83c9faa0-ca85-4e48-9165-707b2f9a324b",
+            m.contains_subset(json!({
                 "relationship": "PackageOf",
-                "purl": [],
-                "cpe": m["cpe"], // long list assume it's correct
                 "name": "SATELLITE-6.15-RHEL-8",
                 "version": "6.15",
-                "deps": [],
-            })
+            }))
         })
         .collect();
 
@@ -683,7 +678,7 @@ async fn spdx_package_of(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
     );
     let request: Request = TestRequest::get().uri(&uri).to_request();
     let response: Value = app.call_and_read_body_json(request).await;
-    log::info!("{}", serde_json::to_string_pretty(&response)?);
+    log::debug!("{}", serde_json::to_string_pretty(&response)?);
 
     let sbom = &response["items"][0];
     let matches: Vec<_> = sbom["ancestors"]
@@ -691,15 +686,11 @@ async fn spdx_package_of(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
         .into_iter()
         .flatten()
         .filter(|m| {
-            m == &&json!({
-              "sbom_id":  sbom["sbom_id"],
-              "node_id": m["node_id"],
+            m.contains_subset(json!({
               "relationship": "PackageOf",
-              "purl": m["purl"], // long list assume it's correct
-              "cpe": m["cpe"], // long list assume it's correct
               "name": "rubygem-google-cloud-compute",
               "version": "0.5.0-1.el8sat"
-            })
+            }))
         })
         .collect();
 
--- a/modules/analysis/src/endpoints/test.rs
+++ b/modules/analysis/src/endpoints/test.rs
@@ -656,21 +656,15 @@ async fn spdx_package_of(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
     let response: Value = app.call_and_read_body_json(request).await;
     log::debug!("{}", serde_json::to_string_pretty(&response)?);
 
-    let sbom = &response["items"][0];
-    let matches: Vec<_> = sbom["deps"]
-        .as_array()
-        .into_iter()
-        .flatten()
-        .filter(|m| {
-            m.contains_subset(json!({
+    assert!(response.contains_deep_subset(json!({
+        "items": [ {
+            "deps": [ {
                 "relationship": "PackageOf",
                 "name": "SATELLITE-6.15-RHEL-8",
                 "version": "6.15",
-            }))
-        })
-        .collect();
-
-    assert_eq!(1, matches.len());
+            }]
+        }]
+    })));
 
     let uri = format!(
         "/api/v2/analysis/root-component?q={}",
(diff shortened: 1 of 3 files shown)


modules/analysis/src/endpoints/test.rs
```diff
@@ -677,5 +677,33 @@ async fn spdx_package_of(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
...
+        .filter(|m| {
+            m == &&json!({
+              "sbom_id":  sbom["sbom_id"],
+              "node_id": m["node_id"],
+              "relationship": "PackageOf",
+              "purl": m["purl"], // long list assume it's correct
+              "cpe": m["cpe"], // long list assume it's correct
+              "name": "rubygem-google-cloud-compute",
+              "version": "0.5.0-1.el8sat"
+            })
```
jcrossley3, line 702: To me, this is a brittle test. I'd want my expectation to be the minimum required to affirm the test. If some future change adds or takes away one of the fields that has nothing to do with this test, it's still gonna break. Is that good or annoying? I'd rather something like this, maybe even omitting `name` and `version`, too:
```suggestion
            m["sbom_id"] == sbom["sbom_id"]
                && m["relationship"] == "PackageOf"
                && m["name"] == "rubygem-google-cloud-compute"
                && m["version"] == "0.5.0-1.el8sat"
```

modules/analysis/src/endpoints/test.rs
```diff
@@ -654,7 +654,7 @@ async fn spdx_package_of(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
     let uri = format!("/api/v2/analysis/dep/{}", urlencoding::encode(purl));
     let request: Request = TestRequest::get().uri(&uri).to_request();
     let response: Value = app.call_and_read_body_json(request).await;
-    log::debug!("{response:#?}");
+    log::debug!("{}", serde_json::to_string_pretty(&response)?);
```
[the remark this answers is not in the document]
chirino: 100%

modules/analysis/src/endpoints/test.rs
chirino: I kinda was thinking about similar lines.  Is there an existing function that can test if an actual Value matches a partial set of fields of another expected Value?  The assertion would become more concise and less brittle then.

jcrossley3: Not that I know of. You can always downcast the `Value` to an `AncestorSummary` or `DepSummary`, but I'm not sure `m["name"]` is all that better or worse than `m.name`.

Another option is to use [`Value::pointer`](https://docs.rs/serde_json/latest/serde_json/enum.Value.html#method.pointer), but I haven't personally tried that.

modules/analysis/src/test.rs
```diff
@@ -3,11 +3,64 @@ use crate::{
...
+            for (key, value_a) in expected {
+                if Some(value_a) != actual.get(key.as_str()) {
+                    return false;
+                }
+            }
+            true
+        }
+        _ => false,
+    }
+}
```
jcrossley3, line 25: This is a very cool idea! I want it to do more, though, so forgive me for challenging you. :smile: 

The function expects `Value`'s, but fails if they're not a specific type of `Value`. Make it work for any kind of `Value`. If it's an object, then use recursion to compare the key/value pairs. Probably want to rename the function to `is_same` or `subset` or `contains` or some such.

You'll essentially have two branches: 
```
pub fn is_subset(actual: &Value, expected: &Value) -> bool {
    if expected.is_object() {
            expected.iter().all(|(k, v)| is_subset(&actual[k], v))
    } else {
        expected == actual
    }
}
``` 
And no, that won't actually compile due to `Value`'s overly-complicated API, but you can make it work!

And this will be useful all over, so let's stick it somewhere in `test-context` maybe?

jcrossley3 (changes requested):
I like where you're going, but you haven't gone far enough!!!

modules/analysis/src/test.rs
jcrossley3: Even better...
```rust
pub trait Contains {
    fn contains(&self, subset: Value) -> bool;
}

impl Contains for Value {
    fn contains(&self, subset: Value) -> bool {
        match (self.as_object(), subset.as_object()) {
            (Some(src), Some(tgt)) => tgt
                .iter()
                .all(|(k, v)| src.get(k).is_some_and(|x| x.contains(v.clone()))),
            _ => subset == *self,
        }
    }
}
```
Accepting a reference to self and taking ownership of `subset` allows you to clean up your calling code a bit, e.g.
```rust
        .filter(|m| {
            m.contains(json!({
              "relationship": "PackageOf",
              "name": "rubygem-google-cloud-compute",
              "version": "0.5.0-1.el8sat"
            }))
        })
```
Deciding to take a reference instead of ownership would be determined by whether we think most of our test subsets will be "one-off's". If we often re-use the same one, we might take a reference to avoid having to `.clone()` the subset each time we pass it. But I tend to think it's more likely we'll call `json!` every time we call `.contains` so taking ownership seems reasonable.

Make sense?

jcrossley3: I couldn't help myself...

If you add a branch for `Value::Array` types...
```rust
impl Contains for Value {
    fn contains(&self, subset: Value) -> bool {
        match (self, &subset) {
            (Value::Object(src), Value::Object(tgt)) => tgt
                .iter()
                .all(|(k, v)| src.get(k).is_some_and(|x| x.contains(v.clone()))),
            (Value::Array(src), Value::Array(tgt)) => tgt
                .iter()
                .all(|v| src.iter().any(|x| x.contains(v.clone()))),
            _ => subset == *self,
        }
    }
}
```
You can assert things like this:
```rust
    assert!(response.contains(json!({
        "items": [
            {
                "deps": [
                    {
                        "relationship": "PackageOf",
                        "name": "SATELLITE-6.15-RHEL-8",
                        "version": "6.15",
                    }
                ]
            }
        ]
    })));
```
There's a tradeoff, though. Sometimes you might want arrays to match on the exact contents, e.g. you might want to assert that `"purls": ["pkg:blah"]` has exactly one element in it, and removing that `Value::Array` arm would do that.

chirino: Great ideas.. my next commit will have a version of it.

test-context/src/lib.rs
```diff
@@ -202,10 +203,49 @@ where
     Ok((tokio::task::spawn_blocking(f).await??, digests))
 }
 
+pub trait ContainsSubset {
+    // Returns true if the value is a subset of the receiver.
+    fn contains_subset(&self, value: Value) -> bool;
+    // Returns true if the value a deep subset of the receiver.
+    fn contains_deep_subset(&self, value: Value) -> bool;
```
chirino, line 210: @jcrossley3 did a shallow and a deep version so that the caller can choose how strict the field matching is.

jcrossley3: Very nice! I might suggest putting the trait, impl and tests in another module, maybe `test-context/src/subset.rs`? `lib.rs` is getting kinda crowded, I think.

test-context/src/subset.rs
```diff
@@ -0,0 +1,90 @@
...
+impl ContainsSubset for Value {
+    fn contains_subset(&self, value: Value) -> bool {
+        match (self, &value) {
+            (Value::Object(src), Value::Object(subset)) => subset
+                .iter()
+                .all(|(k, v)| src.get(k).is_some_and(|x| x == v)),
+
+            (Value::Array(src), Value::Array(subset)) => {
+                subset.iter().all(|v| src.iter().any(|x| x == v))
+            }
```
jcrossley3, line 19: I don't think this is necessary, because `contains_deep_subset` already does this. We want to be able to pass in an array and have it match explicitly, e.g. ` m.contains_subset(json!({"purl": [ "pkg:rpm/redhat/rubygem-google-cloud-env@1.6.0-1.el8sat?arch=src" ]}))`

test-context/src/subset.rs
```diff
@@ -0,0 +1,90 @@
...
+        let actual = json!([1, 2, 3]);
+        assert!(actual.contains_subset(json!([2])));
+
+        // other values can be interleaved.
+        let actual = json!([1, 2, 3]);
+        assert!(actual.contains_subset(json!([1, 3])));
+
+        // case where a value is missing
+        let actual = json!([1, 2, 3]);
+        assert!(!actual.contains_subset(json!([0])));
```
jcrossley3, line 88: Maybe "strict" or "fuzzy" or "partial" are better descriptors than "deep". The crucial idea is that we need both a way to test for an explicit match and a way to ask "are all of these in the other one?"

I think this test is more accurate thusly:
```suggestion
        let actual = json!([1, 2, 3]);

        assert!(actual.contains_subset(json!([1, 2, 3])));
        assert!(!actual.contains_subset(json!([2])));

        assert!(actual.contains_deep_subset(json!([2])));
        assert!(actual.contains_deep_subset(json!([1, 3])));
        assert!(!actual.contains_deep_subset(json!([0])));
```

But again, I'm not sure "deep" is the right word. We should always recurse through a recursive structure.

jcrossley3 (changes requested):
We're getting there!

test-context/src/subset.rs
chirino: This is so that: 

```rust
let actual = json!([{"a":1}]);
actual.contains_subset(json!([{"a":1, "b":2}])); // should return false
```

if we used contains_deep_subset then it would return true.

test-context/src/subset.rs
chirino: Yeah. I'd be happy with any of those options.

modules/analysis/src/endpoints/test.rs
```diff
@@ -654,24 +654,43 @@ async fn spdx_package_of(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
...
+
+    let uri = format!(
+        "/api/v2/analysis/root-component?q={}",
+        urlencoding::encode("SATELLITE-6.15-RHEL-8")
+    );
+    let request: Request = TestRequest::get().uri(&uri).to_request();
+    let response: Value = app.call_and_read_body_json(request).await;
+    log::debug!("{}", serde_json::to_string_pretty(&response)?);
+
+    let sbom = &response["items"][0];
```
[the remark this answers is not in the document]
chirino: your right.. lets do it the easy way.

test-context/src/subset.rs
jcrossley3: I would expect the absence of `b` to cause that to return false. I'm not sure how that `Value::Array` branch would even come into play here. What am I missing?

jcrossley3: confirmed:
```rust
        let actual = json!([{"a":1}]);
        assert!(!actual.contains_subset(json!([{"a":1, "b":2}])));
        assert!(!actual.contains_deep_subset(json!([{"a":1, "b":2}])));
```
Maybe add that to `test_array_subset`?

jcrossley3: This works as I'd expect, too:
```rust
        let actual = json!([{"a":1, "b": 2}]);
        assert!(!actual.contains_subset(json!([{"a":1}])));
        assert!(actual.contains_deep_subset(json!([{"a":1}])));
```
Maybe `contains_partial_subset` seems best in light of those?

jcrossley3:
Looking at it again after my walk, I noticed that `contains_subset` isn't recursive. This makes the "deep" in the other fn name make more sense, but IMO it makes `contains_subset` less useful. I don't see much added value in a non-recursive comparison over just doing the comparisons explicitly inside the `.filter(...)` fn.

I do find the `contains_deep_subset` useful as implemented, and I'd probably just rename that to `contains_subset` and have a single fn in the trait.

But I'll approve the PR and leave it to you. As we integrate the feature into more tests, hopefully The Right Way will reveal itself.

jcrossley3 (approved):

chirino:
@jcrossley3 thanks for all the help and guidance on this PR.  I'll queue to merge as is, but feel free to rename those functions  if you have a better option.  Naming is hard.  Additional options:

- contains_all : for the deep version: the _all makes it more obvious the passed parameter should be a collection.
- contains_all_exact/strict: for the the more strict version.

jcrossley3:
> Naming is hard.

Amen! I enjoyed the collaboration!

PR #1189, opened 2025-01-22
feat: Seeing OTEL traces at development time

helio-frota:
* Adds a compose with Jaeger and OTELCOL
* Fixes a problem with actix-web tracing

### DIFF
--- a/common/infrastructure/Cargo.toml
+++ b/common/infrastructure/Cargo.toml
@@ -23,7 +23,7 @@ mime = { workspace = true }
 openssl = { workspace = true }
 opentelemetry = { workspace = true }
 opentelemetry-otlp = { workspace = true }
-opentelemetry_sdk = { workspace = true, features = ["rt-tokio"] }
+opentelemetry_sdk = { workspace = true, features = ["rt-tokio-current-thread"] }
 parking_lot = { workspace = true }
 prometheus = { workspace = true }
 reqwest = { workspace = true }
--- a/common/infrastructure/src/tracing.rs
+++ b/common/infrastructure/src/tracing.rs
@@ -123,13 +123,13 @@ fn init_otlp(name: &str) {
             "service.name",
             name.to_string(),
         )]))
-        .with_batch_exporter(exporter, opentelemetry_sdk::runtime::Tokio)
+        .with_batch_exporter(exporter, opentelemetry_sdk::runtime::TokioCurrentThread)
         .with_sampler(opentelemetry_sdk::trace::Sampler::ParentBased(Box::new(
             sampler(),
         )))
         .build();
 
-    println!("Using Jaeger tracing.");
+    println!("Using OTEL Collector with Jaeger as the back-end.");
     println!("{:#?}", provider);
 
     let formatting_layer = tracing_subscriber::fmt::Layer::default();
@@ -142,6 +142,7 @@ fn init_otlp(name: &str) {
     {
         eprintln!("Error initializing tracing: {:?}", e);
     }
+    opentelemetry::global::set_tracer_provider(provider);
 }
 
 fn init_no_tracing() {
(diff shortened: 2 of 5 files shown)


common/infrastructure/Cargo.toml
```diff
@@ -23,7 +23,7 @@ mime = { workspace = true }
 openssl = { workspace = true }
 opentelemetry = { workspace = true }
 opentelemetry-otlp = { workspace = true }
-opentelemetry_sdk = { workspace = true, features = ["rt-tokio"] }
+opentelemetry_sdk = { workspace = true, features = ["rt-tokio-current-thread"] }
```
helio-frota, line 26: > rt-tokio-current-thread: Spawn telemetry tasks on a separate runtime so that the main runtime won’t be blocked. [link 1](https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/#crate-feature-flags), [link 2](https://github.com/OutThereLabs/actix-web-opentelemetry?tab=readme-ov-file#exporter-configuration)

common/infrastructure/src/tracing.rs
```diff
@@ -123,15 +123,12 @@ fn init_otlp(name: &str) {
...
             name.to_string(),
         )]))
-        .with_batch_exporter(exporter, opentelemetry_sdk::runtime::Tokio)
+        .with_batch_exporter(exporter, opentelemetry_sdk::runtime::TokioCurrentThread)
         .with_sampler(opentelemetry_sdk::trace::Sampler::ParentBased(Box::new(
             sampler(),
         )))
         .build();
 
-    println!("Using Jaeger tracing.");
```
ctron, line 132: I'd love to keep those. As in the case here, we would not get logs on the console and in the pod logs, we would never know "why". With those, we do.

ctron (changes requested):
I love the idea, and thanks for providing all the info.

One small thing, I'd like to keep the `println` in this case. It's an exception.

common/infrastructure/src/tracing.rs
helio-frota: np I deleted like normal printlns I'll add it back :+1:

docs/design/log_tracing.md
```diff
@@ -135,3 +135,26 @@ However, not all function variants might expose the same behavior. Just because
...
+Database:
+
+```shell
+podman compose -f etc/deploy/compose/compose.yaml up
+```
+
+Trustify with traces:
+
+```shell
+OTEL_TRACES_SAMPLER_ARG=1 OTEL_EXPORTER_OTLP_ENDPOINT="http://localhost:4317" cargo run --bin trustd api --db-password trustify --devmode --auth-disabled --tracing enabled
```
jcrossley3, line 156: How/why is the behavior different without `--devmode`?

helio-frota: * Good question, related to otel-traces it continues the same without `--devmode`.
* The circles in red are related to errors like this:
  * I'm not 100% following the recent discussions and/or changes around pm-mode and devmode
  * "Devmode" in this PR title is in the meaning of looking at the traces during development, instead of deployment (to reduce the pain when ensuring traces are working)
  
```console
2025-01-23T14:18:12.772942Z ERROR trustify_module_fundamental::error: Query Error: error returned from database: relation "sbom" does not exist
2025-01-23T14:18:12.773937Z ERROR trustify_module_fundamental::error: Query Error: error returned from database: relation "advisory" does not exist
```

jcrossley3: If it works the same without `--devmode` please remove that switch from the doc. Only include the options essential for the behavior described by this PR.

helio-frota: PR updated thanks for the review :+1:

ctron:
I think the commit should be reworded as well, as long as it's not about devmode. Which is my current understanding.

helio-frota:
@ctron title and commit message updated :+1:

ctron (approved):

PR #1190, opened 2025-01-23
adr for conceptual view

JimFuller-RedHat:
Rendered version: https://github.com/trustification/trustify/blob/adr-analysis-conceptual/docs/adrs/00002-analysis-graph.md

### DIFF
--- a/docs/adrs/00002-analysis-graph.md
+++ b/docs/adrs/00002-analysis-graph.md
@@ -1,0 +1,262 @@
+# 00002. Analysis graph API in Trustify
+
+Date: 2025-01-23
+
+## Status
+
+ACCEPTED
+
+## Context
+
+This ADR is an addenda to [previous ADR](00001-graph-analytics.md) as an attempt to clarify the differences between the graph
+relationships we capture and the view we want to create from the forest of graphs.
+
+Ingesting an sbom captures a set of trustify relationships, which are instantiated
+in the forest of graphs as;
+
+```mermaid
+graph TD
+    PackageA -->|CONTAINS| PackageOther
+    PackageD -->|CONTAINED_BY| PackageA
+    PackageA -->|DEPENDS_ON| PackageB
+    PackageB -->|DEPENDENCY_OF| PackageA
+    SBOMDOC1 -->|DESCRIBES| PackageA
+    UpstreamComponent -->|ANCESTOR_OF| PackageA
+    image.arch1 -->|VARIANT_OF| ImageIndex1
+    image.arch2 -->|VARIANT_OF| ImageIndex1
+    SBOMDOC2 -->|DESCRIBES| ImageIndex1
+
+    SBOMDOC3 -->|DESCRIBES| srpm_component
+    binarycomponent1 -->|GENERATED_FROM| srpm_component
+    binarycomponent2 -->|GENERATED_FROM| srpm_component
+```
+
+Trustify relationships attempt to put an abstraction over relationships 
+defined by any format of sbom (eg. cyclonedx, spdx). 
+
+This graph encapsulates provenance of sbom relationship though end users are unlikely 
+to directly navigate such graphs, as it would mean forcing concrete understanding of relationship directionality. 
+If we were to _normalise_ all relationships (similar to the rewrite of all CONTAINS into CONTAINED_BY) the 
+resultant DAG is probably ok ex. we could rewrite the following relationships:
+
+  [ _binarycomponent1_ **GENERATED_FROM** _srpm_component_] to [_srpm_component_ **GENERATES** _binarycomponent1_]
+
+  [ _image.arch2_ **VARIANT_OF** _ImageIndex1_] to [_ImageIndex1_ **VARIANTS** _image.arch2_]
+
+So while we make the logical model easier to comprehend it is a good practice to keep logical model separate and not try to overload that model to serve as conceptual model.
+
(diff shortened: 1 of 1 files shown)


docs/adrs/00002-analysis-graph.md
```diff
@@ -0,0 +1,203 @@
...
+**Search for 'PackageA'**
+* component ancestors would be `[UpstreamComponent]`
+* component descendents would be the tree underneath 'PackageA' `[PackageOther,PackageD,PackageB]`
+
+**Search for 'image.arch1'**
+* component ancestors would be `[ImageIndex1]`
+* component descendents would be `[]`
+
+_Note: every node in the graph already knows its relationship to original SBOM so no need
+to enumerate DESCRIBES relationship ... though in the future we may see other artifacts (then sbom)
```
jcrossley3, line 47: than or then? I'm not sure either is correct grammar. :smile: 
```suggestion
to enumerate DESCRIBES relationship ... though in the future we may see other artifacts (than sbom) 
```

docs/adrs/00002-analysis-graph.md
```diff
@@ -0,0 +1,203 @@
...
+      "node_id": "",
+      "relationship": "Variant",
+      "purl": [
+        ""
+      ],
+      "cpe": [],
+      "name": "PackageC",
+      "version": ""
+    }
+  ]
```
jcrossley3, line 97: If these are immediate (one deep), then I'm immediately confused that the `ancestor` and `descendant` elements are arrays, which imply potentially more than one, to me.

jcrossley3: Nevermind. One _LEVEL_ might contain multiple ancestors or descendants. Sorry for the noise!

chirino:
I like it.  We should consider if there are better terms to use than `ancestor` and `descendant` as these could get associated with the `AncestorOf` relationship.  Maybe: 
- `relationship` - `reverse-relationship`
- `forward` - `reverse`
- `normal` and `inverse`

If the `ancestor` and `descendant` fields found in the response are optional then the same return type can be used for all these API endpoints.  If that's the case then I think we could consolidate to a single `api/v2/analysis/relationships` endpoint if it accepts arg filters like:

- ancestors bool
- descendants bool
- depth: int
- realtionshipType: string

JimFuller-RedHat:
> I like it. We should consider if there are better terms to use than `ancestor` and `descendant` as these could get associated with the `AncestorOf` relationship. Maybe:
> 
>     * `relationship` - `reverse-relationship`
> 
>     * `forward` - `reverse`
> 
>     * `normal` and `inverse`

In the graph world this conversation comes up regularly ... I always settle on anc/desc as most people agree what it means. 
 
> If the `ancestor` and `descendant` fields found in the response are optional then the same return type can be used for all these API endpoints. If that's the case then I think we could consolidate to a single `api/v2/analysis/relationships` endpoint if it accepts arg filters like:
> 
>     * ancestors bool
> 
>     * descendants bool
> 
>     * depth: int
> 
>     * realtionshipType: string

I am all for simplifying and/or having less 'moving parts' ... even better when parameterised ... in this case I suggested separation because it was easier to puzzle out ... lets discuss this in meeting.

ctron:
> > I like it. We should consider if there are better terms to use than `ancestor` and `descendant` as these could get associated with the `AncestorOf` relationship. Maybe:
> > ```
> > * `relationship` - `reverse-relationship`
> > 
> > * `forward` - `reverse`
> > 
> > * `normal` and `inverse`
> > ```
> 
> In the graph world this conversation comes up regularly ... I always settle on anc/desc as most people agree what it means.
> 
> > If the `ancestor` and `descendant` fields found in the response are optional then the same return type can be used for all these API endpoints. If that's the case then I think we could consolidate to a single `api/v2/analysis/relationships` endpoint if it accepts arg filters like:
> > ```
> > * ancestors bool
> > 
> > * descendants bool
> > 
> > * depth: int
> > 
> > * realtionshipType: string
> > ```
> 
> I am all for simplifying and/or having less 'moving parts' ... even better when parameterised ... in this case I suggested separation because it was easier to puzzle out ... lets discuss this in meeting.

Maybe, we can have both? A generalized endpoint, with the options described. And opinionated versions, which simply set defaults.

ctron:
One idea, when it comes to relationships. Why not use the following serialization:

```json
{
  "sbom_id": "",
  "node_id": "",
  "ancestors": [
    {
      "relationship": "AncestorOf",
      "node": {
         "sbom_id": "",
         "node_id": "",
         …
      }
    }
  ]
}
```

Taking out the relationship of the actual node data.

IMO, this would separate the node from the relationship and might it a bit clearer when to expect which fields.

One alternative to this could be to have all relationships group in a map:

```json
{
   "node_id": "",
   "ancestors": {
     "AncestorOf": [
       { "node_id": "anc1", … }
       { "node_id": "anc2", … }
      ]
   }
}
```

JimFuller-RedHat:
> One idea, when it comes to relationships. Why not use the following serialization:

I can see why this might be useful, though it also complexifies things - for what significant benefit ? if we can have a response payload that is amenable directly to visualisation I think that is the best goal to drive towards. 

The biggest/last issue we need to define is conceptual view relationships - will try to get something proposed there before our meeting today.

JimFuller-RedHat:
@chirino, @ctron and @jcrossley3 please re review

PR #1191, opened 2025-01-23
add adr for license export

bxf12315:
Rendered preview: https://github.com/bxf12315/trustify/blob/license-doc/docs/adrs/00005-license-export-analytics.md


docs/adrs/license/License_change.MD
```diff
@@ -0,0 +1,100 @@
+# Analysis of the license data structure.
```
JimFuller-RedHat, line 1: Pleaes follow ADR convention eg. no need to nest directory structure ... also use the same heading breakdown ... some inspiration here https://adr.github.io/

bxf12315: updated, please check again.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,114 @@
...
+Date: 2025-01-23
+
+## Status
+
+DRAFT
+
+## Context
+### An SBOM typically contains three types of license information:
+
+#### 1 License information for the entire SBOM
```
ctron, line 12: I think the main topic of this ADR is the license of individual components. So it's worth mentioning this as an exclusion later on. But drop it from the main section, if that's the case.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,114 @@
...
+    pub spdx_license_exceptions: Option&lt;Vec&lt;String&gt;&gt;,
+}
+```
+The text field corresponds to CycloneDX's Id, Name, or complete Expression. When it's an Expression, it will be parsed into a license array and stored in spdx_licenses, while parsing exceptions are stored in spdx_license_exceptions.
+For SPDX, the text field corresponds to the complete Expression. Similarly, the Expression will be parsed into a license array and stored in spdx_licenses, with parsing exceptions stored in spdx_license_exceptions.
+Currently missing SPDX's ExtractedLicensingInfos and SBOM-level licenses.
+
+### 4 License data structure in GUAC
+![erd_guac.png](erd_guac.png)
+GUAC parses all expressions into individual licenses and saves them separately. LicenseRef* represents ExtractedLicensingInfos, with its extractedText stored in the inline field.
```
ctron, line 76: This feels wrong. As a SPDX license expression, to my understanding, is not simply a "list of licenses", but actually, indeed an expression. Which might say "either, or". Or even more complex expressions.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,114 @@
...
+    CYDXLCNAME,
+    #[sea_orm(string_value = "cle")]
+    CYDXLEXPRESSION,
+    #[sea_orm(string_value = "cd")]
+    CLEARLYDEFINED,
+    #[sea_orm(string_value = "o")]
+    OTHER,
+}
+```
+SPDXDECLARED: The license from spdx sbom's declered license.
```
ctron, line 100: I don't understand why we'd need different types here. Why do we differentiate between CycloneDX and SPDX? Why would we attach that "type" information to the table hold general license information?

ctron (changes requested):
I still have a hard time understanding how this would work. Maybe some examples would make sense.

I also don't think I understand why the license table has an `id` field, as well as a `license_id` and `license_ref_id` field? What's the meaning of those different IDs? Why does the `extracted_licensing_info` table have a `id` and `licenseId` field?

docs/adrs/00005-license-export-analytics.md
bxf12315: Currently, the requirement is only for the declared licenses in SPDX and the licenses in CycloneDX. So it is important to know where it comes from, because in SPDX there is also a concluded license, and some of these licenses are derived from clearly defined sources.

docs/adrs/00005-license-export-analytics.md
bxf12315: yes. SPDX license expression includes not only individual license information but also the relationships between these licenses. It's like the shortcomings of this design that I mentioned below.However, there must be a single license name to associate with the license ID in ExtractedLicensingInfos.

bxf12315: If we retain the current license structure and add a new entity to store the individual license name information, how would that work?

ctron: Storing the license, I do understand that you need to store individual licenses. However, this is more about the license a project declares. IIRC, we already have a full list of SPDX licenses (as defined by the SPDX license spec).

If we want to extract the information each component declared, we would need to keep that expression. Which, I think, is what the current implementation does.

docs/adrs/00005-license-export-analytics.md
ctron: In the database we do not store the type of the document (SPDX or CDX). Why is that important to know? From what I see it, the information is attached to the `license` table. Which doesn't have any SBOM information.

bxf12315:
@ctron Updated Please check it.

docs/adrs/00005-license-export-analytics.md
JimFuller-RedHat: agree with @ctron - a license (defined or concluded)  is independent to serialisation format eg. does not matter if it was serialised in CDX or SPDX (or what capabilities to represent in either) ...

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,126 @@
...
+    pub text: String,
+    pub spdx_licenses: Option&lt;Vec&lt;String&gt;&gt;,
+    pub spdx_license_exceptions: Option&lt;Vec&lt;String&gt;&gt;,
+}
+```
+The text field corresponds to CycloneDX's Id, Name, or complete Expression. When it's an Expression, it will be parsed into a license array and stored in spdx_licenses, while parsing exceptions are stored in spdx_license_exceptions.
+For SPDX, the text field corresponds to the complete Expression. Similarly, the Expression will be parsed into a license array and stored in spdx_licenses, with parsing exceptions stored in spdx_license_exceptions.
+Currently missing SPDX's ExtractedLicensingInfos and SBOM-level licenses.
+
+### 4 License data structure in GUAC
```
JimFuller-RedHat, line 77: Why is GUAC license structure provided in this ADR ? Please provide the rationale/context of why you are speaking about GUAC here.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+]
+```
+
+##### 1.2 SPDX SBOM-level License
+```json
+"dataLicense": "CC0-1.0",
+```
+If I understand correctly, CycloneDX's SBOM-level License refers to the license of the package corresponding to the SBOM, while SPDX's SBOM-level License refers to the license of the SBOM document itself.
```
JimFuller-RedHat, line 30: Lets keep the language in this ADR to assertive eg. 

CycloneDX's SBOM-level License refers to the license of the package corresponding to the SBOM, while SPDX's SBOM-level License refers to the license of the SBOM document itself

If you do not understand correctly - the ADR is not the place to question yourself ;)

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,126 @@
+# 00005. Analysis of the license data structure.
+
+Date: 2025-01-23
+
+## Status
+
+DRAFT
+
+## Context
```
JimFuller-RedHat, line 9: Why not a few paragraphs about what and why this is being done ? It just launches into a 'laundry' list of formats ... prepare the reader

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,126 @@
...
+This design currently has two issues:
+1. The complete expression is not saved, as the complete expression includes not only individual license
+   information but also the relationships between these licenses.
+2. SBOM-level licenses are not saved.
+
+### Next design proposal
+![next_erd_license.png](next_erd_license.png)
+
+
+The new design retains the original license structure, but adds a new attribute: license type. It also removes the physical association between the license and ExtractedLicensingInfos when the license is a licenseRef. If needed, we can retrieve this information through conditional queries. Additionally, a new entity, sbom_license, is introduced to store license information at the SBOM level. However, as I mentioned earlier, the SBOM-level licenses in SPDX and CycloneDX have completely different meanings, so a type attribute is required to distinguish between them.
```
JimFuller-RedHat, line 122: Design proposal should outline 'what we are going to do' ... useful context information should be in the context section.

PhilipCattanach:
@bxf12315 
Xiaofeng, A couple of further comments.

It was the right thing to do to remove the link to the Googlesheet that is a mock-up of the export file layout.
But it would be good to have 2 tables that essentially contain the same information. I think it is important that there is something in the ADR that shows the layout of the export files, and the source of the data in each column.
A table is something easy for the reader to digest.
    
What would also be a good idea would be some simple examples. Several SBOM snippets, that show license data for particular components, and corresponding extract file snippets illustrating how that data would be represented in the main extract file.
As well as _regular_ examples, include some edge case component examples too.

A code snippet for the SPDX License-Refs and the corresponding ExtractLicensingInfos export file snippet would be good too.

Finally I would like a section that summarizes the impact of the changes;

1. Data model
2. SBOM ingestion
3. License Extract API
4. UI - License export action on the SBOM List and SBOM Detail screens

bxf12315:
> @bxf12315 Xiaofeng, A couple of further comments.
> 
> It was the right thing to do to remove the link to the Googlesheet that is a mock-up of the export file layout. But it would be good to have 2 tables that essentially contain the same information. I think it is important that there is something in the ADR that shows the layout of the export files, and the source of the data in each column. A table is something easy for the reader to digest.
> 
> What would also be a good idea would be some simple examples. Several SBOM snippets, that show license data for particular components, and corresponding extract file snippets illustrating how that data would be represented in the main extract file. As well as _regular_ examples, include some edge case component examples too.
> 
> A code snippet for the SPDX License-Refs and the corresponding ExtractLicensingInfos export file snippet would be good too.
> 
> Finally I would like a section that summarizes the impact of the changes;
> 
> 1. Data model
> 2. SBOM ingestion
> 3. License Extract API
> 4. UI - License export action on the SBOM List and SBOM Detail screens

Ok will do.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+              "spdx_license_exceptions",
+           )
+           .into_model::<SbomLicenseBase>()
+           .all(&ctx.db)
+           .await?;
+  
+   Ok(())
+}
+```
+As shown in the above example, when the SPDX SBOM was created, the sbom_name and sbom_namespace could not be obtained. Additionally, we are currently storing both licenseDeclared and licenseConcluded, and we are unable to distinguish between them.
```
ctron, line 273: I think this is not correct, as is the previous Rust code. Instead of focusing on code, I believe it would be better to focus on the data.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+At same time, if the SBOM is cycloneDX, sbom_name, component_group, component_version will miss.
```
ctron, line 274: I guess this is mixed up with SPDX.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+
+**The information in the ExtractedLicensingInfos section is completely missing.**
```
ctron, line 276: I believe the information is there, just structured in a different way.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+## Decision
```
ctron, line 277: I still struggle with the section. As there seem to be issues up until this point, I am not sure it makes sense coming to a conclusion at this point.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+
+### Current design proposal
+#### Data model
+The current design requires adding a license_type field in purl_license_assertion, as well as a new entity called extracted_licensing_infos. Additionally, the Sbom entity should include fields such as sbom_name, sbom_namespace, component_group, and component_version.
```
ctron, line 281: Why does it require adding a type?

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+
+SpdxDeclared: The license from spdx sbom's declered license.
+SpdxConcluded: The license from spdx sbom's concluded license.
+CydxLcid: The license from CycloneDX sbom's id.
+CydxLcname: The license from CycloneDX sbom's name.
+CydxLexpression: The license from CycloneDX sbom's Expression.
+
+#### SBOM ingestion
+Add the following actions during the ingestion process.
+1. add sbom_namespace, component_group, and component_version into Entity sbom.
```
ctron, line 313: We already have the namespace and version.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+
+#### License Extract API
+This section of the code can be divided into three parts.
+1. Data Collection
+   - This section of the code can refer to [here](#current-license-data-structure-design-in-trustify)    
+
+ 2. Generate the CSV file and compress it. 
+    - This code can refer version one's code.[here](https://github.com/trustification/trustification/blob/main/spog/api/src/license/license_exporter.rs)
+ 3. Endpoint
+    - HTTP GET **/api/v1/sbom/license/{id}**
```
ctron, line 325: We are using `v2` now.

ctron, line 325: This looks like the `id` of the license to me. I assume you mean the `id` of the SBOM?

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+    
+#### UI
+License export action on the SBOM List and SBOM Detail screens.
+
+## Alternative approaches
```
ctron, line 330: This looks empty.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+1. add sbom_namespace, component_group, and component_version into Entity sbom.
+2. add LicenseCategory into Entity purl_license_assertion.
+3. create new Entity extracted_licensing_infos.
+
+#### License Extract API
+This section of the code can be divided into three parts.
+1. Data Collection
+   - This section of the code can refer to [here](#current-license-data-structure-design-in-trustify)    
+
+ 2. Generate the CSV file and compress it.
```
ctron, line 322: I don't think it makes sense pointing to a another system's code which works quite differently.

ctron (changes requested):
I still can't see a clear path to implementing this feature.

I think there's progress and this version is improved. But there are also still assumptions which don't seem correct to me and have an impact on the decision section.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+      2. [SBOM ingestion](#sbom-ingestion)
+      3. [License Extract API](#license-extract-api)
+      4. [UI](#ui)
+
+
+
+
+
+## The analysis of the license information in the SBOM
+### License information for the entire SBOM
```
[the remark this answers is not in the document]
PhilipCattanach: Yesterday, I suggested the following to provide context at the beginning of ADR. We need this, or something like it that provides an overview.

Suggested.....
The requirement is for a user of the application to be able to download all of the license data associated with an individual SBOM. An action will be available from the SBOM list and SBOM detail page that will allow the user to download all of the license data associated to the SBOM. Because of the how license data is represented in SPDX SBOMs (see below) it is necessary to generate 2 export files; the main license export, with 1 row for each component, license combination, and a supporting ExtractedLicensingInfos export file, that will only be populated if the original SBOM was in SPDX format. As there are always 2 export files generated the user will download a compressed file (tar.gz) containing both CSV files.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+### Requirement Description
+The application must allow users to download all license data associated with a specific SBOM. This functionality will be accessible from both the **SBOM list** and the **SBOM detail page**.
+
+Due to the way license data is structured in **SPDX SBOMs** (as detailed below), two export files need to be generated:
+
+1. **Primary License Export File**
+    - Each row corresponds to a unique component-license combination.
+
+2. **ExtractedLicensingInfos Export File**
+    - This file is only included if the original SBOM follows the SPDX format.
```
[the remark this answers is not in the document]
PhilipCattanach: @ctron - perhaps we can chat about this. As I understand it SPDX generates internal LicenseRef-n identifiers that are referenced at component level. At document level each LicenseRef-n has a description that defines what combination of licenses the LicenseRef-n represents. So the second CSV export file is like a look-up table for the LicenseRef-n Ids on the first CSV file. I don't think that concept applies to CycloneDX?  Thanks

PhilipCattanach:
@bxf12315 - As Jens stated the ADR is getting better.
Thanks for adding the table for the main extract file.

We are looking for consistency with the version 1.3 license export. The layout of that export file is defined in this [Googlesheet](https://docs.google.com/spreadsheets/d/1Ki9hkLl94L3G4A4rfQQ4wuEnQFTPkpNflClVjlk2mQA/edit?gid=0#gid=0) So I was expecting the same number of columns (9) in the same order as the Googlesheet. 
Your table differs from the Googlesheet. 

There isn't a table the _License Mapping_ sheet on the Googlesheet.  That needs to be added too.

docs/adrs/00005-license-export-analytics.md
bxf12315: yes sbom id.

ctron: Then I think it should be swapped around, so that it's clear what the ID refers to: `/api/v2/sbom/{id}/license`.

docs/adrs/00005-license-export-analytics.md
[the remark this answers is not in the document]
ctron: To my understanding, CyclondeDX has a similar concept:

image

But mapped differently. So there's the ability to reference licenses too, in order to re-use it.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+```json
+"dataLicense": "CC0-1.0",
+```
+If I understand correctly, CycloneDX's SBOM-level License refers to the license of the package corresponding to the SBOM, while SPDX's SBOM-level License refers to the license of the SBOM document itself.
+### Package-level licenses
+#### CycloneDX
+Contains three types: Id, name, and Expression
+The Expression follows the SPDX expression standard.
+```json
+"licenses" : [  {    "license" : {      "id" : "EPL-1.0"    }  },  {    "license" : {      "name" : "GNU Lesser General Public License",      "url" : "http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html"    }  }],`
```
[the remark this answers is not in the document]
bxf12315: ok

docs/adrs/00005-license-export-analytics.md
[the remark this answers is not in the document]
bxf12315: Do you mean bom-ref?  I can not find an example of it.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+
+### Detail Info
+
+Below, I will list all the required fields for the license export along with their descriptions.
+
+| Field Name     | Description                                                                                                               | SPDX                                                                                                     | CycloneDX   | 
+|----------------|---------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------|-------------|
+| sbom_name      | sbom_name: it's the name of the SBOM.                                                                                     | yes                                                                                                      | yes         |
+| sbom_namespace | It comes from the SPDX's documentNamespace node.                                                                          | yes                                                                                                      | Unsupported |
```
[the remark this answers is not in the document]
bxf12315: The sbom_namespace corresponds to SPDX's documentNamespace, not CycloneDX's serial ID, so I set it as Unsupported.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+| component_group  | The component's group.                                                                                                    | Unsupported                                                                                              | yes         |
+| component_version| The component's version.                                                                                                  | Unsupported                                                                                              | yes         |
```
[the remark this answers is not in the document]
bxf12315: Yes, but the component_group does not exist.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+| package_name   | This SBOM contains multiple packages or components, and package_name refers to the names of these packages or components. | yes                                                                                                      | yes         |
+| purl      | [1](#detailed-description-of-the-package-in-spdx) [2](#detailed-description-of-the-component-in-cyclonedx)                | yes         | yes         |
+| cpe      | [1](#detailed-description-of-the-package-in-spdx) [2](#detailed-description-of-the-component-in-cyclonedx)                | yes                                                                                                      | yes         |
```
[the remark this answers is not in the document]
bxf12315: They will be listed under column alternate package reference.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+| text      | License content, it's text content, for SDPX it just from licenseDeclared                                                 | yes                                                                                                      | yes         |
+| spdx_licenses      | license ids, it's parsed by License content.                                                                                             | yes                                                                                                      | yes         |
+| spdx_license_exceptions      |                                                                                    | yes                                                                                                      | yes         |
```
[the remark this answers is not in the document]
bxf12315: It is the entire license expression.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+      }
+   ],
+   "cpe": "cpe:2.3:a:\\@gradio\\/accordion:\\@gradio\\/accordion:0.3.4:*:*:*:*:*:*:*",
+   "purl": "pkg:npm/%40gradio/accordion@0.3.4",
+   "properties": [
+      ...
+   ]
+},
+```
+Field text from license of licenses node,spdx_licenses from parsed by license node.
```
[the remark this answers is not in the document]
bxf12315: ok

docs/adrs/00005-license-export-analytics.md
bxf12315: Currently, Trustify does not distinguish between licenseConcluded and licenseDeclared, but the requirement only asks to retrieve licenseDeclared.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+| package_name   | This SBOM contains multiple packages or components, and package_name refers to the names of these packages or components. | yes                                                                                                      | yes         |
+| purl      | [1](#detailed-description-of-the-package-in-spdx) [2](#detailed-description-of-the-component-in-cyclonedx)                | yes         | yes         |
+| cpe      | [1](#detailed-description-of-the-package-in-spdx) [2](#detailed-description-of-the-component-in-cyclonedx)                | yes                                                                                                      | yes         |
+| text      | License content, it's text content, for SDPX it just from licenseDeclared                                                 | yes                                                                                                      | yes         |
```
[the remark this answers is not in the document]
bxf12315: They will be listed under the column .

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+| component_group  | The component's group.                                                                                                    | Unsupported                                                                                              | yes         |
+| component_version| The component's version.                                                                                                  | Unsupported                                                                                              | yes         |
+| package_name   | This SBOM contains multiple packages or components, and package_name refers to the names of these packages or components. | yes                                                                                                      | yes         |
+| purl      | [1](#detailed-description-of-the-package-in-spdx) [2](#detailed-description-of-the-component-in-cyclonedx)                | yes         | yes         |
```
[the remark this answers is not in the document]
bxf12315: As mentioned in my example, if there are multiple records for the purl, generate multiple query records accordingly.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,332 @@
...
+{
+"comment": "External License Info is obtained from a build system which predates the SPDX specification and is not strict in accepting valid SPDX licenses.",
+"extractedText": "The license info found in the package meta data is: GPL-2.0-with-classpath-exception. See the specific package info in this SPDX document or the package itself for more details.",
+"licenseId": "LicenseRef-GPL-2.0-with-classpath-exception",
+"name": "GPL-2.0-with-classpath-exception"
+},
+```
+The licenseId can be associated with the license in the package.
+### Current license export requirements
+### Requirement Description
```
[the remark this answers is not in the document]
bxf12315: ok

docs/adrs/00005-license-export-analytics.md
bxf12315: ok

bxf12315:
> @bxf12315 - As Jens stated the ADR is getting better. Thanks for adding the table for the main extract file.
> 
> We are looking for consistency with the version 1.3 license export. The layout of that export file is defined in this [Googlesheet](https://docs.google.com/spreadsheets/d/1Ki9hkLl94L3G4A4rfQQ4wuEnQFTPkpNflClVjlk2mQA/edit?gid=0#gid=0) So I was expecting the same number of columns (9) in the same order as the Googlesheet. Your table differs from the Googlesheet.
> 
> There isn't a table the _License Mapping_ sheet on the Googlesheet. That needs to be added too.

updated

docs/adrs/00005-license-export-analytics.md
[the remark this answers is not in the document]
ctron: Why is it called "exceptions" then?

docs/adrs/00005-license-export-analytics.md
ctron: Which one does Trustify process at the moment?

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,368 @@
...
+ 2. Generate the CSV file and compress it. 
+    - This code can refer version one's code.
+ 3. Endpoint
+    - HTTP GET **/api/v2/sbom/{id}/license** Id: sbom id
+    
+#### UI
+License export action on the SBOM List and SBOM Detail screens.
+
+## Alternative approaches
+Currently, there is an alternative method. Instead of retrieving relevant information through the database's Entity, it directly obtains the original SBOM file (which already exists in the sbom_file Entity), parses it, and extracts all the information to generate the corresponding CSV files.
```
ctron, line 355: The `sbom_file` entity **does not** contain SBOM files. It does contain the "file" elements from a SPDX SBOM. That's difference.

Also, I don't think this alternative exists at the moment. We could implement it though.

docs/adrs/00005-license-export-analytics.md
```diff
@@ -0,0 +1,368 @@
...
+
+ 2. Generate the CSV file and compress it. 
+    - This code can refer version one's code.
+ 3. Endpoint
+    - HTTP GET **/api/v2/sbom/{id}/license** Id: sbom id
+    
+#### UI
+License export action on the SBOM List and SBOM Detail screens.
+
+## Alternative approaches
```
ctron, line 354: What would be the pros/cons of this alternative?

modules/ingestor/tests/reingest/license.rs
```diff
@@ -0,0 +1,119 @@
+// #![allow(clippy::all)]
```
ctron, line 1: Please remove the code.

ctron (changes requested):

docs/adrs/00005-license-export-analytics.md
bxf12315: I think [1] can got the resource file.
[1] https://github.com/trustification/trustify/blob/52fce44f662cd90729a031cdd1fdd6728ff3b473/modules/fundamental/src/sbom/endpoints/mod.rs#L399

ctron: We do have the file stored. There's an internal method for retrieving it. I don't think we need to go into those details. However, I'd also prefer not having false information in the ADR.

bxf12315: ok

modules/ingestor/tests/reingest/license.rs
bxf12315: Done

JimFuller-RedHat:
@bxf12315 what is the status of this PR - it looks like we started doing work on other PRs before getting this accepted ?

JimFuller-RedHat:
as there is no answer, closing this

bxf12315:
The related PR already merged. Thanks

PR #1195, opened 2025-01-23
Trying to come with more clarity in tests

jcrossley3:


jcrossley3:
@chirino do we hate this? :smile:

modules/analysis/src/endpoints/test.rs
```diff
@@ -656,37 +656,27 @@ async fn spdx_package_of(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
...
-                "version": "6.15",
-                "deps": [],
-            })
-        })
-        .collect();
-
-    assert_eq!(1, matches.len());
+    assert_eq!(
+        response.pointer("/items/0/sbom_id"),
+        response.pointer("/items/0/deps/0/sbom_id")
```
chirino, line 661: Don't hate it but I do think assuming the array position of the element could be considered brittle. Is that PackageOf relationship pointing at the expected package?

jcrossley3: You could include additional assertions about the lengths of the relevant arrays like I do in the test below, e.g.
```
assert!(response.pointer("/items/0/deps/1").is_none());
```
The `PackageOf` relationship is asserted on lines 663-666.

I don't like the pointer path as much as your `ContainsSubset` trait.

PR #1200, opened 2025-01-24
adr for external references

JimFuller-RedHat:


chirino (approved):
LGTM

docs/adrs/00003-external-references.md
```diff
@@ -0,0 +1,414 @@
...
+      "ref": "app-2.0.0",
+      "dependsOn": [
+        "urn:uuid:3e671687-395b-41f5-a30f-a58921a69b79/1.6#acme-application-1.0.0"
+      ]
+    }
+  ],
+  "externalReferences": [
+    {
+      "type": "spdx",
+      "url": "https://example.org/other-sbom.json",
```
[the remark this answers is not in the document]
JimFuller-RedHat: yes this is a potential relationship which is not asserted until query time

docs/adrs/00003-external-references.md
```diff
@@ -0,0 +1,1246 @@
...
+### Implement **sbom_external_node** table
+
+- documentNampace (url)
+- document version
+- Checksum/digest/hash
+- conflated reference
+
+Then a node, in `sbom_node` table, can use the conflated reference: 
+
+    _pkg:rpm/redhat/openssl@3.0.7-18.el9_2?arch=src_ **PACKAGES** _https://example.org/other-sbom.json#openssl-3.0.7-18.el9_2_
```
jcrossley3, line 1220: This line confuses me. What is `**PACKAGES**`?

Can we add a section to this doc that shows the exact data for a row in the new table along with the relevant sections of the two SBOM's?

JimFuller-RedHat: this is a result of docs/adrs/00002-analysis-graph.md eg PACKAGE_OF reverses relationship to PACKAGES ... we want to create a tree with only one direction relationships.

chirino:
Now that we have some example record entries written down,  we should think about the types of queries that the analysis layer will do to find the related nodes for these 3 external ref types.  That may reveal the need for some additional indexed fields.

JimFuller-RedHat:
I can flesh out the analysis graph queries though they will be similar to what we do today aka search **sbom_node** table then do petgraph query ... for external node we would be **external_sbom_package_node** then petgraph query

JimFuller-RedHat:
Note - this ADR is getting merged - there are details to be worked out eg. we know resolving ancestor relationships (eg reverse external ref resolution) is the most difficult bit .. I do think we have enough detail in this ADR to move forward and we can now switch gears and go into implementation mood - I will setup a series of tasks which we can divy up amongst ourselves.

PR #1206, opened 2025-01-27
Simplified relationship tests a bit

jcrossley3:
Removed the non-recursive subset fn as it wasn't being used anywhere. There may yet be a need to introduce some mechanism that ensures json array values are matched exactly, rather than the current "contained" match.


chirino (approved):
lgtm.

PR #1209, opened 2025-01-28
normalise Contains relationships

JimFuller-RedHat:
As defined in #1190 - we want to invert relationships to force a directed graph with all relationships going downwards.

### DIFF
--- a/entity/src/relationship.rs
+++ b/entity/src/relationship.rs
@@ -50,6 +50,34 @@ pub enum Relationship {
     #[sea_orm(num_value = 14)]
     PackageOf,
     #[sea_orm(num_value = 15)]
+    Contains,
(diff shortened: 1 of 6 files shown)


ctron:
I like the PR in general. I just not sure about the migration part. To my understanding, we're adding relationsships, not using the old variants.

Why not rename the relationships, and swap left/right for the ones we change in the migration?

JimFuller-RedHat:
> Why not rename the relationships, and swap left/right for the ones we change in the migration?

oh right the migration should also actually migrate the relationship data as well ;)

JimFuller-RedHat:
superseded by https://github.com/trustification/trustify/pull/1213 ... moving to work over there and closing this

PR #1212, opened 2025-01-29
TC-2174 OSV Python ecosystem management

mrizzi:
https://issues.redhat.com/browse/TC-2174

- Added PLSQL function for Python versions (ref. https://packaging.python.org/en/latest/specifications/version-specifiers/ and https://peps.python.org/pep-0440/) with related SeaORM migration and test in `modules/ingestor/tests/version/pythonver.rs`.
- Managed `Python` and `PyPI` ecosystems in OSV loader (with test)
- Enhanced the `translate` function (with test)
- Furthermore added test `modules/fundamental/tests/sbom/details.rs` for the SBOM details to ensure a GHSA OSV vulnerability for the `PyPI` ecosystem (e.g. [GHSA-45c4-8wx5-qw6w](https://github.com/github/advisory-database/blob/ce72d06562c2b0f4e3af52889f6d9080319324a9/advisories/github-reviewed/2023/07/GHSA-45c4-8wx5-qw6w/GHSA-45c4-8wx5-qw6w.json)) properly correlates with an SBOM

### DIFF
--- a/migration/src/m0000850_python_version.rs
+++ b/migration/src/m0000850_python_version.rs
@@ -1,0 +1,138 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+#[allow(deprecated)]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let db = manager.get_connection();
+
+        db.execute_unprepared(
+            &Query::update()
+                .table(VersionScheme::Table)
+                .value(VersionScheme::Id, "python")
+                .and_where(Expr::col(VersionScheme::Id).eq("pypi"))
+                .to_owned()
+                .to_string(PostgresQueryBuilder),
+        )
+        .await?;
+
+        db.execute_unprepared(include_str!("m0000850_python_version/pythonver_cmp.sql"))
+            .await
+            .map(|_| ())?;
+
+        db.execute_unprepared(include_str!(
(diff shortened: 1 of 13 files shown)


modules/ingestor/src/service/advisory/osv/loader.rs
```diff
@@ -173,6 +173,26 @@ impl<'g> OsvLoader<'g> {
                                 )
                                 .await?;
                             }
+                            (RangeType::Ecosystem, Ecosystem::PyPI) => {
```
dejanb, line 176: Maybe this can be a single match?

`(RangeType::Ecosystem, Ecosystem::PyPI | Ecosystem::Python) => {`

migration/src/m0000850_python_version.rs
```diff
@@ -0,0 +1,138 @@
...
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+#[allow(deprecated)]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let db = manager.get_connection();
+
+        db.execute_unprepared(
```
dejanb, line 12: Can this use `update()` function?

dejanb (approved):
Looks great. Just two nitpicks

modules/ingestor/src/service/advisory/osv/loader.rs
mrizzi: Absolutely, you're right: at least I learnt something new about Rust also today 😉 thanks

PR #1213, opened 2025-01-29
Work towards new API from ADR

ctron:


ctron:
Continue work on #1217

Issue #1201, opened 2025-01-24
How do you debug tests by setting RUST_LOG?


jcrossley3:
For a long time, I chastised @bobmcwhirter for using `println!` in tests, but now I'm coming around to his way of thinking. Because how else can I avoid all the instrumentation noise?

After @ctron added a bunch of `instrument` macros to the analysis service, I have to wait well over a minute to see my `log::debug!` output in my test. Who is this helping? How do others set their RUST_LOG to avoid all the noise in the output? I used to set `RUST_LOG=none,trust=debug`, but now I have to be far more explicit, e.g. `RUST_LOG=none,trustify_module_analysis::endpoints=debug` and even then it's WAY more noisy than I feel is productive. 

What's the right thing to do here? Because it feels like `println!` is the path of least resistance.

And who does all the noise benefit?

bobmcwhirter:
The Crossley Probability Principle does state that every time you complained about my use of `println!`, the probability of your *not* complaining about it increases.

Good job!

ctron:
I don't care that much about what we do in tests. Adding the `instrument` macros does help to figure out things when running in production.

I am more worried about the amount of data we stuff into the traces. Sometimes those messages can become quite big.

Just using `println` doesn't really solve the issue. But basically raises the level of the message so that you're more likely to see it. Why not use `warn` in tests? Also, I am not against using `println`, as long as we're talking about tests. Or other situations, where `log` isn't a good fit.

jcrossley3:
Personally, I care very much about what we do in tests, because I think a project with clear, consistent and "inhabitable" tests can attract contributors. If done well, they can help a developer unfamiliar with the project to learn its features, structure, standards, style, etc.

I think we still have decisions to make regarding which of our functions need to be instrumented and whether DEBUG is the appropriate level for them. Regardless, I'd like to be able to debug my particular test without triggering a flood of logs irrelevant to it.

I don't like `println!` because it ultimately diminishes the S/N of the default test output and therefore the perceived quality of the project. In a word, it looks tacky. And I'll continue to discourage folks from using `println!` in their tests.

**My current solution:**

Instead of this...
```
    log::debug!("{x:#?}");
```
...I'll use this:
```
    tracing::debug!(test = "", "{x:#?}");
```
[...]

PR #1216, opened 2025-01-30
More test cleanup; removing hardcoded array indexing

jcrossley3:
Makes it easier to further constrain the results without assuming any particular order in the output.

### DIFF
--- a/modules/analysis/src/endpoints/test.rs
+++ b/modules/analysis/src/endpoints/test.rs
@@ -1,7 +1,6 @@
 use crate::test::caller;
 use actix_http::Request;
 use actix_web::test::TestRequest;
-use itertools::Itertools;
 use serde_json::{json, Value};
 use test_context::test_context;
 use test_log::test;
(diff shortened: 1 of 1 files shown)


modules/analysis/src/endpoints/test.rs
```diff
@@ -19,36 +18,29 @@ async fn test_simple_retrieve_analysis_endpoint(
...
-            .as_array()
-            .unwrap()
-            .contains(&Value::from("pkg:rpm/redhat/BB@0.0.0"))
-    {
-        assert_eq!(&response["total"], 2);
-    } else {
-        panic!("one of the items component should have matched.");
-    }
-    log::info!("{:?}", response);
+    tracing::debug!(test = "", "{response:#?}");
```
JimFuller-RedHat, line 21: nice is using tracing:: this the 'new way' ?

JimFuller-RedHat (approved):
LGTM!

modules/analysis/src/endpoints/test.rs
jcrossley3: It's _a_ way. It's the first way I found that enabled me to easily debug tests without seeing all the instrumentation logs. More details in #1201

PR #1220, opened 2025-01-30
Use jsonpath in tests

jcrossley3:
Better, worse, or same?


JimFuller-RedHat (approved):
looks great to me (being a fan of xpath I would say that)  ...

helio-frota:
I tend to think this `Better` also because we don't have full control over a typo in the previous approach. example: `.filter(|d| d["relationshiip"] == "GeneratedFron")`

helio-frota (approved):

PR #1221, opened 2025-01-30
fix: add an ai feature so that ai features are not compiled by default.

### LINKED ISSUE #1219: Remove the AI endpoint => applicable for V 2.0
Currently there is an endpoint that exposes AI functionally through the REST endpoints. That endpoint becomes usable when the appropriate Environment Variables are set while running the api server.

For the upcoming release V2.0 that endpoint should not be available even if those env vars are available.

chirino:
Fixes #1219

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,6 @@
 [alias]
 xtask = "run --package xtask --"
+xtask-ai = "run --features ai --package xtask --"
 
 [env]
 CARGO_WORKSPACE_ROOT = { value = "", relative = true }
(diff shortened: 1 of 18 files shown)


chirino:
So this basically moves the creation of the conversations table to another migrations table.  

If someone was running a default dev version of trustd, then he might expect the conversations table to not exist for him after taking this change... but that's not the case.  Should I add a migration to delete the conversations table?

ctron:
This also seems to remove the APIs from the openAPI spec. Is that intentional?

chirino:
> This also seems to remove the APIs from the openAPI spec. Is that intentional?

Yes. They shouldn't be part of the released API until the feature is enabled by default.  If you want to add them back openapi spec doc, say for development purposes, run: `cargo xtask-ai openapi`

ctron (approved):
Note to self: We might want to consider adding https://github.com/frewsxcv/cargo-all-features at some point.

PR #1222, opened 2025-01-31
Don’t benchmark on every PR.  

chirino:
This seems to fail PR ci checks consistently.


jcrossley3 (approved):

helio-frota (approved):

ctron:
maybe it makes sense to fix this instead?

jcrossley3:
> maybe it makes sense to fix this instead?

maybe, but if it takes longer than a few minutes to run, I don't want it on every PR.

Issue #1218, opened 2025-01-30
investigate converting graph hashmap into an MRU/LRU


JimFuller-RedHat:
Which will naturally partition data along the memory and/or temporal axis.

Specifically we want to investigate (maybe an ADR) on using something like this https://crates.io/crates/lru on analysis graph hashmap.

chirino:
did a little searching, and found https://github.com/marmeladema/clru-rs which is a weighted lru.  Does anyone have better options?

JimFuller-RedHat:
comment: this recent fix #1211 fixed graph hashmap to actually persist a graph eg. in some previous work  broke it and were actually (lazily) continually reloading graphs into memory.

chirino:
Took a peek at replacing the Map in the GraphMap with CLruCache.  It was mostly straightforward until you got to the point where we implemented the GraphMap::get :

* since the LRU cache needs to be updated on get to track that it was recently used, the get method needs be changed to take a &mut self ref.  
* this has the ripple effect of needing a write lock instead of read lock in query_graph.  

I don't think we want to do that.  You don't want your read path to serialize on a writelock.  Is there a better way to restructure this?

Or should we be looking at more complex cache implementations like https://github.com/moka-rs/moka ?

PR #1217, opened 2025-01-30
Work towards new API from ADR 00002-analysis-graph

ctron:
This is a rather large set of changes ... mostly related to implementing recommendations from [docs/adrs/00002-analysis-graph](https://github.com/trustification/trustify/tree/main/docs/adrs/00002-analysis-graph.md)

Where we normalise all directional relationships from

```mermaid
graph TD
    PackageA -->|CONTAINS| PackageOther
    PackageD -->|CONTAINED_BY| PackageA
    PackageA -->|DEPENDS_ON| PackageB
    PackageB -->|DEPENDENCY_OF| PackageA
    SBOMDOC1 -->|DESCRIBES| PackageA
    UpstreamComponent -->|ANCESTOR_OF| PackageA
    image.arch1 -->|VARIANT_OF| ImageIndex1
    image.arch2 -->|VARIANT_OF| ImageIndex1
    SBOMDOC2 -->|DESCRIBES| ImageIndex1

    SBOMDOC3 -->|DESCRIBES| srpm_component
    binarycomponent1 -->|GENERATED_FROM| srpm_component
    binarycomponent2 -->|GENERATED_FROM| srpm_component
```
That is in our internal model we will only have one form of the relationship (ex. Describes instead of DescribedBy, Variants instead of VariantOf) resulting in all relationships pointing 'downward' in a directed DAG.

```mermaid
graph TD
    SBOMDOC1 -->|DESCRIBES| PackageA
    PackageA -->|CONTAINS| PackageOther
    PackageA -->|CONTAINS| PackageD
    PackageA -->|DEPENDS| PackageB
    
    SBOMDOC2 -->|DESCRIBES| ImageIndex1
    UpstreamComponent -->|ANCESTOR_OF| PackageA
    ImageIndex1 -->|VARIANTS| image.arch1
    ImageIndex1 -->|VARIANTS| image.arch2

    SBOMDOC3 -->|DESCRIBES| srpm_component
    srpm_component -->|GENERATES| binarycomponent1
    srpm_component -->|GENERATES| binarycomponent2
```

This normalisation to a single downward direction makes the resultant conceptual model easier to understand, easier to query and will be easier to maintain.

* removed analysis graph `api/v2/analysis/root-component` and `api/v2/analysis/deps` endpoints
* added analysis graph `api/v2/analysis/component` endpoint
* made numerous simplifying refactorings in the codebase
* fixed a lot of bugs

Example of new graph analysis /component/ endpoint where we 

```
[...]

PR #1224, opened 2025-01-31
Replace GraphMap’s map with a moka::sync::Cache

chirino:
Fixes issue https://github.com/trustification/trustify/issues/1218

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,12 +54,14 @@ bytes = "1.5"
 bytesize = "1.3"
 chrono = { version = "0.4.35", default-features = false }
 clap = "4"
+moka = "0.12.10"
 concat-idents = "1"
 cpe = "0.1.5"
 criterion = "0.5.1"
 csaf = { version = "0.5.0", default-features = false }
 csaf-walker = { version = "0.10.0", default-features = false }
 cve = "0.3.1"
+deepsize = "0.2.0"
 env_logger = "0.11.0"
 fixedbitset = "0.5.7"
 futures = "0.3.30"
@@ -207,3 +209,6 @@ langchain-rust = { git = "https://github.com/chirino/langchain-rust", branch = "
 
 # to pickup fix: https://github.com/eikendev/sectxt/issues/21
 sectxtlib = { git = "https://github.com/ctron/sectxt", branch = "feature/fix_deps_1" }
+
+# to pickup feat: https://github.com/Aeledfyr/deepsize/pull/41
+deepsize = { git = "https://github.com/chirino/deepsize", branch = "main" }
--- a/modules/analysis/src/config.rs
+++ b/modules/analysis/src/config.rs
@@ -1,0 +1,22 @@
+use bytesize::ByteSize;
+use trustify_common::model::BinaryByteSize;
+
+#[derive(clap::Args, Debug, Clone)]
+pub struct AnalysisConfig {
+    #[arg(
+        id = "max-cache-size",
+        long,
+        env = "TRUSTD_MAX_CACHE_SIZE",
+        default_value = "200 MiB",
+        help = "Maximum size of the graph cache."
+    )]
+    pub max_cache_size: BinaryByteSize,
+}
+
+impl Default for AnalysisConfig {
+    fn default() -> Self {
+        Self {
+            max_cache_size: BinaryByteSize(ByteSize::mib(200)),
+        }
+    }
+}
--- a/modules/analysis/src/model.rs
+++ b/modules/analysis/src/model.rs
@@ -4,8 +4,14 @@ pub use roots::*;
 
 use petgraph::Graph;
 use serde::Serialize;
-use std::ops::{Deref, DerefMut};
-use std::{collections::HashMap, fmt};
+use std::{
+    fmt,
+    ops::{Deref, DerefMut},
+    sync::Arc,
+};
+
+use deepsize::DeepSizeOf;
+use moka::sync::Cache;
 use trustify_common::{cpe::Cpe, purl::Purl};
 use trustify_entity::relationship::Relationship;
 use utoipa::ToSchema;
@@ -24,7 +30,7 @@ impl fmt::Display for AnalysisStatus {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, ToSchema, serde::Serialize)]
+#[derive(Debug, Clone, PartialEq, Eq, ToSchema, serde::Serialize, DeepSizeOf)]
 pub struct PackageNode {
     pub sbom_id: String,
     pub node_id: String,
@@ -109,16 +115,33 @@ impl DerefMut for Node {
     }
 }
 
-#[derive(Debug)]
+pub type PackageGraph = Graph<PackageNode, Relationship, petgraph::Directed>;
+
 pub struct GraphMap {
-    map: HashMap<String, Graph<PackageNode, Relationship, petgraph::Directed>>,
+    map: Cache<String, Arc<PackageGraph>>,
+}
+
+#[allow(clippy::ptr_arg)] // &String is required by Cache::builder().weigher() method
+fn size_of_graph_entry(key: &String, value: &Arc<PackageGraph>) -> u32 {
+    (
+        key.deep_size_of()
+            + value.as_ref().deep_size_of()
+            // Also add in some entry overhead of the cache entry
+            + 20
+        // todo: find a better estimate for the the moka ValueEntry
+    )
+    .try_into()
+    .unwrap_or(u32::MAX)
 }
 
 impl GraphMap {
     // Create a new instance of GraphMap
-    pub fn new() -> Self {
+    pub fn new(cap: u64) -> Self {
         GraphMap {
-            map: HashMap::new(),
+            map: Cache::builder()
+                .weigher(size_of_graph_entry)
+                .max_capacity(cap)
+                .build(),
         }
     }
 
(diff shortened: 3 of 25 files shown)


chirino:
We should add some tests around the cache eviction edge cases:
* Verify old sboms are dropped as the cache fills
* Find out what happens when the cache is smaller a single request requires.

modules/analysis/src/model.rs
```diff
@@ -35,7 +37,35 @@ pub struct PackageNode {
...
+            + self.published.len()
+            + self.document_id.len()
+            + self.product_name.len()
+            + self.product_version.len();
+
+        PackageNode {
+            approximate_memory_size: size.try_into().unwrap_or(u32::MAX),
+            ..self.clone()
+        }
+    }
```
chirino, line 66: Is there a better way to estimate how much memory this thing uses up?

chirino:
@JimFuller-RedHat I think I'm happy with this now.  I'm going to keep it draft until #1217 goes in so you guys don't suffer the conflicts.  But if you have idle time maybe worth a gander in case you think this went in the wrong direction.

modules/analysis/src/model.rs
chirino: Kinda assuming that saving the approximate_memory_size will give us savings later, it might not.

chirino: I switched to using the DeepSizeOf train which has a derive macro and it seems to give much better results.

JimFuller-RedHat:
@chirino this looks really great (you are amazing!) ... and yes its likely to change a little bit after our PR (will review in full once we get that one into main) ... one thought was do we need to expose an env var to control params for easy configuration ?

modules/analysis/src/config.rs
```diff
@@ -0,0 +1,18 @@
...
+    let cfg = parse_size::Config::new().with_binary();
+    cfg.parse_size(s)
+}
+
+#[derive(clap::Args, Debug, Clone)]
+pub struct AnalysisConfig {
+    #[arg(
+        id = "max-cache-size",
+        long,
+        env = "TRUSTD_MAX_CACHE_SIZE",
```
JimFuller-RedHat, line 12: super!

Cargo.toml
```diff
@@ -54,12 +54,14 @@ bytes = "1.5"
 bytesize = "1.3"
 chrono = { version = "0.4.35", default-features = false }
 clap = "4"
+moka = "0.12.10"
```
JimFuller-RedHat, line 57: have you consider also exposing moka expiration policies TTL and TTI as env vars ? I was hoping that we can control both time and mem dimensions ... Moka looks great!

modules/analysis/src/service/load.rs
```diff
@@ -323,19 +327,25 @@ impl AnalysisService {
         // Set the result. A parallel call might have done the same. We wasted some time, but the
         // state is still correct.
 
-        self.graph.write().insert(distinct_sbom_id.to_string(), g);
+        let g = Arc::new(g);
+        self.graph_cache
```
JimFuller-RedHat, line 331: it occurred to me there might be times when we want to suspend caching (might be useful for debugging for example)... not sure if this is important or not to consider for right now

modules/analysis/src/service/mod.rs
```diff
@@ -38,7 +38,7 @@ use uuid::Uuid;
 
 #[derive(Clone)]
 pub struct AnalysisService {
-    graph: Arc<RwLock<GraphMap>>,
+    graph_cache: Arc<GraphMap>,
```
JimFuller-RedHat, line 41: I would wait till @ctron reviews ... previously he reworked things in this area and it was a bit tricky.

Cargo.toml
chirino: Yeah was considering it.   But at least as this PR is, we should be able to avoid blowing up the memory usage which was the main goal of this PR.

modules/analysis/src/service/test.rs
```diff
@@ -354,6 +355,45 @@ async fn test_status_service(ctx: &TrustifyContext) -> Result<(), anyhow::Error>
...
+    let small_sbom_size = service.cache_size_used();
+    assert!(small_sbom_size > 6 * kb);
+    assert!(small_sbom_size < 7 * kb);
+
+    ctx.ingest_documents(["spdx/quarkus-bom-3.2.11.Final-redhat-00001.json"])
+        .await?;
+    let all_graphs = service.load_all_graphs(&ctx.db).await?;
+    assert_eq!(all_graphs.len(), 2);
+
+    let big_sbom_size = service.cache_size_used() - small_sbom_size;
```
JimFuller-RedHat, line 379: very nice !

JimFuller-RedHat (approved):
apart from a few minor observations - looks great to me - as @ctron did the final fixes on the service invoke stuff good to get his eyeballs on it as well (as it was more then a little fiddly).

modules/analysis/src/config.rs
```diff
@@ -0,0 +1,18 @@
...
+        default_value = "200 MB",
+        help = "Maximum size of the graph cache.  You can use binary units like 'kb', 'mb' or 'gb'.",
+        value_parser = u64_from_size
```
ctron, line 15: I'd like to keep this aligned with the existing way to parse byte sizes: https://github.com/trustification/trustify/blob/a7e2855368841cd6e667b35aab5608d411330749/common/infrastructure/src/app/http.rs#L134-L141

modules/analysis/src/model.rs
```diff
@@ -4,8 +4,14 @@ pub use roots::*;
 
 use petgraph::Graph;
 use serde::Serialize;
-use std::ops::{Deref, DerefMut};
-use std::{collections::HashMap, fmt};
+use std::{
+    fmt,
+    ops::{Deref, DerefMut},
+};
+
```
ctron, line 11: Please merge and nest this.

modules/analysis/src/service/mod.rs
```diff
@@ -217,31 +237,24 @@ impl AnalysisService {
     fn query_graphs<'a, F>(
         &self,
         query: impl Into<GraphQuery<'a>> + Debug,
-        distinct_sbom_ids: Vec<String>,
+        graphs: Vec<(String, Arc<PackageGraph>)>,
```
ctron, line 240: It looks like it's not necessary to pass a `Vec`, so maybe use `&[Arc<PackageGraph>]` instead.

modules/analysis/src/service/render.rs
```diff
@@ -1,7 +1,7 @@
 use super::*;
 
 impl AnalysisService {
-    pub fn render_dot(&self, sbom: &str) -> Option<String> {
+    pub fn render_dot(&self, graph: Arc<PackageGraph>) -> Option<String> {
```
ctron, line 4: Is the `Arc` really required? Or is a `&PackageGraph` good enough?

server/src/profile/api.rs
```diff
@@ -289,6 +295,7 @@ impl InitData {
         };
 
         Ok(InitData {
+            analysis: AnalysisService::new_sized(run.analysis.max_cache_size),
```
ctron, line 298: I think it would be great it one could construct an `AnalysisService` from an `AnalysisConfig` directly.

modules/analysis/src/service/mod.rs
```diff
@@ -129,20 +129,48 @@ impl AnalysisService {
...
+    /// Default cache size is 100MB.
+    ///
     /// Therefore, it is ok to create a new instance. However, if you want to make use of the
     /// caching, it is necessary to re-use that instance.
     ///
     /// Also, we do not implement default because of this. As a new instance has the implication
     /// of having its own cache. So creating a new instance should be a deliberate choice.
     #[allow(clippy::new_without_default)]
     pub fn new() -> Self {
+        Self::new_sized(1024 * 1024 * 100)
```
ctron, line 141: I am worries this gets overlooked. It looks like we use this for tests only. One way could be to add `#[cfg(test]`. Or drop it an have a `const` default value for tests.

ctron (changes requested):
Looks good. A few things to iron out.

And I really would to have more nested `use` sections with every PR rather than less :wink:

modules/analysis/src/config.rs
helio-frota: nit mind to add a line on https://github.com/trustification/trustify/blob/main/docs/env-vars.md ?

chirino: will do

modules/analysis/src/config.rs
chirino: will do

modules/analysis/src/model.rs
chirino: ok.  still wish tooling automated this.

modules/analysis/src/service/mod.rs
chirino: I'll just drop it and fix it in the tests.

modules/analysis/src/service/render.rs
chirino: ref is good enough.

modules/analysis/src/service/mod.rs
```diff
@@ -217,31 +220,24 @@ impl AnalysisService {
...
         query: impl Into<GraphQuery<'a>> + Debug,
-        distinct_sbom_ids: Vec<String>,
+        graphs: &[(String, Arc<PackageGraph>)],
         mut f: F,
     ) where
         F: FnMut(&Graph<PackageNode, Relationship>, NodeIndex, &PackageNode, &mut FixedBitSet),
     {
         let query = query.into();
 
         // RwLock for reading hashmap<graph>
```
ctron, line 230: There's not rwlock anymore, is there?

server/src/profile/api.rs
chirino: Will change that around. And give AnalysisConfig a default.

modules/analysis/src/service/mod.rs
chirino: will drop

PR #1225, opened 2025-01-31
fix: insert affected status for versions up to the fixed ones

dejanb:
This PR fixes two things:

1. Adds "affected" purl status in the database for every CSAF "fixed" status up to the reported version. This corresponds to the behavior we see in the corresponding OSV converted files. As this might be Red Hat specific behavior, we do this only if cpe shows it's a red hat advisory
2. Properly sets  version range schema based on the purl type (e.g. rpm)


ctron:
TBH, I neither like this:

```
if cpe_vendor == "redhat" {
```

Nor this:

```
//TODO improve status checking as we call db way too often for this simple task
```

modules/ingestor/src/service/advisory/csaf/creator.rs
```diff
@@ -297,6 +323,34 @@ impl<'a> StatusCreator<'a> {
...
+    fn from_purl_version_scheme(purl: &Purl) -> VersionScheme {
+        match purl.ty.as_str() {
+            "maven" => VersionScheme::Maven,
+            "npm" => VersionScheme::Semver,
+            "python" => VersionScheme::Python,
+            "rpm" => VersionScheme::Rpm,
+            "semver" => VersionScheme::Semver,
+            _ => VersionScheme::Generic,
+        }
+    }
```
mrizzi, line 335: It looks like this function is pretty similar to https://github.com/trustification/trustify/blob/106ee32185925cb367dd61a0046c85b8f8df0c73/modules/ingestor/src/service/advisory/cve/loader.rs#L308-L319
so it might be worth thinking about having some `impl From` in `VersionScheme`.

WDYT?

mrizzi (commented):
Commented about some possible duplication version scheme mapping to collect your feedback.

dejanb:
@ctron I refactored the status id lookups. The first commit, just uses local cache per document. The second one moves it up to the ingestor. It works well, but I don't like it propagates through all the method calls.

I tried with the global static cache and that works for the app, but the tests are failing as there are multiple databases running in parallel with different ids.

One idea is to maybe add these kind to the Graph or Database structs?

Anyhow, give it a look and let me know if you have any better ideas.

ctron:
I'd like to avoid "global static". That always looks easy and then gets messy. Like you mentioned, with testing.

I am ok having an instance, carrying that around. Like we do with the `AnalysisService` too. That's basically one "global" instance per application instance. Assuming we don't run into any "freshness" issues. I didn't check. But I think that the status never changes, right?

I think we miss a pattern in general, which spans optimizations across multiple documents. Which would be ok. But if that shared status cache could help, let's try it. Maybe wrap things in a "context" object to have less parameters on functions.

My question would be: does the last step bring enough performance improvements? Maybe we can measure. Maybe @JimFuller-RedHat has an idea about that.

entity/src/version_scheme.rs
```diff
@@ -28,3 +28,30 @@ pub enum VersionScheme {
...
+///
+/// However, the reality looks quite weird. The following command can be run to get an overview of
+/// what the current state holds. Run from the `cves` directory of the repository from:
+/// <https://github.com/CVEProject/cvelistV5>
+///
+/// ```bash
+/// find -name "CVE-*.json" -exec jq '.containers.cna.affected?[]?.versions?[]?.versionType | select (. != null )' {} \; | sort -u
+/// ```
+
+impl From<&str> for VersionScheme {
```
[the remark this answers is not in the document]
mrizzi: thank you 👍

entity/src/status.rs
```diff
@@ -24,3 +25,24 @@ impl Related<purl_status::Entity> for Entity {
...
+    Eq,
+    Hash,
+    Debug,
+    PartialEq,
+    strum::EnumString,
+    strum::Display,
+    Serialize,
+    Deserialize,
+)]
+#[strum(serialize_all = "camelCase")]
```
ctron, line 41: I think for most of the structs and enums we gave up on camelCase. Unless there's a good reason, we should stick to the default here too. I know, as a Java developer, my heart bleeds as well :grin:

modules/ingestor/src/service/advisory/csaf/creator.rs
```diff
@@ -213,39 +201,56 @@ impl<'a> StatusCreator<'a> {
...
-                    status: product.status,
-                    info,
+                // Insert purl status
+                let spec = match &purl.version {
+                    Some(version) => VersionSpec::Exact(version.clone()),
+                    None => VersionSpec::Range(Version::Unbounded, Version::Unbounded),
                 };
-
-                self.entries.insert(purl_status);
+                self.create_purl_status(&product, purl, scheme, spec, status_id);
```
ctron, line 223: It feels like this could benefit from batch insert too.

ctron: Sorry, looks like that's just an in-memory state.

modules/ingestor/src/service/mod.rs
```diff
@@ -141,9 +141,9 @@ impl ResponseError for Error {
 
 #[derive(Clone)]
 pub struct IngestorService {
-    graph: Graph,
-    storage: DispatchBackend,
-    analysis: Option<AnalysisService>,
+    pub graph: Graph,
```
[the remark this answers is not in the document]
dejanb: Fixed. This remained from the previous approach

entity/src/status.rs
dejanb: Yes, `snake_case` is actually the right one to use here.

dejanb:
I did some more testing of these changes.

1. I ran CSAF 2024 ingestion on both PR branch and main

```
PR
Executed in  262.16 mins    fish           external
   usr time   20.74 mins    0.08 millis   20.74 mins
   sys time    2.55 mins    1.12 millis    2.55 mins

Main
Executed in  222.15 mins    fish           external
   usr time   17.78 mins    0.10 millis   17.78 mins
   sys time    2.71 mins    2.08 millis    2.71 mins
```

The ingestion is a bit slower. As far as I could figure so far, the hashing and producing predicable UUIDs is slower than generating random ones, so that explains it. But using predictive ones is the right logic, so we don't make too much choice really.

Also, this changes ingests order of magnitude less version ranges and product statuses, which should improve querying side of the app.

```
Main

trustify=# select count(*) from version_range ;
  count
---------
 1600120
(1 row)

trustify=# select count(*) from version_range ;
 count
--------
 102044
(1 row)
```

2. On API side, I ran load tests locally against these two datasets with additional ds3 sboms

There's one notable slowdown in search API call that needs to be investigated.
Screenshot 2025-02-13 at 10 43 54
Screenshot 2025-02-13 at 10 43 37

Next steps:

1. OSV and CVE ingestors still don't use batch ingesting. I think that's where we can make the most performance gains
2. CSAF code need to be refactored. There's a lot of improvement that can be done to make code more readable and maintainable
3. Investigate slowness in search

JimFuller-RedHat:
are u basing your analysis on latest main ?

ctron (approved):
I think it's a solid improvement, and we do understand the performance impact it has. It has test cases as well! So I'd say we merge this!

dejanb:
@JimFuller-RedHat the latest main changes fix the search performance

Screenshot 2025-02-13 at 14 03 01

Issue #1229, opened 2025-02-03
Allow other graph rendering formats


ctron:
Today, we allow exporting analysis graphs using `dot` (graphviz) files.

We could (and should) add other formats too.

How would we tell the backend which format we want?

chirino:
Could we make the extension of the endpoint control the format?   Example:

```rust
#[get("/v2/analysis/sbom/{sbom}/render.{ext}")]
pub async fn render_sbom_graph(
```

ctron:
In a case of an unknown/unsupported extension, would we return 404? Or something different?

helio-frota:
> Or something different?

> unknown/unsupported extension

I vote for `unknown/unsupported extension` error message

ctron:
`415 Unsupported Media Type`

chirino:
I would say 404 for an unsupported extension.

chirino:
given the extension is part of the resource path I think 404 is better than the 415.  Now if we have a resource path without an extension and the client chooses the type using the Accept header, then 415 is better.

ctron:
How would you, automatically, differentiate between: "sbom not found" and "unsupported renderer"?

chirino:
I don't think you can with just the status code if you have a URL with the extension being part of it.

PR #1230, opened 2025-02-03
refactor: prepare for having multiple render formats

ctron:
Closes: https://github.com/trustification/trustify/issues/1229


modules/analysis/src/endpoints/mod.rs
```diff
@@ -112,23 +114,31 @@ pub async fn search_component(
...
+    let (sbom, ext) = path.into_inner();
+
+    let Ok(ext) = serde_json::from_value::<Renderer>(json!(ext)) else {
+        return Ok(HttpResponse::UnsupportedMediaType().finish());
+    };
+
     service.load_graph(db.as_ref(), &sbom).await;
 
-    if let Some(data) = service.render_dot(&sbom) {
+    if let Some(data) = service.render(&sbom, ext) {
```
chirino, line 141: might make more sense if render control the content type too:

```suggestion
    if let Some((data, content_type)) = service.render(&sbom, ext) {
```

chirino (approved):
Looks good to me.

Issue #1236, opened 2025-02-04
Don't re-ingest SBOMs


ctron:
In case of re-ingesting an existing SBOM, we create a new instance of it. Although the digest hasn't changed.

Ingesting an SBOM we already ingested (with the exact same digest) we seem to re-create an instance. It looks like this comes from the "document" table, which creates a new entry with a UUID v4 ID. Also, afterwards, we create a new entry in the SBOM table with a v7 UUID.

I think, we should only create a new entry if no entry in the "document" table with that hash exists.

I am not sure if this also applies to other documents, but it might.

Issue #303, opened 2024-05-16
Re-ingesting an SBOM will leave traces of previous version


ctron:
Right now, when ingesting an SBOM (probably advisories too) creates all new components (if they don't yet exist).

However, this pattern will not remove components which got removed.

JimFuller-RedHat:
sounds like we first need a test to confirm this

ctron:
> sounds like we first need a test to confirm this

I am also not sure if that is a the case. As a different SBOM will have a different digest, and that's how we separate that now. Which brings this discussion back to the ID problem.

JimFuller-RedHat:
so this is a kind of optimisation - as things are today they are simple but take up more space (which might impact operation) ... I guess we list such duplicates of packages in the ux as well.

ctron:
> I guess we list such duplicates of packages in the ux as well.

Yes, I think we just should come up with expectations on how this should work. Right now it's simple, costs more storage, and might show duplicates. If that's ok, then it's ok :)

JimFuller-RedHat:
closing this

ctron:
I think we still have this. Re-ingesting an SBOM for which we create synthetic component IDs (like some CDX cases) still suffers from this.

ctron:
IIRC that was fixed a while back.

PR #1237, opened 2025-02-04
Fix issue with duplicate CDX nodes

ctron:
Also: 

* Re-add indices for digest columns on the source document table
* Reduce logging of happy path during ingestion, has a performance impact
* Honor the existence of other documents too, although that didn't cause any issues so far

Closes: #1236, #303

### DIFF
--- a/migration/src/m0000870_source_doc_indexes.rs
+++ b/migration/src/m0000870_source_doc_indexes.rs
@@ -1,0 +1,87 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+#[allow(deprecated)]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .create_index(
+                Index::create()
+                    .table(SourceDocument::Table)
+                    .name(Indexes::Sha256Index.to_string())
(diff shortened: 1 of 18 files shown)


modules/ingestor/src/graph/mod.rs
```diff
@@ -28,3 +29,38 @@ impl Graph {
         Self { db }
     }
 }
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Outcome<T> {
+    Existed(T),
```
JimFuller-RedHat, line 35: minor: I do not fully understand the decomposition of Existed and Added state .... I mean I guess we could infer on True/False on Added ... so we want to sometimes add over an existed ... a few words explaining the intent of Outcome probably would help.

migration/src/m0000870_source_doc_indexes.rs
```diff
@@ -0,0 +1,87 @@
...
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+#[allow(deprecated)]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .create_index(
```
JimFuller-RedHat, line 11: thx for this - I believe we had these before and after the refactoring into source_document .. not sure when/why this happened - glad they are back.

JimFuller-RedHat (approved):
apart from minor - LGTM

PR #1242, opened 2025-02-05
OpenAPI generated files are not committed to UI repository

gildub:
Relates to https://github.com/trustification/trustify-ui/pull/331


chirino:
If the generated client is not checked in won't that mean the project is broken in things like IDEs until you build it?

gildub:
@chirino, not really because the IDE doesn't have to deal with it. But if necessary then one can use the generate task `npm run generate`

chirino (approved):
lgtm

PR #1244, opened 2025-02-05
chore: reduce amount of logs on info level

ctron:


helio-frota:
I'm +1 when thinking on performance based on [these non-scientific tests](https://github.com/trustification/trustify/discussions/756#discussioncomment-11914217)

```
#[instrument(skip_all, ret)],
Your change is similar to skip_all when no error occurs
#[instrument(skip_all)] 
```

helio-frota (approved):

PR #1250, opened 2025-02-05
fix: add index for organizations by name

ctron:
A primary way of finding Organizations is "by name". I think we should have an index for that.


JimFuller-RedHat (approved):
Is this proved  by showing a slow performing query plan improved ? How many organisations will we ever have eg. if its less then a thousand a seq scan is going to be super fast (and internally cached in db)... alternately adding this index will not hurt anything as it will not take a lot of time during ingestion or maintaining so LGTM!

PR #1256, opened 2025-02-05
fix: deserialize relationships filter from query params

### LINKED ISSUE #1232: api/v2/analysis/component/{}?relationships= throws deserialize error
Was expecting the following

```bash
curl -H "Authorization:$(oidc token testing-client --bearer)" "http://localhost:8080/api/v2/analysis/component/BB?ancestors=5&descendants=5&relationships=contains" -v
```

to return both ancestors and descendants with only **contains** relationships.

The error thrown:

```
Query deserialize error: invalid type: string "contains", expected a sequence%
```

seems to indicate this is an issue with relationship url param and our current Query internals.

The underlying service works - as proven out by `test simple_by_name_analysis_service_filter_rel`.

jcrossley3:
Fixes #1232

### DIFF
--- a/entity/src/relationship.rs
+++ b/entity/src/relationship.rs
@@ -60,3 +60,26 @@ impl fmt::Display for Relationship {
         write!(f, "{:?}", self)
     }
 }
+
+impl From<&str> for Relationship {
+    fn from(s: &str) -> Self {
+        match s.to_lowercase().as_str() {
+            "contains" => Self::Contains,
+            "dependency" => Self::Dependency,
+            "devdependency" => Self::DevDependency,
+            "optionaldependency" => Self::OptionalDependency,
+            "provideddependency" => Self::ProvidedDependency,
+            "testdependency" => Self::TestDependency,
+            "runtimedependency" => Self::RuntimeDependency,
+            "example" => Self::Example,
+            "generates" => Self::Generates,
+            "ancestorof" => Self::AncestorOf,
+            "variant" => Self::Variant,
+            "buildtool" => Self::BuildTool,
+            "devtool" => Self::DevTool,
+            "describes" => Self::Describes,
+            "package" => Self::Package,
+            _ => Self::Undefined,
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


chirino (approved):
lgtm.

chirino (dismissed):
So the openapi client will genereate requests that look like: 
```
http://localhost:3000/api/v2/analysis/component/test?ancestors=10&relationships[]=contains&relationships[]=dependency
```

Any chance we could parse that?

chirino:
Ah shit maybe we just need to fix the test to do it the openapi way!..

chirino:
@jcrossley3 I think you can close, https://github.com/trustification/trustify/pull/1257 is the simple fix.

jcrossley3:
> @jcrossley3 I think you can close, #1257 is the simple fix.

I agree that fix is simple, but it changes the test, and it forces me to remember to add `[]` after the field name. Does anyone really want that?

carlosthe19916:
My understanding from the internal chat is that we need to pass an array of values through the query params.

From my experience:
- If the queries get to a very complex level we could consider migrating GET to POST. E.g. `GET /myendpoint?complex_params` to `POST /myendpoint with a body` . Main point, set multiple combination of queries in the body of the POST request
- We can use what we have already in all the search endpoints (my example is without encoding).
```shell
/api/v2/vulnerability?limit=10&offset=0&q=average_severity=low|medium|high|critical
```

Notice that there is a variable `average_severity` that is receiving an array of possible values.

chirino:
@carlosthe19916 so the js openapi client can already pass the array with a js call like: 
```
getComponent({
    path: { key: "test" },
    query: {
      ancestors: 10,
      relationships: ["contains", "dependency"],
    },
  });
```

But sadly this PR breaks the above, so that users can manually construct URLs easier.  

#1257 keeps the above working and also supports constructing the URL easy by adding a new field.

JimFuller-RedHat (approved):
LGTM! and thx for litigating the whole conversation on this!

PR #1257, opened 2025-02-05
fix api/v2/analysis/component/{}?relationships= throws deserialize error

### LINKED ISSUE #1232: api/v2/analysis/component/{}?relationships= throws deserialize error

chirino:
Fixes #1232

Test case just need the right url encoding for the field.

### DIFF
--- a/modules/analysis/src/endpoints/mod.rs
+++ b/modules/analysis/src/endpoints/mod.rs
@@ -60,7 +60,6 @@ pub async fn analysis_status(
     operation_id = "getComponent",
     params(
         ("key" = String, Path, description = "provide component name, URL-encoded pURL, or CPE itself"),
-        Query,
         Paginated,
         QueryOptions,
     ),
--- a/modules/analysis/src/endpoints/test.rs
+++ b/modules/analysis/src/endpoints/test.rs
@@ -786,7 +786,6 @@ async fn spdx_package_of(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
 
 #[test_context(TrustifyContext)]
 #[test(actix_web::test)]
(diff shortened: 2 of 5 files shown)


jcrossley3:
Just to be perfectly clear, for what I believe to be the common case -- passing a single value -- we're going to have to remember to include brackets, e.g. `&relationships[]=contains`, right?

And for the less common case (I think) of passing multiple values, we'd have `&relationships[]=contains&relationships[]=variant` instead of say, `&relationships=contains,variant`.

All to appease auto-generated openapi clients? I don't like http array params. :smiling_face_with_tear:

chirino:
We could a relationship: Option<Realationship> field to the query struct to handle that easy case.

chirino:
@jcrossley3 I've updated the PR to use the simpler `relationship: Option<Relationship>` field.

jcrossley3:
Having both a `relationships` and a `relationship` field feels wrong to me. We're adding complexity without any real value other than overcoming the weirdness of http array params. Just my opinion, I'll yield to the majority.

chirino:
This PR actually does not fix the actual issue, passing `?relationshipsp[]=contains` does not add `contains` to the set.  Will close, in favor other other approaches.

Issue #1255, opened 2025-02-05
OTEL metrics - Infra rust code


helio-frota:
Remove unnecessary crates and refactor the code to enable custom metrics

PR #335, opened 2024-05-23
Nobody else seems to use utoipa-swagger-ui with actix-web-prom

jcrossley3:
Fixes #302

For a very broad definition of "fixes".

Also gives us a straightforward way of setting other logging options as default.

PR #1254, opened 2025-02-05
feat: add custom metrics, and an initial cache size metric

ctron:
Closes: https://github.com/trustification/trustify/issues/1255

Issue #355, opened 2024-06-04
Restarting trustd while an importer is running has the job stuck in `running`


ctron:
Steps to reproduce:

* Start server
* Have an importer task in status `running`
* Terminate the process
* Start the server (again)

The job is in status `running`, but actually is not. And it won't ever go into any other state.

PR #1261, opened 2025-02-06
chore: Remove actix_web_prom from RUST_LOG

helio-frota:
Related to #1255 
Related to #335


jcrossley3:
Why are we doing this? Is this reverting #355?

jcrossley3:
> Why are we doing this? Is this reverting #355?

Ah, removed in #1254

jcrossley3 (approved):

helio-frota:
Yes, that's correct. I referenced the issue because of its description. 👍
as Jens sent the PR first and I created the issue later :sweat_smile:

PR #1267, opened 2025-02-07
chore: add index for DELETE operation

ctron:
During ingestion, do we do clean up descriptions using:

```sql
DELETE FROM "vulnerability_description" WHERE "vulnerability_description"."advisory_id" = $1
```

That shows up in the AWS RDS console as top SQL statement. I think an index might help. But I am guessing.


JimFuller-RedHat (approved):
LGTM

Issue #1207, opened 2025-01-28
Allow running multiple importer jobs at the same time


ctron:
As this issue already has links established to relevant sub-issues, let's re-purpose this issue to be the User Story that defines the expected functionality for the first downstream release.

As a TPA user I want to have 4 importers preconfigured; Github Security Advisory data (enabled), Vulnerability (CVE) data (enabled), Red Hat SBOM (disabled) and Red Hat Vex (disabled) and able to execute in parallel so that I can efficiently populate the application with data from external sources.

Acceptance criteria:

GHSA Importers enabled by default to load data in temporal reverse order (latest first) based on GHSA git repository folders structure
CVE Importer enabled by default to load data in temporal reverse order (latest first) based on CVE git repository folders structure
RH SBOM Importer by default disabled
RH VEX Importer by default disabled
Importers run in parallel

The corresponding downstream issue is [TC-2218](https://issues.redhat.com/browse/TC-2218)
    

The original issue description is below, and this still adds value. 
_A single instance of the importer service should be able to run multiple importer runs at the same time._

_Also, multiple instances of the importer service should be able to distribute importer runs between them._

_I think it is worth solving those two problems at the same time, since the might be related and might use a common foundation of sharing the load._

jcrossley3:
This is now fixed by both/either #1278 and/or #1321

PR #1278, opened 2025-02-10
Make the importer execution more concurrent, i.e. faster

jcrossley3:
WIP for #1207


modules/importer/src/runner/csaf/mod.rs
```diff
@@ -103,7 +103,7 @@ impl super::ImportRunner {
 
         Walker::new(source)
             .with_progress(WalkerProgress(progress))
-            .walk(filter)
+            .walk_parallel(10, filter)
```
ctron, line 106: We should not do this. At least, not force it. Doing this with RH data, you will get IP blocked right away.

ctron (changes requested):
I think we should have a concept of what parallel means.

I think it makes sense to:

* Allow running multiple instances of the importer instance
* Allow running multiple importer runs in parallel

But I don't think it makes sense to run a single importer in a parallel mode.

modules/importer/src/runner/csaf/mod.rs
JimFuller-RedHat: agree with @ctron in that better horizontal scalability running multiple instances and multiple importer is what is most useful .... though for the single instance use case ... we can do this if we default to some reasonable value like 4 and control with env var (and document)

ctron: I'd prefer to control this via the importer configuration. Which can be different per source. But yes.

jcrossley3: Fair points. I'll defer the configuration and revert the `walk_parallel`.

JimFuller-RedHat:
also ... I wonder at what point do we would consider bringing in a job queue crate versus building one ourselves ... I have no experience with any rust crates so maybe there are none fit for purpose ?

jcrossley3:
> also ... I wonder at what point do we would consider bringing in a job queue crate versus building one ourselves ... I have no experience with any rust crates so maybe there are none fit for purpose ?

I'm thinking that threshold is very high. Because if we're requiring a _distributed_ job queue, then I expect any crate would wrap redis/rabbit/etc.

ctron:
I think there's a bunch of way we could deal with this. One way could be a postgres based locking mechanism. `ON LOCK SKIP`, or whatever that was called.

Downside, it only works with postgres. Which might be ok.

Another way could be do have a hash approach: have a max number of importers, and an individual instance number, and then only process those jobs. Should work with a statefulset.

we could also hold a state in the database. But then we'd need to define such things as timeouts, stale detection, etc.

jcrossley3:
If we think StatefulSet is the best option, then I'll re-assign this to you. :)

ctron:
To my understanding, this PR simply checks all importers in parallel. With no limitation. Waiting for the joint outcome of a single check run.

I think this would result in the case that one importer runs for a long time, blocking the next check, and thus not really improving the situation. On an initial startup with pending import runs, yes. But afterwards, no.

I think we'd need a strategy first, with the goal mentioned above, before having an implementation.

jcrossley3:
> I think this would result in the case that one importer runs for a long time, blocking the next check, and thus not really improving the situation. On an initial startup with pending import runs, yes. But afterwards, no.

True. So no worse than what we have, but does address the primary motivation for this PR: that QE will only have to wait as long as it takes the longest enabled importer to run for them all to be finished.

Plus, better logging.
 
> I think we'd need a strategy first, with the goal mentioned above, before having an implementation.

Agree. Still thinking about that.

ctron:
I would not want to merge something which only suits some specific QE task. Potentially causing issues for the overall system. Like the case that most of the times it runs sequential, while sometimes it does run in parallel. Without any way of controlling that behavior.

jcrossley3:
> To my understanding, this PR simply checks all importers in parallel. With no limitation. Waiting for the joint outcome of a single check run.

I added a limitation. Defaults to the old behavior: concurrency=1.

modules/importer/src/runner/common/walker/git.rs
```diff
@@ -294,7 +282,7 @@ where
         }
     }
 
-    #[instrument(skip(self), err)]
+    #[instrument(skip(self))]
```
ctron, line 285: I'd like to keep errors. It's ok to reduce the level to e.g. INFO

ctron (approved):
I personally think that's better. And it should also be possible to easily extend that implementation to a distributed version (running X instances at the same time).

I'd just prefer keeping the `err` instrumentation. Also converting the `ret` to `err` instead. And maybe lowering that level to `Info`, as it's not an application level error. But that's not a blocker to me.

ctron (changes requested):
I am ok with this in general. But it also feels incomplete due to ongoing discussions on the [ADR](https://github.com/trustification/trustify/pull/1306). Let's finish that discussion first, and then merge implementations.

jcrossley3:
> I am ok with this in general. But it also feels incomplete due to ongoing discussions on the [ADR](https://github.com/trustification/trustify/pull/1306). Let's finish that discussion first, and then merge implementations.

Come on, man! They're independent. This PR simply adds the ability for a single instance to run multiple import jobs concurrently. Heck, this PR alone might mitigate the need for distributing work at all. It certainly satisfies the need that motivated the creation of the [original downstream issue](https://issues.redhat.com/browse/TC-2218).

jcrossley3:
> Before we go distributed, we should make sure the DB is not bottlenecking us.

Amen. 100%. I think it's also worth looking at the logic within individual importers. The git-based ones are flawed since they're sync, but they only take a few minutes to run. The sbom importer's rate varies widely during a single run, and the CSAF importer just seems broken -- I've never seen it run to completion.

ctron:
> The git-based ones are flawed since they're sync,

IIRC the run on the tokio worker thread pool. Not blocking anything.

> CSAF importer just seems broken

I have. Multiple times: https://console-openshift-console.apps.cluster.trustification.rocks/search/ns/trustify-dumps?kind=core~v1~Pod&q=app.kubernetes.io%2Fcomponent%3Dcreate-dump

jcrossley3:
> IIRC the run on the tokio worker thread pool. Not blocking anything.

Blocks a clean shutdown when those sync `git clone` functions are running. Only takes a minute the first time, but 30 minutes the second, so it'a a pain to have to manually `kill -9` the trustd process.

ctron:
> > IIRC the run on the tokio worker thread pool. Not blocking anything.
> 
> Blocks a clean shutdown when those sync `git clone` functions are running. Only takes a minute the first time, but 30 minutes the second, so it'a a pain to have to manually `kill -9` the trustd process.

Maybe that's a local problem then. It looks ok on our OCP instance:

image

jcrossley3:
> image

Using the timestamps in your image, if you had attempted to `Ctrl+C` to break out of a `cargo run trustd` process at 2/13 15:30:00, the process would hang, I assume waiting for that 30 minute `git clone` to complete. If instead you hit `Ctrl+C` at 2/13 19:00, no problem, no hang. It'd even be fine at 18:00, still processing, but post `git clone`.

Try it for yourself.

modules/importer/src/server/mod.rs
```diff
@@ -56,84 +61,52 @@ struct Server {
...
-                        None,
-                        last_run,
-                        last_error,
-                        continuation,
-                        report.and_then(|report| serde_json::to_value(report).ok()),
-                    )
-                    .await?;
-            }
+            // Add them to our list of currently running jobs
+            runs.extend(todo.into_iter());
```
chirino, line 109: could the `todo` list contain an importer already in the `runs` list?

jcrossley3: The check on line 97, `runs.contains_key(&i.name)` is meant to prevent that.

chirino:
you might want to include an update to https://github.com/trustification/trustify/blob/main/docs/env-vars.md too.

chirino (approved):
LGTM.

jcrossley3:
> you might want to include an update to https://github.com/trustification/trustify/blob/main/docs/env-vars.md too.

Done, thx

JimFuller-RedHat:
**Suggestion**: 
This change is a 'rubicon' of sorts eg. going from predictable serial processing to something more async was always something we had to cross to get better scalability. Would be useful to use a branch and test a bit more at scale this change. Impl level details (like considering skip locking) could be guided by those tests.

jcrossley3:
> **Suggestion**: So while the changes seem straightforward - would be useful to use a branch and test a bit more at scale to assess impact (maybe this has been done and I am ignorant!).

The default value of the new `concurrency` option is 1, meaning the default "new" behavior should match the "old" sequential behavior. My local tests bear that out, but I've done nothing at scale.

My goal for the PR was to match current behavior by default after merging, so would the `concurrency` option serve the same purpose as a branch?

JimFuller-RedHat (approved):
LGTM with the caveat that testing might reveal need for more work in this area. I think this should go in.

PR #1282, opened 2025-02-11
feat: add db parameter group, with current settings

ctron:


helio-frota (approved):
lgtm ?

no idea, approving because of the blank spaces removal

PR #1283, opened 2025-02-11
chore: reducing the happy path logging

ctron:


helio-frota (approved):
I already agree this is a reduction instead of full return it will deal with the error only.
So yeah approving.

PR #1285, opened 2025-02-11
[release/0.2.z] TC-2240 OSV GHSA - Ecosystem RubyGems

mrizzi:
https://issues.redhat.com/browse/TC-2240

Backport:

- https://github.com/trustification/trustify/pull/1281

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8977,7 +8977,7 @@ dependencies = [
 [[package]]
 name = "trustify-ui"
 version = "0.1.0"
-source = "git+https://github.com/trustification/trustify-ui.git?branch=publish%2Fmain#7764cc9b15f82d3fc7c9112572b1d207b217a7eb"
+source = "git+https://github.com/trustification/trustify-ui.git?branch=publish%2Frelease%2F0.2.z#cd7827e5b8d2fe437298df7ae6174b614b465ac5"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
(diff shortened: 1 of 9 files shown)


ctron (changes requested):
I'm against backporting by default. That turned out to be tricky with the last release. Let's just merge main into the release branch before doing a release,.

mrizzi:
> I'm against backporting by default. That turned out to be tricky with the last release. Let's just merge main into the release branch before doing a release,.

I can see the "all-in" pattern works fine in this phase when no selectivity in what to release (vs what is committed into `main`) is required.
But in the mid-term, how can we think about having this selectivity without it being tricky?
In the past, in other projects, we successfully used a tool that did the backport for us simply applying two labels on the PR (opened on `main` branch) to-be-backported, so really pretty basic with no effort involved but with the great benefit of having selectivity to ensure project maintenance.

Could we give this tool a try in this phase to decide if it fits in this context as well?

@ctron @dejanb @jcrossley3 WDYT?

ctron:
> In the past, in other projects, we successfully used a tool that did the backport for us simply applying two labels on the PR (opened on `main` branch) to-be-backported, so really pretty basic with no effort involved but with the great benefit of having selectivity to ensure project maintenance.
> 
> Could we give this tool a try in this phase to decide if it fits in this context as well?

I'd love that.

mrizzi:
Merged into `release/0.2.z` branch when releasing trustify `0.2.4`

PR #1289, opened 2025-02-12
feat: implement ingestion of SPDX and CDX external references

ctron:


modules/ingestor/src/graph/sbom/common/external.rs
```diff
@@ -0,0 +1,115 @@
...
+        });
+
+        let (r#type, value) = Discriminator::split(discriminator);
+
+        self.externals.push(sbom_external_node::ActiveModel {
+            sbom_id: Set(self.sbom_id),
+            node_id: Set(node_id.into()),
+            external_doc_ref: Set(ext_document_id.into()),
+            external_node_ref: Set(ext_node_id.into()),
+            external_type: Set(ExternalType::SPDX),
```
JimFuller-RedHat, line 67: do u plan to add commits to handle ::CDX ?

ctron: Good point.

JimFuller-RedHat (approved):
apart from clippy, this is a good interim PR

ctron:
Ok, I fixed the limitation with SPDX/CDX. And also added a way to have mode for the tests, which are neither SPDX nor CDX.

PR #1292, opened 2025-02-12
perf- add correct indexes for api/v2/purl?q=

JimFuller-RedHat:


ctron (approved):
LGTM: We just be need to careful to not have `GIN` indices and the `tonic` crate come too close.

PR #1293, opened 2025-02-12
TC-2243 TC-2244 TC-2246 OSV GHSA - Ecosystem Hex (Erlang), Swift, Pub (Dart/Flutter)

mrizzi:
https://issues.redhat.com/browse/TC-2243
https://issues.redhat.com/browse/TC-2244
https://issues.redhat.com/browse/TC-2246

### DIFF
--- a/entity/src/version_scheme.rs
+++ b/entity/src/version_scheme.rs
@@ -32,4 +32,5 @@ pub enum VersionScheme {
     Packagist,
     NuGet,
     Gem,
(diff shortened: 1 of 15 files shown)


mrizzi:
> I think it would be good to have some info in the title of a PR which is publicly available.

Sorry, my bad: I have `enter` instead of

dejanb (approved):

Issue #1241, opened 2025-02-05
Use Tempo instead of Jaeger


helio-frota:
See [Jaeger deprecation note](https://docs.openshift.com/container-platform/4.17/observability/distr_tracing/distr_tracing_jaeger/distr-tracing-jaeger-removing.html)

- [x] Add to the [dev env with podman compose](https://github.com/trustification/trustify/blob/main/docs/otel.md) 
- [ ] Add to the [helm charts](https://github.com/trustification/trustify-helm-charts)

PR #1305, opened 2025-02-14
Adding tempo to podman compose

helio-frota:
Related with #1241


etc/telemetry/otel-tempo-config.yaml
```diff
@@ -0,0 +1,25 @@
+receivers:
```
ctron, line 1: So far we seem to have the pattern of `compose-*.yaml`. I think we should stick to that.

helio-frota: thanks for review, PR updated :+1:

PR #1306, opened 2025-02-14
doc: ADR for importer scalability

jcrossley3:
Relates #1207

Preview: https://github.com/jcrossley3/trustify/blob/adr-importer/docs/adrs/00004-importer-scalability.md

### DIFF
--- a/docs/adrs/00004-importer-scalability.md
+++ b/docs/adrs/00004-importer-scalability.md
@@ -1,0 +1,62 @@
+# 00004. Importer Scalability in Trustify
+
+Date: 2025-02-14
+
+## Status
+
+Draft
+
+## Context
+
+This ADR is concerned with our ability to scale the importer processes
+that ingest documents -- SBOM's and advisories -- from various
+sources.
+
+Currently, the importer process iterates over the enabled importer
+definitions in the database sequentially, running each according to
+its schedule. Executing more importer processes results in redundant
+work, each of them doing the same thing, i.e. there is no coordination
+to distribute the load among them.
+
+### Requirements
+
+* The work should be distributed -- no two importer processes should
+  be doing the same work.
+* The solution should be fault-tolerant -- importer instances may come
+  and go -- any work in process by a crashed/killed one must be
+  eventually picked up by another.
+* In short, adding an importer process should increase the ingestion
+  throughput, i.e. the system should be scalable.
+
+## Decision
+
+Since trustify already depends on PostgreSQL -- the import definitions
+persist there -- we can persist the "coordination state" there, too.
+
+Two columns can be added to the `IMPORTER` table:
+* `runner` -- a string identifying the instance currently running the
+  import, e.g. `{hostname}+{PID}`
+* `heartbeat` -- a timestamp regularly updated by the runner to
+  indicate it's still alive while processing the job.
+
+Each importer process will identify rows in the table representing
+jobs ready to be run by not only comparing the `last_run` field to the
+configured schedule, but also that there is no current `runner`
+associated with the job or the current runner hasn't recently updated
+its `heartbeat`.
+
+Once a row has been identified, the importer process will
+optimistically lock it by attempting to update the `runner` field with
+its own identifier, e.g.
+
+```
+  UPDATE importer SET runner = 'foo.com+12345' WHERE name = 'cve' AND runner IS NULL
+```
+
+If the update succeeds, the runner should then regularly update
+`heartbeat` while processing the job. A failed update indicates some
+other instance took the job.
+
+Once the job is complete, the runner is expected to set both fields
+back to NULL, making the row available to other importer processes
+when the schedule fires again.


ctron:
## Alternative A

I'd try to avoid storing a state in the database. One way to deal with this could be to have a hashing implementation, where each instance of the importers has a number assigned, and knows the total number of importers. With that, each importer can filter for "its own" importer jobs, and only process these. For example:

```
let total = 10; // total number of importer runners
let instance = 5; // this instance number

if hash(importer_name) % total == instance {
 // process it
} else {
 // ignore it
}
```

## Alternative B

We let the DB deal with locking. Using `SKIP LOCKED`: https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE

This way an importer can select an importer for processing, and have the DB lock that importer for us. Other importers would not pick up that locked entry, but "skip" it and pick the next one.

jcrossley3:
> I'd try to avoid storing a state in the database. 

Why?

> One way to deal with this could be to have a hashing implementation, where each instance of the importers has a number assigned, and knows the total number of importers.

Who/what assigns each instance a number, and how does each instance know the total number of importers?

> ## Alternative B
> We let the DB deal with locking. Using `SKIP LOCKED`: https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE

My ignorance of that feature aside, I'd be wary of any pessimistic locking approach for this situation, mostly for dealing with crashed instances.

Using a database-specific feature means current/future maintainers of the code will need to understand the nuances of that feature. If we can avoid having to care about transaction isolation levels to coordinate long-running importers, I think we should.

I'm happy to add the alternative approaches to the ADR, but I'm not convinced either is preferable to the "optimistic heartbeat" strategy described in the doc.

Are comments in this PR the appropriate place to discuss the finer points of each approach to come to a decision on how to proceed?

ctron:
> > I'd try to avoid storing a state in the database.
> 
> Why?

Simply because with that, you're creating your fully fledged own distributed work scheduler. With heartbeats, required suicides of instances if they can't update heartbeats, stale data, …

> 
> > One way to deal with this could be to have a hashing implementation, where each instance of the importers has a number assigned, and knows the total number of importers.
> 
> Who/what assigns each instance a number, and how does each instance know the total number of importers?

That can be done through env-vars. E.g. using ansible, or kubernetes. or, for testing, via the command line.

> 
> > ## Alternative B
> > We let the DB deal with locking. Using `SKIP LOCKED`: https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE
> 
> My ignorance of that feature aside, I'd be wary of any pessimistic locking approach for this situation, mostly for dealing with crashed instances.
> 
> Using a database-specific feature means current/future maintainers of the code will need to understand the nuances of that feature. If we can avoid having to care about transaction isolation levels to coordinate long-running importers, I think we should.

Vs a maintainers needing to understand our own, custom code. I agree, that I'd prefer a simpler solution altogether. Like the instance number alternative above.

But my concern would be more that we tie ourselves to a very specific postgres feature.

> 
> I'm happy to add the alternative approaches to the ADR, but I'm not convinced either is preferable to the "optimistic heartbeat" strategy described in the doc.
> 
> Are comments in this PR the appropriate place to discuss the finer points of each approach to come to a decision on how to proceed?

 I think so.

jcrossley3:
> Simply because with that, you're creating your fully fledged own distributed work scheduler. With heartbeats, required suicides of instances if they can't update heartbeats, stale data, …

There is nothing fully-fledged about the "distributed work scheduler" -- if you can even call it that -- described in this ADR. Remember, we're still dividing the work into very big chunks: one instance will still be ingesting all the docs from a particular source.

Now, if we ever think we should be distributing *THAT* load, *THEN* we can talk about a real distributed work scheduler, but I really don't think we're to that point, and probably never will be. There simply aren't enough advisories created in a day to justify that effort.

> That can be done through env-vars. E.g. using ansible, or kubernetes. or, for testing, via the command line.

Sure, but that's more complex to administer. I think the right goal is to be able to "just fire up another importer instance" -- no further thinking about [re]configuration should be required. 

This simplifies the internal logic of the importer, too. The [`reset_all_jobs`](https://github.com/trustification/trustify/blob/45d00c455a24c9146035dd2ceb7a41141209381e/modules/importer/src/server/mod.rs#L140-L169) function goes away. New instances will naturally take over the work of failed jobs per the algorithm described in the doc.

jcrossley3:
Also, recall that the importers are [already sending a hearbeat](https://github.com/trustification/trustify/blob/45d00c455a24c9146035dd2ceb7a41141209381e/modules/importer/src/service.rs#L376-L397) -- we just need to add a timestamp and an instance identifier to give the selection algorithm the data it needs. So the changes shouldn't be all that invasive or complex.

ctron:
I think in that case, it should be fleshed out. Also before merging the other PR. But it feels wrong to merge an implementation while we're still discussion as design.

That doesn't really qualify as a heartbeat. There's no information about processing (just progress), there's no instance id, there no update on it when performance longer, blocking, tasks. …

jcrossley3:
> That doesn't really qualify as a heartbeat. There's no information about processing (just progress), there's no instance id, there no update on it when performance longer, blocking, tasks. …

It's a heartbeat in that it periodically updates the relevant record in the database. It's a logical spot to add the instance id and timestamp, like I mentioned.

I have no idea what "performance longer, blocking, tasks" refers to, so if it's relevant to the task at hand, please clarify.

chirino:
Implementing the ADR will likely fix #1307

chirino:
I think it's fine if the other PR is merged since it's not trying to address multi-replica coordination (and we didn't have any before either). 

As far as how work distribution occurs between the replicas, I tend to like @ctron's approach of using a hash.

jcrossley3:
> As far as how work distribution occurs between the replicas, I tend to like @ctron's approach of using a hash to more finely distribute the work.

Well then I feel dumb, because I trust y'all's opinion, and you must be seeing some flaw with the heartbeat approach that I'm not seeing. And how would a hash "more finely distribute the work"? How would it be any more distributed than with multiple replicas sending heartbeats.

Unless I'm misunderstanding the hashing approach, I see some obvious problems with it:

- I have to configure each replica with a total and an assigned number, right? So adding another replica (or removing one) means I have to then shut down the others and reconfigure them with a new total, correct?
- If one of the replicas crashes, none of its work gets done even though there may be an idle replica available.

The heartbeat approach requires no configuration. What are the downsides you see with it?

chirino:
As long as all replicas can distribute a consistent list of replica members, determining the total number of replicas is straightforward. You can also identify your assigned number by finding where your replica ID appears in the list.

This could be achieved using a membership table with a lease timeout mechanism, similar to what you proposed in the ADR for importers.  If the membership changes (due to a new member row, or member row timeout exporting) then you have to reconfigure which work items your working on. 

The idea behind hash-based work distribution is that it can lead to a more even load distribution if the work keys hash well and you have lots of work keys.  This also gives us more flexibility.  We could even use this to split up the work of a single importer over multiple replicas since they could hash a set of tasks that the importer has to complete.

jcrossley3:
And how/why is that better than what's proposed in the ADR? Why will that be simpler to understand, implement and evolve?

Do we really want to "split up the work of a single importer over multiple replicas"? Does the cost of doing that justify its value?

Only the initial import of our default sources takes hours to complete. After that, each run takes a second or two. Do we really need a fully-fledged work distribution system to handle the daily torrent of new advisories and SBOM's? Correct me if I'm wrong, but I don't think so.

What is the simplest thing we can do that allows us to remove the `reset_all_jobs` function and ensure multi-replica safety?

chirino:
> And how/why is that better than what's proposed in the ADR? Why will that be simpler to understand, implement and evolve?

I agree, it's not going to be simpler. 

From another point of view, the ADR is mixing up workload distribution concerns with the importer.  That could be important if knowing which replica was assigned to the importer is important.  But it seems that's not the case right now.  

> Do we really want to "split up the work of a single importer over multiple replicas"? Does the benefit of doing that justify its cost?
> Only the initial import of our default sources takes hours to complete. After that, each run takes a second or two. Do we really need a fully-fledged work distribution system to handle the daily torrent of new advisories and SBOM's? Correct me if I'm wrong, but I don't think so.
> 

This could help if we have a large source to import from and want the initial import to be sped up by distributing the import work.  I don't know if we will ever have this case.  I don't even know if the DB would bottleneck and make this impossible.  But, a hash distribution strategy would allow us to implement this in the future without changing the data model.  

> What is the simplest thing we can do that allows us to remove the `reset_all_jobs` function and ensure multi-replica safety?

If you want simpler, you just setup a replica leader election (which can again be done with a replica members table) and only have the leader run the importers.

JimFuller-RedHat:
I was hoping this ADR could provide design guidance on importer scalability. Having a broad overview of how we might scale things up could help us know how far to take existing status quo. 

It would be useful to know how far we go in building scalability into the app layer versus providing documented guidance on infrastructure. 

What code decisions can we make today to help support  infrastructure approaches we employ in the future ?
* #n web pods
* #n importer pods
* pgbouncer routing read/write db calls
* pg write replicas
* pg read replicas 

What scenarios are we trying to address ?
* bulk data loading (good for DR/testing)
* runtime
* any more ????

The ADR spends most its time addressing a specific shortcoming of status quo - I think that 'ship has sailed' in terms of us having already built a baby job queue/scheduler inside of trustify ... fixing existing importer processes to avoid redundant work seems more like a bug fix then a strategy for achieving horizontal scalability. 

I would rather we just fix status quo today and focus this ADR on providing guidance of what we should do next - should we continue building out at the app layer ? Should we adopt/use a crate ? Outline the pitfalls, alternatives etc.

docs/adrs/00004-importer-scalability.md
```diff
@@ -0,0 +1,62 @@
...
+  eventually picked up by another.
+* In short, adding an importer process should increase the ingestion
+  throughput, i.e. the system should be scalable.
+
+## Decision
+
+Since trustify already depends on PostgreSQL -- the import definitions
+persist there -- we can persist the "coordination state" there, too.
+
+Two columns can be added to the `IMPORTER` table:
```
JimFuller-RedHat, line 36: Have we considered a separate database for this kind of thing eg. we can have trustify and trustify-importer served up by single pg instance.

docs/adrs/00004-importer-scalability.md
```diff
@@ -0,0 +1,62 @@
...
+  be doing the same work.
+* The solution should be fault-tolerant -- importer instances may come
+  and go -- any work in process by a crashed/killed one must be
+  eventually picked up by another.
+* In short, adding an importer process should increase the ingestion
+  throughput, i.e. the system should be scalable.
+
+## Decision
+
+Since trustify already depends on PostgreSQL -- the import definitions
```
JimFuller-RedHat, line 33: If/when we have a user that wants to use some other database - I do not think we have 'painted ourselves in a corner' eg. there maybe some small work to get things running elsewhere ... to me this is a small theoretical risk.

docs/adrs/00004-importer-scalability.md
```diff
@@ -0,0 +1,62 @@
...
+Draft
+
+## Context
+
+This ADR is concerned with our ability to scale the importer processes
+that ingest documents -- SBOM's and advisories -- from various
+sources.
+
+Currently, the importer process iterates over the enabled importer
+definitions in the database sequentially, running each according to
```
JimFuller-RedHat, line 16: we should identify what risks there are changing from predictable sequential processing to parallel processing eg. increase in RAM/CPU, db conn ... this change should be well tested against data at scale so we understand impact.

ctron:
> > That doesn't really qualify as a heartbeat. There's no information about processing (just progress), there's no instance id, there no update on it when performance longer, blocking, tasks. …
> 
> It's a heartbeat in that it periodically updates the relevant record in the database. It's a logical spot to add the instance id and timestamp, like I mentioned.

However, there's not guarantee from the code, that update happens with a predictable period. For me, this would be a requirement for a heartbeat.

> 
> I have no idea what "performance longer, blocking, tasks" refers to, so if it's relevant to the task at hand, please clarify.

e.g. during the git clone, there will be no update.

In short: the status update was never designed to function as a heartbeat.

jcrossley3:
> e.g. during the git clone, there will be no update.

Ah, good point. Yes, any heartbeat would need to be independent of any steps in the importer that might take longer than the heartbeat interval.

docs/adrs/00004-importer-scalability.md
jcrossley3: The importer's only reason for existing is to populate trustify's database. It seems very odd to consider giving it its own just to coordinate replicas.

docs/adrs/00004-importer-scalability.md
jcrossley3: FYI, we do use a few pg-specific features in the query stuff, e.g. ILIKE, so something to stay mindful of.

docs/adrs/00004-importer-scalability.md
JimFuller-RedHat: ya ...anything we can do to promote  loose coupling is a good thing - for me an importer app (that only pushes to a single app) is still a different thing but not suggesting we should fully pull out, just was thinking how we might encourage separation of concerns (in code organisation and database). I would put these kind of musings in Alternate section.

jcrossley3:
> The ADR spends most its time addressing a specific shortcoming of status quo - I think that 'ship has sailed' in terms of us having already built a baby job queue/scheduler inside of trustify ... fixing existing importer processes to avoid redundant work seems more like a bug fix (we should fix) then an overarching strategy for achieving horizontal scalability.
> 
> I would rather we fix status quo today and focus this ADR on providing guidance of what we should do next - should we continue building out at the app layer ? Should we adopt/use a crate ? Outline the pitfalls, alternatives etc. Note just having a discussion about this ADR is proving very useful - getting team alignment on this will serve as a good template for any other scalability issues elsewhere.

Agree. I'm going to fix the bug first.

chirino:
@jcrossley3 this is what I was thinking off to do the replica membership stuff: https://github.com/trustification/trustify/commit/5ac76a3c76d167897026d435b8cc94d55dc777dd

The ReplicaManager manager deal with using DB the know who are the live members of the replica set.  The current membership is delivered to the importer Server via an event that looks like.

```rust
pub struct MembershipEvent {
    pub replica_id: Uuid,
    pub members: Vec<Uuid>,
}
```

The members Vec is sorted by age of the member.  So oldest member is the first.  This is then done to do leader election choosing the oldest member to be the leader.  With not much more work we can use a hash key strategy to divide the work up over the replica members.

jcrossley3:
> @jcrossley3 this is what I was thinking off to do the replica membership stuff: [3c4c6f2](https://github.com/trustification/trustify/commit/3c4c6f2e6cdef164f490fcb57aa402b71ee36958)

That is very impressive work, but I'm inclined to think it's overkill until we figure out whether we can (or even should) break up the work of each importer to justify the development of any sort of "replica manager".

It's not yet clear to me how to do that, and I think it's going to vary per importer. For example, will git-based importers clone a repo and then create jobs for each advisory in their clone? Will they send the entire raw advisory in that job? Or just a link? If the latter, do we care that we fetched that doc twice: once when we cloned and then again when requesting the link?

I just want a simple way to allow multiple importers to safely and efficiently run simultaneously. I think I've achieved that with #1321. So I'd like to hold off on introducing a new table and new service until we recognize a need for them.

jcrossley3:
> > Will they send the entire raw advisory in that job? Or just a link? If the latter, do we care that we fetched that doc twice: once when we cloned and then again when requesting the link?
> 
> Not sure I understand who the job would be 'sent' sent to. Every replica will have a copy of the job and they don't connect to each other to coordinate work. They just know which slice of the pie is theirs to work based on the replica membership set.

What's in the "pie slice"? How much info are you storing in each replica's copy of the job in order for them to do their work? Are we storing the raw doc in the database? (I hope not) A link to it? What's the impact on our network/storage resources for fetching the doc twice (once in the clone, once by the replica ingesting it) as opposed to how we currently do it (just once, compressed as a git object).

I'm sure I'm not considering all the possible questions we'll have to answer for each type of importer in order to slice those pies.

chirino:
I updated my branch to actually use key hashing to distribute the work.  

See:
https://github.com/trustification/trustify/compare/main...chirino:trustify:replica-manager#diff-d1480003f091763ba94c686026e676909d780535eb4f4ab6be4f9e8b5adca1e3R36-R42

That's core function the used to figure out if a replica should be running an importer job.  It calculates it by knowing which is his member index in a replica membership list.

> What's in the "pie slice"? 

Another way to think about it.  The Pie being the replica membership list.  If there 2 members, then you should get about 1/2 the pie assigned to you.

> How much info are you storing in each replica's copy of the job in order for them to do their work? 

None.  The job / importer records are NOT updated to assign work.

> Are we storing the raw doc in the database? (I hope not) A link to it? What's the impact on our network/storage resources for fetching the doc twice (once in the clone, once by the replica ingesting it) as opposed to how we currently do it (just once, compressed as a git object).
> 
> I'm sure I'm not considering all the possible questions we'll have to answer for each type of importer in order to slice those pies.

jcrossley3:
> I updated my branch to actually use key hashing to distribute the work.

So it effectively does what #1321 does but with more complexity and more database calls. I like that it guarantees no 2 runners will take the same job, but I wish it did it in less code. :smile: 

> Another way to think about it. The Pie being the replica membership list. If there 2 members, then you should get about 1/2 the pie assigned to you.

That's the problem. Our importers are not created equal. Each does not represent the same amount of work. Whichever replica gets that CSAF job is going to be busy while the rest are idle. Ideally, those idle replicas could be made to help the poor guy. 

For the load we expect to receive, I'm still thinking a single importer process running multiple concurrent jobs is sufficient, so I'd like to take a conservative approach toward implementing bespoke distributed job schedulers. Just my opinion, obviously.

chirino:
> > I like that it guarantees no 2 runners will take the same job, but I wish it did it in less code. 😄

Yeah.  I think I could simplify some of it.  But no point on doing that if the general workload distribution approach is not desired.

> To be honest, I'm not sure it _does_ make that guarantee. I'm giving you the benefit of the doubt. Is it conceivable that some failed heartbeat could result in 2 replicas running the same job?

Yeah, but the same problem applies to both of our implementations.   if a heartbeat fails (say network error), you should stop running any assigned jobs ASAP as you may no longer be the leader for the job.  But this would require passing a signal or atomic bool to jobs to let them know they should stop running.   Right now things are considered dead when they miss 3 heartbeats, but this could be made longer to give more safety margin/time to cancel work.

ctron:
What speaks against just using a hashing method with an externally provided "max" and "instance" number? That sounds way simpler and less error prone.

jcrossley3:
> What speaks against just using a hashing method with an externally provided "max" and "instance" number? That sounds way simpler and less error prone.

Configuration complexity and fault intolerance, as described [above](https://github.com/trustification/trustify/pull/1306#issuecomment-2664140715).

jcrossley3:
I'm trying to avoid taking on the additional complexity you're suggesting.

You and I are both justifiably nervous about relying on specific PostgreSQL features. Yet we apparently disagree about relying on k8s features. It complicates local testing if a developer is required to run k8s, rhel and ansible (or configure the complex CLI options they require) in order to test the app's behavior.

Further, I think it's too simplistic to bounce all instances in order to add/delete one at runtime. Currently, canceled jobs don't pick up where they left off: they start over. So, for example, if one instance is 5 hours into running that CSAF importer job, I'd like to be able to just fire up another instance and have it pick up the other waiting jobs. But in order to give every instance a new "max" number, I'll then lose that 5 hours of work.

Yes, #1321 does introduce some complexity into the system, but IMHO it's justified by the benefit it brings. Why specifically don't you?

ctron:
> You and I are both justifiably nervous about relying on specific PostgreSQL features. Yet we apparently disagree about relying on k8s features. It complicates local testing if a developer is required to run k8s, rhel and ansible (or configure the complex CLI options they require) in order to test the app's behavior.

I am nervous about not having those features. I'd trust them, if available. I'd also feel comfortable using k8s feature, compared to re-implementing this ourselves. And even for local testing, just having a bunch of env-vars feels way simpler to me, compared to an automatic system, which does some magic, and has a lot of gaps (like work that keeps running although it shouldn't). 

> Further, I think it's too simplistic to bounce all instances in order to add/delete one at runtime. Currently, canceled jobs don't pick up where they left off: they start over. So, for example, if one instance is 5 hours into running that CSAF importer job, I'd like to be able to just fire up another instance and have it pick up the other waiting jobs. But in order to give every instance a new "max" number, I'll then lose that 5 hours of work.

Scaling up an RDS instance does just that. I really fail to see why we try to create something rather complex, buggy, hard to test … just to be better than the rest, only in some very rare occasions.

And I think we could properly define a distributed work scheduler, which has all the issues fixed. But I also think that we're putting the focus on a thing that we don't need. Or end up with something half-backed, that doesn't live up to its promises.

jcrossley3:
Closing until our requirements come into focus.

PR #1309, opened 2025-02-18
feat: add RH prod/comp relationships

ctron:


modules/ingestor/src/graph/sbom/common/license.rs
```diff
@@ -90,7 +90,7 @@ impl LicenseCreator {
         });
     }
 
-    #[instrument(skip_all, fields(num = self.licenses.len()), err)]
+    #[instrument(skip_all, fields(num = self.licenses.len()), err(level=tracing::Level::INFO))]
```
JimFuller-RedHat, line 93: just curious how one uses it eg. when RUST_LOG=trace ?

modules/ingestor/src/graph/sbom/mod.rs
```diff
@@ -1,11 +1,12 @@
 //! Support for SBOMs.
 
+pub mod clearly_defined;
 pub mod cyclonedx;
+pub mod processor;
 pub mod spdx;
 
-pub mod clearly_defined;
```
JimFuller-RedHat, line 6: why does this get removed ?

modules/ingestor/src/graph/sbom/processor/rh_prod_comp.rs
```diff
@@ -0,0 +1,213 @@
...
+    package: &sbom_package::ActiveModel,
+    relationships: &[package_relates_to_package::ActiveModel],
+) -> Option<String> {
+    let Set(node_id) = &package.node_id else {
+        return None;
+    };
+
+    // is top level: described by the SBOM
+
+    fn is_top_level(
```
JimFuller-RedHat, line 155: alternately 'is_root' ?

modules/ingestor/src/graph/sbom/mod.rs
ctron: Ups.

modules/ingestor/src/graph/sbom/processor/rh_prod_comp.rs
```diff
@@ -0,0 +1,213 @@
...
+        &mut self,
+        InitContext {
+            suppliers,
+            document_node_id,
+        }: InitContext,
+    ) {
+        self.document_node_id = document_node_id.to_string();
+
+        for supplier in suppliers {
+            if SUPPLIERS.contains(supplier) {
```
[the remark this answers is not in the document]
ctron: Oh, this `contains` is on the list. Not strings. Meaning: does this list contain this exact string.

modules/ingestor/src/graph/sbom/processor/rh_prod_comp.rs
ctron: Will rename.

modules/ingestor/src/graph/sbom/processor/rh_prod_comp.rs
```diff
@@ -0,0 +1,213 @@
...
+    document_node_id: String,
+}
+
+impl RedHatProductComponentRelationships {
+    pub fn new() -> Self {
+        Default::default()
+    }
+}
+
+const SUPPLIERS: &[&str] = &["Red Hat", "Organization: Red Hat"];
```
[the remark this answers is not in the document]
ctron: Yea, I considered making this configurable. Java style: `const SUPPLIERS = System.getProperty("org.trustification.sbom.redhat.processor.suppliers");` … but :shrug:

ctron:
I think for tests we need some query functionality first. We can test that it ingests. And we can also see that non-relevant files still ingest (by using the other tests).

If the data is correct, we can only assert once we got query functionality. And I am ok postponing this PR until that's the case.

JimFuller-RedHat:
> If the data is correct, we can only assert once we got query functionality. And I am ok postponing this PR until that's the case.

chicken<->egg ... and query/resolving would like this into main to test querying ;) ... I think it is fine for this PR to go in ... as we are still pre 2.0.0  it is ok to increment away on main (as long as its safe).

Issue #1307, opened 2025-02-17
Make importers safe to run on multiple replicas.


chirino:
Currently there is no coordination between kube replicas of the trustd service.  This will likely lead to some weird behaviours and/or increased API rate limit errors.

jcrossley3:
To be clear, it's only the `importer` subtask that needs coordination. Multiple replicas of the `api` subtask should be fine.

PR #1321, opened 2025-02-19
feat: enable multiple importer replicas to coordinate their work

jcrossley3:
Fixes: #1307

Importers now use an optimistic locking strategy to self-assign jobs from the IMPORTER table and then regularly update the row with a timestamp (unix time in nanoseconds), keeping other replicas from taking the job as long as the timestamp is within the heartbeat rate of 10 seconds.

This obviates the `reset_all_jobs` function as now stale jobs will eventually be picked up by one of any active importer processes.


modules/importer/src/model/mod.rs
```diff
@@ -239,11 +242,13 @@ impl TryFrom<Model> for Importer {
             progress_total,
             progress_message,
             continuation,
-            revision: _,
+            heartbeat,
+            ..
```
ctron, line 246: The idea of this pattern is to be aware of fields which get added. By a compiler error. So I'd want to keep it.

modules/importer/src/server/mod.rs
```diff
@@ -151,6 +118,11 @@ async fn import(
     importer: Importer,
     service: ImporterService,
 ) -> Result<(), Error> {
+    let locked = Heart::beat(&importer, &runner.db)
+        .await
+        .map_err(|_| Error::MidAirCollision)?;
```
ctron, line 123: Wouldn't that lead to a failed importer run? Reported in the UI?

modules/importer/src/runner/common/heartbeat.rs
```diff
@@ -0,0 +1,66 @@
...
+            let mut interval = interval(Heart::RATE);
+            let mut importer = importer;
+            loop {
+                interval.tick().await;
+                match Self::beat(&importer, &db).await {
+                    Ok(i) => {
+                        log::debug!("{}: {:#?}", i.name, i.data.progress);
+                        importer = i;
+                    }
+                    Err(e) => log::error!("Failed to send heartbeat for '{}': {e}", importer.name),
```
ctron, line 32: What will happen if the heartbeat won't get updated? Wouldn't another instance pick up that task, while this one is still running?

modules/importer/src/runner/common/heartbeat.rs
```diff
@@ -0,0 +1,66 @@
...
+            name: Set(importer.name.to_owned()),
+            heartbeat: Set(Some(t.into())),
+            ..Default::default()
+        };
+        use importer::Column::Heartbeat;
+        let lock = match importer.heartbeat {
+            Some(t) => Expr::col(Heartbeat).eq(Decimal::from_i128_with_scale(t, 0)),
+            None => Expr::col(Heartbeat).is_null(),
+        };
+        match importer::Entity::update(model).filter(lock).exec(db).await {
```
ctron, line 54: I think the expectation is that this function returns a "mid air collision" in the case of a broken optimistic lock. But I can't seem to find the logic for that here.

modules/importer/src/server/mod.rs
```diff
@@ -91,58 +90,26 @@ impl Server {
...
-                        None,
-                        None,
-                    )
-                    .await?;
-            }
+                    .list()
+                    .await?
+                    .into_iter()
+                    .filter(|i| {
+                        !(i.data.configuration.disabled || already_running(i) || can_wait(i))
```
ctron, line 102: I think there's a race condition here. If multiple instances retrieve that at the same time, they would also start those jobs, at the same time. Leading to multiple instances of the same job running.

ctron (dismissed):
I think there a few bugs in there.

modules/importer/src/server/mod.rs
ctron: That means that **any** DB gets turned into a "mid air collision"?

modules/importer/src/runner/common/heartbeat.rs
jcrossley3: That error will occur for one of two reasons:

1. Two instances just happened to call `Heart::beat` in `server::import` at the exact same nanosecond for the same job.
2. There is an unusual problem with the database, e.g. connectivity, malicious data changes, etc

I thought the risk of the first was so low that beers are on me if it ever happens. But if you don't like beer, we can add another column in IMPORTER to uniquely identify the runner and add that id to the optimistic where clause.

The risk of the second could result in multiple runners attempting to run the same job, but they're probably suffering from the same database problem and the logs are filling up with error messages anyway.

Assuming the worst -- multiple instances running the same job simultaneously -- what's the fallout? Don't we check hashes on docs to know whether we've ingested them before? The progress numbers may look janky in the  UI, but there will be errors reported to explain that. Do we need more than that to account for what should be an extremely unlikely scenario?

modules/importer/src/runner/common/heartbeat.rs
jcrossley3: This function, `Heart::beat`, is called from two places. Initially, from the `server::import` function with the sole purpose of obtaining the lock. If it can't, it simply stops processing, and no heartbeat task is created.

Only after it acquires the lock is `Heart::new` called, which then creates the loop to continually call `Heart::beat`.

The "mid air collision" is never reported to the user. It's only used as a coordination mechanism.

modules/importer/src/server/mod.rs
jcrossley3: There is absolutely a race condition here. It's accounted for in the very first line of the `import` function. It's entirely possible that a task spawned from this loop may end immediately, before it processes any data.

modules/importer/src/runner/common/heartbeat.rs
ctron: > Assuming the worst -- multiple instances running the same job simultaneously -- what's the fallout? Don't we check hashes on docs to know whether we've ingested them before?

The fallout would be two importers updating the same importer instance/report. Today. Tomorrow, we don't know. And I'd say, today don't fully know either.

To me, that's a bug.

modules/importer/src/runner/common/heartbeat.rs
ctron: What happens if two instances call that function at roughly the same time? What happens, if one of them does not update that row? What happens if both update the row?

modules/importer/src/server/mod.rs
ctron: What would happen if no row was updated? Would that return an error too?

modules/importer/src/runner/common/heartbeat.rs
jcrossley3: Everything's a potential bug in distributed computing! :wink:

modules/importer/src/runner/common/heartbeat.rs
jcrossley3: There is no "roughly". They would have to call it at the exact same time, in nanoseconds. That's the only way both could update the row. And if that happens, they would each start running the same job. But I don't think it's going to. :smile:

modules/importer/src/server/mod.rs
jcrossley3: Yes. I confirmed that. I was happy to discover that calling `::Entity::update` returns an error if no row is updated, unlike `update_many`.

modules/importer/src/runner/common/heartbeat.rs
jcrossley3: But if you think it might, we can easily add another column containing a unique id for the runner.

modules/importer/src/runner/common/heartbeat.rs
jcrossley3: Thinking more about it, because both the heartbeat timestamp and the progress numbers are stored in the same row, whatever's preventing one from updating affects the other, too. So the main work thread will fail, causing the heartbeat to be dropped out of scope.

modules/importer/src/runner/common/heartbeat.rs
ctron: What would be the outcome if no row would be updated by that update call?

modules/importer/src/server/mod.rs
ctron: Please ignore that question above then. I think it would be good to at least document that in our code. As to me, it's not obvious.

modules/importer/src/runner/common/heartbeat.rs
jcrossley3: If the job has already started, and the heartbeat has been running for a while, and then encounters some error that prevents the update, that would then open the window for other instances to take on that job. Those instances would assume the runner previously executing the job had crashed without that heartbeat update.

modules/importer/src/runner/common/heartbeat.rs
ctron: What about the clone git repository? Wouldn't that potentially case two instances processing that? Running in the same instance?

chirino: Could we make the 2nd concurrent import job fail optimistically?

modules/importer/src/runner/common/heartbeat.rs
chirino: Then fields being added are starting to look a lot like the ones needed to track replica membership.  See:

https://github.com/trustification/trustify/compare/main...chirino:trustify:replica-manager

modules/importer/src/runner/common/heartbeat.rs
jcrossley3: > What about the clone git repository? Wouldn't that potentially case two instances processing that? Running in the same instance?

The lock is acquired before the clone starts, so the heart pumps asynchronously to the clone, and that lock prevents other runners from taking that job, whether in-process or out.

Or am I misunderstanding your question?

jcrossley3: > Could we make the 2nd concurrent import job fail optimistically?

It should already, because `Heart::beat` will return an error at the top of that `server::import` function.

modules/importer/src/runner/common/heartbeat.rs
jcrossley3: I think the logic is simpler in this PR than what's in the replica manager, but they both share the same challenges of distributed computing, so not surprising they need similar fields.

jcrossley3:
@ctron I just pushed the only change you requested and hopefully answered your questions sufficiently. If not, I may have lost the thread, so please ask again.

My main goal with this PR was to make it safe to run more than one importer instance. That we distribute the load a bit is a bonus, but given our expected low frequency of new docs to ingest, it's mainly for redundancy and user convenience.

modules/importer/src/runner/common/heartbeat.rs
chirino: Ideally Heart provides a signal and/or atomic bool to indicate if it has failed.  That way an importer can wait on/monitor those to see if it should exit early.

jcrossley3: I'm thinking along the same lines. Refactoring things now to have `Heart::new` accept a `Future` that it can then `spawn_local` in addition to spawning its own heartbeat loop. Essentially, moving the `spawn_local` call from `run_local` to `Heart::new`. This will allow `Heart` to abort the importer (the Future passed to it) the first time its heartbeat fails, for any reason.

chirino (approved):
The strategy of using task.abort() should work BUT I think it may surprise some folks if your import task has some cleanup work the does not get executed due to the task aborting.

ctron:
> The strategy of using task.abort() should work BUT I think it may surprise some folks if your import task has some cleanup work the does not get executed due to the task aborting.

How does "abort" behave if the task is running on one of the tokio thread workers? My assumption is, it keeps going. Leading to a situation where this would again run twice. Which might be a problem for things like "git clone".

jcrossley3:
> How does "abort" behave if the task is running on one of the tokio thread workers? My assumption is, it keeps going. Leading to a situation where this would again run twice. Which might be a problem for things like "git clone".

Correct. But `abort` will only ever be called if the heartbeat can't be updated. So the risk is negligible, and I expect whatever issue is breaking the heartbeat will break the associated task, too.

ctron:
> Correct. But `abort` will only ever be called if the heartbeat can't be updated. So the risk is negligible, and I expect whatever issue is breaking the heartbeat will break the associated task, too.

How would a broken DB connection break the git clone?

chirino:
So if we passed something like a CancelContext to the git GitWalker then I think we could abort the git operations early using having the git callbacks check the CancelContext.

jcrossley3:
> > Correct. But `abort` will only ever be called if the heartbeat can't be updated. So the risk is negligible, and I expect whatever issue is breaking the heartbeat will break the associated task, too.
> 
> How would a broken DB connection break the git clone?

The git clone is a problem no matter what we do because it's not async. It cannot be canceled in any way other than brute force.

jcrossley3:
> So if we passed something like a CancelContext to the GitWalker then I think we could abort the git operations early using having the git callbacks check the CancelContext.

I looked into that a bit. I would think we might be able to `tokio::select!` between some cancel token and that blocking thread, but the `RunContext` that tracks cancellation isn't passed to the GitWalker. 

If we go that route, I'd like to refactor the cancellation logic to use a [CancellationToken](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) instead.

But since that git `run_sync` fn is a problem no matter what we do, I consider that effort orthogonal to the purpose of this PR: allow 2 importer replicas to coordinate their work.

jcrossley3:
@ctron you're not going to approve this PR because you don't like it, correct? 

You would prefer we pass clap options for `max` and `id` instead? What should the app do if those options are not passed? Exit? Or attempt to coordinate anyway? Maybe a compromise is to do both?

I dunno. I feel like my logic is sound -- at least as sound as what we currently have. You think it's buggy. Fine. But either request specific changes or remove your hold. I'd like to get this off my plate.

jcrossley3:
> After putting some debug statements in what I thought were the relevant fields in that `RemoteCallbacks` struct (`progress` and `sideband_progress`), I'm not optimistic about that. For the bulk of the time it took to clone, I got nothing.
> 
> Hopefully, I was doing it wrong.

Huh! I _was_ doing it wrong! Gonna try and refactor both `Heart` and `CheckCancellation` to use a `CancellationToken` that I can query in the git callbacks.

jcrossley3:
@ctron @chirino I rebased onto main and added a commit that employs a `CancellationToken` from both the `Heart` and the `CheckCancellation` tasks. I added an `is_canceled` method to the git walker's `Handler` trait that is now called by the git progress callback.

Seems to work well in my local testing, with the added benefit of exiting within a few seconds of sending a Ctrl+C signal instead of hanging indefinitely.

ctron:
I personally think this is too much complexity that we adopt for too little benefit. Bringing in too much new issues.

I'm not blocking it either, because if everyone agree that this is the best approach, then that's ok.

I think there are some valuable addition in here, like the cancellation point in the git clone. Which I'd love to see in any case.

jcrossley3:
> I personally think this is too much complexity that we adopt for too little benefit. 

FWIW, it's more complexity than I wanted to introduce, because the nature of the work is so IO-bound that I think a single instance running with `--concurrency=6` is probably sufficient for most purposes. Though that might reveal the db as the bottleneck anyway.

But I disagree that the benefit isn't justified:

- Negligible risk of misconfiguration resulting in duplicate or no importers being run.
- Fault tolerance -- multiple replicas automatically coordinated
- Stateless -- all the state is in the DB, so that `StatefulSet` in our helm chart can be replaced with a simple `Deployment` now.
- The cancellation point you mentioned.

> Bringing in too much new issues.

We'll see. Thanks for your feedback.

jcrossley3:
> After putting some debug statements in what I thought were the relevant fields in that `RemoteCallbacks` struct (`progress` and `sideband_progress`), I'm not optimistic about that. For the bulk of the time it took to clone, I got nothing.
> 
> Hopefully, I was doing it wrong.

BTW, I figured out why I thought I was doing it wrong. I wasn't. The frequency of callbacks invoked is determined by the contents of the repo you're cloning.

Recall that our logic is such that we use a depth of 1 when an importer clones the repo the very first time. This takes about a minute for both `cve` and `osv-github` importers. During that time, you can see the `CheckCancellation` log messages periodically, and if you disable either importer, the run will canceled within a few seconds just as we'd expect.

But the next time those clones are updated, we remove our depth constraint -- we have to in order to see which files have changed since our last run. These fetches take over 30 minutes for both `cve` and `osv` -- both repos are enormous.

Interestingly, during that time, you will still see the `CheckCancellation` log messages for the `cve` importer. And therefore you can cancel out of it fine.

But that's not the case for the `osv` clone. For that one, you won't see the `CheckCancellation` log messages because those  git progress callbacks aren't being called. I have no idea why, but I suspect the very first time I tried putting those debug logs in the callbacks to test out y'all suggestion of checking the cancel there, I just happened to be testing against an `osv` clone.

PR #1325, opened 2025-02-20
feat: introduct e2e tests based on hurl

dejanb:
This is a first attempt to create end to end tests based on [hurl](https://hurl.dev/) (thanks @helio-frota). 

I really like it so far. It was very easy to create a test that uploads ds3 dataset and do some basic responses for the sbom vulnerability matching.

It can also be very nicely integrated into our CI workflow.

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -1,0 +1,7 @@
+## Install
+
+cargo install hurl
+
+## Run tests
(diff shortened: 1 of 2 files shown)


helio-frota (approved):
:rocket:

PR #1327, opened 2025-02-20
chore: adds a new criterion bench

helio-frota:
This is more focused on our tests and to help to investigate something with [jsn sync](https://github.com/trustification/trustify/blob/a090bf872a4e6670f74bdd65d2ac4f5e04a8b185/modules/ingestor/src/service/format.rs#L248) function.

Sending this PR now to share the code first.

`cargo bench --bench bench-format`


chirino (approved):
lgtm

Issue #1316, opened 2025-02-19
Experiment - try to instrument and observe criterion benchmark


helio-frota:
Context:

![Image](https://github.com/user-attachments/assets/4e399f6e-80dd-409b-b151-eba5e8647088)

Now that we have OTEL it would be good to try the benchmark again

* I have no idea if that is possible

helio-frota:
trying to workaround this:

```
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2025-02-19T12:43:45.515260Z  INFO opentelemetry_sdk:  name="BatchSpanProcessor.ThreadStopped"
error: bench failed, to rerun pass `-p trustify-module-fundamental --bench bench`
```

helio-frota:
progress, but for some reason, this is only generating traces for the database setup, ATM

![Image](https://github.com/user-attachments/assets/b5463136-9c16-42b5-a41d-af19f3961527)

helio-frota:
it is working now  : )

![Image](https://github.com/user-attachments/assets/ea5fa34e-93ff-4bdd-bbca-c60df5253569)

PR #1329, opened 2025-02-20
chore: csaf ingestion bech with otel

helio-frota:
This the same code from [bench.rs](https://github.com/trustification/trustify/blob/main/modules/fundamental/benches/bench.rs) + otel bits

[The context](https://github.com/trustification/trustify/issues/1316#issue-2863065044) 

how to run:
`podman compose -f etc/telemetry/compose.yaml up`
`cargo bench --bench bench-otel`


modules/fundamental/Cargo.toml
```diff
@@ -78,6 +78,14 @@ urlencoding = { workspace = true }
 walkdir = { workspace = true }
 zip = { workspace = true }
 
+opentelemetry = { workspace = true }
+tracing-opentelemetry = { workspace = true }
+opentelemetry_sdk = { workspace = true }
+opentelemetry-otlp = { workspace = true, features = ["grpc-tonic"] }
+tracing = { workspace = true, features = ["std"]}
+tracing-core = "0.1" # We are not using this in other OTEL-related code, don't need to add in root Cargo.toml for now.
```
ctron, line 86: Everything should go to the workspace.

modules/fundamental/benches/bench-otel.rs
```diff
@@ -0,0 +1,239 @@
...
+            "base_purl",
+            "versioned_purl",
+            "qualified_purl",
+            "cvss3",
+            "cpe",
+            "version_range",
+            "vulnerability",
+        ] {
+            ctx.db
+                .execute_unprepared(format!("DELETE FROM {table} WHERE 1=1").as_str())
```
ctron, line 218: I think you should be able to just drop the `WHERE 1=1`.

ctron (changes requested):
Minor nit.

Might make sense to put common code to a more common place. But not a blocker.

modules/fundamental/Cargo.toml
helio-frota: sure thing thanks for the review I'll also change the original criterion bench :+1:

modules/fundamental/benches/bench-otel.rs
helio-frota: :+1:

helio-frota:
@ctron PR updated yesteday :+1: 

I'll open an issue for the duplication part including some comments related to the `bench-format.rs`

PR #1331, opened 2025-02-21
TC-2176 Enable Downloading of licenses from single SBOM

bxf12315:


entity/src/extracted_licensing_infos.rs
```diff
@@ -0,0 +1,32 @@
+use sea_orm::entity::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
+#[sea_orm(table_name = "extracted_licensing_infos")]
```
chirino, line 4: such a long name.. how about naming it something shorter like `license_info`

entity/src/sbom_package_license.rs
```diff
@@ -0,0 +1,57 @@
+use sea_orm::entity::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
+#[sea_orm(table_name = "sbom_package_license")]
+pub struct Model {
```
chirino, line 5: The one thing that worries me about this PR is that sbom_package_license seems to be very similar to purl_license_assertion / cpe_license_assertion do we need all 3?

entity/src/sbom_package_license.rs
```diff
@@ -0,0 +1,57 @@
...
+    License,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, EnumIter, DeriveActiveEnum)]
+#[sea_orm(rs_type = "String", db_type = "Enum", enum_name = "license_category")]
+pub enum LicenseCategory {
+    #[sea_orm(string_value = "slc")]
+    SpdxDeclared,
+    #[sea_orm(string_value = "sld")]
+    SpdxConcluded,
```
chirino, line 34: Are these annotations correct?  sld would make more sense to mean declared to me.

chirino (commented):
Looks solid, wonder if normalizing the license table is worth it. Like we could

migration/src/m0000572_sbom_package_license.rs
```diff
@@ -0,0 +1,63 @@
...
+                        LicenseCategory::LicenseCategory,
+                        [
+                            LicenseCategory::SLC,
+                            LicenseCategory::SLD,
+                            LicenseCategory::O,
+                        ],
+                    ))
+                    .to_owned(),
+            )
+            .await?;
```
chirino, line 43: Since these records are owned by sboms, there should be a cascade delete added to delete these records when the sbom is deleted.

migration/src/m0000035_create_extracted_licensing_infos.rs
```diff
@@ -0,0 +1,63 @@
...
+                    .col(
+                        ColumnDef::new(ExtractedLicensingInfos::SbomId)
+                            .uuid()
+                            .not_null(),
+                    )
+                    .col(ColumnDef::new(ExtractedLicensingInfos::ExtractedText).string())
+                    .col(ColumnDef::new(ExtractedLicensingInfos::Comment).string())
+                    .to_owned(),
+            )
+            .await?;
```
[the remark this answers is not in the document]
chirino: see https://github.com/bxf12315/trustify/pull/1

chirino:
I saw that in `ingest_purl_license_assertion` it seems like we ingest package and create purl_license_assertion ... This seems to be called from the clearly defined ingestion.   Should this also be adding a sbom_package_license?

https://github.com/trustification/trustify/blob/main/modules/ingestor/src/graph/sbom/mod.rs#L420

entity/src/sbom_package_license.rs
ctron: I don't think we need the purl and cpe variant. The license is tied to the package and the purl/cpes are simply additional names for those packages.

entity/src/sbom_package_license.rs
```diff
@@ -0,0 +1,57 @@
...
+    )]
+    Sbom,
+    #[sea_orm(has_many = "super::sbom_package::Entity")]
+    Package,
+    #[sea_orm(has_one = "super::license::Entity")]
+    License,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, EnumIter, DeriveActiveEnum)]
+#[sea_orm(rs_type = "String", db_type = "Enum", enum_name = "license_category")]
```
ctron, line 29: We've been using integers so far for enums. Why are we using strings now?

modules/ingestor/src/graph/sbom/common/package.rs
```diff
@@ -109,9 +115,44 @@ impl PackageCreator {
...
+
+        if let Some(cyclonedx) = cyclonedx_licenses {
+            for (uuid, _v) in cyclonedx.licenses {
+                self.sbom_packge_licenses
+                    .push(sbom_package_license::ActiveModel {
+                        id: Default::default(),
+                        sbom_id: Set(self.sbom_id),
+                        node_id: Set(node_id.clone()),
+                        license_id: Set(uuid),
+                        license_type: Set(sbom_package_license::LicenseCategory::Other),
```
ctron, line 152: Why does CDX become "other". To my understanding CDX has declared and concluded as well. Which also means that it doesn't make sense to call the other variants "Spdx*"

bxf12315: The acknowledgemen[1] have types declared and concluded , but I have never seen any CycloneDX having this property. Therefore, we do not know which type of acknowledgement the CycloneDX license belongs to.

[1] https://cyclonedx.org/docs/1.6/json/#components_items_licenses_oneOf_i0_items_license_acknowledgement

entity/src/sbom_package_license.rs
```diff
@@ -0,0 +1,57 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
+#[sea_orm(table_name = "sbom_package_license")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: Uuid,
+    pub sbom_id: Uuid,
+    pub node_id: String,
+    pub license_id: Uuid,
+    pub license_type: LicenseCategory,
```
mrizzi, line 11: Only SPDX implicitly always provides the license category (i.e. though `licenseDeclared` and `licenseConcluded` fields) but CycloneDx introduced `acknowledgement` on in v1.6 as an optional field so I feel like this should change with:
```suggestion
    pub license_type: Option<LicenseCategory>,
```
@ctron I would like to hear your feedback on this proposed change.

entity/src/sbom_package_license.rs
mrizzi: @ctron so are you suggesting about removing the purl and cpe variants, i.e. removing the `purl_license_assertion` and `cpe_license_assertion` table and just keep the `sbom_package_license` one?

migration/src/m0000035_create_extracted_licensing_infos.rs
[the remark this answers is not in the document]
mrizzi: @bxf12315 please merge the referenced PR from Hiram.

migration/src/m0000572_sbom_package_license.rs
mrizzi: @bxf12315 please merge the referenced PR from Hiram.

entity/src/sbom_package_license.rs
mrizzi: @bxf12315 beside being switched as Hiram reported, @ctron also clarified having `Spdx` is useless (ref. https://github.com/trustification/trustify/pull/1331/files#r1967240377) so please rename them in `Declared` and `Concluded` so that they'll be used for both SPDX and CycloneDX SBOMs.

entity/src/extracted_licensing_infos.rs
mrizzi: @bxf12315 it's a fair comment, please rename it.

mrizzi (commented):
Some further comments in keep with what we discussed today.

JimFuller-RedHat:
needs scale testing before it goes into main

modules/ingestor/src/graph/sbom/common/package.rs
```diff
@@ -206,6 +247,23 @@ impl PackageCreator {
                 .await?;
         }
 
+        for batch in &self.sbom_packge_licenses.into_iter().chunked() {
+            sbom_package_license::Entity::insert_many(batch)
+                // .on_conflict(
```
JimFuller-RedHat, line 252: why is this commented out ?

bxf12315: Done

entity/src/sbom_package_license.rs
bxf12315: I agree with Jens' suggestion, but I do have a question regarding the process of querying a package's license. If we were to remove those two entities, users would need to first query all SBOMs containing the package, then retrieve the node ID, and finally query the license. On the other hand, if we keep these entities, users can directly query the license of a package, although this would result in multiple records across different SBOMs. Given that a package might have different license records in various SBOMs, removing these entities seems like a reasonable approach.

mrizzi: https://github.com/trustification/trustify/discussions/1396

modules/ingestor/src/graph/sbom/common/package.rs
mrizzi: This relates to https://github.com/trustification/trustify/pull/1331#discussion_r1969547113

modules/fundamental/src/license/endpoints/mod.rs
```diff
@@ -24,70 +21,114 @@ pub fn configure(config: &mut utoipa_actix_web::service_config::ServiceConfig, d
...
     ),
 )]
-#[get("/v2/license/{uuid}/purl")]
-/// Retrieve pURLs covered by a license
-pub async fn get_license_purls(
-    state: web::Data<LicenseService>,
-    uuid: web::Path<String>,
-    web::Query(search): web::Query<Query>,
-    web::Query(paginated): web::Query<Paginated>,
+#[get("/v2/sbom/{id}/license.zip")]
```
mrizzi, line 101:
```suggestion
#[get("/v2/sbom/{id}/license-export")]
```

mrizzi (commented):
Please update the endpoint definition.

Cargo.toml
```diff
@@ -149,6 +149,10 @@ walkdir = "2.5"
 walker-common = "0.12.0"
 walker-extras = "0.12.0"
 zip = "2.2.0"
+csv = "1.3.0"
```
ctron, line 152: Please sort all dependencies alphabetically

entity/src/sbom_package_license.rs
ctron: IIRC, we wanted to reach out to the spec authors would the absence of that information means.

modules/ingestor/src/graph/sbom/mod.rs
```diff
@@ -411,15 +411,10 @@ impl SbomContext {
 
     pub async fn ingest_purl_license_assertion<C: ConnectionTrait>(
         &self,
-        purl: &Purl,
+        _purl: &Purl,
```
ctron, line 414: Is there a reason to keep this?

openapi.yaml
```diff
@@ -1646,6 +1529,31 @@ paths:
...
+        in: path
+        description: Digest/hash of the document, prefixed by hash type, such as 'sha256:<hash>' or 'urn:uuid:<uuid>'
+        required: true
+        schema:
+          type: string
+      responses:
+        '200':
+          description: license zip file
+          content:
+            application/octet-stream:
```
ctron, line 1548: We should change this to the actually returned type.

ctron (changes requested):
Looks great. A few things to iron out.

modules/fundamental/src/error.rs
```diff
@@ -40,6 +40,12 @@ pub enum Error {
     AgentError(AgentError),
     #[error(transparent)]
     ChainError(ChainError),
+    #[error("error from  csv write: {0}")]
```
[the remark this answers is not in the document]
bxf12315: I do not understand this one. Could you please show an example?

modules/fundamental/src/license/service/license_export.rs
```diff
@@ -0,0 +1,159 @@
...
+            archive.finish()?;
+        }
+        Ok(compressed_data)
+    }
+
+    fn generate_csvs(&self) -> Result<CSVs, Error> {
+        let mut wtr_sbom = WriterBuilder::new()
+            .delimiter(b'\t')
+            .quote_style(csv::QuoteStyle::Always)
+            .has_headers(true) // Set delimiter to tab
```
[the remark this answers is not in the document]
bxf12315: The csv needs a header.

modules/fundamental/src/license/endpoints/mod.rs
```diff
@@ -24,70 +20,53 @@ pub fn configure(config: &mut utoipa_actix_web::service_config::ServiceConfig, d
...
+    tag = "sbom",
+    operation_id = "getLicenseExport",
     params(
-        Query,
-        Paginated,
+        ("id" = String, Path, description = "Digest/hash of the document, prefixed by hash type, such as 'sha256:<hash>' or 'urn:uuid:<uuid>'"),
     ),
     responses(
-        (status = 200, description = "Matching licenses", body = PaginatedResults<LicenseSummary>),
+    (status = 200, description = "license zip file", body = Vec<u8>),
```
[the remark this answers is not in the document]
bxf12315: it's a gzip.

modules/fundamental/tests/sbom/license.rs
```diff
@@ -0,0 +1,289 @@
...
+}
+
+#[test_context(TrustifyContext)]
+#[test(tokio::test)]
+async fn test_spdx(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
+    let _result = ctx
+        .ingest_document("spdx/SATELLITE-6.15-RHEL-8.json")
+        .await?;
+
+    let result_sbom: Option<Sbom> = sbom::Entity::find()
```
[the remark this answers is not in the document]
bxf12315: I do not understand this one. It is only used to fetch the current unique record.

modules/ingestor/src/graph/sbom/clearly_defined.rs
```diff
@@ -18,40 +16,39 @@ impl SbomContext {
         let mut purls = PurlCreator::new();
         let mut licenses = LicenseCreator::new();
 
-        let mut assertions = Vec::new();
+        // let mut assertions = Vec::new();
```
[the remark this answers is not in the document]
bxf12315: Here, the absence of a node id prevents the creation of sbom_package_license, so after discussing with @mrizzi , we decided to leave them as they are.

modules/ingestor/src/graph/sbom/common/package.rs
```diff
@@ -17,15 +16,17 @@ pub struct PackageCreator {
...
 }
 
 pub struct NodeInfoParam {
     pub node_id: String,
     pub name: String,
     pub group: Option<String>,
     pub version: Option<String>,
+    pub declared_licenses: Option<LicenseInfo>,
+    pub concluded_licenses: Option<LicenseInfo>,
+    pub cyclonedx_licenses: Option<Vec<Uuid>>,
```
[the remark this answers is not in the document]
bxf12315: what's suggestion?

openapi.yaml
bxf12315: GPT told me that the best way to handle gzip is with body = Vec<u8>, and this is generated via cargo xtask generate-schemas.

modules/fundamental/src/license/service/mod.rs
```diff
@@ -202,6 +207,3 @@ impl LicenseService {
         Ok(None)
     }
 }
-
-#[cfg(test)]
-mod test;
```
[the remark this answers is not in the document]
bxf12315: https://github.com/trustification/trustify/pull/1331/files#diff-d570a56c519d162fb367738b04e59b3e58ac3aada22b39141ca1b74ae2665291R115, There already has some unit test for license.

modules/fundamental/src/error.rs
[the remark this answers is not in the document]
ctron: Please see the variant one above.

modules/fundamental/src/license/endpoints/mod.rs
```diff
@@ -1,93 +1,72 @@
...
 
 #[utoipa::path(
-    tag = "license",
-    operation_id = "listLicenses",
+    tag = "sbom",
+    operation_id = "getLicenseExport",
     params(
-        Query,
-        Paginated,
+        ("id" = String, Path, description = "Digest/hash of the document, prefixed by hash type, such as 'sha256:<hash>' or 'urn:uuid:<uuid>'"),
```
ctron, line 30: Because I just changed this in the rest of the code base, please use `Id` as the type. And remove the description, as it's part of the `Id` type now.

modules/fundamental/src/license/endpoints/mod.rs
```diff
@@ -1,93 +1,72 @@
...
+    (status = 404, description = "The document could not be found"),
     ),
 )]
-#[get("/v2/license")]
-/// List licenses
-pub async fn list_licenses(
-    state: web::Data<LicenseService>,
-    web::Query(search): web::Query<Query>,
-    web::Query(paginated): web::Query<Paginated>,
+#[get("/v2/sbom/{id}/license-export")]
```
ctron, line 37: If this is an `sbom` API, then it should also be in the `sbom` module. Not in the `license` module.

modules/fundamental/src/license/service/license_export.rs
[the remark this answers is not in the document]
ctron: Sorry, I wanted to mention that this line has nothing to with with "delimiters", as the comment suggests.

modules/ingestor/src/graph/sbom/clearly_defined.rs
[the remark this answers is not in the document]
ctron: then there should be some "TODO" in the code, explaning what has been done and why. For larger tasks, there should be a follow up task being referenced.

modules/ingestor/src/graph/sbom/common/licensing_info.rs
```diff
@@ -0,0 +1,84 @@
...
+            },
+        });
+    }
+
+    #[instrument(skip_all, fields(num = self.license_refs.len()), err)]
+    pub async fn create<'g, C>(self, db: &C) -> Result<(), DbErr>
+    where
+        C: ConnectionTrait,
+    {
+        licensing_infos::Entity::insert_many(self.license_refs)
```
ctron, line 70: Please batch the creation. Otherwise it may fail due too many parameters.

modules/ingestor/src/graph/sbom/common/package.rs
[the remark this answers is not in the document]
ctron: Remove the "cdx" variant and sort it in the appropriate field. Which one that is, we need to figure out.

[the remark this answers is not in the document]
ctron: I am wondering, why are we adding three distinct fields? The database stores this as a combination of "license" plus a "type". Why don't we just do the same here?

modules/ingestor/src/graph/sbom/common/licensing_info.rs
bxf12315: what's " batch the creation"?

ctron: For example: https://github.com/trustification/trustify/blob/ad79c1bf9825f063163ca649734ead27abe83b48/modules/ingestor/src/graph/sbom/common/package.rs#L137-L150

bxf12315: Done

modules/ingestor/src/graph/sbom/common/package.rs
```diff
@@ -1,12 +1,12 @@
 use crate::graph::sbom::common::node::NodeCreator;
-use crate::graph::sbom::{Checksum, LicenseInfo, ReferenceSource};
+use crate::graph::sbom::{Checksum, ReferenceSource};
 use sea_orm::{ActiveValue::Set, ConnectionTrait, DbErr, EntityTrait};
 use sea_query::OnConflict;
 use tracing::instrument;
 use trustify_common::db::chunk::EntityChunkedIter;
+use trustify_entity::sbom_package_license::LicenseCategory;
```
ctron, line 7: Please nest.

modules/fundamental/src/license/service/mod.rs
```diff
@@ -1,142 +1,147 @@
...
                 JoinType::Join,
-                versioned_purl::Relation::LicenseAssertions.def(),
+                package_relates_to_package::Relation::RightPackage.def(),
             )
+            .select_only()
+            .column_as(sbom_node::Column::Name, "sbom_name")
+            .column_as(sbom_package::Column::Group, "sbom_group")
+            .column_as(sbom_package::Column::Version, "sbom_version")
+            .into_model::<SbomNameGroupVersion>()
+            .one(connection)
```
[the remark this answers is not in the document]
mrizzi: @bxf12315 please take a look at https://github.com/bxf12315/trustify/pull/7

openapi.yaml
mrizzi: @bxf12315 please take a look at https://github.com/bxf12315/trustify/pull/7

modules/fundamental/src/license/service/mod.rs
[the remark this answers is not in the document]
mrizzi: Here, compared to the approach implemented in trustification, the issue comes from having this query trying to cover both the SPDX and CycloneDX situations.

In fact CycloneDX format allows for only one `component` object within the `metadata` field so there's just one package but the SPDX specs allow for multiple packages to be described by the SBOM itself.

The latter situation isn't affecting the implementation in trustification because the license export from an SPDX SBOM just has the `name` and the `namespace` values (both single for an SBOM) and no info about group and version (that is where the problem with the multiple describes would come from).

Hence, for consistency, this query should be divided into two queries, one for each format for which the license export has to work with.

I'm going to open a further PR to this PR to implement it.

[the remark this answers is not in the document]
mrizzi: Implemented in https://github.com/bxf12315/trustify/pull/8

modules/ingestor/src/graph/sbom/clearly_defined.rs
```diff
@@ -18,40 +16,40 @@ impl SbomContext {
         let mut purls = PurlCreator::new();
         let mut licenses = LicenseCreator::new();
 
-        let mut assertions = Vec::new();
+        // TODO: Since the node id cannot be obtained here, it’s not possible to replace purl_license_assertion with sbom_package_license.
```
ctron, line 19: It's ok to keep a TODO. But I'd appreciate if we could add a tracking issue for this.

ctron (approved):
Yay! Great to see the completed. There's a minor thing, adding a tracking issue for the commented out code. But I'd not consider this a blocker.

Thanks for all the work.

modules/ingestor/src/graph/sbom/clearly_defined.rs
mrizzi: https://github.com/trustification/trustify/issues/1419

mrizzi:
Load tests show consistent results compared to `main` branch.

report-pr1331-e1d878e-0 html

PR #1333, opened 2025-02-21
Revert "chore: revert stack size modification, doesn't work"

chirino:
This reverts commit 09ca2991dd2b079af4e676528b038b15152b77ce 

The stack size modification helps the test pass on a mac m2 box.


ctron (changes requested):
This actually doesn't work, that's why I removed it. You need to set the `RUST_MIN_STACK`.

The stack overflow happens on the test runner's thread, which is not on the tokio work pool.

This reproducibly fixes the issue locally and on the CI.

PR #1334, opened 2025-02-21
Simplify the LicenseInfo type, the refs: HashMap fields is not needed.

chirino:


modules/ingestor/src/graph/sbom/common/license.rs
```diff
@@ -14,21 +14,12 @@ const NAMESPACE: Uuid = Uuid::from_bytes([
 #[derive(Default, Debug, Clone, PartialEq, Eq)]
 pub struct LicenseInfo {
     pub license: String,
-    pub refs: HashMap<String, String>,
 }
 
 impl LicenseInfo {
     pub fn uuid(&self) -> Uuid {
-        let mut text = self.license.clone();
```
bxf12315, line 22: In this hashmap, the key corresponds to the licenseId in hasExtractedLicensingInfos, and the value corresponds to the name. The licenseId is only valid within the current SBOM file. The name is described in the specification as "Provide a common name of the license that is not on the SPDX list. If the License Name field is not present for a license, it implies an equivalent meaning to NOASSERTION.". Using the name as a UUID better reflects its uniqueness.
https://spdx.github.io/spdx-spec/v2.3/other-licensing-information-detected/#1031-description

chirino:
Can you write a unit test that causes test failure due to this PR that corresponds to your statement?

modules/ingestor/src/graph/sbom/spdx.rs
```diff
@@ -138,24 +138,16 @@ impl SbomContext {
...
-            .map(|e| (e.license_identifier.clone(), e.license_name.clone()))
-            .collect::<HashMap<_, _>>();
-
         let mut packages =
             PackageCreator::with_capacity(self.sbom.sbom_id, sbom_data.package_information.len());
 
         for package in &sbom_data.package_information {
             let declared_license_info = package.declared_license.as_ref().map(|e| LicenseInfo {
                 license: e.to_string(),
-                refs: license_refs.clone(),
```
ctron, line 153: If there's a license ref being used in the declared license (or concluded license), shouldn't that ref then be replaced with the license name?

ctron (changes requested):
Looks good, but there's a thing for me to understand about the license ref during ingestion.

bxf12315:
> Can you write a unit test that causes test failure due to this PR that corresponds to your statement?
I believe this change will not cause any errors; it simply represents a difference in the choice between name and licenseId. From my perspective, using name is more reasonable. For example, you can find the code from [1]
`{
            "comment": "External License Info is obtained from a build system which predates the SPDX specification and is not strict in accepting valid SPDX licenses.",
            "extractedText": "The license info found in the package meta data is: GPLv2+. See the specific package info in this SPDX document or the package itself for more details.",
            "licenseId": "LicenseRef-0",
            "name": "GPLv2+"
        },`
and another code from [2]
`{
            "comment": "External License Info is obtained from a build system which predates the SPDX specification and is not strict in accepting valid SPDX licenses.",
            "extractedText": "The license info found in the package meta data is: GPLv2+. See the specific package info in this SPDX document or the package itself for more details.",
            "licenseId": "LicenseRef-2",
            "name": "GPLv2+"
        },`
They contain the same different license IDs but the same name, and their contents are identical. If the license ID is used as a UUID, they would be two separate database records. 

Additionally, you can find the same license ID in these two different SBOM files, but their contents are completely different. In this case, they would be only one record in the data where we use license id as UUID. However, this does not lead to errors because purl_license_assertion includes the sbom_id.
code from [1] 
`{
            "comment": "External License Info is obtained from a build system which predates the SPDX specification and is not strict in accepting valid SPDX licenses.",
            "extractedText": "The license info found in the package meta data is: MIT/X License, GPL/CDDL, ASL2. See the specific package info in this SPDX document or the package itself for more details.",
            "licenseId": "LicenseRef-1",
            "name": "MIT/X License, GPL/CDDL, ASL2"
        },`

code from [2]
`{
            "comment": "External License Info is obtained from a build system which predates the SPDX specification and is not strict in accepting valid SPDX licenses.",
            "extractedText": "The license info found in the package meta data is: AFL and BSD and (BSD or GPLv2+) and BSD with advertising and Boost and GFDL and GPL and GPL+ and (GPL+ or Artistic) and GPLv1+ and GPLv2 and (GPLv2 or BSD) and (GPLv2 or GPLv3) and GPLv2 with exceptions and GPLv2+ and (GPLv2+ or AFL) and GPLv2+ with exceptions and GPLv3 and GPLv3+ and GPLv3+ with exceptions and IJG and ISC and LGPLv2 and LGPLv2+ and (LGPLv2+ or BSD) and (LGPLv2+ or MIT) and LGPLv2+ with exceptions and LGPLv2/GPLv2 and LGPLv3 and LGPLv3+ and MIT and (MIT or LGPLv2+ or BSD) and (MPLv1.1 or GPLv2+ or LGPLv2+) and Open Publication and OpenLDAP and OpenSSL and Public Domain and Python and (Python or ZPLv2.0) and Rdisc and Redistributable, no modification permitted and SISSL and Vim and zlib. See the specific package info in this SPDX document or the package itself for more details.",
            "licenseId": "LicenseRef-1",
            "name": "AFL and BSD and (BSD or GPLv2+) and BSD with advertising and Boost and GFDL and GPL and GPL+ and (GPL+ or Artistic) and GPLv1+ and GPLv2 and (GPLv2 or BSD) and (GPLv2 or GPLv3) and GPLv2 with exceptions and GPLv2+ and (GPLv2+ or AFL) and GPLv2+ with exceptions and GPLv3 and GPLv3+ and GPLv3+ with exceptions and IJG and ISC and LGPLv2 and LGPLv2+ and (LGPLv2+ or BSD) and (LGPLv2+ or MIT) and LGPLv2+ with exceptions and LGPLv2/GPLv2 and LGPLv3 and LGPLv3+ and MIT and (MIT or LGPLv2+ or BSD) and (MPLv1.1 or GPLv2+ or LGPLv2+) and Open Publication and OpenLDAP and OpenSSL and Public Domain and Python and (Python or ZPLv2.0) and Rdisc and Redistributable, no modification permitted and SISSL and Vim and zlib"
        },`
[1] https://github.com/trustification/trustify/blob/main/etc/test-data/spdx/OCP-TOOLS-4.11-RHEL-8.json
[2] https://github.com/trustification/trustify/blob/main/etc/test-data/spdx/SATELLITE-6.15-RHEL-8.json

modules/ingestor/src/graph/sbom/spdx.rs
chirino: Where does that replacement show up in the data model?

chirino:
If this DB record had fields contained the resolved lic names, I could see why it would want to hold on to the refs field.  But since it doesn't, I don't see the point of including it in the hash calculation.

bxf12315 (approved):

modules/ingestor/src/graph/sbom/cyclonedx.rs
```diff
@@ -457,10 +457,7 @@ impl<'a> ComponentCreator<'a> {
                             continue 'l;
                         };
 
-                        let license = LicenseInfo {
-                            license,
-                            refs: Default::default(),
-                        };
+                        let license = LicenseInfo { license };
```
ctron, line 460: To my understanding, in the case of the `id` being used, we'd need to resolve that first into a name? Right?

ctron (approved):
I think the current state is a bit weird. If the PR doesn't break anything, then it's either ok or we are missing test cases.

PR #1338, opened 2025-02-24
chore: Renaming variables and refactoring code a bit

helio-frota:
> [!NOTE] 
> This pull request does not resolve any issue but aims to improve clarity and avoid confusion during benchmarks.

* Initially, I suspected performance issues with the `csaf` library, but that was not the case. I realized this by commenting and un-commenting lines while running benchmarks multiple times.
* I had unsuccessful attempts with `tokio::spawn_blocking` and `rayon` trying to improve performance, that means I need to learn more. 
* The screenshot below (**if I’m not mistaken again**) indicates that the function was busy for `35ms` and waiting for `129ms`, which suggests that something might be blocking execution.

ingest_product_statuses_before

* The criterion bench triggered by this command: `cargo bench --bench bench-format` shows that we have no much difference in performance with this pull request
  * [result](https://gist.github.com/helio-frota/4058b859cd2ddd12f7f391750fb0d40c)

> [!WARNING]  
> I trust the tests indicate that this code works, though I'm not 100% sure if these changes might have an impact. While we have git revert, I appreciate and welcome the review.

### DIFF
--- a/modules/ingestor/src/service/advisory/csaf/creator.rs
+++ b/modules/ingestor/src/service/advisory/csaf/creator.rs
@@ -49,38 +49,36 @@ impl<'a> StatusCreator<'a> {
         }
     }
 
-    pub fn add_all(&mut self, ps: &Option<Vec<ProductIdT>>, status: &'static str) {
-        for r in ps.iter().flatten() {
-            let mut product = ProductStatus {
-                status,
-                ..Default::default()
-            };
-            let mut product_ids = vec![];
-            match self.cache.get_relationship(&r.0) {
-                Some(rel) => {
-                    let inner_id: &ProductIdT = &rel.product_reference;
-                    let context = &rel.relates_to_product_reference;
+    pub fn add_all(&mut self, csaf_product_ids: &Option<Vec<ProductIdT>>, status: &'static str) {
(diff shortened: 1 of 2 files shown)


modules/ingestor/src/service/advisory/csaf/creator.rs
```diff
@@ -49,38 +49,36 @@ impl<'a> StatusCreator<'a> {
         }
     }
 
-    pub fn add_all(&mut self, ps: &Option<Vec<ProductIdT>>, status: &'static str) {
-        for r in ps.iter().flatten() {
```
ctron, line 53: I'd prefer keeping the `.flatten()` approach. Less indentation levels.

ctron:
I do prefer the for loop changes. But I also think the existing variable name are good enough.

helio-frota:
> But I also think the existing variable name are good enough.

yeah I have nothing against moving back to `.flatten()`. 

some of the changes where mainly thinking on rayon (not even sure if we need this, probably needs to be solved with spawn_blocking later) and inlay hints

helio-frota:
@ctron I'm happy to close this PR better than things get lost in the chat.
the intention was mainly to share with code in case others want to run the tests they can came up with a better version tying to reduce the idle time :+1:

PR #1344, opened 2025-02-24
Skip sbom_details_cyclonedx_osv test if RUST_MIN_STACK=50000000 is not set.

chirino:


jcrossley3 (approved):

ctron:
I think skipping it is the wrong action to take. We should fail the test, giving proper instructions what is required.

If we introduce a regression now, we might miss that. And then wonder "why".

chirino:
> I think skipping it is the wrong action to take. We should fail the test, giving proper instructions what is required.
> 

I think having a unit test fail for a user running the typical `cargo test` is the wrong action for something we don't think is test failure.  

> If we introduce a regression now, we might miss that. And then wonder "why".

The regression wont pass CI checks.  You'll see the test that is failing and when you try to run that test locally, you'll see a message saying why the test was skipped.

ctron:
> > I think skipping it is the wrong action to take. We should fail the test, giving proper instructions what is required.
> 
> I think having a unit test fail for a user running the typical `cargo test` is the wrong action for something we don't think is test failure.

Well do consider this would be a test failure. The only thing is, that we require some additional "infrastructure" (env-var). I had a bunch of projects which annoyed the hell out of me, because the CI failed and locally it worked. We shouldn't do that.

> 
> > If we introduce a regression now, we might miss that. And then wonder "why".
> 
> The regression wont pass CI checks. You'll see the test that is failing and when you try to run that test locally, you'll see a message saying why the test was skipped.

Unless someone removed that env-var from the CI. Nothing would happen in such a case.

jcrossley3:
Don't we think it might be a bug that we require 50 mb to do something when every other thing we do takes no more than 2 mb? Isn't that at least worth creating an issue to research and understand why?

chirino:
How do debug deep stacks in rust?

helio-frota:
> How do debug deep stacks in rust?

/me no idea but I found this https://crates.io/crates/stacker although started to use https://docs.rs/stacker/0.1.19/stacker/fn.remaining_stack.html and then fighting with `println!` on tests and I gave up

PR #1345, opened 2025-02-24
Squash all migrations.

### LINKED ISSUE #1343: Squash the db migrations
This should speed up initializing the DB, reduce binary size, simplify things a bit.

chirino:
Fixes #1343


chirino:
It was relatively easy to generate the up and down .sql files.  So maybe I just keep this in draft until the license DDL changes are committed so that the squash can pick that up too?

JimFuller-RedHat (approved):
LGTM the sooner this goes in the sooner we know what issues happen ;)

ctron:
This means that we will need to re-create all existing instances and create new dumps.

I am ok with this, just mentioning it.

ctron:
One thing that just came to my mind. Assuming it's not too much effort, can you create a `.md` / `.adoc` file which explains what needs to be done in getting this. Note to future self? :wink:

Issue #1352, opened 2025-02-25
Delete Containerfile.pmmode


helio-frota:
Initial comment/question from @jcrossley3  [here](https://github.com/trustification/trustify/pull/766#issuecomment-2347071958). 
Indeed we have no `package` like the [xtask containerized package](https://github.com/trustification/trustify/pkgs/container/xtask).

PR #1359, opened 2025-02-26
chore: removes unused Containerfile.pmmode

helio-frota:


ctron:
What's the motivation of deleting it?

helio-frota:
@ctron https://github.com/trustification/trustify/issues/1352 :+1:

ctron:
I personally think it's ok keeping it. If people disagree, I am not against removing it.

helio-frota:
Ok I found something do to with this, let's keep and then I'll find a way to compare trustify tags performance (not at scale) :+1:

Issue #551, opened 2024-07-16
CycloneDX ingestion, check for invalid bom-refs


ctron:
Currently we simply accept the `bom-ref` references as they are. However, it may be, that there are broken refs or duplicates, like with SPDX.

We should provide a user friendly error in such a case, and allow any import process to continue, still raising the issue in a report or response. Following what we already do for SPDX.

jcrossley3:
Still relevant in light of recent 1.6 upgrade?

ctron:
Yes it is. Right now this still triggers an internal error IIRC.

ctron:
I changed the priority to P2. Simply because this will not create faulty data, but "only" provide the user with a bad error message.

JimFuller-RedHat:
might be useful to have a test demonstrating this issue

ctron:
I'll create a test and maybe add a fix for it too.

PR #1363, opened 2025-02-26
fix(sbom): validate CDX files during ingestion

ctron:
This allows to catch some errors, which might lead to internal errors. Instead of failing with a DB error, we now return a more user friendly explanation of what went wrong. We also return warnings, if they don't prevent the ingestion of the document.

Closes: #551

### DIFF
--- a/modules/fundamental/tests/sbom/cyclonedx/mod.rs
+++ b/modules/fundamental/tests/sbom/cyclonedx/mod.rs
@@ -1,3 +1,4 @@
+mod corner_cases;
 mod cpe;
 mod external;
 mod purl;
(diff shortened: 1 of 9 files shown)


modules/fundamental/tests/sbom/cyclonedx/mod.rs
```diff
@@ -146,7 +147,9 @@ where
                 serde_cyclonedx::cyclonedx::v_1_6::CycloneDx,
             >(data)?)
         },
-        |ctx, sbom, tx| Box::pin(async move { ctx.ingest_cyclonedx(sbom.clone(), tx).await }),
+        |ctx, sbom, tx| {
+            Box::pin(async move { ctx.ingest_cyclonedx(sbom.clone(), &Discard, tx).await })
```
JimFuller-RedHat, line 151: nice!

JimFuller-RedHat (approved):
great stuff, LGTM

PR #1364, opened 2025-02-26
fix(terraform): only ensure the tag exists, don't require a value

ctron:
Before this change, we did check for the internal subnet by filtering for kubernetes.io/role/internal-elb having an empty value. That worked for cluster we set up in the past. However, it looks like newer clusters to have a value of "1" for this tag. This might come from openshift installers of a newer version.

As we don't require a specific value, but only look for the existence of the tag, this change perform this change. Allowing both variants.


ctron:
@queria maybe you can test this

Issue #1356, opened 2025-02-25
Investigate and decide about the necessity of `bench-format.rs`


helio-frota:
While analyzing (in a noob way) which sync functions might be blocking the async runtime, `bench-format.rs` was created specifically to observe the behavior of the [masked function](https://github.com/trustification/trustify/blob/2f72e0efd3f84e69c068b087b86f909135e64eee/modules/ingestor/src/service/format.rs#L248) from the `jsn` crate.

During test execution, it was noted that this kind of test could generally help to identify places to improve the performance of our tests that use the ingest_document[s] function[s] from `trustify_test_context`.

As this bench focuses on the ingestion sequence, emphasizing file processing, migration, and database ingestion (without directly involving business logic) it would be nice to know if this approach really helps, in case not, it would be nice to have other alternative.

edit: also I don't like this name `bench-format.rs`

helio-frota:
progress,  jsn-masked-sync-function is not causing problems:

`... inserting document 6789...`

```
2025-02-26T10:41:36.022926Z  INFO bench_format::trustify_benches: inserting document 6789...
2025-02-26T10:41:36.023117Z  INFO ingest{format=Unknown labels=("source", "TrustifyContext") issuer=None}:from_bytes:advisory_from_bytes: trustify_module_ingestor::service::format: return=Ok(CSAF)
2025-02-26T10:41:36.023126Z  INFO ingest{format=Unknown labels=("source", "TrustifyContext") issuer=None}:from_bytes: trustify_module_ingestor::service::format: return=Ok(CSAF)
ingest_documents        time:   [272.78 µs 293.18 µs 315.20 µs]
``` 

* with this result the name `bench-format.rs` makes no sense anymore
* will continue on the second part

helio-frota:
this issue can be closed, the file can be deleted later

helio-frota:
this comment `from the tech part ended up on the database creation for TrustifyContext` 

Clarification:

this means I used the good old approach: comment -> bench, uncomment -> bench... and 
I ended up on the postgresql embedded creation as something that causes a good amount of time usage for our tests.
  * We need that postgresql embedded creation to isolate the test
  * That means all the other previous parts of our test-framework are really fast according to the criterion results that compares current and previous runs

PR #1365, opened 2025-02-26
chore: removes unnecessary bench-format.rs bench

helio-frota:
Related with #1356


helio-frota:
:thinking: 

I'm removing 1 file 
🤷
```
Run cargo fmt --check
Error: file `/home/runner/work/trustify/trustify/modules/fundamental/benches/bench-format.rs` does not exist
```

helio-frota:
This is happening locally also, I'll close the PR and investigate

PR #1372, opened 2025-02-27
ci: Adds a new build doing the same as current

helio-frota:
* We probably need to merge and send a fake PR later to see the cache behavior.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,64 @@
+name: Rust
+
+on:
+  push:
+    branches:
+      - main
+      - release/**
+  merge_group:
+    types:
+      - checks_requested
+  pull_request:
+    branches:
+      - main
+      - release/**
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Cache Theseus Postgresql Installation
+        uses: actions/cache@v4
+        with:
+          path: ~/.theseus/postgresql
+          key: ${{ runner.os }}-theseus-postgresql-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Format
+        run: cargo fmt --check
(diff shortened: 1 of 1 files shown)


ctron:
I am sorry, but I fail to see the difference to the current CI build.

helio-frota:
:+1:  

the github starters-workflow expects a file called `rust.yml` https://github.com/actions/starter-workflows/tree/main?tab=readme-ov-file#starter-workflows, this is the first difference.

and then:

* it already setup rust (cargo etc) without `Swatinem/rust-cache@v2` 
* during tests using this, even not removing java and etc the disk usage was smaller at the end, you can see the difference [here](https://github.com/helio-frota/bench-action-tests/actions)
*  you can compare both files [here](https://github.com/helio-frota/bench-action-tests/tree/main/.github/workflows)
* and the cache size difference [here](https://github.com/helio-frota/bench-action-tests/actions/caches)

helio-frota:
yay we have cache size improvement !

helio-frota:
@ctron next step is to merge and send a fake PR to see the timing

helio-frota:
> IIRC the du command will be run in a sub-directory, not at the root of the filesystem. Which means, that you most likely will not observe any change caused by the deletion of the other frameworks (like Java).

that's correct. both shows `/home/runner/work/bench-action-tests/bench-action-tests`, but the `rust` version returns smaller result even without deletion of the other frameworks, which means if I remove I'll save more disk space.

> To my understanding, the Rust caching action is more specialized, and will collect build information from other, unused, dependencies. Meaning, it gets bigger overtime.

tbqh I have no idea about that.

> You have more runs on the "ci" workflow than you have on the "rust" workflow. (Which actually leads to the name having an impact, but not that the name is "rust").

yea that makes sense.

> I think you can just delete the caches, and re-run both workflows again, once more. And I'd bet that you end up with the same disk usage and cache size.

good, I'll do it thanks for the review :+1:

helio-frota:
> And I'd bet that you end up with the same disk usage and cache size.

that's correct:



and both du `5.9G	.`

thanks for the review, I'll close and start the other tasks :+1:

PR #1377, opened 2025-02-28
chore: Remove unused internal dependency

helio-frota:
Noticed that this is now part of the trustify-server crate. Removing unused internal dependency improves build speed, reduces cache usage in CI, and enhances the experience with rust-analyzer and the editor (though I don't have numbers to share).


helio-frota:
I deleted my branch by mistake but not problem a new PR can be sent after this 
https://github.com/trustification/trustify/issues/1381 :+1:

PR #1383, opened 2025-03-03
fix: analysis graph needs to consult graph_cache for external sboms

JimFuller-RedHat:
This PR addresses a few analysis graph issues - the first being that external sbom was not being 'found' due to it either not being loaded into graph_cache or not being in 'slice' of graph_cache that we select for each query.

- [x] added more tests to analysis graph endpoints and services
- [x] query graph_cache for external sbom lookup
- [x] lots of minor fixes, reorganised test-data, re-organised tests
- [x] fix supplier issue with cdx (eventually red hat generated sboms should consistently define **metadata.supplier** key)

There is two open questions (which may also be fixed by this PR) which I will follow up with @ctron 

* currently we seem to take a 'slice' from graph_cache ... maybe its simpler to just use graph_cache direct ?
* heuristics for cdx is failing (due to supplier not being set, as well as another issue)


ctron (approved):
Nice work!

PR #1387, opened 2025-03-03
feat: filter components by partial matches of a purl and/or cpe

### LINKED ISSUE #1280: api/v2/analysis/components?q=purl does not resolve
purl should be escaped ... but does not seem to work consistently ... also wondering what partial match would look like

jcrossley3:
Fixes #1280

This required translating CPE queries to their constituent fields in the CPE table.

We converted query::Value from an enum to a trait to query both purl's and cpe's because simple string comparisons won't cut it for both the loading of relevant objects *and* the comparisons within the analysis graph.

So we delegate the comparison logic to the disparate objects being queried -- each must now implement the Value trait and corresponding trait objects are then used to find matches.

### DIFF
--- a/common/src/db/query.rs
+++ b/common/src/db/query.rs
@@ -16,6 +16,7 @@ use std::collections::HashMap;
 use std::str::FromStr;
 use std::sync::OnceLock;
 use utoipa::{IntoParams, ToSchema};
+use value::OrderedValue;
 
 /// Convenience function for creating a search Query
 ///
@@ -79,23 +80,26 @@ impl Query {
     /// returning true if the context is successfully matched by the
     /// query, by either a filter or a full-text search of all the
     /// values.
-    pub fn apply(&self, context: &HashMap<&'static str, Value>) -> bool {
+    pub fn apply(&self, context: &HashMap<&str, Box<&dyn Value>>) -> bool {
         use Operator::*;
         self.parse().iter().all(|c| match c {
             Constraint {
                 field: Some(f),
                 op: Some(o),
                 value: vs,
-            } => context.get(f.as_str()).is_some_and(|field| match o {
-                Equal => vs.iter().any(|v| field.eq(v)),
-                NotEqual => vs.iter().all(|v| field.ne(v)),
-                Like => vs.iter().any(|v| field.contains(v)),
-                NotLike => vs.iter().all(|v| !field.contains(v)),
-                GreaterThan => vs.iter().all(|v| field.gt(v)),
-                GreaterThanOrEqual => vs.iter().all(|v| field.ge(v)),
-                LessThan => vs.iter().all(|v| field.lt(v)),
-                LessThanOrEqual => vs.iter().all(|v| field.le(v)),
-                _ => false,
+            } => context.get(f.as_str()).is_some_and(|field| {
+                let field = OrderedValue(**field);
+                match o {
+                    Equal => vs.iter().any(|v| field.eq(v)),
+                    NotEqual => vs.iter().all(|v| field.ne(v)),
+                    Like => vs.iter().any(|v| field.0.like(v)),
+                    NotLike => vs.iter().all(|v| !field.0.like(v)),
+                    GreaterThan => vs.iter().all(|v| field.gt(v)),
+                    GreaterThanOrEqual => vs.iter().all(|v| field.ge(v)),
+                    LessThan => vs.iter().all(|v| field.lt(v)),
+                    LessThanOrEqual => vs.iter().all(|v| field.le(v)),
+                    _ => false,
+                }
             }),
             Constraint {
                 field: None,
@@ -103,7 +107,7 @@ impl Query {
                 ..
             } => context
                 .values()
-                .any(|field| vs.iter().any(|v| field.contains(v))),
+                .any(|field| vs.iter().any(|v| field.like(v))),
             _ => false,
         })
     }
--- a/common/src/db/query/value.rs
+++ b/common/src/db/query/value.rs
@@ -1,102 +1,108 @@
-use chrono::{Local, NaiveDateTime};
-use human_date_parser::{ParseResult, from_human_time};
 use std::cmp::Ordering;
 use time::OffsetDateTime;
 
-pub enum Value<'a> {
-    String(&'a str),
-    Int(i32),
-    Float(f64),
-    Date(&'a OffsetDateTime),
-    Array(Vec<Value<'a>>),
+pub trait Value {
+    fn like(&self, pat: &str) -> bool;
+    fn compare(&self, other: &str) -> Option<Ordering>;
 }
 
-impl Value<'_> {
-    pub fn contains(&self, pat: &str) -> bool {
-        match self {
-            Self::String(s) => s.contains(pat),
-            Self::Date(d) => d.to_string().contains(pat),
-            Self::Int(_) | Self::Float(_) => false,
-            Self::Array(a) => a.iter().any(|v| v.contains(pat)),
+impl Value for &str {
+    fn like(&self, pat: &str) -> bool {
+        str::contains(self, pat)
+    }
+    fn compare(&self, other: &str) -> Option<Ordering> {
+        (*self).partial_cmp(other)
+    }
+}
+
+impl Value for String {
+    fn like(&self, pat: &str) -> bool {
+        self.as_str().contains(pat)
+    }
+    fn compare(&self, other: &str) -> Option<Ordering> {
+        self.as_str().partial_cmp(other)
+    }
+}
+
+impl<T: Value> Value for Vec<T> {
+    fn like(&self, pat: &str) -> bool {
+        self.iter().any(|v| v.like(pat))
+    }
+    fn compare(&self, other: &str) -> Option<Ordering> {
+        if self
+            .iter()
+            .any(|v| matches!(v.compare(other), Some(Ordering::Equal)))
+        {
+            Some(Ordering::Equal)
+        } else {
+            None
         }
     }
 }
 
-impl PartialEq<String> for Value<'_> {
-    fn eq(&self, rhs: &String) -> bool {
-        match self {
-            Self::String(s) => s.eq(rhs),
-            Self::Int(v) => match rhs.parse::<i32>() {
-                Ok(i) => v.eq(&i),
-                _ => false,
-            },
-            Self::Float(v) => match rhs.parse::<f64>() {
-                Ok(i) => v.eq(&i),
-                _ => false,
-            },
-            Self::Date(v) => match from_human_time(&v.to_string()) {
-                Ok(ParseResult::DateTime(field)) => match from_human_time(rhs) {
-                    Ok(ParseResult::DateTime(other)) => field.eq(&other),
-                    Ok(ParseResult::Date(d)) => {
-                        let other = NaiveDateTime::new(d, field.time())
-                            .and_local_timezone(Local)
-                            .unwrap();
-                        field.eq(&other)
-                    }
-                    Ok(ParseResult::Time(t)) => {
-                        let other = NaiveDateTime::new(field.date_naive(), t)
-                            .and_local_timezone(Local)
-                            .unwrap();
-                        field.eq(&other)
-                    }
-                    _ => false,
-                },
-                _ => false,
-            },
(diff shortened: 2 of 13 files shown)


modules/analysis/src/service/load.rs
```diff
@@ -236,7 +240,22 @@ impl AnalysisService {
...
             GraphQuery::Query(query) => sbom_node::Entity::find()
-                .filtering(query.clone())?
+                // TODO: The following translation effectively ignores
+                // any fields named 'purl' or 'cpe', potentially
+                // selecting every SbomId in the db, without any other
+                // fields in the query. We need help constructing the
+                // SELECT statement that would enable us to find
+                // partial matches of the purl's stored within
+                // entity::qualified_purl::Column::Purl, i.e. the
+                // 'PURL' column of the 'QUALIFIED_PURL' table.
```
jcrossley3, line 250: @JimFuller-RedHat help?

JimFuller-RedHat: maybe something like

`SELECT * FROM qualified_purl WHERE get_purl(id)::text LIKE '%test%';`

JimFuller-RedHat: I think 'we got the graph side covered' does the needful with respect to this bug ... the formal definition of url and uri, order of url params are not significant.

modules/analysis/src/service/load.rs
```diff
@@ -236,7 +240,22 @@ impl AnalysisService {
...
+                .filtering_with(
+                    query.clone(),
+                    sbom_node::Entity.columns().translator(|f, _, _| match f {
```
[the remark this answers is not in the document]
jcrossley3: This is a temporary hack. It essentially translates any `purl` filter to "unfiltered", effectively returning all SBOM id's in the db.

modules/analysis/src/service/load.rs
jcrossley3: I'm sorry, I don't understand.

modules/analysis/src/service/load.rs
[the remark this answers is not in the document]
jcrossley3: So it ignores the `purl` filter on the DB fetch but uses it in the graph query, if that makes sense. I wanted something that enabled the test to pass, so that when we refine the filter for the fetch, the graph query continues to work as well.

modules/analysis/src/service/load.rs
jcrossley3: Maybe we just compare the versioned purl, i.e. all but the qualifiers, in the load? And then provide extra logic in the graph query to account for the qualifier order.

jcrossley3: But that won't pick up partial matches in the qualifiers, e.g. `purl~curl` would match but `purl~i386` would not.

JimFuller-RedHat: I am not proposing fixing filter queries for partial match on purl in postgres db with this PR ... sounds more complicated then just fixing analysis graph (though admittedly not as satisfying!).

JimFuller-RedHat: the one where people use 'api/v2/analysis/component'

jcrossley3: No, I meant what are the uses of that URL that need to work and currently don't? I think I'm asking for an answer to [this question you posed](https://github.com/trustification/trustify/issues/1280#issuecomment-2685100672). I think I can make the globbing work (alternative 1) using your suggestion of `get_purl` with caveats, in particular that a glob that includes some portion of a versioned purl *AND* one or more qualifiers may return a false negative.

You have a better idea of what globs users will search for than me to know whether that's acceptable.

modules/analysis/src/endpoints/tests/mod.rs
```diff
@@ -335,6 +333,52 @@ async fn quarkus_component_by_purl(ctx: &TrustifyContext) -> Result<(), anyhow::
...
+    // query(&app, format!("purl={}", PURL.replace("&", r"\&"))).await;
+
+    // This will work, but I was unable to include the
+    // `repository_url` qualifier in the query due to unexpected
+    // urlencoding, either in the database or the get_purl function.
+    query(
+        &app,
+        "purl~pkg:maven/com.redhat.quarkus.platform/quarkus-bom@3.2.11.Final-redhat-00001&purl~type=pom",
+    )
+    .await;
```
jcrossley3, line 377: @JimFuller-RedHat we need to decide whether this is acceptable. And if we think it's too much to expect users to remember the `~` (ILIKE) operator, we can easily make `=` behave like `~`. I'm thinking users may like that, but :shrug:

JimFuller-RedHat:
thx for this - I propose we hold off merging this - feel free to hand off the issue to me - I will investigate further exactly what is needed as well as have a go at implementing.

jcrossley3:
> thx for this - I propose we hold off merging this - feel free to hand off the issue to me - I will investigate further exactly what is needed as well as have a go at implementing.

Is #1280 no longer required?

modules/analysis/src/endpoints/tests/mod.rs
```diff
@@ -335,6 +333,50 @@ async fn quarkus_component_by_purl(ctx: &TrustifyContext) -> Result<(), anyhow::
...
+
+    for each in [
+        "purl=pkg:maven/com.redhat.quarkus.platform/quarkus-bom@3.2.11.Final-redhat-00001?type=pom\\&repository_url=https%3a%2f%2fmaven.repository.redhat.com%2fga%2f",
+        "purl~pkg:maven/com.redhat.quarkus.platform/quarkus-bom@3.2.11.Final-redhat-00001&purl~type=pom&purl~repository_url=https%3A%2F%2Fmaven.repository.redhat.com%2Fga%2F",
+        "purl~quarkus-bom",
+        "cpe=cpe:/a:redhat:quarkus:3.2::el8",
+        "cpe~cpe:/a:redhat:quarkus:3.2::el8",
+        "cpe~cpe:/a:redhat:quarkus:3.2",
+        "cpe~cpe:/a::quarkus",
+        "cpe~redhat",
```
jcrossley3, line 367: @JimFuller-RedHat I have these queries working now. Please let me know if I missed any that you need.

modules/analysis/src/endpoints/tests/mod.rs
```diff
@@ -335,6 +333,50 @@ async fn quarkus_component_by_purl(ctx: &TrustifyContext) -> Result<(), anyhow::
...
+        tracing::debug!(test = "", "{response:#?}");
+        response
+    }
+
+    let app = caller(ctx).await?;
+
+    for each in [
+        "purl=pkg:maven/com.redhat.quarkus.platform/quarkus-bom@3.2.11.Final-redhat-00001?type=pom\\&repository_url=https%3a%2f%2fmaven.repository.redhat.com%2fga%2f",
+        "purl~pkg:maven/com.redhat.quarkus.platform/quarkus-bom@3.2.11.Final-redhat-00001&purl~type=pom&purl~repository_url=https%3A%2F%2Fmaven.repository.redhat.com%2Fga%2F",
+        "purl~quarkus-bom",
```
JimFuller-RedHat, line 362: cool - so we are using ~ to indicate partial matches desired ?

modules/analysis/src/service/load.rs
```diff
@@ -237,9 +248,59 @@ impl AnalysisService {
                 .distinct()
                 .into_query(),
             GraphQuery::Query(query) => sbom_node::Entity::find()
-                .filtering(query.clone())?
+                .join(JoinType::Join, sbom_node::Relation::Package.def())
```
JimFuller-RedHat, line 251: I can see why we want to narrow down the set of sboms first ... though concerned about performance ... does this set of joins perform well ?

modules/analysis/src/service/load.rs
```diff
@@ -237,9 +248,59 @@ impl AnalysisService {
...
+                                ("cpe", "~", Err(_)) => Some(v.into()),
+                                ("cpe", _, Err(e)) => Some(e.to_string()),
+                                ("cpe", _, _) => Some("illegal operation for cpe".into()),
+                                _ => None,
+                            }
+                        })
+                        .add_expr(
+                            "purl",
+                            SimpleExpr::FunctionCall(
+                                Func::cust("get_purl".into_identity())
```
JimFuller-RedHat, line 298: FWIW work @ctron has recently done probably means we no longer need to use **get_purl**

modules/analysis/src/service/load.rs
```diff
@@ -237,9 +248,59 @@ impl AnalysisService {
...
+                )
                 .select_only()
                 .column(sbom_node::Column::SbomId)
+                .filtering_with(
+                    query.clone(),
+                    sbom_node::Entity
+                        .columns()
+                        .add_columns(cpe::Entity.columns())
+                        .translator(|f, op, v| {
+                            match (f, op, OwnedUri::from_str(v)) {
```
JimFuller-RedHat, line 266: I cannot tell what the resultant sql code looks like from this code block ... it might be we need to do a pg func.

JimFuller-RedHat (approved):
Thx for persevering through this ... I can see the pros/cons being balanced - the desire to update the core query mechanisms to encapsulate what is needed for analysis graph expands the scope of this PR but cannot see how we can avoid ... **LGTM** with the caveat that we need to see impact on performance (I have added a few analysis graph queries to scale load testing).

common/src/db/query.rs
```diff
@@ -79,31 +80,34 @@ impl Query {
...
+                }
             }),
             Constraint {
                 field: None,
                 value: vs,
                 ..
             } => context
                 .values()
-                .any(|field| vs.iter().any(|v| field.contains(v))),
+                .any(|field| vs.iter().any(|v| field.like(v))),
```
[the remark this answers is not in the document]
jcrossley3: This was merely a name change of a function on `query::Value` which used to be an enum but with this PR is now a trait. This simplified the avoidance of name conflicts for things implementing `query::Value` that already had a `contains()` defined.

To answer your question, no, this particular change has no impact on performance. But other changes in this PR might. :smile:

common/src/db/query/value.rs
```diff
@@ -1,102 +1,108 @@
...
+    }
+    fn compare(&self, other: &str) -> Option<Ordering> {
+        match other.parse::<f64>() {
+            Ok(i) => self.partial_cmp(&i),
+            _ => None,
+        }
+    }
+}
+
+impl Value for OffsetDateTime {
```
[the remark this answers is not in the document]
jcrossley3: Correct. Because we changed Value from an enum to a trait, we had to implement `Value` for all the variants of the old enum. Not directly related to anything regarding the implementation of `Value` for `Purl` and `Cpe`.

We could've introduced variants for `Purl` and `Cpe` in the old enum, but it seemed "cleaner" to convert it to a trait. This does introduce a potential perf impact, though. Using trait objects involves a vtable lookup that enum's don't have to do. I wouldn't expect that impact to be significant but :shrug:

modules/analysis/src/endpoints/tests/mod.rs
jcrossley3: Yes, that's always been how the `q=` stuff works. But until now, it's only the UX guys who know -- they're the only ones who call the api. :smile:

modules/analysis/src/service/load.rs
jcrossley3: Oh? I didn't see any related changes beneath `migrations/` -- are we no longer storing purl's in a json blob?

modules/analysis/src/service/load.rs
jcrossley3: I should also point out that the triggering of a full text search by an invalid CPE only affects what's loaded into the graph. But the querying of the loaded graph will still be accurate. The user will never see whatever false positives were initially loaded by the full-text search.

This is an extreme edge case, obviously. :wink:

modules/analysis/src/service/load.rs
JimFuller-RedHat: we are, but I think the full serialisation is byproduct of the model now

JimFuller-RedHat: lets keep this how you have it for now

jcrossley3:
@JimFuller-RedHat I think I found a better alternative to this. The trait objects are just too awkward to use. I'm gonna create another PR for comparison. Closing this one for now.

PR #1398, opened 2025-03-05
test: create more tests for image_index

ctron:


JimFuller-RedHat (approved):
you might consider adding document_id to the test assert as that is the main (stable) indicator) when we transition to external document in the returned REST API results, otherwise LGTM

Issue #1395, opened 2025-03-05
Uploading the same SBOM multiple times in parallel creates multiple equivalent SBOMs


carlosthe19916:
How to reproduce it:
- Clone this repository and start it without Auth enabled. E.g. `AUTH_DISABLED=true cargo run --bin trustd`
- Upload the same SBOM multiple times in parallel.
  - In my case I used this command `cat ./etc/datasets/ds3/spdx/quarkus-bom-2.13.8.Final-redhat-00004.json.bz2 | http POST localhost:8080/api/v2/sbom Content-Type:application/json+bzip2` and execute it in parallel by:

```shell
seq 10 | parallel -j10 my_script
```

Finally you can see that although only one SBOM was uploaded repeatedly, there are 10 instances of the same SBOM:
```shell
http GET http://localhost:8080/api/v2/sbom | jq
```

https://github.com/user-attachments/assets/109703e7-b557-4ac0-a9c1-ea6a69dd7fc5

Note: I discovered this while executing the e2e tests of the UI as automation can trigger independent tests in parallel.

ctron:
This looks like a bug. As you seem to have that scripted, could you do a quick check to see if the same is true for advisories?

carlosthe19916:
> This looks like a bug. As you seem to have that scripted, could you do a quick check to see if the same is true for advisories?

I confirm. Same results with Advisories

ctron:
There's an easy way to reproduce this in a test.

ctron:
I think this should be closed by #1399 and  #1403

PR #1399, opened 2025-03-05
fix: don't create duplicate documents when ingesting in parallel (spdx)

ctron:
fixes partially #1395

### DIFF
--- a/migration/src/m0000010_init_up.sql
+++ b/migration/src/m0000010_init_up.sql
@@ -3688,21 +3688,21 @@ CREATE INDEX sbomnodenameginidx ON public.sbom_node USING gin (((name)::text) pu
 -- Name: sha256_index; Type: INDEX; Schema: public; Owner: -
 --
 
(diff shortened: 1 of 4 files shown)


ctron:
This works for the SPDX,  but is still missing the fix for other document types.

migration/src/m0000010_init_up.sql
```diff
@@ -3688,21 +3688,21 @@ CREATE INDEX sbomnodenameginidx ON public.sbom_node USING gin (((name)::text) pu
 -- Name: sha256_index; Type: INDEX; Schema: public; Owner: -
 --
 
-CREATE INDEX sha256_index ON public.source_document USING btree (sha256);
+CREATE UNIQUE INDEX sha256_index ON public.source_document USING btree (sha256);
```
JimFuller-RedHat, line 3691: with simple indexes we used sea_orm syntax ...

JimFuller-RedHat (approved):
LGTM ... I can do the cdx PR if you like

PR #1401, opened 2025-03-06
fix: improve performance of sbom/advisory endpoint

dejanb:
This PR improves fetching vulnerabilities for SBOM, by improving two things:
- Pre-selecting CPEs related to the SBOM and using them in a query. This significantly improves the performance of the query in large datasets.
- Fetching missing data directly in the query, instead of running separate queries in the post-query loop. Querying things in the loop can be costly when dealing with large number of rows. This is now done for advisory issuer organisation and vulnerability description.

With these two changes the speed of the endpoint for ds4 on a laptop is increased tenfold

```
before
Executed in  276.15 secs    fish           external
   usr time    3.24 secs    0.10 millis    3.24 secs
   sys time    0.42 secs    1.36 millis    0.42 secs

after
Executed in   28.02 secs    fish           external
   usr time    2.44 secs    0.08 millis    2.44 secs
   sys time    0.39 secs    1.11 millis    0.39 secs   
```

and it makes ds4 quite usable even in the local environment.

This is just a first step in optimizing this endpoint without doing big refactoring and API changes.

Two things that needs to be addressed further:

- We still calculate score for each vulnerability in a loop which is inefficient in all cases, but especially when we are dealing with hundreds or thousands of vulnerabilities. I think this is the first case where we would want to introduce some kind of pre-processing of data and caching.
- The logic for getting all CPEs related to SBOM is a bit crude at the moment. For example, it will not cover the cases of supersets. It needs more work to filter all these CPEs properly in the future. So far I haven't found an advisory that will be affected by this.

Both of these should be addressed in separate issues.

### DIFF
--- a/etc/datasets/ds3/csaf/2024/cve-2024-28834.json
+++ b/etc/datasets/ds3/csaf/2024/cve-2024-28834.json
@@ -1,0 +1,12198 @@
+{
+  "document": {
+    "aggregate_severity": {
+      "namespace": "https://access.redhat.com/security/updates/classification/",
+      "text": "Moderate"
+    },
+    "category": "csaf_vex",
+    "csaf_version": "2.0",
+    "distribution": {
+      "text": "Copyright © Red Hat, Inc. All rights reserved.",
+      "tlp": {
+        "label": "WHITE",
+        "url": "https://www.first.org/tlp/"
+      }
+    },
+    "lang": "en",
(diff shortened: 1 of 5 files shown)


JimFuller-RedHat (approved):
LGTM and yes I can see where we can improve things further but this is good to get in.

PR #1402, opened 2025-03-06
RDS : extension handled by master user

gildub:
For security reasons AWS RDS handles extensions with different restrictions.
The `pg_trgm` extension must be created and granted to public schema from the Master user (RDS super-super user).

I initially thought the extension had to be moved out of the migration to avoid breaking it.
Meanwhile SeaORM migration can execute admin user level commands.

This is easier than having to insert those commands in postgres post installations.

See https://github.com/trustification/trustify-helm-charts/issues/4 for more details


chirino:
What runs the 000005 sql files?

chirino:
And commit message should include note for the release notes that this is a breaking change in that existing dbs will need to be rest.

migration/src/m0000010_init.rs
```diff
@@ -15,12 +15,18 @@ impl MigrationTrait for Migration {
         // Delete the SET statements at the start of the file
         // Move the drop statements at the start to m0000010_init_down.sql
 
-        let x = manager
+        manager
+            .get_connection()
+            .execute_unprepared(include_str!("m0000005_admin_up.sql"))
+            .await
+            .map(|_| ())?;
+
```
chirino, line 23: How is it that this executes 'admin level' commands, but it doesn't in the next statement for m0000010_init_up.sql?

gildub: @chirino, 

That's a good point. I was focused on running that part as "manager" but didn't realize that the other migrations are running at the same level !

The SeaOrm's manager has only the privileges of the application user's roles allocated to it. Therefore that won't work and the extension creation must be removed from the migration.

PR #1403, opened 2025-03-07
Ensure no duplication of documents during parallel upload in more locations

ctron:
This fixes issue #1395 also for advisories and adds a test for CDX. Which was already covered by a previous PR, but didn't have a test.


JimFuller-RedHat (approved):
if CI passes, LGTM

PR #1406, opened 2025-03-10
fix: imageindex>imagevariant external sboms

### LINKED ISSUE #1392: imageindex>variant with external multisbom does not work
This is not working due to a combination of factors:

* cdx **pedigree.variants** in parent sbom (imageindex) needs to result in inserted rows into **sbom_external_node**
* **bom-ref** and **checksums** in child sboms (variants) are denoted slightly differently

Also - we need to make sure we get a latest set of examples to ensure we are aligned.

JimFuller-RedHat:
RH imagevariants are a bit 'special' as they currently do not encode **checksum** natively in spdx/cdx ... opting to set _version=sha256###############_. 

This means that we have to handle (for now) with alternate heuristics for determining their relationships.

- [x] add more tests for imagevariant
- [x] lookup for external sbom imageindex>imagevariant
- [x] lookup for ancestor queries with external sbom imageindex>imagevariant
- [x] minor fixes

fixes #1392

**Note: In the future we may remove this special case if/when rh imagevariant start denoting checksums**

### DIFF
--- a/modules/ingestor/src/graph/sbom/processor/rh_prod_comp.rs
+++ b/modules/ingestor/src/graph/sbom/processor/rh_prod_comp.rs
@@ -103,13 +103,15 @@ impl super::Processor for RedHatProductComponentRelationships {
                 );
             }
 
-            // if it is a top level component, described by the SBOM, packaging components,
-            // then we process those
-            if !((relationship == &Relationship::Generates
-                || relationship == &Relationship::Package)
-                && top_level.contains(prod_node_id))
-            {
-                continue;
+            if relationship != &Relationship::Variant {
+                // if it is a top level component, described by the SBOM, packaging components,
+                // then we process those
+                if !((relationship == &Relationship::Generates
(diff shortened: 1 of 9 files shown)


modules/ingestor/src/graph/sbom/processor/rh_prod_comp.rs
```diff
@@ -103,13 +103,15 @@ impl super::Processor for RedHatProductComponentRelationships {
...
             }
 
-            // if it is a top level component, described by the SBOM, packaging components,
-            // then we process those
-            if !((relationship == &Relationship::Generates
-                || relationship == &Relationship::Package)
-                && top_level.contains(prod_node_id))
-            {
-                continue;
+            if !(relationship == &Relationship::Variant) {
```
ctron, line 106:
```suggestion
            if relationship != &Relationship::Variant {
```

It feels easier to read.

ctron (approved):

Issue #1322, opened 2025-02-20
`test_sbom_details_cyclonedx_osv` cause stack overflow errors


ctron:
`test_sbom_details_cyclonedx_osv` causes a stack overflow:

```
 thread 'sbom::details::test_sbom_details_cyclonedx_osv' has overflowed its stack
fatal runtime error: stack overflow
```

helio-frota:
can we close this ? 👍

helio-frota:
has relation with https://github.com/trustification/trustify/pull/1344 but I'm not sure if that is the same

ctron:
That PR simply hides this issue, but doesn't resolve it.

helio-frota:
@ctron @chirino @jcrossley3 

I think I found the places where stack overflows occur, they are commented together, as it is a stack error that must be at the limit... un-commenting one or the other will cause the error. The cause may be these functions or other previous ones that used up stack space.

As you can see now I have assertion error:

```console
──── STDERR:             trustify-module-fundamental::fundamental sbom::details::sbom_details_cyclonedx_osv

thread 'sbom::details::sbom_details_cyclonedx_osv' panicked at modules/fundamental/tests/sbom/details.rs:90:5:
assertion `left == right` failed
  left: 11
 right: 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

  Cancelling due to test failure
────────────
     Summary [   1.358s] 1 test run: 0 passed, 1 failed, 389 skipped
        FAIL [   1.348s] trustify-module-fundamental::fundamental sbom::details::sbom_details_cyclonedx_osv
error: test run failed
```

```diff
➜  trustify git:(test-stack) ✗ git diff
diff --git a/modules/fundamental/tests/sbom/details.rs b/modules/fundamental/tests/sbom/details.rs
index 6f093e27..93fefce7 100644
--- a/modules/fundamental/tests/sbom/details.rs
+++ b/modules/fundamental/tests/sbom/details.rs
@@ -12,13 +12,13 @@ use trustify_test_context::TrustifyContext;
 #[instrument]
 async fn sbom_details_cyclonedx_osv(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
     // get the env var RUST_MIN_STACK and make sure it set to 50000000
-    let stack_size = env::var("RUST_MIN_STACK").unwrap_or("".to_string());
-    if stack_size != "50000000" {
-        println!(
[...]

PR #1407, opened 2025-03-10
fix: prevent stack overflow

ctron:
Instead of keeping the CycloneDX structure on the stack, we move it to the heap (using Box). This prevents the stack overflow.

This is done both for the deserialized struct, as well as of the recursively created component creators.

Closes: #1322


JimFuller-RedHat (approved):
LGTM

Issue #1405, opened 2025-03-07
Intermittent index failure when uploading the same document in parallel


ctron:
When uploading the same document in parallel, it may happen that instead of accepting the document, one gets an internal error, due do an index constraint violation.

When inserting the "source document" entry, there's three indices involved, which are unique (the three SHAs). Only the first one (sha256) is part of the "on conflict" statement. If a duplicate entry is being created, an index violation is triggered, an "nothing" is done, keeping the existing document.

However, it might happen that the sha256 index update succeeds, and the sha384 update fails. As that one isn't covered by the "on conflict" statement, it will trigger an error. Unfortunately, it seems that the "on conflict" cannot be written in a way to say "either one of those columns", only "all of those columns".

PR #1408, opened 2025-03-10
fix: ensure we always detect a duplicate source document

ctron:
Having multiple unique indices on the source document table, it might happen that the sha256 is not the first one being reported. This may lead to the issue that the "on conflict" statement will not be respected and that we still get an internal error reported. As, right now, it is not possible to use "on conflict" with multiple constraints independently, we need to manually check for this error and also use nested transactions, in order to continue with the transaction.

Closes: #1405


common/src/db/mod.rs
```diff
@@ -176,6 +181,54 @@ impl ConnectionTrait for Database {
     }
 }
 
+#[async_trait::async_trait]
```
JimFuller-RedHat, line 184: this is nice

JimFuller-RedHat (approved):
LGTM and we should check out tomorrow's load report to see impact on perf

PR #1412, opened 2025-03-10
feat: adds metric of running importers

ctron:


modules/importer/src/server/mod.rs
```diff
@@ -92,6 +96,9 @@ impl Server {
             runs.retain(|heart| heart.is_beating());
             let count = runs.len();
 
+            // Update metrics
+            running_importers.record(count as _, &[]);
```
jcrossley3, line 100: When there are multiple replicas, won't the idle ones be updating this value to 0 every second?

ctron: Each instance has its own instance ID.

jcrossley3 (approved):

PR #1415, opened 2025-03-12
perf: advisory and vuln endpoints

JimFuller-RedHat:
This PR helps speed up _api/v2/advisory_ and _api/v2/vulnerability_ endpoints.

After applying this it is good to run:
```sql
VACUUM ANALYZE;
REINDEX database trustify;
```
to start at theoretical baseline performance.

### DIFF
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -1,13 +1,13 @@
 pub use sea_orm_migration::prelude::*;
 
-mod m0000010_init;
-mod m0000020_add_sbom_group;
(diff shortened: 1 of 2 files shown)


ctron (approved):

dejanb:
This looks good. I think one of the problems in these endpoints is all the queries we are executing in the loop after the "main" queries to fetch additional data (like vulnerability description). It's the similar issue to the one tackled in https://github.com/trustification/trustify/pull/1401. I'll try to come up with some improvements in that area.

Issue #1417, opened 2025-03-13
/api/v2/sbom/{id}/packages?limit=0&offset=0 contains duplicate PURLs


carlosthe19916:
In Response to https://github.com/trustification/trustify-ui/issues/392

- Uploaded [ANSIBLE-AUTOMATION-PLATFORM-2.0-RHEL-8.json](https://github.com/user-attachments/files/19207310/ANSIBLE-AUTOMATION-PLATFORM-2.0-RHEL-8.json)
- Hit `GET /api/v2/sbom/{id}/packages?limit=0&offset=0` and the response will be [response](https://gist.github.com/carlosthe19916/8617ec4d994153ee4a5680a6b8bd2905)

If you see the lines between [206-269](https://gist.github.com/carlosthe19916/8617ec4d994153ee4a5680a6b8bd2905#file-gistfile1-txt-L206-L269) then you will see this:

![Image](https://github.com/user-attachments/assets/8476ea50-a047-4063-be51-ff75d5223e48)

Multiple PURLs with the same UUID are reported.

ctron:
Interesting, I can reproduce this in a test case. Will investigate.

JimFuller-RedHat:
I was going to ask if you started from scratch (eg. remove .trustify) and reingest... but it looks like @ctron can reproduce.

PR #1418, opened 2025-03-13
fix: prevent duplicate PURLs or CPEs being returned

ctron:
The query of encoding a list of PURLs (and CPEs) into a single field did create duplicates. This is now properly being aggregated. One requirement of this the ability to compare PURLs, which isn't possible using the JSON type in PSQL. It works with strings or rows though.

Therefore, and also because it was overdue, this PR switches to extracting the PURL from the `purl` field, instead of the three layer table structure. Which might have a positive performance effect too.

Closes: #1417


carlosthe19916 (approved):
LGTM! I confirm, I do not see duplicate values any longer.

JimFuller-RedHat (approved):
glad u did this (I was not brave enough to contemplate), LGTM

PR #1422, opened 2025-03-13
feat: add endpoint to process multiple purls

### LINKED ISSUE #1425: Allow to query multiple purls
In order to allow the Exhort service to retrieve vulnerabilities from TPA it has to expose a REST endpoint that returns the vulnerabilities associated with a collection of PURLs.

This is required in order to integrate Dependency Analytics with a self managed instance of TPA.

Downstream JIRA [TC-2304](https://issues.redhat.com/browse/TC-2304)

ruromero:
Allow retrieving multiple PurlDetails by either purl or uuid

Fix #1425

### DIFF
--- a/modules/fundamental/src/ai/service/tools/package_info.rs
+++ b/modules/fundamental/src/ai/service/tools/package_info.rs
@@ -67,19 +67,13 @@ Input: The package name, its Identifier URI, or UUID.
             .ok_or("Input should be a string")?
             .to_string();
 
-        // Try lookup as a PURL
-        let mut purl_details = match Purl::try_from(input.clone()) {
+        let mut purl_details = match service
+            .fetch_purl_details(&vec![input.clone()], Deprecation::Ignore, db)
+            .await
+        {
             Err(_) => None,
-            Ok(purl) => service.purl_by_purl(&purl, Deprecation::Ignore, db).await?,
+            Ok(details) => details.get(&input.clone()).cloned(),
         };
-
-        // Try lookup as a UUID
-        if purl_details.is_none() {
-            purl_details = match Uuid::parse_str(input.as_str()) {
-                Err(_) => None,
-                Ok(uuid) => service.purl_by_uuid(&uuid, Deprecation::Ignore, db).await?,
-            };
-        }
 
         // Fallback to search
         if purl_details.is_none() {
@@ -98,9 +92,16 @@ Input: The package name, its Identifier URI, or UUID.
             purl_details = match results.items.len() {
                 0 => None,
                 1 => {
-                    service
-                        .purl_by_uuid(&results.items[0].head.uuid, Deprecation::Ignore, db)
-                        .await?
+                    let details = service
+                        .fetch_purl_details(
+                            &vec![results.items[0].head.uuid.to_string()],
+                            Deprecation::Ignore,
+                            db,
+                        )
+                        .await?;
+                    details
+                        .get(&results.items[0].head.uuid.to_string())
+                        .cloned()
                 }
                 _ => {
                     #[derive(Serialize)]
--- a/modules/fundamental/src/ai/service/tools/package_info.rs
+++ b/modules/fundamental/src/ai/service/tools/package_info.rs
@@ -68,7 +68,7 @@ Input: The package name, its Identifier URI, or UUID.
             .to_string();
 
         let mut purl_details = match service
-            .fetch_purl_details(&vec![input.clone()], Deprecation::Ignore, db)
+            .fetch_purl_details(&[input.as_str()], Deprecation::Ignore, db)
             .await
         {
             Err(_) => None,
@@ -94,7 +94,7 @@ Input: The package name, its Identifier URI, or UUID.
                 1 => {
                     let details = service
                         .fetch_purl_details(
-                            &vec![results.items[0].head.uuid.to_string()],
+                            &[results.items[0].head.uuid.to_string().as_str()],
                             Deprecation::Ignore,
                             db,
                         )
--- a/modules/fundamental/src/ai/service/tools/package_info.rs
+++ b/modules/fundamental/src/ai/service/tools/package_info.rs
@@ -68,11 +68,11 @@ Input: The package name, its Identifier URI, or UUID.
             .to_string();
 
         let mut purl_details = match service
-            .fetch_purl_details(&[input.as_str()], Deprecation::Ignore, db)
+            .fetch_purl_details(&[&input], Deprecation::Ignore, db)
             .await
         {
             Err(_) => None,
-            Ok(details) => details.get(&input.clone()).cloned(),
+            Ok(details) => details.get(&input).cloned(),
         };
 
         // Fallback to search
@@ -94,7 +94,7 @@ Input: The package name, its Identifier URI, or UUID.
                 1 => {
                     let details = service
                         .fetch_purl_details(
-                            &[results.items[0].head.uuid.to_string().as_str()],
+                            &[results.items[0].head.uuid.to_string()],
                             Deprecation::Ignore,
                             db,
                         )
--- a/modules/fundamental/src/purl/endpoints/mod.rs
+++ b/modules/fundamental/src/purl/endpoints/mod.rs
@@ -6,12 +6,8 @@ use crate::{
     },
 };
 use actix_web::{HttpResponse, Responder, get, post, web};
-use sea_orm::prelude::Uuid;
-use std::{collections::HashMap, str::FromStr};
 use trustify_auth::{ReadSbom, authorizer::Require};
-use trustify_common::{
-    db::Database, db::query::Query, model::Paginated, model::PaginatedResults, purl::Purl,
-};
+use trustify_common::{db::Database, db::query::Query, model::Paginated, model::PaginatedResults};
 
 use super::model::details::purl::{PurlsRequest, PurlsResponse};
 
@@ -37,41 +33,6 @@ pub fn configure(config: &mut utoipa_actix_web::service_config::ServiceConfig, d
         .service(all);
 }
 
-async fn fetch_purl_details(
-    service: &PurlService,
-    db: &Database,
-    deprecated: trustify_module_ingestor::common::Deprecation,
-    identifiers: &[String],
-) -> Result<HashMap<String, PurlDetails>, String> {
-    let mut purls: Vec<Purl> = Vec::new();
-    let mut uuids: Vec<Uuid> = Vec::new();
-    for key in identifiers {
-        if key.starts_with("pkg:") {
-            let purl = Purl::from_str(key).map_err(|e| format!("Invalid purl '{}': {}", key, e))?;
-            purls.push(purl);
-        } else {
-            let id = Uuid::from_str(key).map_err(|e| format!("Invalid UUID '{}': {}", key, e))?;
-            uuids.push(id);
-        }
-    }
-    let mut result: HashMap<String, PurlDetails> = HashMap::new();
-    let purl_details = service
-        .purls_by_purl(purls, deprecated, db)
-        .await
-        .map_err(|e| format!("Failed to fetch purl details by purl: {}", e))?;
-    for detail in purl_details {
-        result.insert(detail.head.purl.to_string(), detail);
-    }
-    let uuid_details = service
-        .purls_by_uuid(uuids, deprecated, db)
-        .await
-        .map_err(|e| format!("Failed to fetch purl details by uuid: {}", e))?;
-    for detail in uuid_details {
-        result.insert(detail.head.uuid.to_string(), detail);
-    }
-    Ok(result)
-}
-
 #[utoipa::path(
     operation_id = "getPurl",
     tag = "purl",
@@ -94,7 +55,10 @@ pub async fn get(
 ) -> actix_web::Result<impl Responder> {
     let result_key = key.into_inner();
     let identifiers = vec![result_key.clone()];
-    match fetch_purl_details(&service, &db, deprecated, &identifiers).await {
+    match service
+        .fetch_purl_details(&identifiers, deprecated, db.as_ref())
+        .await
+    {
         Ok(details) => match details.get(&result_key) {
             Some(detail) => Ok(HttpResponse::Ok().json(detail)),
             None => Ok(HttpResponse::NotFound().body("Identifier not found")),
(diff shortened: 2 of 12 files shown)


jcrossley3:
Is this feature in support of some new user request or reported problem?

ruromero:
> Is this feature in support of some new user request or reported problem?

@jcrossley3 I have linked the related Jira

ruromero:
Got it. I just went ahead and did it without asking 😄

modules/fundamental/src/purl/endpoints/mod.rs
```diff
@@ -33,9 +33,45 @@ pub fn configure(config: &mut utoipa_actix_web::service_config::ServiceConfig, d
...
+    }
+    let uuid_details = service
+        .purls_by_uuid(uuids, deprecated, db)
+        .await
+        .map_err(|e| format!("Failed to fetch purl details by uuid: {}", e))?;
+    for detail in uuid_details {
+        result.insert(detail.head.uuid.to_string(), detail);
+    }
+    Ok(result)
+}
```
jcrossley3, line 73: I think this belongs in the service rather than the endpoint. And return an `Error` instead of a `String`, similar to how the other service functions do, e.g. `purls_by_purl`.

Style feedback: a for loop after a mutable collection is a smell to me. It looks like java code. :smile: There are almost always cleaner, safer alternatives in the `Iterator` trait, e.g. `fold`, `map`, `partition`, etc. It can sometimes be tricky to do in async code, but I think it's worth familiarizing yourself with `Iterator` to take advantage of it when you can.

modules/fundamental/src/purl/endpoints/test.rs
```diff
@@ -350,3 +350,43 @@ async fn package_with_status(ctx: &TrustifyContext) -> Result<(), anyhow::Error>
 
     Ok(())
 }
+
+#[test_context(TrustifyContext)]
+#[test(actix_web::test)]
```
jcrossley3, line 355: Way to include a test! ;)

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -230,6 +230,30 @@ impl PurlService {
...
+            .all(connection)
+            .await?;
+
+        let mut details = Vec::with_capacity(items.len());
+        for purl in items {
+            details
+                .push(PurlDetails::from_entity(None, None, &purl, deprecation, connection).await?);
+        }
+        Ok(details)
+    }
```
jcrossley3, line 255: I don't love that both this fn and `purl_by_purl` exist. They're mostly redundant. I'd say we can probably delete `purl_by_purl` in favor of this one, and any current caller of `purl_by_purl` can just grab the first element returned by this one. Make sense?

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -249,6 +273,25 @@ impl PurlService {
...
```
jcrossley3, line 293: Same applies to this one as my comment above.

jcrossley3 (changes requested):

modules/fundamental/src/purl/endpoints/mod.rs
```diff
@@ -33,9 +33,45 @@ pub fn configure(config: &mut utoipa_actix_web::service_config::ServiceConfig, d
...
         .service(get)
+        .service(get_multiple)
         .service(all);
 }
 
+async fn fetch_purl_details(
+    service: &PurlService,
+    db: &Database,
+    deprecated: trustify_module_ingestor::common::Deprecation,
+    identifiers: &[String],
```
ctron, line 44: Welcome to Rust :) Here's a thought:

`&[String]` is an immutable reference to a type considered for providing mutable string processing. Most likely, you're ok with just using `&str` and thus `&[&str]`.

However, you can also use a trait, and let the compiler figure it out: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=76815c841727331234f2e7bf437bda44

```rust
fn do_it(i: impl IntoIterator<Item=impl AsRef<str>>) {
    for i in i {
        println!("{}", i.as_ref());
    }
}

fn main() {
    do_it(["foo", "bar"]);
    do_it(["foo".to_string(), "bar".to_string()]);
    do_it(vec!["foo".to_string(), "bar".to_string()]);
}
```

Which basically says: give me something that I can turn into an iterator which gives me something that can turn into `&str`.

modules/fundamental/src/purl/endpoints/mod.rs
ruromero: Guilty of being a Java developer 😅

modules/fundamental/src/purl/service/mod.rs
ruromero: Yes, perfect sense.

modules/fundamental/src/purl/service/mod.rs
ruromero: done

modules/fundamental/src/purl/endpoints/mod.rs
ruromero: I have made the changes but I'm not sure if the way I'm creating the `uuids` and `purls` is more efficient as it has to iterate twice, given that I can't directly throw errors from closures.
Please, let me know if there's a better way to achieve something similar.

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -230,41 +232,90 @@ impl PurlService {
...
+            .into_iter()
+            .map(|detail| (detail.head.uuid.to_string(), detail));
+
+        let result: HashMap<String, PurlDetails> = purls.chain(uuids).collect();
+        Ok(result)
+    }
+
+    async fn purls_by_purl<C: ConnectionTrait>(
+        &self,
+        purls: Vec<Purl>,
```
ctron, line 280: It doesn't look like you actually need (to own) a `Vec` here.

You can either use `&[Purl]`, or consume the `purls` later, using:

```rust
        let canonical: Vec<CanonicalPurl> = purls
            .into_iter()
            .map(|purl| CanonicalPurl::from(purl))
            .collect();
```

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -230,41 +232,90 @@ impl PurlService {
         }
     }
 
-    pub async fn purl_by_purl<C: ConnectionTrait>(
+    #[instrument(skip(self, connection), err(level=tracing::Level::INFO))]
+    pub async fn fetch_purl_details<C: ConnectionTrait>(
         &self,
-        purl: &Purl,
+        identifiers: &[String],
```
ctron, line 238: I think this is still valid:

`&[String]` is an immutable reference to a type considered for providing mutable string processing. Most likely, you're ok with just using `&str` and thus `&[&str]`.

However, you can also use a trait, and let the compiler figure it out: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=76815c841727331234f2e7bf437bda44

```rust
fn do_it(i: impl IntoIterator<Item=impl AsRef<str>>) {
    for i in i {
        println!("{}", i.as_ref());
    }
}

fn main() {
    do_it(["foo", "bar"]);
    do_it(["foo".to_string(), "bar".to_string()]);
    do_it(vec!["foo".to_string(), "bar".to_string()]);
}
```

Which basically says: give me something that I can turn into an iterator which gives me something that can turn into `&str`.

modules/fundamental/src/purl/endpoints/mod.rs
jcrossley3: Take a gander at the commit I just pushed, and let's discuss.

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -230,41 +231,84 @@ impl PurlService {
...
+                    .await?
+                    .into_iter()
+                    .map(|detail| (detail.head.uuid.to_string(), detail)),
+            )
+            .collect();
+
+        Ok(details)
+    }
+
+    async fn purls_by_purl<C: ConnectionTrait>(
```
jcrossley3, line 271: Nice attention to detail making this and `purls_by_uuid` private. I like that the only entry point is through `fetch_purl_details`

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -230,41 +231,84 @@ impl PurlService {
...
+            .purls_by_purl(&purls, deprecated, connection)
             .await?
-        {
-            Some(purl) => Ok(Some(
-                PurlDetails::from_entity(None, None, &purl, deprecation, connection).await?,
-            )),
-            None => Ok(None),
+            .into_iter()
+            .map(|detail| (detail.head.purl.to_string(), detail))
+            .chain(
```
jcrossley3, line 260: I liked your use of `chain` here, even though I refactored it a bit.

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -230,41 +231,84 @@ impl PurlService {
...
         connection: &C,
-    ) -> Result<Option<PurlDetails>, Error> {
-        let canonical = CanonicalPurl::from(purl.clone());
-        match qualified_purl::Entity::find()
-            .filter(qualified_purl::Column::Purl.eq(canonical))
-            .one(connection)
+    ) -> Result<HashMap<String, PurlDetails>, Error> {
+        let (purls, uuids): (Vec<_>, Vec<_>) = identifiers
+            .iter()
+            .partition(|key| key.as_ref().starts_with("pkg:"));
```
jcrossley3, line 243: I didn't love the `partition_map(|either| either)` use before. A closure doing nothing other than returning its argument is a smell.

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -230,41 +231,84 @@ impl PurlService {
         }
     }
 
-    pub async fn purl_by_purl<C: ConnectionTrait>(
+    #[instrument(skip(self, connection), err(level=tracing::Level::INFO))]
+    pub async fn fetch_purl_details<C: ConnectionTrait, I: AsRef<str> + Debug>(
```
jcrossley3, line 235: Accepting an `AsRef<str>` makes callers lives simpler as far as needing a `String`, a `&String` or a `&str`. They all implement `AsRef<str>` so they can all be passed. `Debug` is only there because of the `instrument` macro above.

jcrossley3 (approved):
Hope my changes make sense. Happy to discuss them further.

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -233,47 +232,40 @@ impl PurlService {
...
 
     #[instrument(skip(self, connection), err(level=tracing::Level::INFO))]
-    pub async fn fetch_purl_details<C: ConnectionTrait>(
+    pub async fn fetch_purl_details<C: ConnectionTrait, I: AsRef<str> + Debug>(
         &self,
-        identifiers: &[&str],
+        identifiers: &[I],
         deprecated: Deprecation,
         connection: &C,
-    ) -> Result<HashMap<String, PurlDetails>, String> {
```
jcrossley3: Pretty much every module has an `Error` struct that does convenient macro-y things. Rarely (if at all) within trustify are our Results returning some type of `Error`, not a `String`.

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -230,41 +231,84 @@ impl PurlService {
...
+        Ok(details)
+    }
+
+    async fn purls_by_purl<C: ConnectionTrait>(
+        &self,
+        purls: &[Purl],
+        deprecation: Deprecation,
+        connection: &C,
+    ) -> Result<Vec<PurlDetails>, Error> {
+        let canonical: Vec<CanonicalPurl> = purls
```
jcrossley3, line 277:
```suggestion
    ) -> Result<Vec<PurlDetails>, Error> {
        if purls.is_empty() {
            return Ok(Default::default());
        }
        let canonical: Vec<CanonicalPurl> = purls
```
Should probably put in a short-circuit return for an empty list...

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -230,41 +231,84 @@ impl PurlService {
...
-        match qualified_purl::Entity::find_by_id(*purl_uuid)
-            .one(connection)
-            .await?
-        {
-            Some(pkg) => Ok(Some(
-                PurlDetails::from_entity(None, None, &pkg, deprecation, connection).await?,
-            )),
-            None => Ok(None),
+    ) -> Result<Vec<PurlDetails>, Error> {
+        let items = qualified_purl::Entity::find()
```
jcrossley3, line 301:
```suggestion
    ) -> Result<Vec<PurlDetails>, Error> {
        if uuids.is_empty() {
            return Ok(Default::default());
        }
        let items = qualified_purl::Entity::find()
```
Here too...

modules/fundamental/src/purl/service/mod.rs
ruromero: Thanks!

modules/fundamental/src/purl/service/mod.rs
ruromero: I will read more about loops vs iterators. Thanks

modules/fundamental/src/purl/service/mod.rs
ruromero: Why is it better over iterating an empty Vector?

ruromero:
@jcrossley3 thank you very much for your changes and comments. Let me know if something is missing.

modules/fundamental/src/purl/service/mod.rs
jcrossley3: It's just to prevent hitting the database with that `qualified_purl::Entity::find()` call on line 282. I think that's going to execute whether the list is empty or not, and our db is busy enough as it is. :smile:

ruromero: Makes sense. Change applied.

modules/fundamental/src/purl/service/mod.rs
ruromero: updated

modules/fundamental/src/ai/service/tools/package_info.rs
```diff
@@ -67,20 +67,14 @@ Input: The package name, its Identifier URI, or UUID.
...
 
-        // Try lookup as a PURL
-        let mut purl_details = match Purl::try_from(input.clone()) {
+        let mut purl_details = match service
+            .fetch_purl_details(&[input.as_str()], Deprecation::Ignore, db)
+            .await
+        {
             Err(_) => None,
-            Ok(purl) => service.purl_by_purl(&purl, Deprecation::Ignore, db).await?,
+            Ok(details) => details.get(&input.clone()).cloned(),
```
jcrossley3, line 75: I'm actually surprised clippy didn't catch this.
```suggestion
            Ok(details) => details.get(&input).cloned(),
```

jcrossley3 (commented):
Just a couple more nits

PR #1426, opened 2025-03-14
refactor: remove three later PURL processing in more places

ctron:
Instead of using the three PURL layers (base, version, qualifier) for building result PURLs, we only use the `purl` field from the qualified result, and evaluate the rest inline.

The allows us to reduce load on the database even more. I'll run a scale-test next.

### DIFF
--- a/modules/fundamental/src/sbom/model/details.rs
+++ b/modules/fundamental/src/sbom/model/details.rs
@@ -150,13 +150,6 @@ impl SbomDetails {
                 "vulnerability"."modified" AS "vulnerability$modified",
                 "vulnerability"."withdrawn" AS "vulnerability$withdrawn",
                 "vulnerability"."cwes" AS "vulnerability$cwes",
-                "base_purl"."id" AS "base_purl$id",
-                "base_purl"."type" AS "base_purl$type",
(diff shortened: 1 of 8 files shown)


JimFuller-RedHat (approved):
LGTM

mrizzi (approved):
@ctron thanks 👍

modules/fundamental/src/sbom/model/details.rs
```diff
@@ -356,11 +349,7 @@ impl SbomAdvisory {
                 id: each.sbom_package.node_id.clone(),
```
dejanb, line 349: I think we can optimize this further by not returning the whole package info as I don't think it's useful in this context. That would also allow us to join less tables in the `relevant_advisory_info` query. But we can do that in a separate PR

ctron: Yea, I agree, there's more potential. But I didn't want to go all crazy right before a release. :wink:

ctron:
Either I did it wrong, or it doesn't impact performance:

image

ctron:
Ah, I did it wrong indeed. :facepalm: Will re-run.

PR #1428, opened 2025-03-14
perf: drop advisory_vulnerability_vulnerability_id_gist index for now

dejanb:
The `advisory_vulnerability_vulnerability_id_gist` considerably slows down `sbom/{id}/advisory` endpoint. Let's disable it for now, until we can further investigate it.


ctron:
Just checking, is dropping it good enough? Or do we need the original one back?

JimFuller-RedHat (approved):
LGTM!

ctron:
Ah, indeed: https://github.com/search?q=repo%3Atrustification%2Ftrustify%20advisory_vulnerability_vulnerability_id_gist&type=code

The `DELETE` there made me assume we already had one.

ctron (approved):

PR #1436, opened 2025-03-17
feat: filter components by partial matches of a purl and/or cpe

jcrossley3:
Fixes: #1280

Added the proper logic to translate CPE queries into their constituent columns and Purl queries into comparisons against the `get_purl()` proc, which stringifies the associated JSON blob in the database.

This accounts for the loading of the graph. The fact that both `Cpe` and `Purl` structs now implement Valuable accounts for querying the graph, i.e. applying the query to a context constructed from the graph.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8392,6 +8392,7 @@ dependencies = [
  "tracing",
  "trustify-migration",
  "trustify-test-context",
+ "urlencoding",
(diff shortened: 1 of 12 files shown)


JimFuller-RedHat (approved):
this looks great, and faster ;) LGTM

PR #1450, opened 2025-03-21
feat: add analyze endpoint for purl-based vulnerability scan

### LINKED ISSUE #1449: Expose an endpoint to analyze the vulnerabilities associated with the given pURLs
As a user I want to know if the pURLs I'm using are affected by any known vulnerabilities.

```
{
    "purls": [
        "pkg:maven/jakarta.enterprise/jakarta.enterprise.cdi-api@2.0.2?type=jar",
        "pkg:maven/io.quarkus/quarkus-hibernate-orm@2.13.5.Final?type=jar",
        "pkg:maven/jakarta.interceptor/jakarta.interceptor-api@1.2.5?type=jar",
        "pkg:maven/io.quarkus/quarkus-narayana-jta@2.13.5.Final?type=jar",
        "pkg:maven/io.quarkus/quarkus-jdbc-postgresql@2.13.5.Final?type=jar",
        "pkg:maven/com.fasterxml.jackson.core/jackson-databind@2.13.1?type=jar",
        "pkg:maven/jakarta.el/jakarta.el-api@3.0.3?type=jar",
        "pkg:maven/org.postgresql/postgresql@42.5.0?type=jar",
        "pkg:maven/io.quarkus/quarkus-core@2.13.5.Final?type=jar"
    ]
}
```

This will help the Dependency Analytics - Exhort backend to create a report based on the SBOM generated from any given package manager file.

ruromero:
Fix #1449 

The proposed solution evaluates if the provided purls are within the range of the vulnerable versions available in the database.

It returns a HashMap where the key is the purl and the value is a Vector containing the vulnerabilities.

Drop endpoint added it #1425 because it is not needed from any client and this endpoint implements what the other endpoint was meant for.

### DIFF
--- a/modules/fundamental/src/vulnerability/endpoints/mod.rs
+++ b/modules/fundamental/src/vulnerability/endpoints/mod.rs
@@ -2,15 +2,14 @@
 mod test;
 
 use crate::{
-    Error,
-    Error::Internal,
+    Error::{self, Internal},
     endpoints::Deprecation,
     vulnerability::{
-        model::{VulnerabilityDetails, VulnerabilitySummary},
+        model::{AnalysisRequest, AnalysisResponse, VulnerabilityDetails, VulnerabilitySummary},
         service::VulnerabilityService,
     },
 };
-use actix_web::{HttpResponse, Responder, delete, get, web};
+use actix_web::{HttpResponse, Responder, ResponseError, delete, get, post, web};
 use sea_orm::TransactionTrait;
 use trustify_auth::{DeleteVulnerability, ReadAdvisory, authorizer::Require};
 use trustify_common::{
@@ -25,7 +24,8 @@ pub fn configure(config: &mut utoipa_actix_web::service_config::ServiceConfig, d
         .app_data(web::Data::new(db))
         .service(all)
         .service(delete)
-        .service(get);
+        .service(get)
+        .service(analyze);
 }
 
 #[utoipa::path(
@@ -130,5 +130,32 @@ pub async fn delete(
         }
     } else {
         Ok(HttpResponse::NotFound().finish())
+    }
+}
+
+#[utoipa::path(
+  operation_id = "analyze",
+  tag = "vulnerability",
+  request_body = AnalysisRequest,
+  responses(
+      (status = 200, description = "Analyze the provided purls to search for known vulnerabilities", body = AnalysisResponse),
+  ),
+)]
+#[post("/v2/vulnerability/analyze")]
+/// Analyze the provided purls for the known vulnerabilities
+pub async fn analyze(
+    service: web::Data<VulnerabilityService>,
+    db: web::Data<Database>,
+    request: web::Json<AnalysisRequest>,
+    _: Require<ReadAdvisory>,
+) -> actix_web::Result<impl Responder> {
+    let purls: Vec<&str> = request.purls.iter().map(|s| s.as_str()).collect();
+    match service.analyze_purls(purls, db.as_ref()).await {
+        Ok(details) => Ok(HttpResponse::Ok().json(details)),
+        Err(error) => match error {
+            Error::BadRequest(_) => Ok(error.error_response()),
+            _ => Ok(HttpResponse::InternalServerError()
+                .body(format!("Error analyzing purls: {}", error))),
+        },
     }
 }
--- a/modules/fundamental/src/vulnerability/service/mod.rs
+++ b/modules/fundamental/src/vulnerability/service/mod.rs
@@ -1,8 +1,14 @@
+use std::{collections::HashMap, str::FromStr};
+
 use crate::{
     Error,
     vulnerability::model::{VulnerabilityDetails, VulnerabilitySummary},
 };
-use sea_orm::{EntityTrait, FromQueryResult, IntoIdentity, QuerySelect, QueryTrait, prelude::*};
+use futures_util::{TryFutureExt, TryStreamExt};
+use sea_orm::{
+    EntityTrait, FromQueryResult, IntoIdentity, QuerySelect, QueryTrait, Statement, StreamTrait,
+    prelude::*,
+};
 use sea_query::{ColumnRef, Func, IntoColumnRef, IntoIden, SimpleExpr};
 use trustify_common::{
     db::{
@@ -11,6 +17,7 @@ use trustify_common::{
         query::{Columns, Filtering, Query},
     },
     model::{Paginated, PaginatedResults},
+    purl::{Purl, PurlErr},
 };
 use trustify_entity::{
     cvss3::{self, Severity},
@@ -154,6 +161,103 @@ impl VulnerabilityService {
         let result = query.exec(connection).await?;
 
         Ok(result.rows_affected)
+    }
+
+    pub async fn analyze_purls<C>(
+        &self,
+        purls: Vec<&str>,
+        connection: &C,
+    ) -> Result<HashMap<String, Vec<VulnerabilityDetails>>, Error>
+    where
+        C: ConnectionTrait + StreamTrait,
+    {
+        let query = purls
+            .iter()
+            .map(|p| {
+                let purl = Purl::from_str(p)?;
+
+                let ns_condition = match &purl.namespace {
+                    Some(namespace) => {
+                        let sql = "base_purl.namespace = $1";
+                        Statement::from_sql_and_values(
+                            connection.get_database_backend(),
+                            sql,
+                            [namespace.into()],
+                        )
+                    }
+                    None => Statement::from_string(
+                        connection.get_database_backend(),
+                        "base_purl.namespace IS NULL",
+                    ),
+                };
+
+                let values = [
+                    p.to_string().into(),
+                    purl.name.into(),
+                    purl.ty.into(),
+                    purl.version.ok_or_else(|| PurlErr::MissingVersion(p.to_string()))?.into(),
+                ];
+                let sql = format!(r#"
+                  SELECT $1 as requested_purl,
+                    vulnerability.id, vulnerability.title, vulnerability.reserved,
+                    vulnerability.published, vulnerability.modified, vulnerability.withdrawn, vulnerability.cwes
+                  FROM base_purl
+                    LEFT JOIN purl_status ON base_purl.id = purl_status.base_purl_id
+                    INNER JOIN version_range ON purl_status.version_range_id = version_range.id
+                    LEFT JOIN vulnerability ON purl_status.vulnerability_id = vulnerability.id
+                    INNER JOIN status ON purl_status.status_id = status.id
+                  WHERE {ns_condition}
+                    AND base_purl.name = $2
+                    AND base_purl.type = $3
+                    AND version_matches($4, version_range.*) = TRUE
+                    AND status.slug != 'fixed'"#);
+                let query =
+                    Statement::from_sql_and_values(connection.get_database_backend(), &sql, values);
+                Ok(query.to_string())
+            })
+            .collect::<Result<Vec<String>, Error>>()?
+            .join(" UNION ALL ");
+
+        let stmt = Statement::from_string(connection.get_database_backend(), query);
+        let result = connection.stream(stmt).map_err(Error::from).await?;
+        let result = result
+            .map_err(Error::from)
+            .and_then(|row| self.to_vuln(row, connection))
+            .try_fold(
+                HashMap::new(),
+                |mut acc: HashMap<String, Vec<VulnerabilityDetails>>,
+                 (requested_purl, vuln_details)| async {
+                    acc.entry(requested_purl).or_default().push(vuln_details);
+                    Ok(acc)
+                },
+            )
+            .await?;
+
+        Ok(result)
+    }
+
+    async fn to_vuln<C: ConnectionTrait>(
+        &self,
+        row: QueryResult,
+        connection: &C,
+    ) -> Result<(String, VulnerabilityDetails), Error> {
+        let requested_purl: String = row.try_get("", "requested_purl")?;
+        let vulnerability = vulnerability::Model {
+            id: row.try_get("", "id")?,
+            title: row.try_get("", "title")?,
+            reserved: row.try_get("", "reserved")?,
+            published: row.try_get("", "published")?,
+            modified: row.try_get("", "modified")?,
+            withdrawn: row.try_get("", "withdrawn")?,
+            cwes: row.try_get("", "cwes")?,
+        };
+        let vuln_details =
+            VulnerabilityDetails::from_entity(&vulnerability, Deprecation::Ignore, connection)
+                .await;
+        match vuln_details {
+            Ok(details) => Ok((requested_purl, details)),
+            Err(e) => Err(e),
+        }
     }
 }
 
--- a/modules/fundamental/src/vulnerability/service/test.rs
+++ b/modules/fundamental/src/vulnerability/service/test.rs
@@ -423,3 +423,52 @@ async fn vulnerability_queries(ctx: &TrustifyContext) -> Result<(), anyhow::Erro
 
     Ok(())
 }
+
+#[test_context(TrustifyContext)]
+#[test(actix_web::test)]
+async fn analyze_purls(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
+    let service = VulnerabilityService::new();
+
+    ctx.ingest_documents(["osv/RUSTSEC-2021-0079.json", "osv/GHSA-2ccf-ffrj-m4qw.json"])
+        .await?;
+
+    let result = service.analyze_purls(vec![], &ctx.db).await?;
+
+    assert!(result.is_empty());
+
+    for purl in ["this is not valid", "pkg:npm/missing.version"].iter() {
+        let result = service.analyze_purls(vec![purl], &ctx.db).await;
+        assert!(result.is_err());
+    }
+
+    let expected = [
+        "pkg:npm/%40fastify/passport@1.0.0", // ECOSYSTEM:afected
+        "pkg:cargo/hyper@0.14.9",            // SEMVER:affected - no namespace
+    ];
+
+    let not_found = [
+        "pkg:npm/%40fastify/passport@2.3.0", // ECOSYSTEM:fixed
+        "pkg:golang/github.com/metal3-io/baremetal-operator/apis@0.8.0", // Missing
+        "pkg:cargo/hyper@0.14.11",           // SEMVER:fixed - no namespace
+    ];
+
+    let items: Vec<&str> = expected.iter().chain(&not_found).copied().collect();
+
+    let result = service.analyze_purls(items, &ctx.db).await?;
+
+    expected.iter().for_each(|&item| {
+        assert!(
+            result.contains_key(item),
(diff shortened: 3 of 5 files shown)


modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +162,88 @@ impl VulnerabilityService {
...
+          };
+
+          let name = canonical.name;
+          let ty = canonical.ty;
+          let ns_condition = match canonical.namespace.clone() {
+            Some(namespace) => format!(r#"base_purl.namespace = '{namespace}'"#),
+            None => String::from("base_purl.namespace is null")
+          };
+
+          Some(format!(r#"
```
ctron, line 198: We should try to avoid using string injection of values, due to obvious issues with SQL injection stuff.

You can use database level parameters like this: https://github.com/trustification/trustify/blob/5e1075324e8821c8a636725c7da004127882cd2f/modules/analysis/src/service/load.rs#L167-L168

modules/fundamental/src/vulnerability/service/test.rs
```diff
@@ -427,3 +427,48 @@ async fn vulnerability_queries(ctx: &TrustifyContext) -> Result<(), anyhow::Erro
...
+
+    ctx.ingest_documents(["osv/RUSTSEC-2021-0079.json", "osv/GHSA-2ccf-ffrj-m4qw.json"])
+        .await?;
+
+    let mut result = service.analyze_purls(vec![], &ctx.db).await?;
+
+    assert!(result.is_empty());
+
+    let expected = [
+        "pkg:npm/@fastify/passport@1.0.0", // ECOSYSTEM:afected
```
jcrossley3, line 444: This is an invalid purl. It should've been percent-encoded [here](https://github.com/trustification/trustify/blob/5e1075324e8821c8a636725c7da004127882cd2f/common/src/purl.rs#L289), or better yet, we should create a `new` fn for our [Purl struct](https://github.com/trustification/trustify/blob/5e1075324e8821c8a636725c7da004127882cd2f/common/src/purl.rs#L33-L39) and hide all its fields.

modules/fundamental/src/vulnerability/endpoints/mod.rs
```diff
@@ -132,3 +132,29 @@ pub async fn delete(
...
+) -> actix_web::Result<impl Responder> {
+    let purls: Vec<&str> = request.purls.iter().map(|s| s.as_str()).collect();
+    match service.analyze_purls(purls, db.as_ref()).await {
+        Ok(details) => Ok(HttpResponse::Ok().json(details)),
+        Err(error) => {
+            Ok(HttpResponse::InternalServerError()
+                .body(format!("Error analyzing purls: {}", error)))
+        }
+    }
+}
```
jcrossley3, line 160: To me, this belongs in the `purl` module, not vulnerability. We already have a [similar api](https://github.com/trustification/trustify/blob/5e1075324e8821c8a636725c7da004127882cd2f/modules/fundamental/src/purl/endpoints/version.rs#L11-L36) that I don't think is used by the UX, but I would prefer you make that work for you instead of adding a completely new api here.

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +162,88 @@ impl VulnerabilityService {
...
+              INNER JOIN version_range ON purl_status.version_range_id = version_range.id
+              LEFT JOIN vulnerability ON purl_status.vulnerability_id = vulnerability.id
+              INNER JOIN status ON purl_status.status_id = status.id
+            WHERE {ns_condition}
+              AND base_purl.name = '{name}'
+              AND base_purl.type = '{ty}'
+              AND version_matches('{version}', version_range.*) = TRUE
+              AND status.slug != 'fixed'"#))
+        }).collect::<Vec<String>>()
+        .join(" UNION ALL ");
```
jcrossley3, line 213: I really don't want to make a habit of raw sql in our services. Feels like php! :smile:

jcrossley3 (changes requested):
I hate to be Nancy Negative, but...

modules/fundamental/src/vulnerability/endpoints/mod.rs
ruromero: As this endpoint returns vulnerabilities based on purls I considered to use this one and drop the other one. Wouldn't it make more sense?

modules/fundamental/src/vulnerability/service/mod.rs
ruromero: I wasn't able to make it work using SeaORM. I could use some help here

modules/fundamental/src/vulnerability/service/mod.rs
ruromero: You're right. Fixed.

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +162,88 @@ impl VulnerabilityService {
...
+        let stmt = Statement::from_string(connection.get_database_backend(), query);
+        let result = connection.query_all(stmt).await?;
```
ctron, line 233: As you're using stream anyway (which is great), why not use `stream` instead of `query_all`, so it's all streaming!

It may be that you need to use `C: ConnectionTrait + StreamTrait` and rebase to the most recent main.

modules/fundamental/src/vulnerability/service/mod.rs
ctron: Especially for more complex queries, I'd prefer having some raw SQL (with proper parameters) over some hard to understanding sea_orm chained calls. It also makes it easier to debug the SQL, or work with it in other tools.

modules/fundamental/src/vulnerability/service/test.rs
ruromero: updated.

ctron (commented):
:+1: from me, but I'll leave it to @jcrossley3 to approve

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +163,111 @@ impl VulnerabilityService {
...
+        if purls.is_empty() {
+            return Ok(Default::default());
+        }
+
+        let query = purls
+            .iter()
+            .filter_map(|p| {
+                let purl = Purl::from_str(p).map_err(Error::Purl);
+                let canonical = match purl {
+                    Ok(purl) => CanonicalPurl::from(purl),
```
jcrossley3, line 184: What is `CanonicalPurl` doing for you that `Purl` can't?

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +163,111 @@ impl VulnerabilityService {
...
+                    Err(_) => return None,
```
jcrossley3, line 185: So you're ignoring bad purls? Silently failing?

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +163,111 @@ impl VulnerabilityService {
...
+                };
+
+                let version = match canonical.version.as_ref() {
+                    Some(v) => v,
+                    None => return None,
```
jcrossley3, line 190: Another silent fail? Personally, I'd change `filter_map` to `map` where the closure returns a `Result` that `collect` can turn into a `Vec` or return an error for any invalid purls or valid ones without a version.

modules/fundamental/src/purl/endpoints/mod.rs
```diff
@@ -68,38 +65,6 @@ pub async fn get(
...
-pub async fn get_multiple(
-    service: web::Data<PurlService>,
-    db: web::Data<Database>,
-    request: web::Json<PurlsRequest>,
-    web::Query(Deprecation { deprecated }): web::Query<Deprecation>,
-    _: Require<ReadSbom>,
-) -> actix_web::Result<impl Responder> {
-    let items: Vec<&str> = request.items.iter().map(|s| s.as_str()).collect();
-    match service
-        .fetch_purl_details(&items, deprecated, db.as_ref())
```
jcrossley3, line 93: So we went to a lot of effort to change `fetch_purl_details` to take a list only to pass a single element to it all the time, going forward. If I'm looking at this code 6 months from now, I'm probably going to wonder why, especially if there's any performance impact to a sql 'IN' clause over an `=`.

jcrossley3 (changes requested):
There's a huge semantic difference between "analize" and "analyze". I'm not sure we want the former in our release notes. :stuck_out_tongue_winking_eye:

modules/fundamental/src/purl/endpoints/mod.rs
ruromero: Agree, it's better to revert the whole PR. I'll open a separate one for that.

jcrossley3: That's fine, but wouldn't it make more sense to delete your old multi-purl endpoint in that separate PR, too?

ruromero: Indeed. That was my intention. To revert all the changes. Will do it ASAP

ruromero: See https://github.com/trustification/trustify/pull/1462

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +163,111 @@ impl VulnerabilityService {
...
+                            [namespace.into()],
+                        )
+                    }
+                    None => Statement::from_string(
+                        connection.get_database_backend(),
+                        "base_purl.namespace IS NULL",
+                    ),
+                };
+                let name = canonical.name;
+                let ty = canonical.ty;
```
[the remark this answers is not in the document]
ruromero: Uhm. It seems I can't do that because it's a String

modules/fundamental/src/vulnerability/service/mod.rs
ruromero: I'll update the response to return `vulnerabilities` and `errors`

modules/fundamental/src/vulnerability/service/mod.rs
ruromero: Apparently nothing. I'll replace it. Thanks

ruromero:
@jcrossley3 I have rebased and applied your suggestions. Let me know if there's something pending.
Thanks again for your time

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +162,141 @@ impl VulnerabilityService {
...
+        purls: Vec<&str>,
+        connection: &C,
+    ) -> Result<HashMap<String, Vec<VulnerabilityDetails>>, Error>
+    where
+        C: ConnectionTrait + StreamTrait,
+    {
+        if purls.is_empty() {
+            return Ok(Default::default());
+        }
+
```
jcrossley3, line 177: You shouldn't need this. We used it before because both the `by_uuid` and `by_purl` fn's were always being called, whether they were needed or not.

In this case, our query will be "" if there aren't any purl's, and sea_orm should be smart enough to know it doesn't need to execute an empty string. :smile:

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +162,141 @@ impl VulnerabilityService {
...
+        let queries: Vec<Result<String, Error>> = purls
```
jcrossley3, line 178: My instinct wouldn't be to collect all the errors for every purl passed in. I'd be inclined to just return the first one. If you insist on returning all of them, there's a [more idiomatic pattern](https://doc.rust-lang.org/rust-by-example/error/iter_result.html#collect-the-failed-items-with-map_err-and-filter_map) for doing so, but for the remainder of this review, I'm going to just return the first error.
```suggestion
        let query = purls
```

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +162,141 @@ impl VulnerabilityService {
...
+            .iter()
+            .map(|p| {
+                let purl = match Purl::from_str(p) {
+                    Ok(purl) => purl,
+                    Err(e) => return Err(Error::Purl(e)),
+                };
```
jcrossley3, line 184: The `?` is your friend here
```suggestion
                let purl = Purl::from_str(p)?;
```

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +162,141 @@ impl VulnerabilityService {
...
+
+                let version = match purl.version.as_ref() {
+                    Some(v) => v,
+                    None => return Err(Error::Purl(PurlErr::MissingVersion(p.to_string()))),
+                };
```
jcrossley3, line 189: You don't need another variable here. You can just refer to `purl`'s fields when you need them.
```suggestion
                if purl.version.is_none() {
                    return Err(Error::Purl(PurlErr::MissingVersion(p.to_string())));
```

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +162,141 @@ impl VulnerabilityService {
...
+
+                let ns_condition = match purl.namespace.clone() {
```
jcrossley3, line 191: No need for a clone here
```suggestion
                let ns_condition = match &purl.namespace {
```

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +162,141 @@ impl VulnerabilityService {
...
+                };
+
+                let name = purl.name;
+                let ty = purl.ty;
+                let values = [
+                    p.to_string().into(),
+                    name.into(),
+                    ty.into(),
+                    version.to_string().into(),
+                ];
```
jcrossley3, line 213: These extra vars aren't needed
```suggestion
                let values = [
                    p.to_string().into(),
                    purl.name.into(),
                    purl.ty.into(),
                    purl.version.into(),
                ];
```
You also shouldn't have to use `p` here. You should be able to use `purl.to_string().into()` but for a bug that's fixed by #1460 once someone approves it.

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +162,141 @@ impl VulnerabilityService {
...
+        let query = queries
+            .into_iter()
+            .filter_map(|x| match x {
+                Ok(r) => Some(r),
+                Err(e) => {
+                    errors.push(e);
+                    None
+                }
+            })
+            .collect::<Vec<String>>()
```
jcrossley3, line 245:
```suggestion
            .collect::<Result<Vec<_>,_>>()?
```
This is why I renamed `queries` to `query` above. `collect` is powerful. It can turn a vector of results into a result of vectors. And that `?` is what's returning the first error, if any. I'm using [this idiom](https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect) here.

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +162,141 @@ impl VulnerabilityService {
...
+        if !errors.is_empty() {
+            return Err(Error::BadRequest(format!(
+                "Invalid purls received: [{}]",
+                errors
+                    .iter()
+                    .map(|e| e.to_string())
+                    .collect::<Vec<_>>()
+                    .join(", ")
+            )));
+        }
```
jcrossley3, line 257: I don't love constructing errors like this in the code. I'd rather use macros in the Error struct, but in this case, I think the `BadRequest` is a little overkill. I think the common case is that all the purl's will be valid, so it's fine to return one invalid one at a time. If you think it's more common than I think, and the caller would prefer to handle all at once rather than one at a time, you can leave it, but construct the mutable errors array at the top of the fn and use map_err to fill it like the pattern I linked to above, please.
```suggestion
```

modules/fundamental/src/vulnerability/service/test.rs
```diff
@@ -423,3 +424,58 @@ async fn vulnerability_queries(ctx: &TrustifyContext) -> Result<(), anyhow::Erro
...
+    let fail = ["this is not valid", "pkg:npm/missing.version"];
+
+    let result = service.analyze_purls(fail.to_vec(), &ctx.db).await;
+
+    assert!(result.is_err());
+
+    match result {
+        Err(Error::BadRequest(msg)) => assert!(msg.contains("Invalid purls received")),
+        _ => panic!("Expected BadRequest error"),
+    }
```
jcrossley3, line 449:
```suggestion
    assert!(result.is_err());
```
I don't love verifying error strings in tests. Feels brittle to me. For what we're doing, I think `.is_err()` is enough.

modules/fundamental/src/vulnerability/service/test.rs
```diff
@@ -423,3 +424,58 @@ async fn vulnerability_queries(ctx: &TrustifyContext) -> Result<(), anyhow::Erro
...
+#[test(actix_web::test)]
+async fn analyze_purls(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
+    let service = VulnerabilityService::new();
+
+    ctx.ingest_documents(["osv/RUSTSEC-2021-0079.json", "osv/GHSA-2ccf-ffrj-m4qw.json"])
+        .await?;
+
+    let result = service.analyze_purls(vec![], &ctx.db).await?;
+
+    assert!(result.is_empty());
```
jcrossley3, line 438: Good test. It'll affirm it's fine to remove that guard I mentioned above.

jcrossley3 (changes requested):
We're getting there!

Remember release notes are taken from commit messages, not PR titles, so you may still want to reword that "analize" commit. Or not. It's funny. :smile:

ruromero:
What a review! Thanks a lot @jcrossley3 
I have squashed the commits and updated the commit message. No more analize.

modules/fundamental/src/vulnerability/service/mod.rs
jcrossley3: You should be able to do anything i can do. Where did you move `purl.ty` and why? could you have used a reference instead?

jcrossley3: If you're still assigning`ty = purl.ty` above that `values` assignment, then yeah, that'll move it and cause that error. But if you delete those ty/name lets, you should be good.

modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -155,6 +162,105 @@ impl VulnerabilityService {
...
+                    ),
+                };
+
+                let name = purl.name;
+                let ty = purl.ty;
+                let values = [
+                    p.to_string().into(),
+                    name.into(),
+                    ty.into(),
+                    purl.version.ok_or_else(|| PurlErr::MissingVersion(p.to_string()))?.into(),
```
jcrossley3, line 200: nice!

modules/fundamental/src/vulnerability/service/mod.rs
ruromero: Noob's mistake. Apologies.

jcrossley3 (approved):
Thanks for your patience!

Issue #1453, opened 2025-03-21
Give some real thought to our API


jcrossley3:
Parts of our API are confusing. For example, having "versioned purl" and "base purl" in there seems to reveal internal implementation details that will only confuse consumers of the API.

I propose we figure out which API's are being used by the UX and delete all the rest. And going forward, we raise the bar for determining whether a new endpoint should be added.

helio-frota:
> I propose we figure out which API's are being used by the UX and delete all the rest. And going forward, we raise the bar for determining whether a new endpoint should be added.

+1

ctron:
I am not sure that this is the right approach. As, to my understanding, Atlas/Camp doesn't use any UI. So if "UX" excludes "people using curl and command line tools", then this approach seems wrong to me.

helio-frota:
>  As, to my understanding, Atlas/Camp doesn't use any UI.

thanks for the extra info 👍 

> So if "UX" excludes "people using curl and command line tools", then this approach seems wrong to me.

yeah +1

PR #1467, opened 2025-03-24
clean up the api endpoint namespace a bit

jcrossley3:
relates #1453

Moves license-related endpoints beneath "licences"

Collects admin-related endpoints beneath "admin"

As most of these are currently unused, I think, we can move them around as the UX coagulates.

### DIFF
--- a/modules/fundamental/src/license/endpoints/spdx.rs
+++ b/modules/fundamental/src/license/endpoints/spdx.rs
@@ -8,7 +8,7 @@ use trustify_common::{
 };
 
 #[utoipa::path(
-    tag = "licenses",
+    tag = "license",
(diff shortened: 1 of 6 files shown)


ctron:
I noticed two things:

* we using the singular for e.g. advisory, so I'd expect a singular of `license` too
* User preferences are not really "admin" stuff. But, "user" stuff.

jcrossley3:
> * User preferences are not really "admin" stuff. But, "user" stuff.

What's a better name than "admin" for both user prefs and organization? They seem related.

ctron:
Not sure if they belong into the same box. Organizations are entities we extract for the whole "product" thing. While "user" is really "logged in user" stuff.

PR #1472, opened 2025-03-25
ci: add branch cherry-pick automation

carlosthe19916:
This PR should connect this repository to the release-tools cherry-pick automation across branches.

Generally speaking this is how it should work:
- Every PR that needs to be backport its changes to another branch e.g. `release/0.2.z.` needs to have a label indicating the desire to do so. 

Setting this PR as draft just because I'd like double-check the status of the release-tools repository. And set exact instructions and avoid confusion

### DIFF
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -1,0 +1,17 @@
+name: backport
+
+on:
+  pull_request_target:
+    branches:
+      - main
+    types:
+      - closed
+
+jobs:
+  cherry_pick_job:
+    permissions:
+      pull-requests: write
+      contents: write
+    if: github.event.pull_request.merged == true
+    secrets: inherit
+    uses: trustification/release-tools/.github/workflows/backport.yaml@main


carlosthe19916:
Ok, so I am ready to share what we have now:

- If this PR is merged, then the backport feature will be handled by https://github.com/trustification/release-tools
- Each PR that needs to be backported need to have the label `cherry-pick/{branch_name}`. 
  - The user can choose whether to do "squash and merge" or "Create a merge commit" and all commits will be included in the backported PR.
  - The original title of the PR will be reused but the backported PR will have a suffix `[Backport {branch_name}]`

You can see an example of a backport PR with multiple commits here https://github.com/trustification/release-tools/pull/31

Things to be aware of:
- Depending of the merge strategy we choose we might or might not generate conflicts while doing the backports.
  - In case the tool cannot generate a backport for any reason (e.g. conflicts) then a message/comment will always be added to the PR saying something like "We could not create a backport do it manually"
  - The backport feature has nothing to do with PR title, commit prefix validation. That is another topic and feature the release-tools repository has but it is independent from the backport feature.


### Things to customize

@ctron expressed his preference to change the label "cherry-pick/{branch_name}" by "backport {branch_name}". We can certainly do it, we just need to agree on it. And since this is not the only repository that makes use of the release-tools repo, the other repositories will need to follow the same pattern.


### Additional things
The release tools has some other tools that can be reused like:

- Issue automation: we need to opt-in manually just like in this PR
- generation of same labels across repos described at https://github.com/trustification/release-tools/blob/main/pkg/config/config.yaml . Each repo also needs to opt-in.
- PR title verification. 
 
These are things we might be interested in the future but lest's leave it for when we need it.

ctron:
As the whole initial comment of the issue (which I very much like) uses "backport", even in the title of PRs … I think using "backport" as a label makes more sense. Otherwise we'll just create yet another confusing situation.

Alternatively, we could change "backport" to "cherry-pick" in all those locations. However, as Carlos mentioned, I'd prefer "backport". 

Still, it should be aligned.

carlosthe19916:
My apologies for the very late reply on this PR.
@ctron The https://github.com/trustification/release-tools repository has been updated and as you requested the label for backporting is: `backport {branch_name}`. E.g. `backport release/0.2.z`.

The UI Repository will also use this same pattern to backport changes across branches.

We can always keep fine tuning the backport workflows or disable it in case it does not match expectations.

.github/workflows/pr-closed.yaml
```diff
@@ -0,0 +1,17 @@
...
+      - closed
+
+jobs:
+  cherry_pick_job:
+    permissions:
+      pull-requests: write
+      contents: write
+    if: github.event.pull_request.merged == true
+    secrets: inherit
+    uses: trustification/release-tools/.github/workflows/cherry-pick.yaml@main
```
ctron, line 17: Can we rename that too to `backport` to keep it aligned?

carlosthe19916: Done. The workflow is not named `backport.yaml` and the name of it is `name: backport`

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,17 @@
...
+      - closed
+
+jobs:
+  cherry_pick_job:
+    permissions:
+      pull-requests: write
+      contents: write
+    if: github.event.pull_request.merged == true
+    secrets: inherit
+    uses: trustification/release-tools/.github/workflows/cherry-pick.yaml@main
```
ctron, line 17: I can still see the name `cherry-pick.yaml` here. Am I missing something?

ctron (approved):

.github/workflows/backport.yaml
carlosthe19916: Sorry, I missed that.

carlosthe19916:
@ctron this PR has not been merged yet. I didn't want to merge PRs without permission.
Seems it is safe to merge this one now.

PR #1479, opened 2025-03-26
refactor: common code for criterion benches

### LINKED ISSUE #1354: Improve the use of criterion bench
@chirino has created a sort of micro-framework that not only applies the benchmark but also handles database setup and cleanup based on the existing test infrastructure that uses `trustify_test_context`.

Currently, when creating a new benchmark, we end up duplicating parts of the micro-framework's structure.

It would be nice to adopt this [suggestion](https://github.com/trustification/trustify/pull/1329#pullrequestreview-2636264060) of having a common code for it.

helio-frota:
Closes #1354

### DIFF
--- a/modules/fundamental/benches/bench-otel.rs
+++ b/modules/fundamental/benches/bench-otel.rs
@@ -2,29 +2,16 @@ use std::time::Duration;
 
 use criterion::{Criterion, criterion_group, criterion_main};
 
+mod common;
+
 #[allow(clippy::expect_used, clippy::unwrap_used)]
 pub(crate) mod trustify_benches {
-    use std::io::Error;
-    use std::ops::Add;
-    use std::str::FromStr;
-    use std::sync::Arc;
     use std::time::{Duration, Instant};
 
-    use bytes::Bytes;
-    use cpe::cpe::Cpe;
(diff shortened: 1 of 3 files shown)


helio-frota:
> [!NOTE]
> Current bench tests are not working anymore see bellow

```console
     Running benches/bench.rs (target/release/deps/bench-fcbaae3b8575d960)
Gnuplot not found, using plotters backend
Benchmarking ingestion_csaf: Warming up for 3.0000 s
thread 'main' panicked at modules/fundamental/benches/bench.rs:50:34:
ingest ok: Graph(Database(Custom("document vanished"))) <---------------------------------------------
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

error: bench failed, to rerun pass `-p trustify-module-fundamental --bench bench`
➜  trustify git:(main) <---------------------------------------------
```

Maybe something related with [this commit](https://github.com/trustification/trustify/commit/325b0fcb0d4e9cfab923a751a4e5781c7e3c8774) and that means we need to change the bench code to avoid ingesting the same document multiple times. This can be done in a different pull request later, as this one is extracting common code to a common place.

ctron (approved):

Issue #1374, opened 2025-02-27
Null severity in advisory results in "None" severity in vulnerability


jcrossley3:
https://github.com/trustification/trustify/issues/1362#issuecomment-2688066647

PR #1480, opened 2025-03-26
refactor: prelculate advisory and vulnerability average scores and severities

dejanb:
This commit introduces pre-calculated averages for scores and severities. The averages are pre-calculated during ingestion. The change simplifies querying significantly (along with performances) as there's no need to do recalculation on every query and often in the loop for a large number of entities.

1. It add average rows in the vulnerability and advisory tables
2. It defines two PLPGSQL functions that can be used to pre-calculate one or all entries
3. Modifies queries and services to remove query-time calculations

### Ingestion

Currently, there's a trigger that will update advisory and vulnerability each time the cvss3 score is inserted. Volume of data here is not that big ~50K even for the load dataset. Updating all entries takes locally around 10 sec
My main concern was if this will impact the ingestion. In the little testing that I have done, I didn't see any changes in ingestion time during this change.
In theory we can move this to the application logic and recalculate all after ingestion, but I wound't like to do it if it is not necessary.

### Querying

I ran load test locally and things looks promising in general

image

I saw even better improvements on the most problematic advisory endpoint in the previous runs. And although I didn't run the latest version of the scale tests that test sbom/advisory endpoint, manually testing shows at least 50% performance improvements.

Let's see what the automated load testing will show.

There's still more things to change here as some of the queries are still can be improved.

### DIFF
--- a/modules/fundamental/src/advisory/model/details/advisory_vulnerability.rs
+++ b/modules/fundamental/src/advisory/model/details/advisory_vulnerability.rs
@@ -2,8 +2,8 @@ use crate::{Error, vulnerability::model::VulnerabilityHead};
 use sea_orm::{ColumnTrait, ConnectionTrait, EntityTrait, LoaderTrait, QueryFilter};
 use serde::{Deserialize, Serialize};
 use trustify_common::memo::Memo;
+use trustify_cvss::cvss3::Cvss3Base;
 use trustify_cvss::cvss3::severity::Severity;
-use trustify_cvss::{cvss3::Cvss3Base, cvss3::score::Score};
 use trustify_entity::{advisory, advisory_vulnerability, cvss3, vulnerability};
 use utoipa::ToSchema;
 
@@ -34,14 +34,6 @@ impl AdvisoryVulnerabilityHead {
         vulnerability: &vulnerability::Model,
         tx: &C,
     ) -> Result<Self, Error> {
-        let cvss3 = cvss3::Entity::find()
-            .filter(cvss3::Column::AdvisoryId.eq(advisory.id))
-            .filter(cvss3::Column::VulnerabilityId.eq(&vulnerability.id))
-            .all(tx)
-            .await?;
-
-        let score = Score::from_iter(cvss3.iter().map(Cvss3Base::from));
-
         let advisory_vuln = advisory_vulnerability::Entity::find()
             .filter(advisory_vulnerability::Column::AdvisoryId.eq(advisory.id))
             .filter(advisory_vulnerability::Column::VulnerabilityId.eq(&vulnerability.id))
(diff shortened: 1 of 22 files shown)


JimFuller-RedHat:
before I review this - I just want to state that adding triggers at this stage is risky ... event driven handling in databases introduces a next level of complexity - would rather avoid that. Is there any way to do this on ingestion instead of using triggers ?

modules/fundamental/src/advisory/model/details/advisory_vulnerability.rs
```diff
@@ -57,8 +49,11 @@ impl AdvisoryVulnerabilityHead {
             };
             Ok(AdvisoryVulnerabilityHead {
                 head,
-                severity: score.severity(),
-                score: score.value(),
+                severity: advisory
+                    .average_severity
+                    .map(|sev| sev.into())
+                    .unwrap_or(Severity::None),
+                score: advisory.average_score.unwrap_or(0.0),
```
jcrossley3, line 56: This is actually a bug: #1374 

Can we just change these to Options in `AdvisoryVulnerabilityHead` and kill 2 birds with 1 stone? :smile:

jcrossley3 (commented):
I vote for doing this at ingestion rather than a trigger, too.

dejanb:
@JimFuller-RedHat  @jcrossley3 Yeah, trigger was the easiest way to get things going. I will rework it to do it from the code.

modules/fundamental/src/advisory/model/details/advisory_vulnerability.rs
dejanb: Yes. It bothered me as well, but didn't want to change API on my own. I can include it as well.

dejanb:
> As I just added a bunch of tests to the scale test, could you re-run that again, with a recent checkout of the scale-testing repository?

@ctron I had some issues running the latest changes in scale-testing, both errors for the analysis tests and querying the database for large sboms. So, I reverted to the old one, just to get the first feeling of the performance. It'd be good to be able to run the action against PR as we discussed before.

dejanb:
@ctron: I got 

```
Error: error returned from database: column "num" does not exist

Caused by:
    column "num" does not exist
```

But I didn't look too much into it, so I might be doing something weird.
Another nit: having `DATABASE_URL` to be mandatory (even for trying to get `--help`) is a counter-intuitive

ctron:
> @ctron: I got
> 
> ```
> Error: error returned from database: column "num" does not exist
> 
> Caused by:
>     column "num" does not exist
> ```

Ah, you might want to try latest `main` for that.

> 
> But I didn't look too much into it, so I might be doing something weird. Another nit: having `DATABASE_URL` to be mandatory (even for trying to get `--help`) is a counter-intuitive

Yea, that's a problem I'm not sure we can easily solve. Goose uses it's own argument system. That's why we rely on env-vars for rest. `--help` comes from that arg system. So it's processes only after we processed our stuff. Not sure what's an easy way out of this.

We forked Goose already, maybe migrating to clap is an idea. But that's expensive.

modules/fundamental/src/advisory/model/details/advisory_vulnerability.rs
jcrossley3: Not sure when this is getting merged, so went ahead and pushed #1591

PR #1481, opened 2025-03-27
chore: simplify the internal API, add instrumentation

ctron:
This drops a bunch of "async" functions and Result returns, which aren't being used anyway.

It also adds a bit more instrumentation, and some test tooling, allowing one to enable flamechart traces for tests.


modules/fundamental/tests/issues/oom.rs
```diff
@@ -2,22 +2,22 @@
...
-use tracing::instrument;
 use trustify_common::id::Id;
 use trustify_module_fundamental::sbom::service::SbomService;
-use trustify_test_context::TrustifyContext;
+use trustify_test_context::{TrustifyContext, flame::setup_global_subscriber};
 
 #[test_context(TrustifyContext, skip_teardown)]
-#[test(tokio::test)]
-#[instrument]
+#[tokio::test]
```
JimFuller-RedHat, line 10: just wondering out load if we could make test dependent on dataset (maybe with feature enabling via cfg annotation)

JimFuller-RedHat (approved):
LGTM

helio-frota (approved):
approving related to unused async part thanks for removing them !

modules/fundamental/tests/issues/oom.rs
ctron: Yea, that would be nice.

PR #1420, opened 2025-03-13
perf: change cvss related pg funcs to immutable parallel safe and a few key indexes

JimFuller-RedHat:
This PR should help advisory/vuln endpoints by enabling some pg functions to run in parallel. Also a few key indexes are added.

PR #1482, opened 2025-03-27
chore: add back advisory_vuln index

JimFuller-RedHat:
adding back index originally set in #1420


ctron (approved):
LTGM. But it might be worth running the scale tests

PR #1490, opened 2025-03-29
chore: refine analysis productA/productB>component

JimFuller-RedHat:


ctron (approved):
Looks good. I don't think we have covered anything in that area with scale tests, but I'll run them anyway. Maybe wait merging until they finished.

PR #1491, opened 2025-03-31
let sbom_package_license match clearyly defined

bxf12315:


modules/ingestor/src/graph/sbom/clearly_defined.rs
```diff
@@ -1,55 +1,62 @@
...
+use uuid::Uuid;
 
 impl SbomContext {
     #[instrument(skip(db, curation), err)]
     pub async fn ingest_clearly_defined_curation<C: ConnectionTrait>(
         &self,
         curation: Curation,
         db: &C,
+        sbom_id: Uuid,
+        node_id: String,
```
ctron, line 19: To my understanding clearly defined doesn't have a real "node ID", right? If that's the case, then we could use a static/magic/const value instead. Like we do it with the CyclondeDX root node. In the context of the "SBOM", that would always be unique.

bxf12315: It was based on this code that I started using the document ID as the node ID.
https://github.com/trustification/trustify/blob/0224bef3af458bca3be3ce30763915dd0d6789f3/modules/ingestor/src/graph/sbom/clearly_defined.rs#L92

bxf12315: I believe it is also acceptable, but it would be better to use the same method in the same place.

ctron: What's the benefit?

bxf12315: It is strange to use two different node id definitions in the same clear-defined .

bxf12315: https://github.com/trustification/trustify/blob/0224bef3af458bca3be3ce30763915dd0d6789f3/modules/ingestor/src/graph/sbom/clearly_defined.rs#L92

bxf12315: At this point, the document incorporates the content of the coordinates from the curation, such as "crate/cratesio/-/chrono." In my view, this approach is more meaningful in practice.

ctron: The "node id" is an unique ID in the context of an SBOM. Having a static node id makes it predictable. As the value itself is not important, it wouldn't matter.

Unless there are reasons against it, I'd like to use a static value.

bxf12315: What static value should we use? “ClearlyDefinedCuration”.

ctron: Sounds reasonable.

modules/ingestor/src/graph/sbom/clearly_defined.rs
```diff
@@ -1,55 +1,61 @@
...
 
-            // assertions.push(purl_license_assertion::ActiveModel {
-            //     id: Default::default(),
-            //     license_id: Set(license_info.uuid()),
-            //     versioned_purl_id: Set(purl.version_uuid()),
-            //     sbom_id: Set(self.sbom.sbom_id),
-            // });
+            sbom_package_license_list.push(sbom_package_license::ActiveModel {
+                sbom_id: Set(self.sbom.sbom_id),
+                node_id: Set("ClearlyDefinedCuration".to_string()),
```
ctron, line 34: Can you pull that out into a `const` please.

modules/ingestor/src/graph/sbom/clearly_defined.rs
```diff
@@ -1,55 +1,61 @@
...
     pub async fn ingest_clearly_defined_curation<C: ConnectionTrait>(
         &self,
         curation: Curation,
         db: &C,
+        sbom_id: Uuid,
     ) -> Result<(), anyhow::Error> {
         let mut purls = PurlCreator::new();
         let mut licenses = LicenseCreator::new();
 
         // TODO: Since the node id cannot be obtained here, it’s not possible to replace purl_license_assertion with sbom_package_license.
```
ctron, line 23: Is that comment still valid?

modules/ingestor/src/graph/sbom/clearly_defined.rs
```diff
@@ -1,55 +1,61 @@
...
-        // let mut assertions = Vec::new();
+        let mut sbom_package_license_list = Vec::new();
 
+        let _a: SbomInformation = (&curation).into();
```
ctron, line 26: That seems to be dead code.

Issue #1323, opened 2025-02-20
Would be good to make sure we are using Cargo-feature-inheritance properly


helio-frota:
We add and remove dependencies over time, but the reality is that we have no clear insight into whether:

* We are correctly propagating features to sub-crates ?
  * If not, does this impact build performance and/or binary size ?
* How can we ensure everything is in order ?
  * We can probably make use of `cargo tree --depth 2 -e features` to compare old and new results [?]

helio-frota:
maybe https://github.com/taiki-e/cargo-hack can help us on this task

helio-frota:
@chirino I found this error when running `cargo-hack` but I have no idea if that is `wip` or something

```
cargo hack check --each-feature --no-dev-deps
info: --no-dev-deps modifies real `Cargo.toml` while cargo-hack is running and restores it when finished
info: running `cargo check --no-default-features` on trustify-common (1/38)
error: feature `ai` includes `trustify-common/ai`, but `trustify-common` is not a dependency
  --> migration/Cargo.toml:13:6
   |
13 | ai = ["trustify-common/ai"]
   |      ^^^^^^^^^^^^^^^^^^^^^^
```

helio-frota:
> If not, does this impact build performance and/or binary size ?

super basic test with this [repo](https://github.com/helio-frota/cascade-feat) which contains only `clap` as dependency, shows that the binary size continues the same but we have a bit ~`0.2` seconds improvement when disabling `default-features`.

```
clap = "4"
Total units:	19
Max concurrency:	10 (jobs=12 ncpu=12)
Build start:	2025-02-24T20:46:42Z
Total time:	2.6s
-----
clap = { version = "4", default-features = false, features = ["std"] }
Total units:	12
Max concurrency:	5 (jobs=12 ncpu=12)
Build start:	2025-02-24T20:49:58Z
Total time:	2.4s
```

helio-frota:
Noticed that we can skip [this](https://github.com/trustification/trustify/issues/1323#issuecomment-2679951735) with `cargo hack check --each-feature --skip ai` 👍

helio-frota:
@i386x  I think we can change that initial confusing issue description to:

_Can we use `default-features = false` on all dependencies and ensure that everything still works? If not, we need to add the exact feature(s) in the dependency declaration to compile._

I found a repo which contains a good amount of workspace members and it is using this technique:
https://github.com/kpreid/all-is-cubes/blob/main/Cargo.toml

👍

[...]

PR #1498, opened 2025-03-31
build: saves ~11 seconds build time

helio-frota:
Please let me know with :+1:  or :-1: or PR review rejection/approval if this worth to have a bit verbose `Cargo.toml` to save ~15 seconds during build.

 More details you can see here: https://github.com/trustification/trustify/issues/1323#issuecomment-2766839176

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -532,12 +532,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ascii_utils"
-version = "0.9.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "71938f30533e4d95a6d17aa530939da3842c2ab6f4f84b9dae68447e4129f74a"
-
-[[package]]
 name = "assert-json-diff"
 version = "2.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


helio-frota:
Update: now 170.3s (2m 50.3s) `~11 seconds` got tests broken , interestingly the build pass and we have a runtime error in tests...

JimFuller-RedHat:
I would first fix tests and whatever runtime error (maybe good to raise issue for each as well).

helio-frota:
thanks for the feedback folks

> I would first fix tests and whatever runtime error (maybe good to raise issue for each as well).

yeah that was fixed yesterday. It was caused by:

```diff
- git2 = { version = "0.20.0", default-features = false, features = ["ssh"] }
+ git2 = { version = "0.20.0", features = ["ssh"] }
```
probably related to the missing [https feature](https://github.com/rust-lang/git2-rs/blob/master/Cargo.toml#L38) and the runtime error in tests fixed yesterday removing `default-features = false`

> I'm not sure saving 11 seconds build time really is that important, adding that extra complexity of maintaining all those feature flags.

yeah that's why I :-1: 

> On the other side, it may be more for a --release build. And it may have a positive impact on the release binary? I think those things should be checked as well.

We have the same result without `--release`

Before:


After:


@JimFuller-RedHat @ctron let's close the the PR and the issue then?

helio-frota:
ok closing :+1: 

```
➜  trustify git:(issue-1323) ls -sh target/release/trustd
114M target/release/trustd*
➜  trustify git:(issue-1323) cargo build --no-default-features --release
```

PR #1500, opened 2025-04-01
fix: set title for vunerabilities with non typical description language code

### LINKED ISSUE #1495: CVE lang codes
We assume `en` to be the code for english description in CVE files and then use them as default titles and description of vulnerabilities.

However, there are cases when subvariants with regions are used as well

| lang  | count|
|-------|--------|
| en    | 280481|
| de    |   5738|
| en-US |   5613|
| en_US |     46|

For example, 
https://github.com/CVEProject/cvelistV5/blob/f2162c4dfd357958aca89723656fdece7abcd123/cves/2022/29xxx/CVE-2022-29123.json#L233

is one such case.
Unfortunately, the `en_US` is non-standard but it's used in only a small set of rejected vulnerabilities. For these vulnerabilities titles will ne be shown.

dejanb:
Fixes #1495

I opted for a simple `starts_with` check instead of using BCP47 parser, as it would work anyways for the non standard `en_US` cases. We can improve this in the future if needed.

I also added missing CVE files for DS3

### DIFF
--- a/modules/ingestor/src/service/advisory/cve/loader.rs
+++ b/modules/ingestor/src/service/advisory/cve/loader.rs
@@ -33,8 +33,6 @@ pub struct CveLoader<'g> {
     graph: &'g Graph,
 }
 
-const DESCRIPTION_EN: &str = "en";
-
 impl<'g> CveLoader<'g> {
     pub fn new(graph: &'g Graph) -> Self {
         Self { graph }
@@ -80,7 +78,8 @@ impl<'g> CveLoader<'g> {
             .ingest_vulnerability(id, information, &tx)
             .await?;
 
-        let (entries, english_description) = Self::build_descriptions(descriptions);
+        let entries = Self::build_descriptions(descriptions);
+        let english_description = Self::find_best_description_for_title(descriptions);
 
         let advisory = self
             .graph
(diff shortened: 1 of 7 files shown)


ctron:
I'm not sure this is the right approach. I think we should follow the approach of finding the "best match", which be would: `en-US` first, then `en`, then default.

dejanb:
I reworked the code to prioritize finding `en-US` and `en_US` before `en`, although there's no such cases in data yet. All vulnerabilities that have multiple descriptions are [`en`, `de`] combination.

modules/ingestor/src/service/advisory/cve/loader.rs
```diff
@@ -177,28 +176,21 @@ impl<'g> CveLoader<'g> {
...
     }
 
     /// Quicker version to find the best description as an alternative when not having a title.
     fn find_best_description_for_title(descriptions: &[Description]) -> Option<&str> {
-        // Currently, we simply choose the first english description.
         descriptions
             .iter()
-            .find(|desc| desc.language == DESCRIPTION_EN)
-            .map(|desc| &*desc.value)
+            .find(|desc| matches!(desc.language.as_str(), "en-US" | "en_US"))
```
ctron, line 191: I love your use of `matches` here :)

Issue #1440, opened 2025-03-18
get_purl incorrectly assumes all purls have namespaces


jcrossley3:
https://github.com/trustification/trustify/blob/772b053b8f7e40682d7bd9e7657fa9a845aebbde/migration/src/m0000010_init_up.sql#L637

`get_purl` returns 2 slashes for purls without namespaces, e.g.

`pkg:cargo/rand@0.7.2`

becomes

`pkg:cargo//rand@0.7.2`

jcrossley3:
I reverted the fix for this thinking we'd soon get rid of `get_purl`.

In addition to the double slashes, get_purl also fails to encode names, namespaces, and versions properly.

I think we'd all be happier if we replace `get_purl` with a redundant field in `qualified_purl` that is the properly string-ified representation of the corresponding json blob in that table, populated either from ingested docs or generated by `common::Purl::to_string()`, which handles missing namespaces and encoding correctly.

JimFuller-RedHat:
adding a purl_string that materialises the purl in all its glory is good ... though lets not drop the jsonb ... or put another way we want to somehow refactor our way out of the three table stuff and I think the jsonb will probably be needed.

jcrossley3:
No intention of deleting the jsonb -- that's what I meant to imply by the word, "redundant". Is it ok if I rename it?

PR #1503, opened 2025-04-01
fix: get_purl now deals with missing ns properly

### LINKED ISSUE #1456: allow filtering component analysis by base purl
In a typical Product Security workflow a security analyst would look for base purls matching a vulnerability report, then try to find all the (latest) products with that base purl. In order to support that work flow we should be able to filter by base_purl and arch.

Today we can use a partial match on purl to do a similar thing, but it returns too many results, which then need to be filtered out by the search client. It would be more efficient to only return matching base purls and archs.

For example:

`api/v2/analysis/component?q=purl~pkg:rpm/redhat/webkit2gtk3`

Returns the expected result, like:

`pkg:rpm/redhat/webkit2gtk`

but also purls other packages with other trailing characters, (binary RPMS) eg:

`pkg:rpm/redhat/webkit2gtk3-debuginfo`

Even if the base_purl matched exactly, there are often multiple arch for the same result, eg:

`pkg:rpm/redhat/webkit2gtk3@2.42.5-1.el9?arch=src`
`pkg:rpm/redhat/webkit2gtk3@2.42.5-1.el9?arch=aarch64`

so being able to filter by arch would also be nice:

`api/v2/analysis/component?q=purl~pkg:rpm/redhat/webkit2gtk3&arch=src`

In fact, being able to filter to arch would probably give us what we need most of the time, even without filtering by base_purl.

jcrossley3:
Fixes #1456

We're gonna need get_purl to work properly to migrate the data for #1440 anyway.

### DIFF
--- a/modules/analysis/src/endpoints/tests/mod.rs
+++ b/modules/analysis/src/endpoints/tests/mod.rs
@@ -10,6 +10,7 @@ use serde_json::{Value, json};
 use test_context::test_context;
 use test_log::test;
 use trustify_test_context::{TrustifyContext, call::CallService, subset::ContainsSubset};
+use urlencoding::encode;
 
 #[test_context(TrustifyContext)]
 #[test(actix_web::test)]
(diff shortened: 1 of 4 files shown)


modules/analysis/src/endpoints/tests/mod.rs
```diff
@@ -422,6 +419,38 @@ async fn find_component_by_query(ctx: &TrustifyContext) -> Result<(), anyhow::Er
...
+    ] {
+        assert!(query(each).await.contains_subset(json!({
+            "items": [{
+                "purl": [ PURL ],
+            }]
+        })));
+    }
+
+    Ok(())
+}
```
jcrossley3, line 452: @jasinner thanks for the unit test! :smile:

JimFuller-RedHat (approved):
LGTM!

PR #1512, opened 2025-04-02
fix: TC-2388 OSV range with last_affected

mrizzi:
https://issues.redhat.com/browse/TC-2388

OSV `last_affected` field from `ranges` events management.

### DIFF
--- a/modules/ingestor/src/service/advisory/osv/loader.rs
+++ b/modules/ingestor/src/service/advisory/osv/loader.rs
@@ -433,13 +433,21 @@ async fn create_package_status<C: ConnectionTrait>(
             Version::Inclusive(start.clone()),
             Version::Unbounded,
         )),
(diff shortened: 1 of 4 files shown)


modules/ingestor/src/service/advisory/osv/loader.rs
```diff
@@ -459,7 +467,7 @@ async fn create_package_status<C: ConnectionTrait>(
             .await?;
     }
 
-    if let (_, Some(fixed)) = &parsed_range {
+    if let (_, Some((fixed, false))) = &parsed_range {
```
dejanb, line 470: Are we ingesting both "affected" and "fixed" statuses every time we find high version? This is not a topic for this PR, but I think we should rethink this in future as it seems like redundant information.

dejanb (approved):
Looks good. I have only one comment regarding statuses that we might consider in a separate issue.

PR #1517, opened 2025-04-03
fix: mavenver_cmp function fail to compare versions with build numbers properly

dejanb:
In Maven version scheme, the version with qualifiers are considered lesser than similar ones without a qualifier, e.g.

1.2.3-alpha < 1.2.3

But in case, the version have a "build number" field, the qualifier should mean a higher version, e.g.

4.1.108.Final > 4.1.108.Final-redhat-0001

In this case, the "Final" is the "build number" and that means that qualifiers should be treated as post release version (higher).


mrizzi (approved):
@dejanb thanks a lot, really well done 👏

PR #1523, opened 2025-04-04
fix: pythonver_cmp function to properly handle local versions

dejanb:
Current implementation didn't properly handled local version of python packages, like `pkg:pypi/torch@2.0.1+cu118`

https://peps.python.org/pep-0440/#local-version-identifiers

I think we need more work on this function to handle all the possible cases, but I'll raise a separate issue for that.


mrizzi (approved):
@dejanb thanks a lot, it looks great!

ctron:
We merge it anyway. And I'll try to see if that function is in the code path of `search_purls`. If it is not, then it's not a risk to backport this to the release branch.

PR #1524, opened 2025-04-06
fix: reap crashed jobs so they don't appear "stuck" in the UI

### LINKED ISSUE #1499: Importer job stuck
It looks like the job cleanup doesn't work any more. I'm sure it needs some more investigation, but I just noticed that killing the pod during an import leads the following situation:

Image

The import is now running `osv-github` and `cve` seems stuck.

jcrossley3:
Fixes #1499

### DIFF
--- a/modules/importer/src/server/mod.rs
+++ b/modules/importer/src/server/mod.rs
@@ -2,7 +2,7 @@ pub mod context;
 pub(crate) mod progress;
 
 use crate::{
-    model::Importer,
+    model::{Importer, State},
     runner::{
         ImportRunner,
         common::heartbeat::Heart,
@@ -99,11 +99,14 @@ impl Server {
             // Update metrics
             running_importers.record(count as _, &[]);
 
+            let importers = service.list().await?;
+
+            // Update any importers that we assume have crashed
+            reap(&importers, &service).await?;
+
             // Asynchronously fire off new jobs subject to max concurrency
             runs.extend(
-                service
-                    .list()
-                    .await?
+                importers
                     .into_iter()
                     .filter(|i| i.is_enabled() && i.is_due() && !i.is_running())
                     .take(self.concurrency - count)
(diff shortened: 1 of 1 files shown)


ctron (approved):
Feels like that's adding a race condition:

* reap
* run (reaped)

If those two steps happen simultaneously with two different processes, but slightly different timings, one might reap the freshly started one.

I'll approve it anyway, if you think that's an issue, I'd be happy to see that fixed before merging.

jcrossley3:
> Feels like that's adding a race condition:
> 
> * reap
> * run (reaped)

I'm not sure I follow. I want to say this line accounts for the race condition: 
```
.filter(|i| i.data.state == State::Running && !i.is_running())
```
Only those jobs in the running state that haven't updated their heartbeat recently (within 2 expected periods) will be reaped.

Can you clarify where the race condition might occur?

ctron:
> > Feels like that's adding a race condition:
> > 
> > * reap
> > * run (reaped)
> 
> I'm not sure I follow. I want to say this line accounts for the race condition:
> 
> ```
> .filter(|i| i.data.state == State::Running && !i.is_running())
> ```
> 
> Only those jobs in the running state that haven't updated their heartbeat recently (within 2 expected periods) will be reaped.
> 
> Can you clarify where the race condition might occur?

You have one job in that state. To pods run their loop. Depending on timing, they could both reap and start the job. Where it might be, that the one process reaps the newly started one.

jcrossley3:
> You have one job in that state. To pods run their loop. Depending on timing, they could both reap and start the job. Where it might be, that the one process reaps the newly started one.

That is theoretically possible, but only one will be able to obtain the optimistic lock.

jcrossley3:
> You're passing in `None` as expected revision. Unless I got that wrong. How does the lock work?

All the [process coordination is handled](https://github.com/trustification/trustify/blob/3458975dea1099dd4bb4c39d29e9a96615d8226a/modules/importer/src/runner/common/heartbeat.rs#L76-L94) with the `HEARTBEAT` column of the `IMPORTER` table. I left the `REVISION` stuff alone as it seemed to be intended for ETag headers, and I wanted a microsecond timestamp for the optimistic lock.

PR #1526, opened 2025-04-07
feat: a new dataset containing a few sboms and osv advisories

dejanb:
These examples showed some issues in our version matching and correlation and could be a good base to create an integration test.


ctron:
Right now, we did number datasets. Having that pattern, I think we should keep using it, or completely switch it. On the pro side of just numbering them, we don't get into issues when when we change content in a way that invalidates the name.

dejanb:
@ctron makes sense. Renamed the directory

ctron (approved):

PR #1528, opened 2025-04-07
chore: update petgraph and drop forked deepsize

ctron:


JimFuller-RedHat (approved):
I recently tested with petgraph 0.8.0 which looked good - and this LGTM

Issue #1444, opened 2025-03-19
Vulnerabilities endpoint is slow with ~500k Advisories ingested


queria:
In Trustify (TPA build) v0.2.10 setup:

- with AWS RDS psql (db.m7g.xlarge, bigger than default .large)
- when having default/recommended importers (from helm charts repo values-importers.yaml) enabled for almost a week
- at the end there is show to be around half a million Advisories ingested

Accessing areas in UI where advisory/vulnerability-advisory information gets involved seems to be very slow (observed can be e.g. 13sec single response from API) or even not processable (gw timeout).

Example URL of what never loads:

- UI: /vulnerabilities/CVE-2021-3620
- -> API: /api/v2/vulnerability/CVE-2021-3620 => HTTP 504 (gw timeout)
- after it failed UI retries the API call 3 more times (total 4) before it gives up and shows the Unable to connect message

This is likely related to #1298 .

—

Configured importers (each had 7 executions before disabling):

![Image](https://github.com/user-attachments/assets/a151364d-22e1-416b-9fe4-3fd413620c60)

Counts of ingested items (as reported in UI Search):

![Image](https://github.com/user-attachments/assets/2ad9d293-a9d1-498a-a84a-439dc29d8d28)

Outcome when attempting to view CVE-2021-3620 Vulnerability details:

![Image](https://github.com/user-attachments/assets/bce4385e-d1da-4712-968f-b51a8dc4224a)

JimFuller-RedHat:
thx for update - FWIW issue with vulnerability is known ( I believe @dejanb is working on it).

helio-frota:
I'm searching more information around this issue (mainly the rust part) 👍 
now getting 87+ seconds, no vacuum analyze  and no reindexing done after import
 
```
➜  ~ curl -w "%{time_total}s\n" -s -o /dev/null localhost:8080/api/v2/vulnerability/CVE-2023-31084
87.581250s
➜  ~ curl -w "%{time_total}s\n" -s -o /dev/null localhost:8080/api/v2/vulnerability/CVE-2023-31084
87.128461s
```

from telemetry part I'll play with adding temporary `instrument` macro to see more functions around this endpoint as atm it shows only `from_advisory`

![Image](https://github.com/user-attachments/assets/5f190df4-c765-4bfd-95e9-76b8da6436a4)

![Image](https://github.com/user-attachments/assets/7f5f80a9-d0dc-48bc-a8c1-d13ff8c7e159)

(I'm running pm-mode)

PR #1533, opened 2025-04-07
chore: lowered otel-collector logs

helio-frota:
level detailed was useful during development but is causing span drops

I wasn't caring about this warning:

```
2025-04-07T19:01:33.028587Z  WARN
get:fetch_vulnerability:from_entity:from_entities:from_query_result:from_query_result_multi_model:
opentelemetry_sdk:  name="BatchSpanProcessor.SpanDroppingStarted"
BatchSpanProcessor dropped a Span due to queue full/internal errors. No
further internal log will be emitted for further drops until Shutdown.
During Shutdown time, a log will be emitted with exact count of total
Spans dropped.
```

Until notice the difference of half the spans being dropped

### DIFF
--- a/etc/telemetry/config-otel-tempo.yaml
+++ b/etc/telemetry/config-otel-tempo.yaml
@@ -12,7 +12,7 @@ exporters:
     tls:
       insecure: true
   debug:
-    verbosity: detailed
+    verbosity: basic
 
 processors:
   batch: {}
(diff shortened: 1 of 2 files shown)


helio-frota:
verbosity -> detailed:
14k



verbosity -> basic:
31k
https://github.com/trustification/trustify/issues/1444#issuecomment-2784205689

ctron (approved):

PR #1534, opened 2025-04-07
Remove `get_purl` stored proc to DRY our PURL encoding rules

### LINKED ISSUE #1440: get_purl incorrectly assumes all purls have namespaces

jcrossley3:
Fixes #1440

We only need to maintain evolving encoding specifications in a single
place now: our Purl object.

Most queries are still supported with only slight changes using the
new "nested field" query support, e.g. `purl:version`,
`qualifiers:respository_url`, etc.


migration/src/m0001030_add_original_purl_column.rs
```diff
@@ -0,0 +1,48 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
```
jcrossley3, line 7: @JimFuller-RedHat should i create/drop indexes for the new `original` column in here? If so, can you give me the exact syntax for each, please?

ctron (changes requested):
Noted a few things I'd like to understand.

I also don't think, the order of qualifiers is a real issue (https://issues.redhat.com/browse/TC-2407). So I'm not pushing for having this behavior. But is also wouldn't not really be "the original" PURL from the document.

migration/src/m0001030_add_original_purl_column.rs
JimFuller-RedHat: if we are doing searches on this column - which include LIKE or partial searches, something like

```
                    DROP INDEX IF EXISTS qualified_purl_original_gist;
                    CREATE INDEX qualified_purl_original_gist ON qualified_purl USING GIST (original gist_trgm_ops);
```

should do it (we cannot represent gist with sea_orm syntax).

modules/ingestor/src/graph/purl/creator.rs
```diff
@@ -67,7 +68,8 @@ impl PurlCreator {
                     id: Set(qualified),
                     versioned_purl_id: Set(version),
                     qualifiers: Set(Qualifiers(purl.qualifiers)),
-                    purl: Set(cp),
+                    canonical: Set(cp),
+                    purl: Set(original),
```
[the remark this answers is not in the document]
jcrossley3: Good point. I'll see if I can't shuffle things around to expose the original.

modules/analysis/src/service/load.rs
```diff
@@ -104,9 +104,11 @@ purl_ref AS (
     SELECT
         sbom_id,
         node_id,
-        array_agg(get_purl(qualified_purl_id)) AS purls
+        array_agg(qualified_purl.purl) AS purls
     FROM
         sbom_package_purl_ref
+    LEFT JOIN
+        qualified_purl ON (sbom_package_purl_ref.qualified_purl_id = qualified_purl.id)
```
jcrossley3, line 111: @ctron @JimFuller-RedHat Do either of you know how I can preserve this logic without either the `get_purl` proc or the stringified purl field?

This is the only place in the codebase that currently uses `get_purl`.

JimFuller-RedHat: I guess this has to be rewritten to join with sbom_package_purl_ref

leave that with me (As subject of a different PR)

jcrossley3: I can't finish this one without doing it myself, if we agree that's the way to go.

ctron: I think you can use the recipe of the `cpe`, from the next few lines.

JimFuller-RedHat: ya agreed then ;)

jcrossley3: I can do that, but if we go that route, half of [these tests](https://github.com/trustification/trustify/blob/0a75ac4fcb56759876c0d4a22c48ab808fbf5bd8/modules/analysis/src/endpoints/tests/mod.rs#L399) will break (any referencing a purl), and I think prodsec is already relying on them.

I'd like to proceed with adding a `purl: String` field to `sbom_package_purl_ref` if y'all are ok with that.

JimFuller-RedHat: maybe I am getting confused ... and a little nervous - we want all the info in the purl, not restricted.

ctron: Why would using the cpe pattern force us dropping any information?

jcrossley3: It doesn't. I'm conflating two issues, sorry. The cpe pattern works fine for removing `get_purl` here. I forgot that we currently use `get_purl` in [another place](https://github.com/trustification/trustify/blob/16d6066d7e2f4ad7ae8a6f54fe94b4171971c3bc/modules/analysis/src/service/load.rs#L295) -- that I'd already removed in this PR -- and removing that is what will break the tests mentioned above.

modules/analysis/src/endpoints/tests/mod.rs
```diff
@@ -398,8 +398,8 @@ async fn find_component_by_query(ctx: &TrustifyContext) -> Result<(), anyhow::Er
     };
 
     for each in [
```
jcrossley3, line 400: @ctron for these tests to pass, the ones including a `purl` field require comparisons against a stringified representation of a PURL, which used to come from the `get_purl` proc but now come from `sbom_package_purl_ref.purl`.

modules/fundamental/src/purl/endpoints/test.rs
```diff
@@ -215,3 +217,36 @@ async fn purl_queries(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
...
+
+    // Because these purl's were created as a result of ingesting an
+    // SBOM, we can successfully filter by purl in our queries
+    for each in [
+        r"purl=pkg:maven/com.redhat.quarkus.platform/quarkus-bom@3.2.11.Final-redhat-00001?repository_url=https://maven.repository.redhat.com/ga/\&type=pom",
+        "purl~pkg:maven/com.redhat.quarkus.platform/quarkus-bom@3.2.11.Final-redhat-00001&purl~type=pom&purl~repository_url=https://maven.repository.redhat.com/ga/",
+        "purl~quarkus-bom@3.2.11.Final-redhat",
+        "purl~quarkus-bom@3.2.11.Final-redhat&purl~type=pom&type=maven", // note the two types
+        "quarkus-bom&purl~type=pom&type=maven", // full text search filtered to match 1
+    ] {
```
jcrossley3, line 247: @ctron same as here. Note they used to be commented out in the above test.

entity/src/sbom_package_purl_ref.rs
```diff
@@ -12,6 +12,8 @@ pub struct Model {
 
     #[sea_orm(primary_key)]
     pub qualified_purl_id: Uuid,
+
+    pub purl: String,
```
ctron, line 16: To my understanding, we decided against adding this. Why is that necessary?

jcrossley3: That's why I mentioned you in the above two comments. It's those 2 unit tests that relied on the `get_purl` stored proc, which I'd like to remove, because I don't want to have to implement evolving encoding specs in two places. Without `get_purl`, we need a stringified representation of the purl in the db for comparisons that prodsec is already using.

modules/analysis/src/endpoints/tests/mod.rs
jcrossley3: @ctron here

modules/fundamental/src/purl/endpoints/test.rs
jcrossley3: @ctron and here

entity/src/sbom_package_purl_ref.rs
ctron: Why aren't we using the pattern of the CPE approach? That doesn't require a stringified version, does it?

jcrossley3: I used the "CPE approach" to get around the `get_purl` usage in the raw sql mentioned in another comment.

But I stopped short of applying the [translation stuff](https://github.com/trustification/trustify/blob/a5ecedad2b5d37b2ee6469400ef1f526a90874be/modules/analysis/src/service/load.rs#L262-L291) to PURL's because I don't think it's very user-friendly. A user is expected to provide a valid CPE, e.g. `OwnedUri::from_str` to query any part of it.

I didn't want to impose the same restriction on PURL's, mostly because we already have internal prodsec people using it. So removing that functionality may break existing external apps.

I could also restore `get_purl`, but I really don't want to.

I felt like the added column was the solution with the least tradeoffs.

ctron: I'm not sure I understand. That code is there to identify the SBOMs which need to be loaded. Shouldn't that be the same code as we already have for searching SBOMs? We do have the PURL in JSON format in the database. Shouldn't that allow for doing a very specific search on parts of the purl? What kind of restriction would there be?

jcrossley3: > I'm not sure I understand. That code is there to identify the SBOMs which need to be loaded. 

What code?

> We do have the PURL in JSON format in the database. Shouldn't that allow for doing a very specific search on parts of the purl? 

Sure. But that's all you can do. You can only search on specific parts of the purl, e.g. name, namespace, version and a few hard-coded qualifiers.

> What kind of restriction would there be?

An inability to search by purl, as is demonstrated in the unit tests I keep referring you to. Any test that includes "purl" in the query string would have to be removed. And prodsec is already searching by "purl".  You seem determined to remove that feature.

ctron: > > I'm not sure I understand. That code is there to identify the SBOMs which need to be loaded.
> 
> What code?
The code you linked to.

> 
> > We do have the PURL in JSON format in the database. Shouldn't that allow for doing a very specific search on parts of the purl?
> 
> Sure. But that's all you can do. You can only search on specific parts of the purl, e.g. name, namespace, version and a few hard-coded qualifiers.

What else (aside from searching all qualifiers) would be necessary?

> 
> > What kind of restriction would there be?
> 
> An inability to search by purl, as is demonstrated in the unit tests I keep referring you to. Any test that includes "purl" in the query string would have to be removed. And prodsec is already searching by "purl". You seem determined to remove that feature.

I still don't understand why it would need to be removed.

jcrossley3: > > An inability to search by purl, as is demonstrated in the unit tests I keep referring you to. Any test that includes "purl" in the query string would have to be removed. And prodsec is already searching by "purl". You seem determined to remove that feature.
> 
> I still don't understand why it would need to be removed.

With the redundant `PURL` column in `SBOM_PACKAGE_PURL_REF`, users can query something like `q=purl~quarkus@3`, because it's just a simple string comparison. Without the `PURL` column, we could add translation logic similar to what we're doing for CPE, but because `quarkus@3` isn't a valid PURL, we would effectively be issuing a full-text search, i.e. `q=quarkus@3`, resulting in an ILIKE clause for every column fetched by the SBOM SELECT statement, including all the JSONB keys for both CPE's and PURL's. That might be fine for `quarkus@3` but it might return way more SBOM's than we'd need for something like `q=purl~redhat`.

jcrossley3: I should mention that `q=quarkus@3` won't return anything, because none of the component fields would contain that string. But `q=redhat` might return a lot, way more that what the user might expect from `q=purl~redhat` because CPE's might match that string instead of PURL's.

The only way we'd be able to match PURL component fields (name, namespace, version, etc) successfully would be to require the user to pass valid purl's as the values in `q=purl{op}{value}` and that, to me, isn't user-friendly.

ctron: Wouldn't string matching also match parts of the PURL that might not be that obvious? Searching for `quarkus` would show up PURLs with `qualifier=some-quarkus-1`, or `quarkus=2.0`? That would be quite unexpected from my point of view.

I'd rather do: `purl:name=quarkus` or `purl:qualifier:my-qualifier=quarkus` instead. Or `purl=<full purl>`. And wouldn't that also solve the issues described above? But be more predictable?

jcrossley3: No

JimFuller-RedHat: actually I believe if we decompose the query purl into its parts and then do eq or LIKE on each ... as the following sql query 
```
                    explain analyze   SELECT id
                       FROM qualified_purl
                       WHERE (qualified_purl.purl->>'ty' = $1
                          AND qualified_purl.purl->>'name' LIKE $2
                          AND qualified_purl.purl->>'namespace' = $3)
                     )
```
we get the advantage of using existing indexes  (e.g. gist indexes we originally created) so its fast

materialising the purl into a string into its own column is not really required - unless I am missing something.

jcrossley3: > materialising the purl into a string into its own column is not really required - unless I am missing something.

The only thing I can think of is giving up the ability to search "across" purl component boundaries, e.g. `q=purl~/quarkus@3`. That query (and any other "partial yet invalid" purl queries) won't work without the redundant stringified purl column. Instead, to get _similar_ results, users will have to pass `q=name=quarkus&version~3`.

I say _similar_ because you won't get the exact results. Being able to pass `@3` says show me all the purls with MAJOR version = 3. But passing `version~3` will return purls with MINOR and PATCH versions equal to 3 as well.

But I'll do it how y'all want. You guys know your users better than me.

jcrossley3: I found another thing we'll miss without `get_purl` or any column holding the purl.

A purl filter, e.g. `q=purl=pkg:nuget/NGX@31.0.15.5356` will work, because the purl is parsed into its components, which are then used in the query.

But a full-text search containing only a purl, e.g. `q=pkg:nuget/NGX@31.0.15.5356` will no longer work, because nothing among the selected columns has that string, which is pretty counter-intuitive for a "full text search". :)

I'm pretty sure this behavior will surprise the ProdSec guys, but there is a workaround.

jcrossley3: Ok, I've removed the stored procs without adding a column. This required a number of changes to our query language, all for the better, I hope.

JimFuller-RedHat: doing this means we have pushed processing from very fast db to much slower rust ... I see in scale test report that things have slowed down in search_purls (but those have bumped around) but we do not have any tests yet in scale testing for analysis graph ... next week I will see the blast radius of this ... at worst we might have to keep get_purl for analysis graph ...

jcrossley3: > very fast db to much slower rust

That would surprise me. We're asking the DB to do less work, so I'd expect it to be faster.

ctron:
I'm a bit confused by the syntax. I saw `purl:name` and `qualifier:foo`. Is "qualifier" now part of the purl to? Should that then be `purl:qualifier:foo`?

jcrossley3:
> I'm a bit confused by the syntax. I saw `purl:name` and `qualifier:foo`. Is "qualifier" now part of the purl to? Should that then be `purl:qualifier:foo`?

That reflects [our model](https://github.com/trustification/trustify/blob/a3afdfed5485ad763cd7d5df142fe3db57cc97b9/migration/src/m0000010_init_up.sql#L1872-L1873). The qualifiers field is a subset of the purl field.

Right now, only 1 colon is allowed in a field name, i.e. we only go 1 level deep. I can look at extending that if we think it'd be useful.

jcrossley3:
@ctron It's a fairly significant change. And this PR is already large and works well, I think, for our most common queries. Can you approve this one, and we can do the deeply nested query fields in a separate PR?

JimFuller-RedHat:
that last scale test significantly blew up .. and that still does not have any analysis graph specific tests.

jcrossley3:
How can I run that scale test locally?

jcrossley3:
> I think that having `purl:qualifier:foo` is more logical.

[Done](https://github.com/trustification/trustify/pull/1534/commits/700926228d40faaeb43ed4f8013efa080157df88)

ctron (approved):
Thanks for making the changes. I think that still leaves `type` as a conflicting column. I assume that's for keeping it compatible with the existing implementation. Maybe it makes sense documenting this as "deprecated", so that we can remove it in 2.1 or beyond.

JimFuller-RedHat:
scale tests blows up with these changes ...

jcrossley3:
So unless I'm reading that report wrong, it appears most of the negative perf impact is in `list_packages`?

JimFuller-RedHat:
the requests per seconds is still too low eg. dropped from .34 to .3 ... but fine to merge, it just might mean we have  more work to get back more speed - a mystery for next week.

PR #1538, opened 2025-04-08
chore: update declared pg-embedded to 0.18.2

helio-frota:
~As an attempt to reduce github api rate limit~

edit: Just a chore PR to align declared dependencies with lock file


ctron:
As the lock file didn't change, it seems we already picked up a newer version.

helio-frota:
yeah correct thanks :+1:  
actually 0.18.2 https://github.com/trustification/trustify/blob/main/Cargo.lock#L5341

 similar to https://github.com/trustification/trustify/pull/1138#issuecomment-2589583382
 and I sent the PR without check the existence of .lock file again
 
 let me change the commit message and PR tittle

helio-frota:
done :+1:

PR #1539, opened 2025-04-08
test: add parallel test, getting stuck

ctron:


ctron:
Run with:

```shell
podman compose -f etc/deploy/compose/compose.yaml up
env RUST_LOG=debug EXTERNAL_TEST_DB=true DB_USER=postgres DB_PASSWORD=trustify EXTERNAL_TEST_DB_BOOTSTRAP=true cargo test --package trustify-module-fundamental --test fundamental advisory::csaf::timeout::timeout -- --exact --nocapture
```

modules/ingestor/src/service/mod.rs
```diff
@@ -230,8 +230,8 @@ impl IngestorService {
             };
         }
 
-        let duration = Instant::now() - start;
-        log::debug!(
+        let duration = start.elapsed();
+        log::info!(
```
dejanb, line 234: Should we return this to debug?

dejanb (approved):

modules/ingestor/src/service/mod.rs
ctron: Should be back to `debug`.

JimFuller-RedHat (approved):
LGTM

PR #1542, opened 2025-04-08
perf: TC-2428 add gin index to advisory

mrizzi:
https://issues.redhat.com/browse/TC-2428

within the context of the advisory search:
- the columns from the `advisory` table have been added to a `gin` index to improve search performances
- the digests columns from `source_document` table have been removed to make the execution plan more reliable


dejanb (approved):
The changes are good and perform well locally.

Adding a scale test for this use case 
https://github.com/trustification/scale-testing/pull/51
should prove that and guard the performance in the future.

PR #1545, opened 2025-04-09
feat: implement api/v2/analysis/latest/component

### LINKED ISSUE #1231: Implement naive latest filtering
When retrieving sboms (or components contained  from an sbom) we often need to discriminate between the latest instance of an sbom and all the rest.  For example when searching `api/v2/analysis/component?q=curl` we may only want to retrieve _curl_ matches from **latest** sboms. but the main thing is we want to get the latest set of root packages.

Here is a specific example of this in CDX: 

_Given two product manifests that identify a product as cpe:/a:redhat:rhel_eus:9.2::baseos:_
    
* https://github.com/RedHatProductSecurity/security-data-guidelines/blob/main/sbom/examples/product/rhel-9.2-main%2Beus.cdx.json
* https://github.com/RedHatProductSecurity/security-data-guidelines/blob/main/sbom/examples/product/rhel-9.2-eus.cdx.json
   
 Search query for **pkg:rpm/redhat/openssl** should return the latest of the two:
* Yes: `pkg:rpm/redhat/openssl@3.0.7-18.el9_2?arch=src`
* No: `pkg:rpm/redhat/openssl@3.0.7-17.el9_2?arch=src`

Though things get complicated because there are ecosystem specific rules for determining **latest**.

We  first need to identify stable heuristics  defining (across ecosystem types) what is a  'latest' sbom. 

We may need the ability to arbitrarily _override_  heuristics eg. say on ingestion denote an SBOM as **latest** ?  It occurs to me that, in this scenario, maybe the term **latest** is not entirely correct and means 'this SBOM is the most authoratative' ... needs some more thinking.

There are some potential  performance optimisations implementing this heuristic/filter:

* We may only ever need to load into analysis graph the **latest** instance of SBOM (and evict all others). 
* We could _partition_ data (and/or create partial indexes) on the _axis_ of **latest**.

Lastly, having the latest view set as default (in the ux and rest api) would help general users reducing the 'many truths' and  dramatically cut down on the number of duplicate (yet valid) data we currently retrieve.

JimFuller-RedHat:
fixes #1231


ctron (approved):

ctron:
I can't really review the SQL itself, I'd ask James for help :grinning: but all the other stuff looks good. Also doesn't change existing code paths and adds new tests.

PR #1546, opened 2025-04-09
feat: allow upload gzip compressed files

ctron:


JimFuller-RedHat (approved):
ah now I understand walker-common update ... LGTM

PR #1551, opened 2025-04-10
feat: filter SBOM's and Advisories by labels

### LINKED ISSUE #491: Search by labels
Now that we have labels, I think would make sense to allow searching advisories and SBOM also by labels.

As a first step, I would want to search using labels like this:

* Give me all documents having `foo=bar`
* Give me all documents having `foo=bar` and `bar=baz`
* Give me all documents having `foo=bar` or `foo=baz`
* Give me all documents having a label of `foo`.

jcrossley3:
Fixes #491

### DIFF
--- a/modules/fundamental/src/advisory/endpoints/test.rs
+++ b/modules/fundamental/src/advisory/endpoints/test.rs
@@ -17,8 +17,11 @@ use trustify_cvss::cvss3::{
     PrivilegesRequired, Scope, UserInteraction,
 };
 use trustify_entity::labels::Labels;
-use trustify_module_ingestor::{graph::advisory::AdvisoryInformation, model::IngestResult};
+use trustify_module_ingestor::{
+    graph::advisory::AdvisoryInformation, model::IngestResult, service::Format,
+};
 use trustify_test_context::{TrustifyContext, call::CallService, document_bytes};
+use urlencoding::encode;
 use uuid::Uuid;
 
 #[test_context(TrustifyContext)]
@@ -609,6 +612,53 @@ async fn delete_advisory(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
 
     log::debug!("Code: {}", response.status());
     assert_eq!(response.status(), StatusCode::NOT_FOUND);
+
+    Ok(())
+}
+
+#[test_context(TrustifyContext)]
+#[test(actix_web::test)]
+async fn query_advisories_by_label(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
+    let query = async |q| {
+        let app = caller(ctx).await.unwrap();
+        let uri = format!("/api/v2/advisory?q={}", encode(q));
+        let req = TestRequest::get().uri(&uri).to_request();
+        let response: Value = app.call_and_read_body_json(req).await;
+        assert_eq!(1, response["total"], "for {q}");
+    };
+    ctx.ingest_document_as(
+        DOC,
+        Format::CSAF,
+        [
+            ("type", "csaf"),
+            ("source", "test"),
+            ("importer", "none"),
+            ("file", "cve-2023-33201.json"),
+            ("datasetFile", "none"),
+            ("foo", "bar"),
+        ],
+    )
+    .await?;
+
+    query("labels:type!=spdx").await;
+    query("labels:type!~zap").await;
+    query("labels:type~af").await;
+    query("labels:type=csaf").await;
+    query("labels:type=csaf&labels:source=test").await;
+    query("labels:type=csaf&labels:source=test&labels:importer=none").await;
+    query(
+        "labels:type=csaf&labels:source=test&labels:importer=none&labels:file=cve-2023-33201.json",
+    )
+    .await;
+    query("labels:type=csaf&labels:source=test&labels:importer=none&labels:file=cve-2023-33201.json&labels:datasetFile=none").await;
+    query("labels:file>aaah.json").await;
+    query("labels:datasetFile<zilch").await;
+    query("label:foo=bar").await;
+    query("label:type=csaf").await;
+    query("label:importer=some|none").await;
+    query("label:type!=spdx").await;
+    query("labels:type~af&labels:foo>aah").await;
+    query("labels:importer~one&label:file~33201").await;
 
     Ok(())
 }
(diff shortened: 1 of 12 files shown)


modules/fundamental/src/advisory/service/mod.rs
```diff
@@ -117,7 +117,11 @@ impl AdvisoryService {
                         // v = "" for all sort fields
                         ("average_severity", "") => Some(format!("average_score:{op}")),
                         _ => None,
-                    }),
+                    })
+                    .json_keys(
+                        "labels",
+                        &["type", "source", "importer", "file", "datasetFile"],
```
ctron, line 123: A user can apply any label and value. Will it be possible to e.g. search for `ci=my-build-job` too?

modules/fundamental/src/sbom/endpoints/test.rs
```diff
@@ -306,6 +306,42 @@ async fn query_sboms_by_ingested_time(ctx: &TrustifyContext) -> Result<(), anyho
...
+            ("type", "cyclonedx"),
+            ("source", "test"),
+            ("importer", "none"),
+            ("file", "zoo.json"),
+            ("datasetFile", "none"),
+        ],
+    )
+    .await?;
+
+    query("type!=spdx").await;
```
ctron, line 332: It looks to me as if there's no indication that this is a label. What would be the case if there's a column `namespace` from PURLs, and the user applies a label `namespace`?

modules/fundamental/src/advisory/service/mod.rs
jcrossley3: Not as a result of merging this PR. That's why I was careful to "relate" #491, because this PR doesn't fully "fix" it. But it does give @JimFuller-RedHat a way to filter SBOM's by type.

modules/fundamental/src/sbom/endpoints/test.rs
jcrossley3: Dunno. What do you think should be the case?

ctron: I would want to query for labels explicitly.

modules/fundamental/src/advisory/service/mod.rs
ctron: I don't think we should hardcode any labels like that. If we allow filtering, then by any label.

modules/fundamental/src/sbom/endpoints/test.rs
jcrossley3: Be specific. Propose your desired syntax for querying labels.

modules/fundamental/src/advisory/service/mod.rs
jcrossley3: We're already hardcoding the labels. Where do you think I got those from?

modules/fundamental/src/sbom/endpoints/test.rs
ctron: I don't really know the current syntax. All I would like to to is to say "all items with label `foo` equals/contains/… `bar`. Not conflicting with any other column named `foo`.

jcrossley3: > I don't really know the current syntax. 

I know. Suggest a new one then. Be constructive.

> All I would like to to is to say "all items with label `foo` equals/contains/… `bar`. Not conflicting with any other column named `foo`.

I understand. This PR doesn't provide that capability. Which is why it's only marginally related to #491 -- it does not fully address it and therefore won't close it.

jcrossley3: That's easy enough, but it requires [one of the commits](https://github.com/trustification/trustify/pull/1534/commits/f600eca2ea0d3cc6a920f8eea715cfab03d854fc) in #1534, so if you're gonna reject that one, I'll need to pull it out into its own PR.

jcrossley3: Done

test-context/src/lib.rs
```diff
@@ -78,22 +78,21 @@ impl TrustifyContext {
...
-    /// Ingest a document with a specific format.
+    /// Ingest a document with a specific format and labels
     ///
     /// The path is relative to `<workspace>/etc/test-data`.
-    pub async fn ingest_document_as(
+    pub async fn ingest_document_as<const N: usize>(
         &self,
         path: &str,
         format: Format,
+        labels: [(&str, &str); N],
```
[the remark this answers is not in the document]
jcrossley3: Compiler doesn't like it. `ingest` is expecting `impl Into<Labels>`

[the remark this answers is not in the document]
ctron: then maybe use that?

jcrossley3:
> @jcrossley3 @ctron Thanks for working on this! Once merged I will update #1555.

Great! Feel free to approve!

modules/fundamental/src/advisory/endpoints/test.rs
```diff
@@ -612,3 +615,50 @@ async fn delete_advisory(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
...
+            ("source", "test"),
+            ("importer", "none"),
+            ("file", "cve-2023-33201.json"),
+            ("datasetFile", "none"),
+            ("foo", "bar"),
+        ],
+    )
+    .await?;
+
+    query("labels:type!=spdx").await;
```
ctron, line 643: Just looking what others do. GitHub is using the singular, which also makes more sense to me:

image

ctron: Hm. I also notices some other examples using the singular. Feels a bit confusing to me. Is both allowed?

ctron (approved):
I think it's good to have it. I think it would be good to document it and have `label` (instead of `labels`, or point out that both mean the same?).

modules/fundamental/src/advisory/endpoints/test.rs
jcrossley3: [yes](https://github.com/trustification/trustify/blob/7bcb9939d3387b99e7d364dd0f5285f143623df4/modules/fundamental/src/sbom/service/sbom.rs#L140)

PR #1552, opened 2025-04-10
feat: return a list of valid fields in a query error message

jcrossley3:
It's the least we can do

### DIFF
--- a/common/src/db/query.rs
+++ b/common/src/db/query.rs
@@ -172,7 +172,7 @@ pub struct Query {
 
 #[derive(Debug, PartialEq, thiserror::Error)]
 pub enum Error {
-    #[error("query syntax error: {0}")]
+    #[error("Query syntax error: {0}")]
     SearchSyntax(String),
 }
 
--- a/common/src/db/query/columns.rs
+++ b/common/src/db/query/columns.rs
@@ -196,8 +196,8 @@ impl Columns {
                         })
                 })
                 .ok_or(Error::SearchSyntax(format!(
-                    "'{field}' is an invalid field. Try {:?}",
-                    self.fields()
+                    "'{field}' is an invalid field. Try [{}]",
+                    self.fields().join(", ")
                 )))
         }
     }
--- a/common/src/db/query/columns.rs
+++ b/common/src/db/query/columns.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::{Display, Formatter};
 
 use sea_orm::entity::ColumnDef;
@@ -196,7 +196,8 @@ impl Columns {
                         })
                 })
                 .ok_or(Error::SearchSyntax(format!(
-                    "Invalid field name: '{field}'"
+                    "'{field}' is an invalid field. Try {:?}",
+                    self.fields()
                 )))
         }
     }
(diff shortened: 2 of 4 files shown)


ctron (commented):
This is a great enhancement. Does the message end up in the "human readable details" field? That seems appropriate.

jcrossley3:
> Does the message end up in the "human readable details" field? That seems appropriate.

I'm not sure what that is. If it's a UX thing, then probably not. The UX already knows the valid fields.

But as we have more people using the openapi endpoints, I thought it'd be nice to show them the valid fields when they guess wrong.

ctron:
> > Does the message end up in the "human readable details" field? That seems appropriate.
> 
> I'm not sure what that is. If it's a UX thing, then probably not. The UX already knows the valid fields.
> 
> But as we have more people using the openapi endpoints, I thought it'd be nice to show them the valid fields when they guess wrong.

https://github.com/trustification/trustify/blob/a5ecedad2b5d37b2ee6469400ef1f526a90874be/common/src/error.rs#L11-L13

jcrossley3:
> https://github.com/trustification/trustify/blob/a5ecedad2b5d37b2ee6469400ef1f526a90874be/common/src/error.rs#L11-L13

Cool. What exactly are you asking me to do?

ctron:
> > https://github.com/trustification/trustify/blob/a5ecedad2b5d37b2ee6469400ef1f526a90874be/common/src/error.rs#L11-L13
> 
> Cool. What exactly are you asking me to do?

Ensure the explanation ends up in the `details` field.

jcrossley3:
> Ensure the explanation ends up in the `details` field.

The error message, with all the valid fields, is already being passed back to the UX in the `message` field [here](https://github.com/trustification/trustify/blob/a5ecedad2b5d37b2ee6469400ef1f526a90874be/modules/fundamental/src/error.rs#L70-L72). You want it in the `details` field, too? So that both `message` and `details` contain the same text? That seems redundant, so I just want to make sure.

ctron:
> > Ensure the explanation ends up in the `details` field.
> 
> The error message, with all the valid fields, is already being passed back to the UX in the `message` field [here](https://github.com/trustification/trustify/blob/a5ecedad2b5d37b2ee6469400ef1f526a90874be/modules/fundamental/src/error.rs#L70-L72). You want it in the `details` field, too? So that both `message` and `details` contain the same text? That seems redundant, so I just want to make sure.

The idea behind the three fields is:

* One field which is machine readable (`error`)
* One field which is human readable (`message`)
* One field which carries additional, human readable details (`details`)

The additional information about available fields should to into the `details` field, not in any of the other fields.

ctron:
The `details` field is not a UI field. It's a "human" field, which gets exposed on the API too.

jcrossley3:
I stand corrected: it's closer to 99% of the time it's not used. It's used in two places. I'd argue that the [first one](https://github.com/trustification/trustify/blob/6316078533a93a424fd3c2fc4f643e616ee73a6b/modules/ingestor/src/service/mod.rs#L122-L123) is just wrong, and the [second one](https://github.com/trustification/trustify/blob/6316078533a93a424fd3c2fc4f643e616ee73a6b/modules/fundamental/src/sbom/endpoints/query.rs#L34-L38) could easily be reconstituted as a single error message.

I think it'd clean up the code significantly to remove `details`. I can do that in a separate PR that I'm sure you'll reject, but I'd like to keep this one as is.

bobmcwhirter:
LGTM

Issue #1547, opened 2025-04-09
Panick inserting a CSAF document with an invalid CVSS


ctron:
```
thread 'main' panicked at /cargo/git/checkouts/csaf-rs-9ea2a917fa6c455d/4148969/src/vulnerability.rs:219:14:
You should not have been able to construct a cvss_json::Cvss3 except from a cvss::v3::Base which should always have a valid vector string: UnknownMetric { name: "E" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #1553, opened 2025-04-11
fix: ensure that an invalid CVSS does not panic

ctron:
Closes: #1547


helio-frota (approved):
approving moving the workflow forward

PR #1555, opened 2025-04-11
docs: Add v1 to v2 migration guide

i386x:
Resolves: TC-2296

Currently contain only one known documented use case from the Jira ticket above but more can come/be added. ~I will left this PR in draft mode and switch it to reviewable state after the reporter of the TC-2296 is satisfied with the result.~

~The part of this PR is also an Antora extension for checking the correctness of API calls (see also added `README.md`).~

### DIFF
--- a/docs/book/modules/migration/pages/index.adoc
+++ b/docs/book/modules/migration/pages/index.adoc
@@ -1,0 +1,33 @@
+= TPA v1 to v2 Migration Guide
+
+Do you have to try TPA v2 (a.k.a. Trustify) but your applications and use cases
+still depend on TPA v1 (a.k.a. Trustification)? No problem! In this guide, we
+show you how to migrate your applications and use cases to TPA v2. Just choose
+from the sections below, depending on what part of TPA v1 you wish to migrate.
+
+== SBOM API
+
+=== Ingesting an SBOM File
+
+In TPA v1, you upload SBOM file like this:
+----
+POST /api/v1/sbom?id={my_id}
+Headers: "Content-Type: application/json"
+Payload (file): my_sbom.json
+----
+
+In TPA v2, using `id` to name your SBOMs is no longer possible. When an SBOM is
+uploaded, the special UUID is generated and attached to it. This UUID is
+returned as a part of the response, so you can use it later for downloading the
+SBOM.
+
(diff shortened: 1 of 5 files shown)


docs/book/modules/migration/pages/index.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+(async () => {
+  await bombastic.post(`/api/v1/sbom?id=${sbomId}`, payload, {
+    headers: { "Content-Type": "application/json" }
+  });
+
+  const { data } = await bombastic.get(`/api/v1/sbom?id=${sbomId}`);
+
+  assert.deepStrictEqual(data, JSON.parse(payload));
+})();
```
carlosthe19916, line 40: The file `migration/pages/index.adoc` is the main contribution in this PR. It is very much Programming Language centrist IMO (Javascript). I think we should focus on the API definitions, and only then give examples in multiple programming languages. E.g.

V1 allows files to be uploaded through:
```
POST /api/v1/sbom?id={my_id}
Headers: "Content-Type: application/json"
```

While in v2 we should upload files like this:
```
POST /api/v2/sbom
```

My previous example might not be accurate as I wrote it just for the sake of giving an example.
The point is that for describing and comparing an endpoint in v1 and v2 I did not need to write actual code in any programming language to describe them.

The original request is to document v1 vs v2 endpoints right? IMHO we could focus on just describing with text the endpoints.
-  If we are going to write an example in a programming language we should also understand in which language our clients have written their integrations. E.g. The CAMP team, in which language their tools are written and then write examples giving priority to the language of those clients

carlosthe19916 (changes requested):
@i386x some observations:

-  This PR introduces some NPM dependencies and also include some tools like Eslint. While I appreciate the effort, I am not sure it is the right move.
   - There needs to be a person continuously maintaining those things. I think the voice of the main contributors will be important in this case.

I also added an observation about the documenting the rest apis in pure programming language when I think we could focus on just text descriptions and only after jump into programming language examples.

> Note. If the PR is in draft mode, it might be overlooked by the other maintainers of this repository. Changing it to "ready" would help the others to look at this.

carlosthe19916 (approved):
@i386x Given the fact that you are not modifying the package.json file, perhaps the package-lock.json file should not be changed either. Besides it, LGTM.

PR #1556, opened 2025-04-14
fix: Python versions PLSQL comparator (TC-2469)

mrizzi:
https://issues.redhat.com/browse/TC-2469

[PEP-0440](https://peps.python.org/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions) allowed a more permissive convention for pre-release versions allowing to have a `-` between the version and the pre-release marker, e.g. `1.2.3-rc0`.

So the regex to retrieve the `{left,right}_pre{_num}` fields have been updated consistently and three tests to cover this scenario has been added swapping left and right in order to ensure the same patterns are applied to both left and right fields.

### DIFF
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -21,6 +21,7 @@ mod m0001000_sbom_non_null_suppliers;
 mod m0001010_alter_mavenver_cmp;
 mod m0001020_alter_pythonver_cmp;
 mod m0001030_perf_adv_gin_index;
+mod m0001040_alter_pythonver_cmp;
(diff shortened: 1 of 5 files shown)


dejanb (approved):
Looks good. It'd be good to add tests to cover these new cases.

mrizzi:
> Looks good. It'd be good to add tests to cover these new cases.

My bad, you're right: I mentioned three tests in the PR's description but I forgot to commit them 🤦 
I've just added them to the PR.

Issue #1535, opened 2025-04-08
Something panicked around YAML


ctron:
Just noticed the following in the logs, but I can't reproduce it:

```
thread 'actix-server worker 0' panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libyml-0.0.5/src/scanner.rs:2798:17:
String join would overflow memory bounds
thread 'actix-server worker 1' panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libyml-0.0.5/src/scanner.rs:2798:17:
String join would overflow memory bounds
```

helio-frota:
```
thread 'actix-server worker 4' panicked at /home/heliofrota/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libyml-0.0.5/src/scanner.rs:2798:17:
String join would overflow memory bounds
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
is clearly_defined ????????????????????
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```

[is_clearly_defined](https://github.com/trustification/trustify/blob/main/modules/ingestor/src/service/format.rs#L197-L199) called from [sbom_from_bytes](https://github.com/trustification/trustify/blob/main/modules/ingestor/src/service/format.rs#L135) is using the bugged version of `serde_yml`. And this case is different from [this bug](https://users.rust-lang.org/t/serde-yml-bug-which-causes-a-panic/120489) as we don't have `serde_yml::to_string` call before the `serde_yml::from_slice` call.

PR #1557, opened 2025-04-14
fix: prevent panic when handling non YAML content

helio-frota:


jcrossley3 (commented):
If `serde_yml` is buggy, can we avoid using it? Even with your change, couldn't we hit the panic if the uploaded file is yaml?

What happens when you make this change in the root Cargo.toml?

```diff
-serde_yml = "0.0.12"
+serde_yaml_ng = "0.10"
+serde_yml = { package = "serde_yaml_ng", version = "0.10" }
```

ctron (changes requested):
A panic is different then returning `Result::Err`. I don't think this will work. Maybe it makes sense adding a test for this too.

helio-frota:
> If serde_yml is buggy, can we avoid using it?

I had thought about that, but I was more focused on reproducing the issue and addressing the panic.

> Even with your change, couldn't we hit the panic if the uploaded file is yaml?

That's the strange situation I mentioned in the chat, we can't upload YAML files.


And we have a function [that checks yamls](https://github.com/trustification/trustify/blob/main/modules/ingestor/src/service/format.rs#L135). Then when we upload the `json` and it enters on that function as a `json`. On this PR I'm checking if that is a `json` do nothing and return false.

> A panic is different then returning Result::Err. I don't think this will work. Maybe it makes sense adding a test for this too.

I'm not seeing the panic anymore, but I agree that a test would be good.
But if we can't upload YAML, then why have a YAML check?

helio-frota:
> What happens when you make this change in the root Cargo.toml?

It worked :+1: 

@ctron @jcrossley3 I'll change the PR to use this crate then. 

Otherwise we still have this: 

`if we can't upload YAML, then why to have a YAML check?`
and
https://users.rust-lang.org/t/serde-yml-bug-which-causes-a-panic/120489/11

Cargo.toml
```diff
@@ -117,7 +117,8 @@ serde = "1.0.183"
 serde-cyclonedx = "0.9.1"
 serde_json = "1.0.114"
 serde_with = "3.11.0"
-serde_yml = "0.0.12"
+serde_yaml_ng = "0.10"
```
ctron, line 120: Do we need that one? I'd expect the second line to be sufficient.

ctron (approved):

helio-frota:
@ctron yah for this case we already have a test https://github.com/trustification/trustify/blob/main/modules/fundamental/tests/advisory/osv/ingest.rs#L10 :+1: 

the panic situation happened when selecting `vex` `.json` files to be uploaded on `sbom` tab
 [link](https://github.com/trustification/trustify/issues/1535#issuecomment-2801566775)

[then the buggy serde_yml received `bytes(json)` to parse as yaml](https://github.com/trustification/trustify/issues/1535#issuecomment-2801665857)

helio-frota:
@ctron ok sounds good  :+1:

PR #1563, opened 2025-04-15
fix: ensure that container build aligns with deployment architecture

ctron:


helio-frota (approved):
[ probably related to https://docs.aws.amazon.com/eks/latest/userguide/sample-deployment.html ]

PR #1564, opened 2025-04-16
refactor: move advisory status cpe filtering to the queries

dejanb:
This PR refactors SBOM/advisory correlation logic to correctly filter all SBOM-related CPEs in the query, moving it from the post-query loop.

This helps the performance by properly and early selecting statuses that are only related to the SBOM in question. It should lead to much less rows pulled from the database and filtered later (which is one of the problems with large SBOMs).

It also makes this code a bit more maintainable and suitable for future refactoring.

It still doesn't solve all performance issue with the endpoint. RHOSE-4.14 SBOM doesn't fail for me any more locally, but it still take a few minutes (That's why I didn't include load test yet). I don't expect it to have impact on other endpoints.

This is a first step toward performance of SBOM correlation logic for large SBOMs.

### DIFF
--- a/modules/fundamental/src/sbom/model/raw_sql.rs
+++ b/modules/fundamental/src/sbom/model/raw_sql.rs
@@ -1,0 +1,126 @@
+/// This constant is a SQL subquery that filters the context_cpe_id
+/// based on the given sbom_id. It checks if the context_cpe_id is null
+/// or if it is in the list of CPEs that are related to the packages
+/// that describes the SBOM. The additional logic allows us to find
+/// superset of generalized CPEs that don't include subfields like edition
+/// and find "stream" releases based on the major version.
+pub const CONTEXT_CPE_FILTER_SQL: &str = r#"
+(
+    context_cpe_id IS NULL OR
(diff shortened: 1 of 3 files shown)


modules/fundamental/src/sbom/model/raw_sql.rs
```diff
@@ -0,0 +1,129 @@
+/// This function returns SQL subquery that filters the context_cpe_id
+/// based on the given sbom_id. It checks if the context_cpe_id is null
+/// or if it is in the list of CPEs that are related to the packages
+/// that describes the SBOM. The additional logic allows us to find
+/// superset of generalized CPEs that don't include subfields like edition
+/// and find "zstream" releases based on the major version.
+pub fn context_cpe_filter_sql() -> &'static str {
```
ctron, line 7: I guess (not sure) you can make those `fn`s `const`. So that they would be run a compile time.

dejanb: Cool. Let me give it a try

ctron (approved):

PR #1568, opened 2025-04-22
docs: ADR for signatures, sigstore and GPG

ctron:
Rendered version: https://github.com/ctron/trustify/blob/feature/adr_check_sig_1/docs/adrs/00005-signatures.md


carlosthe19916:
Looks great.

- In terms of how it would look like for users, I think we could somehow copy the Github way of rendering signed commits.
  - I also wonder if in the UI there should be visibility of the keys used to verify signature? Something only an "Admin" should be able to see. Asking just in case a dedicated REST API would be needed to expose this information.
- In regards of the Upload REST API. I wonder how an SBOM and its signature will be uploaded? I might be wrong but I guess the SBOM and its signature are 2 different files, should/can they be uploaded together?

PR #1569, opened 2025-04-22
fix: when deleting an entity, delete also a source document

dejanb:
When deleting an entity (advisory or sbom), the source document is left in the database. This makes uploading the same file again (after deleting) fails.

This PR fixes this. And additionally adds foreign keys to the CVSS tables to clean them as well after deleting an advisory.

### DIFF
--- a/e2e/ds3.hurl
+++ b/e2e/ds3.hurl
@@ -7,10 +7,13 @@ HTTP 201
 [Asserts]
 jsonpath "$.files.['spdx/quarkus-bom-2.13.8.Final-redhat-00004.json.bz2'].id" exists
 jsonpath "$.files.['spdx/ubi8-8.8-1067.json.bz2'].id" exists
+jsonpath "$.files.['csaf/2023/cve-2023-0044.json'].id" exists
 
 [Captures]
 quarkus_sbom_id: jsonpath "$.files.['spdx/quarkus-bom-2.13.8.Final-redhat-00004.json.bz2'].id"
 ubi8_sbom_id: jsonpath "$.files.['spdx/ubi8-8.8-1067.json.bz2'].id"
+advisory_id: jsonpath "$.files.['csaf/2023/cve-2023-0044.json'].id"
+
 
 # Check vulnerabilities for UBI 8 SBOM
 GET http://localhost:8080/api/v2/sbom/{{ubi8_sbom_id}}/advisory
@@ -34,3 +37,13 @@ HTTP 200
 
 [Asserts]
 jsonpath "$.title" not isEmpty
+
+## Test delete and re-upload advisory
+DELETE http://localhost:8080/api/v2/advisory/{{advisory_id}}
+
+HTTP 200
+
+POST http://localhost:8080/api/v2/advisory
+file,etc/datasets/ds3/csaf/2023/cve-2023-0044.json;
+
+HTTP 201
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -22,6 +22,7 @@ mod m0001010_alter_mavenver_cmp;
 mod m0001020_alter_pythonver_cmp;
 mod m0001030_perf_adv_gin_index;
 mod m0001040_alter_pythonver_cmp;
+mod m0001050_foreign_key_cascade;
 
 pub struct Migrator;
 
@@ -47,6 +48,7 @@ impl MigratorTrait for Migrator {
             Box::new(m0001020_alter_pythonver_cmp::Migration),
             Box::new(m0001030_perf_adv_gin_index::Migration),
             Box::new(m0001040_alter_pythonver_cmp::Migration),
+            Box::new(m0001050_foreign_key_cascade::Migration),
         ]
     }
 }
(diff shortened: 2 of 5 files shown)


modules/fundamental/src/advisory/endpoints/mod.rs
```diff
@@ -139,6 +140,7 @@ pub async fn delete(
         match rows_affected {
             0 => Ok(HttpResponse::NotFound().finish()),
             1 => {
+                fetched.delete_source_document(&tx).await?;
```
ctron, line 143: Can't we do a "cascade delete"?

dejanb: No (or at least I didn't find a way to do it) as the relationship is reversed. We can add cascade to delete advisory when source document is deleted, but there's no reference from source document table to sbom/advisory tables.

ctron: Ah, ok.

modules/fundamental/src/source_document/model/mod.rs
```diff
@@ -56,3 +65,34 @@ impl TryInto<StorageKey> for SourceDocument {
...
 }
+
+#[async_trait::async_trait]
+pub trait DeleteSourceDocument {
+    async fn delete_source_document(&self, db: &DatabaseTransaction) -> Result<(), DbErr>;
+}
+
+async fn delete_source_document_by_id(
+    db: &DatabaseTransaction,
+    source_document: Option<SourceDocument>,
```
ctron, line 76: Wouldn't `Option<&SourceDocument>` be enough. Reducing the `clone`?

dejanb: Yup. Fixed!

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -303,6 +306,7 @@ pub async fn delete(
             match rows_affected {
                 0 => Ok(HttpResponse::NotFound().finish()),
                 1 => {
+                    v.delete_source_document(&tx).await?;
```
ctron, line 309: I think that should be part of `delete_sbom`.

modules/fundamental/src/source_document/model/mod.rs
```diff
@@ -1,13 +1,22 @@
+use sea_orm::DatabaseTransaction;
+use sea_orm::DbErr;
+use sea_orm::EntityTrait;
```
ctron, line 3: Please nest.

modules/fundamental/src/source_document/model/mod.rs
```diff
@@ -56,3 +66,34 @@ impl TryInto<StorageKey> for SourceDocument {
         (&self).try_into()
     }
 }
+
+#[async_trait::async_trait]
```
ctron, line 70: Not sure we actually need this. I think this should be part of the actual "delete" operations, and not really exposed to the user of the (service) API.

modules/fundamental/src/advisory/endpoints/mod.rs
ctron, line 143: I think this should be part of the `delete_advisory` operation.

ctron (changes requested):
I think the part of deleting the source document shouldn't be leaked outside the service APIs. But handles as part of `delete_*`.

dejanb:
The logic has been rewritten to move the logic in the `delete_` methods of the appropriate entities. I also fixed `product_version->sbom` foreign key to cascade on delete.

jcrossley3:
> The logic has been rewritten to move the logic in the `delete_` methods of the appropriate entities. I also fixed `product_version->sbom` foreign key to cascade on delete.

Nice! I was wondering why it was SET NULL before. I've been testing SBOM's oblivious to the fact you were already working on it.

modules/fundamental/src/source_document/model/mod.rs
```diff
@@ -5,9 +5,12 @@ use trustify_common::id::{Id, IdError};
 use trustify_entity::source_document;
 use trustify_module_storage::service::StorageKey;
 use utoipa::ToSchema;
+use uuid::Uuid;
 
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
 pub struct SourceDocument {
+    #[serde(skip)]
+    pub id: Uuid,
```
ctron, line 13: Do we still need this now?

ctron (changes requested):

modules/fundamental/src/source_document/model/mod.rs
ctron: The `skip` attribute made me wonder. It's just an internal thing, not visible on the outside. So why have it? Doesn't feel right.

ctron (approved):
I'd appreciate a cleanup of the "source document ID", either way. But not a blocker.

dejanb:
> I'd appreciate a cleanup of the "source document ID", either way. But not a blocker.

OK, removed. Just for the future reference "skip" was there because the structs is flattened into the SbomDetails and there it clashes with SBOM id.

PR #1574, opened 2025-04-28
Speed up ingestion

ctron:
This PR has a few improvements for SBOM ingestion, which also should improve other documents ingestion with it.

* [x] Allow providing a format during the upload, working around "auto detection"
* [x] Don't push work on the worker thread pool if there is none (decompression)
* [x] Queue analysis graph loading after ingestion, instead of waiting for it

### DIFF
--- a/modules/ingestor/src/service/mod.rs
+++ b/modules/ingestor/src/service/mod.rs
@@ -146,6 +146,28 @@ impl ResponseError for Error {
     }
 }
 
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Default, serde::Deserialize, utoipa::ToSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum Cache {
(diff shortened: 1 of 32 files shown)


modules/ingestor/src/service/mod.rs
```diff
@@ -212,12 +212,12 @@ impl IngestorService {
         if let Some(analysis) = &self.analysis {
             match fmt {
                 Format::SPDX | Format::CycloneDX => {
-                    if result.id.to_string().starts_with("urn:uuid:") {
+                    if let Id::Uuid(id) = result.id {
```
JimFuller-RedHat, line 215: ah Id::Uuid  does the needful, thx learned something

JimFuller-RedHat (approved):
so far LGTM though I think you need to regen openapi - also expecting more code for queuing ?

PR #1575, opened 2025-04-28
docs: adr for sbom dashboard

bxf12315:


docs/adrs/00007-sbom-dashbord.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Status
+
+DRAFT
+
+## Context
+This ADR document is intended to define and implement the backend logic for the heard of SBOM dashboard in the UI.
+
+Its mockup design document is as follows.
+![header-dashbord.png](header-dashbord.png)
```
ctron, line 13: please move graphics into an ADR specific subfolder.

docs/adrs/00007-sbom-dashbord.md
```diff
@@ -0,0 +1,65 @@
...
+- sbom state
+  -  **HTTP GET api//v2/sbom/{id}/sbom-status**
+  -  Reponse playload
+  ```json
+  {
+    "total_packages": "0",
+    "total_licenses": "0"
+  }
+  ```
+- Vulnerabilities state
```
ctron, line 38: What's the reason for having three "state" endpoints? Wouldn't it be simpler to have one?

docs/adrs/00007-sbom-dashbord.md
```diff
@@ -0,0 +1,65 @@
...
+  -  **HTTP GET api//v2/sbom/{id}/vulnerabilities-status**
```
ctron, line 39: the double slash looks wrong

docs/adrs/00007-sbom-dashbord.md
```diff
@@ -0,0 +1,65 @@
...
+- License info
+![sbom-info.png](sbom-info.png)
+  It consists of two parts:
+    - SBOM information (name, version, published date).
+    - License grouping - aggregating all unique licenses and visualizing them in a pie chart.
+
+## Decision
+Design an endpoint for each of these parts.
+- sbom state
+  -  **HTTP GET api//v2/sbom/{id}/sbom-status**
```
ctron, line 30: the double slash looks wrong

ctron, line 30: So far, we don't use `kebab-case`, why now? Some for the others.

docs/adrs/00007-sbom-dashbord.md
```diff
@@ -0,0 +1,65 @@
...
+    "total_low": "0"
+  }
+  ```
+
+- license state
+  -  **HTTP GET api//v2/sbom/{id}/license-status*
+  -  Reponse playload
+  ```json
+  {
+    "name": "Sbom-name",
```
ctron, line 56: Why do we have additional SBOM information (name, version, …) here. But not in the others?

docs/adrs/00007-sbom-dashbord.md
```diff
@@ -0,0 +1,65 @@
...
+    "version": "0",
+    "published_date": "",
+    "licenses": [
+      { "name:": "Apache-2.0", "count": "10"},
```
ctron, line 60: To my understanding some SPDX licenses are actually license expressions. How do we handle those?

docs/adrs/00007-sbom-dashbord.md
```diff
@@ -0,0 +1,65 @@
...
+- Vulnerabilities state
+  -  **HTTP GET api//v2/sbom/{id}/vulnerabilities-status**
+  - -  Reponse playload
+  ```json
+  {
+    "total_vulnerabilities": "0",
+    "total_critical": "0",
+    "total_high": "0",
+    "total_medium": "0",
+    "total_low": "0"
```
ctron, line 47: What about "none"?

docs/adrs/00007-sbom-dashbord.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Decision
+Design an endpoint for each of these parts.
+- sbom state
+  -  **HTTP GET api//v2/sbom/{id}/sbom-status**
+  -  Reponse playload
+  ```json
+  {
+    "total_packages": "0",
+    "total_licenses": "0"
```
ctron, line 35: Wouldn't that information make more sense in the license state endpoint?

ctron (changes requested):
I am missing the following information in this document:

* What the meaning/definition of the different "total" values? How can we produce this information?
* How do we handle custom licenses and complex SPDX expressions?
* What's the `id` path variable exactly? We have different ways of addressing SBOMs at the moment. Which one do we use here?
* 

I also think that we might want to think about:

* Who's eligible to retrieve that information (authorization)
* Does it make sense caching this information? If so, how?

docs/adrs/00007-sbom-dashbord.md
bxf12315: We already have `spdx_licenses` to store the standard SPDX IDs parsed from license expressions.

bxf12315: Indicates how many times the MIT license appears.

ctron: So `MIT OR LGPL-2.1-only` would show up as `MIT` and `LGPL-2.1-only`? That feels wrong.

docs/adrs/00007-sbom-dashbord.md
ctron: To my understanding `none` is a dedicated severity. So there can 5 vulnerabilities of type `none`.

docs/adrs/00007-sbom-dashbord.md
bxf12315: Why is it wrong?

docs/adrs/00007-sbom-dashbord.md
bxf12315: you are right.

PR #1576, opened 2025-04-28
fix: remove advisory_vulnerability gist index

dejanb:
The gist index on advisory_vulnerability(vulnerability_id) slows down significantly (4-5x) any join with that table on vulnerability_id. The database chooses to use gist index over btree one no matter what I tried.
As a result the correlating advisories with a large SBOM is now much faster (e.g. 260s -> 60s). It's still not great, but it's a big improvement.
On the other hand I don't see slowdown in other endpoints when this index is dropped. I don't think we ever directly search this table.


JimFuller-RedHat (approved):
LGTM

PR #1578, opened 2025-04-28
ci: allow backports with comments

carlosthe19916:
The main branch backport.yaml workflow requires the appropriate label to be added before the PR is merged.

If for any reason we forgot to add the label and we still need to backport a PR after it has been merged then:
- Add the appropriate label to indicate to which branch we need to backport
- Add a comment with `/backport` and it should trigger a backport even though the PR has been merged already.

Since the PR has been merged already hence CLOSED I think it is better for the user to make a conscious decision typing `/backport` to trigger a backport

### DIFF
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -6,12 +6,30 @@ on:
       - main
     types:
       - closed
+  issue_comment:
+    types:
+      - created
 
 jobs:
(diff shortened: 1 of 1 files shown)


ctron (approved):
I think that making the decision post-merge is actually the most common way we do it right now.

Issue #1543, opened 2025-04-09
create rest-api tutorial


JimFuller-RedHat:
There are/will be a lot of questions about how to use the REST API - unfortunately our generated rest api docs is far too coarse grained eg. there is no comprehensive description of q= ?

I propose:

1) we update where possible openapi descriptions
2) create a docs/rest-api-tutorial.md

jcrossley3:
This seems like a good place to put links for the (admittedly terse) docs we do have:

- The [original issue](https://github.com/trustification/trustify/issues/9) requesting it
- The [rustdocs](https://github.com/trustification/trustify/blob/16d6066d7e2f4ad7ae8a6f54fe94b4171971c3bc/common/src/db/query.rs#L33-L71) implementing it

i386x:
@JimFuller-RedHat Not exactly a tutorial, but I am slowly started to document an API differences. First attempt: #1555

PR #1585, opened 2025-04-29
feat: allow string arrays to be queried in the q= syntax

### LINKED ISSUE #1558: The query language should probably support arrays
Something along the lines of `WHERE "sbom"."authors" @> '{"Organization: Anchore, Inc"}'`

Might need to call `array_to_string` for the LIKE operator.

Not sure if the [regular array operators](https://www.postgresql.org/docs/9.1/functions-array.html) make sense or [behave like `Value::Array`](https://github.com/trustification/trustify/blob/a5ecedad2b5d37b2ee6469400ef1f526a90874be/common/src/db/query/value.rs#L129-L154).

jcrossley3:
Fixes #1558

### DIFF
--- a/common/src/db/query/columns.rs
+++ b/common/src/db/query/columns.rs
@@ -1,11 +1,10 @@
 use std::collections::{BTreeMap, BTreeSet};
-use std::fmt::{Display, Formatter};
+use std::fmt::Display;
 
 use chrono::Local;
 use human_date_parser::{ParseResult, from_human_time};
 use sea_orm::{
-    ColumnTrait, ColumnType, EntityTrait, IntoIdentity, Iterable, Value as SeaValue,
-    entity::ColumnDef, sea_query,
+    ColumnTrait, ColumnType, EntityTrait, IntoIdentity, Iterable, Value as SeaValue, sea_query,
 };
 use sea_query::{
     Alias, ColumnRef, Expr, ExprTrait, Func, IntoColumnRef, IntoIden, SimpleExpr,
@@ -25,35 +24,6 @@ pub struct Columns {
     translator: Option<Translator>,
     json_keys: BTreeMap<&'static str, ColumnRef>,
     exprs: BTreeMap<&'static str, (SimpleExpr, ColumnType)>,
-}
-
-impl Display for Columns {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        for (r, ty) in &self.columns {
-            writeln!(f)?;
-            match r {
-                ColumnRef::SchemaTableColumn(_, t, c) | ColumnRef::TableColumn(t, c) => {
-                    write!(f, "  \"{}\".\"{}\"", t.to_string(), c.to_string())?
-                }
-                ColumnRef::Column(c) => write!(f, "  \"{}\"", c.to_string())?,
-                _ => write!(f, "  {r:?}")?,
-            }
-            write!(f, " : ")?;
-            match ty {
-                ColumnType::Text | ColumnType::String(_) | ColumnType::Char(_) => {
-                    write!(f, "String")?
-                }
-                ColumnType::Enum { name, variants } => write!(
-                    f,
-                    "Enum({}) {:?}",
-                    name.to_string(),
-                    variants.iter().map(|v| v.to_string()).collect::<Vec<_>>()
-                )?,
-                t => write!(f, "  {t:?}")?,
-            }
-        }
-        Ok(())
-    }
 }
 
 pub trait IntoColumns {
@@ -94,11 +64,9 @@ impl Columns {
     }
 
     /// Add an arbitrary column into the context.
-    pub fn add_column<I: IntoIdentity>(mut self, name: I, def: ColumnDef) -> Self {
-        self.columns.push((
-            name.into_identity().into_column_ref(),
-            def.get_column_type().clone(),
-        ));
+    pub fn add_column<I: IntoIdentity>(mut self, name: I, ty: ColumnType) -> Self {
+        self.columns
+            .push((name.into_identity().into_column_ref(), ty));
         self
     }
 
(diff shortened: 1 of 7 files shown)


common/src/db/query/columns.rs
```diff
@@ -190,12 +158,18 @@ impl Columns {
...
     ) -> Result<SimpleExpr, Error> {
         self.for_field(field).and_then(|(lhs, ct)| {
-            match (value.to_lowercase().as_str(), operator) {
-                ("null", Operator::Equal) => Ok(lhs.is_null()),
-                ("null", Operator::NotEqual) => Ok(lhs.is_not_null()),
-                (_, Operator::Like) => Ok(lhs.ilike(like(value))),
-                (_, Operator::NotLike) => Ok(lhs.not_ilike(like(value))),
-                _ => parse(value, &ct).map(|rhs| lhs.binary(operator, rhs)),
+            match (value.to_lowercase().as_str(), operator, &ct) {
+                ("null", Operator::Equal, _) => Ok(lhs.is_null()),
```
ctron, line 162: I know the PR doesn't change this, but this feels like a massive mistake: https://www.wired.com/story/null-license-plate-landed-one-hacker-ticket-hell/

ctron (changes requested):
I'd like to have the functionality, but it feels like the syntax is moving more into a problematic direction. Creating all kinds of corner cases.

* How would I search for `foo` equal "empty string"
* How do I search for `foo` equal `"null"`
* What will happen when the user enters `%`?
* What will happen when the user enters `|` in the UI? What needs to happen to support searching for `|`?

The search syntax seems to get more corner cases and special characters. With no documentation on how to escape the special meaning.

I think there should be a more "top down" approach. Before we're making the existing parsing and syntax more complex.
*

jcrossley3:
> I'd like to have the functionality, but it feels like the syntax is moving more into a problematic direction. Creating all kinds of corner cases.

This PR doesn't change the syntax at all.

> * How would I search for `foo` equal "empty string"

I wasn't aware of that use case. Do we expect empty strings within arrays -- the subject of this PR -- returned from the api? If so, what would you suggest as a way to accomplish that? 

> * How do I search for `foo` equal `"null"`

At present, we treat a value of "null" -- `q=foo=null` -- special when the operator is either `=` or `!=`. We have anecdotal evidence that it's intuitive for the user. Searching for strings containing the literal value `"null"` is possible with either a full text search -- `q=null` -- or a LIKE filter -- `q=foo~null`, but if you feel strongly that a field containing the literal string `"null"` should also be found with the `=` operator, create an issue, but that new behavior is beyond the scope of this PR.

> * What will happen when the user enters `%`?

Nothing surprising. Characters treated specially for the SQL ILIKE operator are escaped prior to being included in a SQL statement. We have unit tests demonstrating this.

> * What will happen when the user enters `|` in the UI? What needs to happen to support searching for `|`?

Users will need to prefix any of our query language's special characters with a backslash, e.g. `q=foo\|bar`

> The search syntax seems to get more corner cases and special characters. With no documentation on how to escape the special meaning.

This PR adds no new syntax. And there is already an [open issue](https://github.com/trustification/trustify/issues/1543) for the lacking docs.

> I think there should be a more "top down" approach. Before we're making the existing parsing and syntax more complex. *

Your hatred of the query language from its inception is well known and not entirely unwarranted. I wish your criticism was more constructive and less fear-based. I wish you would acknowledge the benefits of our design choices as well as the costs we traded for them.

carlosthe19916:
@ctron @jcrossley3 my two cents about the question "How would I search for foo equal "empty string""

There are 2 types of labels:
- Single String labels (we do not support this in Trustify)
- Key-Value labels

### String based labels
- If we think in our Gmail inbox. We label emails using a SINGLE String.
image

### Key-Value based labels
Trustify is using this approach.

### How to search for foo equal "empty string"

This should go to the first category: "string based labels". so the query does not need to have things like `foo=null` bot rather something like `foo=something`.

I admit `foo=something` does not answer the original question but that makes me think that we cannot answer that kind of question in Gmail either. We cannot search all emails that do not have the label "mylabel". Instead, we search for emails that do that the label `foo=something`.

I guess we can create really complex questions and implement them, but unless we are sure we might be solving a problem that does not exist.

jcrossley3:
> I guess we can create really complex questions and implement them, but unless we are sure we might be solving a problem that does not exist.

:100:

ctron:
I am sure I'd want to search for "label equals empty string", that's basically a "tag". I also think I'd like to search for `label=null`, as that might just happen by people assigning `null`. Sure, you can tell me now I don't want this. Alright.

As of what's new in this syntax and what's existing, I don't know. As there's no documentation on how it works today. From the title of the PR and the code changes, I assumed it is a new feature.

I think instead of adding more and more features, which I understand this PR does, we should take a step back and define what we need. Without the need for a lot of escaping, and magic characters. Something which you can write in a text editor, and defines a search model, from which you could even deduct a documentation of individual fields that search supports. And which supports things like grouping of expressions (like: either "expression A" or not ("expression B") ).

And most likely, that is already resolved by an existing syntax. Like carlos mentioned, like we had in v1.

jcrossley3:
All this PR does is extend existing behavior to a new type: arrays of strings. I'm just anticipating a potential bug report when someone tries to filter on a field of type `Vec<String>`.

In the interest of "completeness", I'd like to merge this PR. You've effectively blocked that merge without requesting any specific changes. 

Blocking this PR in order to "step back and define what we need" is an abuse of our process. Please create a separate issue for that.

jcrossley3:
Others? Are there others?

bobmcwhirter:
LGTM!

bobmcwhirter (approved):

jcrossley3:
> I am sure I'd want to search for "label equals empty string", that's basically a "tag". I also think I'd like to search for `label=null`, as that might just happen by people assigning `null`. Sure, you can tell me now I don't want this. Alright.

I just looked into how we might support searching for empty strings, and realized that we already can: use `label=`. So you have a way to search for either empty strings or nulls. I added tests/docs for it in #1615

PR #1588, opened 2025-04-30
fix: ensure middleware sboms work with analysis graph and optimise graph cache

### LINKED ISSUE #1559: Slow latest query for Quarkus / Domino generated SBOMs
Using the latest commit of Trustify [a5ecedad](https://github.com/trustification/trustify/commit/a5ecedad2b5d37b2ee6469400ef1f526a90874be) with the [Quarkus SBOMs](https://github.com/RedHatProductSecurity/trustshell/tree/c0fcea4a585fb0107561bd43127938aefc6b2652/docs/sboms/maven/quarkus/) uploaded from trustshell examples the follow query takes around 12 secs on my machine:

`api/v2/analysis/latest/component?q=pkg%3Amaven%2Fio.vertx%2Fvertx-core%40&limit=10`

Compared with the non-latest version of the query, which returns in less than 2 seconds:

`api/v2/analysis/component?q=pkg%3Amaven%2Fio.vertx%2Fvertx-core%40&limit=10`

Using HTTPX Python library with the default (10s) timeout I get the following Exception:

`httpcore.ReadTimeout: timed out`

JimFuller-RedHat:
fixes #1559 because the previous q= query was a hack (against node_id) and middleware sboms dont generate spdx ids in the same way (so we could no longer 'cheat').

* optimised some analysis graph latest queries using postgres **PARTITION_BY**
* latest queries ensures only the sboms that are needed are loaded into analysis graph
* converted sql into sea_orm so we can use trustify query capabilities
* fixed some tests

### DIFF
--- a/modules/analysis/src/service/load.rs
+++ b/modules/analysis/src/service/load.rs
@@ -269,6 +269,7 @@ impl InnerService {
                 .join(JoinType::Join, sbom_package::Relation::Purl.def())
                 .filter(sbom_package_purl_ref::Column::QualifiedPurlId.eq(purl.qualifier_uuid()))
                 .select_only()
+                .distinct()
                 .column(sbom_node::Column::SbomId)
                 .into_query(),
             GraphQuery::Component(ComponentReference::Cpe(cpe)) => sbom_node::Entity::find()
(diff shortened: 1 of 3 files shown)


modules/analysis/src/service/load.rs
```diff
@@ -346,6 +347,267 @@ impl InnerService {
...
+                  SELECT distinct sbom_id
+                  FROM (
+                      SELECT sbom.sbom_id, sbom.published, product.id, cpe.id,
+                      RANK() OVER (PARTITION BY cpe.id ORDER BY sbom.published DESC) AS rank
+                      FROM sbom_node
+                      LEFT JOIN sbom ON sbom.sbom_id = sbom_node.sbom_id
+                      LEFT JOIN product_version on sbom.sbom_id = product_version.sbom_id
+                      LEFT JOIN product on product.id = product_version.product_id
+                      LEFT JOIN sbom_package_cpe_ref on sbom.sbom_id = sbom_package_cpe_ref.sbom_id
+                      LEFT JOIN cpe ON sbom_package_cpe_ref.cpe_id = cpe.id
```
jcrossley3, line 368: Are all these SELECT's identical except for the WHERE clause? I'd prefer we keep all the redundant parts in a single variable.

modules/analysis/src/service/load.rs
```diff
@@ -346,6 +347,267 @@ impl InnerService {
...
+                    ptype: Option<&'a str>,
+                    namespace: Option<&'a str>, // Represents the full namespace string
+                    name: Option<&'a str>,
+                    version: Option<&'a str>,
+                    qualifiers_str: Option<&'a str>, // Qualifiers as a single raw string
+                    subpath: Option<&'a str>,
+                }
+
+                /// WARNING: highly non-compliant parsing of partial pURL.
+                fn basic_non_compliant_parse_purl(purl_str: &str) -> Option<BasicPurlParts> {
```
jcrossley3, line 487: I'm not really clear on what you're trying to achieve here. Isn't this behavior already captured in `Purl::from_str()`?

jcrossley3 (commented):
So `RANK()` is the critical change here, correct? I'm not really familiar with it. I don't love all the redundant SQL, and I'm not clear on why we're replicating `Purl::from_str` logic.

Help me help you. :wink:

modules/analysis/src/service/load.rs
JimFuller-RedHat: I dont think so - basically we try to optimistically parse purl string ... we do not fail, we just try to crack into parts based on what we see

modules/analysis/src/service/load.rs
JimFuller-RedHat: I was thinking the optimisation here would be to rewrite all those into sea_orm ... how I have no clue ... less enthusiastic of DRY on SQL statements themselves (and attendent string munging).

modules/analysis/src/service/load.rs
jcrossley3: I'm happy to help with the sea_orm/RANK stuff, because I know that's a pain.

I hate that we've gotten into a situation where we don't now agree on which requirements we agreed on then, but I understood quite clearly that "partial purl searches" were not a requirement based on [these comments](https://github.com/trustification/trustify/pull/1534#discussion_r2038890651), which were later confirmed in a video chat with you and @ctron. 

We can ammend those requirements, of course, but if we do, I'd like for both the `/latest/component` and `/component` endpoints to work the exact same way. The corpus for the latter is [here](https://github.com/trustification/trustify/blob/0b4efc5f7364d9d34a53f590783004d57770edfc/modules/analysis/src/endpoints/tests/mod.rs#L394-L406), so we should refactor the code such that those (and any others we need) are used for both endpoints.

JimFuller-RedHat: I think we (more like I) are mis-communicating.

1) we want status quo as today with component searches

2) we may (in addition) want behavior on q=partialpurl which attempts to translate 

```
q=somecomponent@12.0 
```
into 
```
q=name~somecomponent&version~12.0
```

you might view this potentially new requirement as if the existing implementation does it wrong ... it does not ... you may also think the query implementation would be better if we did a straight string search in the classic tables - not needed as we already have that in graphs (as we store purls as strings). Note that this query is a pre filter to decide which sboms to ensure are loaded into graphs. 

As mentioned - its likely we can get the client to do the following call

```
q=name~somecomponent&version~12.0
```

As mentioned - I think that just means we need help on converting the following sql queries using RANK into sea_orm (name query example) so we can also use existing query mechanisms:

```
                let sql = r#"
                  SELECT distinct sbom_id
                  FROM (
                      SELECT sbom.sbom_id, sbom.published, product.id, cpe.id,
                      RANK() OVER (PARTITION BY cpe.id ORDER BY sbom.published DESC) AS rank
                      FROM sbom_node
                      LEFT JOIN sbom ON sbom.sbom_id = sbom_node.sbom_id
                      LEFT JOIN product_version on sbom.sbom_id = product_version.sbom_id
                      LEFT JOIN product on product.id = product_version.product_id
                      LEFT JOIN sbom_package_cpe_ref on sbom.sbom_id = sbom_package_cpe_ref.sbom_id
                      LEFT JOIN cpe ON sbom_package_cpe_ref.cpe_id = cpe.id
                      WHERE sbom_node.name = $1
                  ) AS subquery
                  WHERE rank = 1;
                  "#;
                let stmt =
                    Statement::from_sql_and_values(DatabaseBackend::Postgres, sql, [name.into()]);
                let rows = connection.query_all(stmt).await?;
                rows.into_iter()
                    .filter_map(|row| {
                        row.try_get_by_index::<Uuid>(0)
                            .ok()
                            .map(|sbom_id| sbom_id.to_string())
                    })
                    .collect::<Vec<String>>()
```

hoping that is helpful

modules/analysis/src/service/load.rs
```diff
@@ -346,6 +347,267 @@ impl InnerService {
...
+                  FROM (
+                      SELECT sbom.sbom_id, sbom.published, product.id, cpe.id,
+                      RANK() OVER (PARTITION BY cpe.id ORDER BY sbom.published DESC) AS rank
+                      FROM sbom_node
+                      LEFT JOIN sbom ON sbom.sbom_id = sbom_node.sbom_id
+                      LEFT JOIN product_version on sbom.sbom_id = product_version.sbom_id
+                      LEFT JOIN product on product.id = product_version.product_id
+                      LEFT JOIN sbom_package_cpe_ref on sbom.sbom_id = sbom_package_cpe_ref.sbom_id
+                      LEFT JOIN cpe ON sbom_package_cpe_ref.cpe_id = cpe.id
+                      WHERE sbom_node.name = $1
```
jcrossley3, line 399: It doesn't seem like we need the product-related joins in any of these variants, do we? Tests pass without them.

```suggestion
                      SELECT sbom.sbom_id, sbom.published, cpe.id,
                      RANK() OVER (PARTITION BY cpe.id ORDER BY sbom.published DESC) AS rank
                      FROM sbom_node
                      LEFT JOIN sbom ON sbom.sbom_id = sbom_node.sbom_id
                      LEFT JOIN sbom_package_cpe_ref on sbom.sbom_id = sbom_package_cpe_ref.sbom_id
                      LEFT JOIN cpe ON sbom_package_cpe_ref.cpe_id = cpe.id
                      WHERE sbom_node.name = $1
```

modules/analysis/src/service/load.rs
jcrossley3: Honestly, it would take me a while to figure out how to map that SQL into sea_orm, if I even could, so it's not really fair of me to ask you to do it.

modules/analysis/src/service/load.rs
```diff
@@ -346,6 +347,267 @@ impl InnerService {
...
+                    name: Option<&'a str>,
+                    version: Option<&'a str>,
+                    qualifiers_str: Option<&'a str>, // Qualifiers as a single raw string
+                    subpath: Option<&'a str>,
+                }
+
+                /// WARNING: highly non-compliant parsing of partial pURL.
+                fn basic_non_compliant_parse_purl(purl_str: &str) -> Option<BasicPurlParts> {
+                    // start with "pkg:" if it exists
+                    let remaining = purl_str.strip_prefix("pkg:")?;
```
jcrossley3, line 489: This line effectively requires all `q=` values to start with "pkg:". This entirely defeats the purpose of the q parameter, and I think it'll lead to confusion. The logic prevents me from querying say `q=quarkus&purl:version>2`. 

We already have a variant for these endpoints that takes an actual purl, i.e. something that _must_ start with "pkg:". Perhaps we change the logic of that variant to have the behavior you're attempting to implement here, a kind of "loose" parsing of a purl?

jcrossley3: Or maybe introduce a new parameter, say "partial", that does this logic? I'd prefer we not abuse the 'q' parameter. Since it's used throughout the app, I think it's behavior should be consistent.

JimFuller-RedHat: as prev mentioned 'forget about all that' ... will force push and remove - once I puzzle out how to get windowing with RANK into sea_orm (or how to use filtering_with on statements).

modules/analysis/src/service/load.rs
JimFuller-RedHat: we can do it together ;)

ctron (approved):
I'll follow up with a PR, clearing out some duplicate code.

PR #1600, opened 2025-05-05
chore: clean up labels API, add tests, handle removing, add docs

ctron:
While documenting some examples, I noted that tags with empty values are a problem when patch-updating. Fixing that, adding tests, align tests between sbom and advisories. And give some examples.


JimFuller-RedHat (approved):
quick scan, LGTM

PR #1604, opened 2025-05-06
doc: adr for handling vulnerability scores and severities

dejanb:
This PR contains ADR for handling vulnerability scores and severities. It covers current state of things in specs that we are using and proposing API changes to properly reflect data that users are interested in.

### DIFF
--- a/docs/adrs/00004-advisory-scores.md
+++ b/docs/adrs/00004-advisory-scores.md
@@ -1,0 +1,125 @@
+# 00004. Vulnerability scores
+
+Date: 2025-05-06
+
+## Status
+PROPOSED
+
+## Context
+
+There are various vulnerability scores format versions used in different advisories. The purpose of this document is to summarize all these options and propose what Trustify need to store and return through its endpoints.
+
+## Scores
+
+Common Vulnerability Scoring System (CVSS) scores are used to determine severity of the vulnerabilities. The system evolved over they years, so we have multiple versions of it in use today.
+
+The version 2 was used primarily until late 2015. Since then version 3.0 was mainly used until 2019, when it was upgraded to 3.1. Staring late 2024, CVE project started using CVSS 4.0 version of scoring.
+
+## Advisory format
+
+CVE JSON Schema version 5 and OSV schema that Trustify uses to ingest data from these sources support all aforementioned CVSS score formats.
+
+OSV format supported only v2 and v3 until version 1.6.2 which was released in 2024. Most of the documents Trustify is importing today used 1.4.0 of the OSV schema.
+
+CSAF version 2.0 (still official version), supports only v2 and v3 of CVSS standard, while the support for v4 will be introduced in 2.1 version of the CSAF standard.
+
+## Vulnerability score
+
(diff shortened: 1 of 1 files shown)


docs/adrs/00004-advisory-scores.md
```diff
@@ -0,0 +1,119 @@
...
+
+Advisories can "override" the original vulnerability score, depending on the further analysis or the context. Both OSV and CSAF support this. Each vulnerability described by the advisory can be attached multiple scores (but that is rare in practice).
+
+There's no rush in supporting CVSS v4 for these standards as they are still not widely used. Again, CVSS v2 would be good to have.
+
+## Aggregate severity
+
+OSV record describes only one CVE vulnerability, while CSAF can describe multiple vulnerabilities in a single document.
+
+That's why CSAF introduces the concept of the "aggregate severity" considering the whole document, providing the useful information on how to deal with the whole advisory. It is independent of the CVSS standard and provides just a textual representation (the usual values are "Critical", "Important", "Moderate").
```
ctron, line 40: I don't think that's correct. I think that the spec describes a vendor independent scheme. Allowing a vendor to define their own list. The `namespace` making it clear, which vendor defined it.

Also see: https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#3212-document-property---aggregate-severity

ctron:
I absolutely like the detail of this ADR. The context and history! Awesome!

I am wondering why the data model doesn't have a "type" field (v2, v3, v4?) Did I overlook it? Isn't it necessary?

dejanb:
> I absolutely like the detail of this ADR. The context and history! Awesome!
> 
> I am wondering why the data model doesn't have a "type" field (v2, v3, v4?) Did I overlook it? Isn't it necessary?

We definitely need it. I added it (version) in the CVSSVector object just to make it clear. Note that this is not the exact data model proposal, but just proposed pseudo-API changes.

ctron (approved):

PhilipCattanach:
@dejanb - what's not clear to me is what is going to change/

I think the Vulnerability score is straight forward as it is a 1:1

The Advisory can have a 1:M to Vulnerability so are we still intending on presenting the Aggregate score on the Advisory List and Detail Screen? 

Or should each Advisory List row be expandable to show the individual CVEs and their scores from that Advisory (essentially what the Advisory Vulnerability list scree does currently)?

A couple of concrete examples for the AS IS and the TO BE illustrating what the changes will provide would be super helpful.

dejanb:
@PhilipCattanach I was thinking to keep the ADR focused on describing how things will be in the future, as this transition is just a temporary.

I created a project to detail of the work we are doing to make it work

https://github.com/orgs/trustification/projects/11/views/1?template_dialog_tab=featured&layout_template=table

As for the concrete topics.

Vulnerabilities will only get scores from CVE files.
CSAF advisories should have "aggregate severity" displayed, but it doesn't make sense for others. All advisories will have all scores in them.

Note also, that this is nothing to with UI, just how we store and return data using API. UI is another discussion completely.

mrizzi (approved):
@dejanb thanks a lot: the ADR is precise and clear and it reads very well.

Issue #1589, opened 2025-04-30
Got ~26 seconds to delete this advisory `/api/v2/advisory/urn:uuid:ed944a21-a2a2-4d4d-811b-f4427318b6c2`


helio-frota:
* Using [this](https://github.com/trustification/trustify-load-test-runs/blob/main/compose.yaml#L106) dump 
* `deleteAdvisory.hurl: Success (1 request(s) in 26669 ms)`
* Running on pm-mode

PR #1617, opened 2025-05-08
fix: remove duplicate UNION in gc_purls SELECT

helio-frota:
Note: This does not address the performance issue described in https://github.com/trustification/trustify/issues/1589, but the duplication still needed to be resolved.

### DIFF
--- a/modules/fundamental/src/purl/service/gc_purls.sql
+++ b/modules/fundamental/src/purl/service/gc_purls.sql
@@ -62,7 +62,7 @@ WITH
         (
             SELECT * from deleted_base_purl
         ) UNION (
-            SELECT * from deleted_base_purl
+            SELECT * from deleted_versioned_purl
         ) UNION (
             SELECT * from deleted_qualified_purl
         )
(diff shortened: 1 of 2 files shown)


helio-frota:
fixed the test adding the required magic number to make it pass

ctron:
I did disble auto-merge so that we can run (and await) a scale test run.

helio-frota:
Just noticed this is being used only in `delete`  for `sbom` and `advisory` endpoints

```
➜  trustify git:(fix-gc-purls-sql) rg "purl_service\.gc_purls\(" -B 4 --glob "!test.rs"
modules/fundamental/src/sbom/endpoints/mod.rs
303-            let rows_affected = service.delete_sbom(v.head.id, &tx).await?;
304-            match rows_affected {
305-                0 => Ok(HttpResponse::NotFound().finish()),
306-                1 => {
307:                    let _ = purl_service.gc_purls(&tx).await; // ignore gc failure..

modules/fundamental/src/advisory/endpoints/mod.rs
138-        let rows_affected = state.delete_advisory(fetched.head.uuid, &tx).await?;
139-        match rows_affected {
140-            0 => Ok(HttpResponse::NotFound().finish()),
141-            1 => {
142:                let _ = purl_service.gc_purls(&tx).await; // ignore gc failure..
```

PR #1616, opened 2025-05-08
Revert #1585

ctron:
Tonight's performance regression seems to be cause by #1585:

```
HEAD: 
 3: test                 
   1: logon               |        6.90 |          6 |           9 |          7
   2: search_advisory     |        4919 |      4,769 |       5,173 |      4,769

minus b27418b4306f458e23665c70869a1b19ae1abd99,d914a349bc2342ad2f659c5c370f903561acf2d6:
 3: test                 
   1: logon               |        6.98 |          6 |          17 |          7
   2: search_advisory     |      301.93 |        193 |         460 |        300
```

PR #1618, opened 2025-05-08
Reapply "feat: allow string arrays to be queried in the q= syntax"

jcrossley3:
This reverts commit 4e8589b6099325b35c4be6e998bfce90970068bd.

### DIFF
--- a/modules/fundamental/src/advisory/service/mod.rs
+++ b/modules/fundamental/src/advisory/service/mod.rs
@@ -3,9 +3,9 @@ use crate::{
     advisory::model::{AdvisoryDetails, AdvisorySummary},
 };
 use sea_orm::{
-    ActiveModelTrait, ActiveValue::Set, ColumnTrait, ColumnTypeTrait, ConnectionTrait,
-    DatabaseBackend, DbErr, EntityTrait, FromQueryResult, IntoActiveModel, IntoIdentity,
-    QueryResult, QuerySelect, QueryTrait, RelationTrait, Select, Statement, TransactionTrait,
+    ActiveModelTrait, ActiveValue::Set, ConnectionTrait, DatabaseBackend, DbErr, EntityTrait,
+    FromQueryResult, IntoActiveModel, IntoIdentity, QueryResult, QuerySelect, QueryTrait,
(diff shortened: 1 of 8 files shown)


modules/fundamental/src/advisory/service/mod.rs
```diff
@@ -94,7 +94,10 @@ impl AdvisoryService {
             .filtering_with(
                 search,
                 Columns::from_entity::<advisory::Entity>()
-                    .add_columns(source_document::Entity)
+                    .add_column(
+                        source_document::Column::Ingested,
+                        ColumnType::TimestampWithTimeZone,
+                    )
```
jcrossley3, line 100: I believe this is performance culprit. I lazily added all the `source_document::Entity` columns instead of just the one we need for this query, `source_document::Column::Ingested`

jcrossley3:
The original, now-reverted PR had 2 commits. One was adding the ability to query string arrays, the essential change. As evidenced by the first scale test above, it has minimal impact on perf:
```
GET	search_advisory	102 (+62)	0	1191.47 (-20816.60)	580 (-14645)	1812 (-24476)	0.34 (+0.21)	0.00 (+0.00)
```
The second commit was a refactoring of the `db::query::columns::Columns` api in which I inadvertently added more columns than were needed to the advisory query. After adding that change to this PR and running a scale test, we got our perf problem back:
```
GET	search_advisory	40 (0)	0	22836.88 (+828.80)	18754 (+3529)	28072 (+1784)	0.13 (+0.00)	0.00 (+0.00)
```
The third commit in this PR is the fix for this, as shown by the results of the 3rd scale test:
```
GET	search_advisory	98 (+58)	0	1143.34 (-20864.74)	401 (-14824)	1648 (-24640)	0.33 (+0.19)	0.00 (+0.00)
```
So I think this PR restores the feature without the perf impact.

mrizzi (approved):
Executed the test locally like it has been done in https://github.com/trustification/trustify/pull/1616#pullrequestreview-2824472830 to confirm the results and the scale tests result is successfully confirmed locally.
Thanks @jcrossley3 👍

PR #1619, opened 2025-05-08
fix: delete vulnerability and related product_status (TC-2519)

mrizzi:
https://issues.redhat.com/browse/TC-2519


mrizzi:
> I was looking at this a bit yesterday and it seems to me that it needs a bit more things to do. At least we need to clean up `advisory_vulnerability`. What I'm not 100% sure yet is if we should delete all related advisories as well, as it does not make too much sense to have advisory without underlying vulnerability.

I fully get what you mean.
I have to admit that I didn't apply any change on the `advisory_vulnerability` because, even if `vulnerability_id` column is part of the primary key (together with `advisory_id`) of that table, it isn't defined with the foreign key towards the `vulnerability` table and, in this I might have been wrong, I thought it had been done on purpose to let "intact" the advisory when a vulnerability is removed. https://github.com/trustification/trustify/blob/72f189f6ee03fc5b1a57cc93d5c1827c784f4105/migration/src/m0000010_init_up.sql#L3743

But now that you mentioned it, I checked the entity `advisory_vulnerability`'s definition https://github.com/trustification/trustify/blob/12b73235b54ae22d7a02dbc5c96e0e1e0b74125a/entity/src/advisory_vulnerability.rs#L32 and it looks like having the foreign key on vulnerability (just like it has one towards advisory) so maybe it's a further bug we've just spotted.

I think an advisory could be removed only if ALL the related vulnerabilities have been deleted.

@dejanb WDYT?

dejanb (approved):
LGTM

mrizzi:
After a deep review with @dejanb and @PhilipCattanach, it turned out that a vulnerability can not be inserted on its own (there's no `POST` verb support for the `vulnerability` resource) so it doesn't make any sense to have a way to delete a resource that can not be added/inserted.
Hence the decision to deprecate the endpoint in the next z-stream release with the goal of removing it in the next y-stream release.

modules/fundamental/src/vulnerability/endpoints/mod.rs
```diff
@@ -93,44 +92,22 @@ pub async fn get(
...
-            1 => {
-                tx.commit().await?;
-                Ok(HttpResponse::Ok().json(vuln))
-            }
-            _ => Err(Internal("Unexpected number of rows affected".into())),
-        }
-    } else {
-        Ok(HttpResponse::NotFound().finish())
-    }
+    Ok(HttpResponse::NotImplemented().finish())
```
ctron, line 110: I'm not against this. But "deprecation" would mean to me, that it is still there, but will be removed in a future version.

This effectively removes it (the functionality) today.

Maybe we can be bit more clear in the description of this: it never really worked, we disabled it for now?

PR #1635, opened 2025-05-15
chore: ignore spans shorter than 1ms

helio-frota:
before:

after:


helio-frota:
I think we have some ways to solve that issue with functions shorter than `1ms`:

* Remove the `#[instrument]` macro from the function (easy and intrusive)
* Create a custom spanExporter in rust (hard and I already spent ~2 hours yesterday with no success (also can be impacted with the evolution of otel-rust apis) ) 
* Delegate that to collector (easy and we are using the collector anyway so why not)

dejanb (approved):

PR #1638, opened 2025-05-16
fix: add sorting and pagination to importer report endpoint

### LINKED ISSUE #1636: GET "/v2/importer/{name}/report" sorting does not work
### Requirement:

I think The endpoint `GET /v2/importer/{name}/report` should have:

- Pagination: `GET /v2/importer/{name}/report?offset=0&limit=10`
- Sorting: `GET /v2/importer/{name}/sort?creation:asc`

But given the nature of the reports, I don't think we need filtering.

### Issue

- Sorting does not work.- I send the appropriate GET request and although there is no error generated, the result does not change the order.
- Pagination and Sorting are not exposed in the Openapi.yaml definition.

jcrossley3:
Fixes #1636

### DIFF
--- a/modules/importer/src/endpoints.rs
+++ b/modules/importer/src/endpoints.rs
@@ -10,7 +10,7 @@ use std::convert::Infallible;
 use trustify_auth::authorizer::Require;
 use trustify_auth::{CreateImporter, DeleteImporter, ReadImporter, UpdateImporter};
 use trustify_common::{
-    db::Database,
+    db::{Database, query::Query},
     model::{Paginated, PaginatedResults, Revisioned},
 };
 
@@ -281,6 +281,10 @@ async fn delete(
 #[utoipa::path(
     tag = "importer",
     operation_id = "listImporterReports",
+    params(
+        Query,
+        Paginated,
+    ),
     responses(
         (status = 200, description = "Retrieved importer reports", body = PaginatedResults<ImporterReport>),
     )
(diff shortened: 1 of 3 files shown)


carlosthe19916 (approved):

modules/importer/src/endpoints.rs
```diff
@@ -281,6 +281,10 @@ async fn delete(
 #[utoipa::path(
     tag = "importer",
     operation_id = "listImporterReports",
+    params(
+        Query,
+        Paginated,
+    ),
```
carlosthe19916, line 287: Just out of curiosity why is it that this block of code does not need something like:
```
params(
        ("name", Path, description = "Name of the Importer"),
        Query,
        Paginated,
    ),
```

I saw similar endpoints like https://github.com/trustification/trustify/blob/main/modules/fundamental/src/sbom/endpoints/mod.rs#L323 

and those endpoints use `("id", Path, description = "ID of the SBOM to get packages for"),` but not this endpoint.

I am not saying there is anything wrong, I actually tested the openapi.yaml generated by this PR and it worked like a charm, just asking to fix my ignorance on these matters :)

jcrossley3: I honestly don't know. I just copied it from another working endpoint. :smile:

jcrossley3: Might be because the `name` is contained within the `#[get("/v2/importer/{name}/report")]` macro already, so the endpoint wouldn't even be resolved without it? Not sure. But since it seems to be working, I'm gonna merge.

PR #1639, opened 2025-05-18
TC-1902 Add license support to the sbpm package query conditions.

bxf12315:


entity/src/sbom_package_license.rs
```diff
@@ -27,13 +28,31 @@ pub enum Relation {
...
-#[derive(Copy, Clone, Debug, strum::Display, Eq, PartialEq, EnumIter, DeriveActiveEnum)]
+#[derive(Copy, Clone, Debug, strum::Display, Eq, PartialEq, EnumIter, Enum, DeriveActiveEnum)]
 #[sea_orm(rs_type = "i32", db_type = "Integer")]
 pub enum LicenseCategory {
     Declared = 0,
     Concluded = 1,
 }
 
+impl LicenseCategory {
+    pub fn to_i32(self) -> i32 {
```
ctron, line 39: I don't think we need this one. I think you can just do:

```rust
LicenseCategory::Declared as i32
```

entity/src/sbom_package_license.rs
```diff
@@ -27,13 +28,31 @@ pub enum Relation {
...
+        self as i32
+    }
+}
+
+impl TryFrom<i32> for LicenseCategory {
+    type Error = &'static str;
```
ctron, line 45: `&'static str` isn't a very good error type. With `thiserror` you can easily accomplish the same, but with a proper error type.

modules/fundamental/src/sbom/endpoints/test.rs
```diff
@@ -14,6 +14,50 @@ use trustify_module_ingestor::{model::IngestResult, service::Format};
...
+    assert_eq!(result.total, 5);
+
+    let id = ctx
+        .ingest_document("spdx/OCP-TOOLS-4.11-RHEL-8.json")
+        .await?
+        .id
+        .to_string();
+
+    let result = query(&app, &id, "name=resolve&Text~MIT").await;
+    assert!(result.items.iter().any(|p| p.name == "resolve"
```
ctron, line 45: I think we have some "contains subset" functions, which make it clearer what the tests are expecting. Unless they don't work, please use those instead.

modules/fundamental/src/sbom/endpoints/test.rs
```diff
@@ -104,14 +148,14 @@ async fn filter_packages(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
         .to_string();
 
     let result = query(&app, &id, "").await;
-    assert_eq!(result.total, 41);
+    assert_eq!(result.total, 51);
```
ctron, line 151: Why do we get more results now?

ctron (changes requested):
I am worried about tests just expected more results know. I'd like to understand first "why" that's the case.

bxf12315:
> I am worried about tests just expected more results know. I'd like to understand first "why" that's the case.

I already add the detailed on last comment of https://issues.redhat.com/browse/TC-1902. It explains why the count number is change.

ctron:
> > I am worried about tests just expected more results know. I'd like to understand first "why" that's the case.
> 
> I already add the detailed on last comment of https://issues.redhat.com/browse/TC-1902. It explains why the count number is change.

This isn't information which is accessible upstream. Please provide it here too.

ctron:
> > > > I am worried about tests just expected more results know. I'd like to understand first "why" that's the case.
> > > 
> > > 
> > > I already add the detailed on last comment of https://issues.redhat.com/browse/TC-1902. It explains why the count number is change.
> > 
> > 
> > This isn't information which is accessible upstream. Please provide it here too.
> 
> In order to align with license export, I need to add two new columns to the package table to indicate the license type and license expression. This will introduce some changes:
> 
> When processing SPDX, each package will generate two records, because there are two types of license: Concluded and Declared. For CycloneDX, a package may contain multiple license records, so multiple records will also be generated.

That's a major issue. A package has a meaning. Duplicate packages, because some values aren't scalar, but lists, is reporting wrong information. If a package has multiple licenses/license types, then the information must be presented as such.

bxf12315:
@ctron @mrizzi please check again.

mrizzi:
@ctron every comment should have been addressed, the PR has been rebased on the latest `main` (adopting the feature flag for graphql attribute) and the scale test has been executed with results consistent with current `main` branch.

Do you mind having a further check for this PR please?

ctron (approved):
Looks great. Thanks! :heart:

PR #1640, opened 2025-05-19
fix: set vulnerability score from cve advisory

dejanb:
This PR changes how vulnerability scoring is set. Instead of calculating an average
from all advisories, it is now parsed from the CVE file. The score and severity values
are stored in the vulnerability database.
I tried at first looking for an cve advisory and retrieving the value, but the query gets even more complicated and slow than calculating an average. And as it is the integral part of the vulnerability, I think the value belongs in the table.
With this PR I didn't change the current API, so it's still called `average_score` and `average_severity`. I will tackle that in a separate PR (I need to think of how deprecating current values will work).
The search query is changed though. Please see the comment below for the details.

### DIFF
--- a/modules/fundamental/src/vulnerability/service/mod.rs
+++ b/modules/fundamental/src/vulnerability/service/mod.rs
@@ -5,24 +5,16 @@ use crate::{
     vulnerability::model::{VulnerabilityDetails, VulnerabilitySummary},
 };
 use futures_util::{TryFutureExt, TryStreamExt};
-use sea_orm::{
-    EntityTrait, FromQueryResult, IntoIdentity, QuerySelect, QueryTrait, Statement, StreamTrait,
-    prelude::*,
-};
-use sea_query::{ColumnRef, Func, IntoColumnRef, IntoIden, SimpleExpr};
+use sea_orm::{EntityTrait, Statement, StreamTrait, prelude::*};
 use trustify_common::{
     db::{
-        limiter::LimiterAsModelTrait,
-        multi_model::{FromQueryResultMultiModel, SelectIntoMultiModel},
+        limiter::LimiterTrait,
         query::{Columns, Filtering, Query},
     },
     model::{Paginated, PaginatedResults},
     purl::{Purl, PurlErr},
 };
-use trustify_entity::{
-    cvss3::{self, Severity},
-    vulnerability,
-};
+use trustify_entity::vulnerability;
 use trustify_module_ingestor::common::Deprecation;
 
 #[derive(Default)]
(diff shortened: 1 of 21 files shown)


modules/fundamental/src/vulnerability/service/mod.rs
```diff
@@ -35,100 +27,25 @@ impl VulnerabilityService {
...
-                    SimpleExpr::FunctionCall(Func::avg(SimpleExpr::Column(
-                        trustify_entity::cvss3::Column::Score.into_column_ref(),
-                    ))),
-                )),
-                "average_severity",
-            )
-            .group_by(vulnerability::Column::Id);
-
-        let mut outer_query = vulnerability::Entity::find();
+        search.q = search.q.replace("average_severity", "base_severity");
```
dejanb, line 35: This is a temporary workaround, until we update UI to use new values.
@jcrossley3 I tried playing with the aliases, but failed miserably. If you have any idea on how it could be done, please let me know. Although, this should be removed altogether soon, so I wouldn't spend too much time on it.

dejanb: OK, removed the "aliasing" and created UI PR (https://github.com/trustification/trustify-ui/pull/499).

It is definitely a breaking change, but I didn't see those notes used in trustify project. Any pointers on how it should look?

On the other note, this is just one change in the whole set that will change the API and need data reimport. So we'll need custom release notes once it's done.

modules/ingestor/src/service/advisory/cve/loader.rs
```diff
@@ -288,6 +274,44 @@ impl<'g> CveLoader<'g> {
...
         };
 
+        let mut scores = vec![];
+        let mut base_score = None;
+        let mut base_severity = None;
+        if let Cve::Published(published) = cve.clone() {
+            let all_metrics = published.containers.cna.metrics.iter().chain(
+                published
+                    .containers
+                    .adp
```
ctron, line 284: I am not sure adding the ADP scores like that is the right approach. ADPs can amend/augment what the CNA published. But the following code doesn't seem to distinguish that any more. So we are mixing stuff again, which we should not.

ctron (commented):
Just picking "the first one" seems wrong.

More information about CNAs and APDs: https://www.cve.org/ProgramOrganization/ADPs

modules/ingestor/src/service/advisory/cve/loader.rs
dejanb: I added this as I found in our tests an example where CVE had adp score and not the cna one.

If I'm reading the document correctly cna and adp scores should be mutually exclusive and if they are not (which should be exception)

> In the rare event that there is a CVSS, CWE, or CPE string provided by the originating CNA and the CISA ADP, the originating CNA’s data should take precedence for any decision making.

I think the current logic handles this properly: iterating over cna iterator first and use it as a value if it's found. Also, cna can have multiple scores, which is used to provide different versions (e.g. v3 and v4), but the resulting score should be the same.

I think in the future we should add a "label" or something to properly "display" this information on the details page.

I updated comment to reflect this conversation. Let me know what you think.

modules/fundamental/src/vulnerability/service/mod.rs
dejanb: I found the BREAKING_CHANGE example and added a description for this commit. Thanks for the pointer.

dejanb:
I ran a scale test locally to check the slowdown in the advisory endpoint showed here

image

It's not seen locally.

modules/ingestor/src/service/advisory/cve/loader.rs
ctron: I'm ok adding this in general. And I think the order (taking cna first) is ok. And it's documented.

My concern is that we don't know (once ingested) where that value came from. If we're correcting the existing issue, I think we should spend the time and also sort this one out too.

Right now, it looks like we don't have a real representation in our datamodel for this. And the `base_` fields are basically extracted "caches" for having that information pre-evaluated (and stored). Which makes sense.

But can't we have the full information somewhere in our database? Assuming we want to do better in the future, we'd have to re-ingest all documents in order to get to the same result.

dejanb: Here's the issue to track this https://github.com/trustification/trustify/issues/1656

As we need to work further on CVSS v4, which will require changes to the database in this area and reimporting the vulnerabilities, this could be implemented during that effort.

ctron: Could you add a comment in the code, referencing that issue.

dejanb: Done!

PR #1654, opened 2025-05-21
docs(adr): Proposal for a "PURL report"

ctron:
Rendered preview: https://github.com/ctron/trustify/blob/feature/adr_da_check_1/docs/adrs/00007-purl-report.md

### DIFF
--- a/docs/adrs/00007-purl-report.md
+++ b/docs/adrs/00007-purl-report.md
@@ -1,0 +1,168 @@
+# 00007. PURL vulnerability report
+
+Date: 2025-05-21
+
+## Status
+
+ACCEPTED
+
+## Context
+
+The UI should have a feature for users to "upload" their SBOM, and get a report about vulnerabilities of contained
+components. The term "upload" may be misleading, as the backend must never store the document.
+
+"Dependency Analytics" (DA) requires a way to check a set of PURLs for their vulnerabilities too. Ideally, we can
+somehow re-use that functionality.
+
+As processing the correlation information may require some time, we may encounter request timeouts, like we sometimes
+see for larger SBOMs. We do need a way to either cut down work items into smaller pieces or ensure the connection
+between the frontend and the backend stays open for long-running operations.
+
+### Assumptions
+
+This ADR makes the following assumptions:
+
+* The provided SBOM will only be scanned for PURLs, not for CPEs, hashes/digests or other references
+
+## Decision
+
+There's a two-step process. First, upload (but don't store) the SBOM to the backend. The backend will parse the
+document and extract all PURLs, returning them to the client. This process, depending on the size of the file, is
+rather fast. The request shouldn't time out, as either the document or the response is being transferred. The actual
+processing, without any transfer, should be quick.
+
+The client can also list the result, once returned, to the user. To show the user what was understood from the SBOM.
+But that's an optional approach.
+
+Next, the client will send batches of PURLs to the backend, asking for vulnerability information attached to them. The
+backend might take some time to come up with this information, so the client has to be careful not to use "too big"
+chunks. There should be a way to tweak this for a running system (configuration).
+
+The client uses the returned information for rendering the result.
+
+```mermaid
+sequenceDiagram
+
+    participant UI
+    participant Backend
+
+    UI->>Backend: extract PURLs from SBOM
+    Backend->>UI: PURLs
+
+    loop Batch Requests
+      UI->>Backend: correlate PURLs
+      Backend->>UI: vulnerabilities for PURLs
+    end
+
+    UI->>UI: Render report
+```
+
+DA can use the same API, but processes data a bit different upfront:
+
+```mermaid
+sequenceDiagram
+  participant DA
+  participant Backend
+
+  DA->>DA: Generate hierarchy of PURLs
+
+  loop Batch Requests
+    DA->>Backend: correlate PURLs
+    Backend->>DA: vulnerabilities for PURLs
+  end
+
+  DA->>DA: Render report
+```
+
+The flow of data is like this:
+
+```mermaid
+flowchart TD
+
+    A[DA] --> |array of PURLs|B[/Correlate PURLs/]
+
+    C[UI] --> |SBOM data|D[/Extract PURLs/]
+    D --> |array of PURLs|B
+    B --> E[[JSON result]]
+```
+
+## API
+
+* `POST /api/v2/ui/extractSbomPurls`: Send an SBOM, retrieve all PURLs
+
+  Response (`200 OK`):
+
+  ```json5
+  {
+    "purls": [
+      "pkg:maven/org.apache.xmlgraphics/batik-anim@1.9.1?packaging=sources",
+      "pkg:npm/%40angular/animation@12.3.1"
+    ],
+    "relationships": [
+      {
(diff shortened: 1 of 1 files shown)


docs/adrs/00007-purl-report.md
```diff
@@ -0,0 +1,104 @@
...
+
+There's a two-step process. First, upload (but don't store) the SBOM to the backend. The backend will parse the
+document and extract all PURLs, returning them to the client. This process, depending on the size of the file, is
+rather fast. The request shouldn't time out, as either the document or the response is being transferred. The actual
+processing, without any transfer, should be quick.
+
+The client can also list the result, once returned, to the user. To show the user what was understood from the SBOM.
+But that's an optional approach.
+
+Next, the client will open a websocket connection to the backend. It can then send as many PURLs as it wants to the
```
ruromero, line 37: Another solution is to send requests in batches, this is how other vulnerability providers are implemented (examples are OSS Sonatype or OSV, if I recall correctly). Maybe batch requests will be more efficient than 1 by 1 queries.

DA Exhort can easily adapt the current implementation into batch requests as it has already been done for the other providers.

docs/adrs/00007-purl-report.md
```diff
@@ -0,0 +1,104 @@
...
+
+The client can also list the result, once returned, to the user. To show the user what was understood from the SBOM.
+But that's an optional approach.
+
+Next, the client will open a websocket connection to the backend. It can then send as many PURLs as it wants to the 
+backend, which will look up vulnerabilities one by one and report them back to the client once correlated.
+
+## API
+
+* `POST /api/v2/extractPurlsFromSbom`: Send an SBOM, retrieve all PURLs
```
ruromero, line 42: as the domain is `sbom` you can first think of `POST /api/v2/sboms/extract/purls`allowing other extractions like `dependencyTree` or `components`, etc.

docs/adrs/00007-purl-report.md
```diff
@@ -0,0 +1,104 @@
...
+  }
+  ```
+
+## Open items
+
+* Find a better path/name for the "extract PURLs" endpoint 
+
+## Alternative approaches
+
+Have the browser extract the PURLs and send those to the backend for checking.
```
ruromero, line 88: It depends on the requirements (see my previous comment). But as mentioned below I'd say that local processing is more efficient in most cases.

docs/adrs/00007-purl-report.md
```diff
@@ -0,0 +1,104 @@
...
+* 👍 The data transmitted to the backend would be way less. This would also need less CPU and memory on the backend.
+* 👍 The full data would never end up on the backend, so we'd never get in contact with information we don't need
+* It's unclear (without measuring) what's faster from a user perspective. Processing the file locally requires resources
+  and takes some time. But transmitting a larger document does too.
+* 👎 There's a notion of consolidating logic on the backend.
+
+Have the browser use a websocket API instead of the DA HTTP API:
+
+* 👎 Add "websockets" as a technology
+* 👎 Duplicate functionality of the dependency analytics API
```
ruromero, line 99: Agree.
However, this also means that if the IDE sends the SBOM to DA and DA sends it to TPA that might sound excessive for such a trivial task (again, depending on the type of extraction).
IMO I would not touch the current DA functionality.
Also don't forget about the current JSON and HTML reports generated in DA (in case you are considering adding something similar to the UI)

docs/adrs/00007-purl-report.md
```diff
@@ -0,0 +1,104 @@
...
+## Status
+
+PROPOSED
+
+## Context
+
+The UI should have a feature for users to "upload" their SBOM, and get a report about vulnerabilities of contained
+components. The term "upload" may be misleading, as the backend must never store the document.
+
+"Dependency Analytics" requires a way to check a set of PURLs for their vulnerabilities too. Ideally, we can somehow
```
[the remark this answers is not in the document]
ctron: Good point. The scope was, to my understanding, generating a UI report.

@carlosthe19916 should that include the dependency hierarchy too? Or just a list of packages?

docs/adrs/00007-purl-report.md
ctron: Ah yes. That's already what I changed it to. But didn't seem to push that change. I'd prefer that too.

docs/adrs/00007-purl-report.md
ctron: We already have `/api/v2/sbom` (singular). Trying to follow a REST approach.

However, that prefix works with "stored" SBOMs. I'd like to separate this, making it clear that we are not having "stored" SBOMs in this case.

docs/adrs/00007-purl-report.md
ctron: For DA, it's ok keeping it that. This ADR is about having the browser generate a report and visualizing this. @carlosthe19916 needs to let us know what the UI requires for that from the backend.

I'd prefer the browser to extract that information too. As I think it removes unnecessary load from the backend. And, as you said, DA duplicates that already to some degree.

Which brings up the question: could we unify that between the browser and DA? Could we come up with a backend endpoint supporting both. Minimum the JSON. At best, also the HTML. :point_left: @carlosthe19916 @ruromero

ctron: Added this to the open items section.

docs/adrs/00007-purl-report.md
ruromero: Ah yes, the API is in singular. 

Well the thing is that it belongs to the same _domain_ and domains don't necessarily imply state, that's why the verbs should be for. I think it can be a good solution to consider `purl-extraction` as a subresource of `sbom` and have `/api/v2/sbom/purl-extraction` or if you think it belongs more to the `purl` domain then have `/api/v2/purl/sbom-extraction` but it looks less intuitive.

docs/adrs/00007-purl-report.md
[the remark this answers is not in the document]
carlosthe19916: Not sure. Seems like we are discussing two things
- DA use case
- UI Scan SBOM use case

and I am not sure if visually both DA and ScanSBOM should render the same report. But since DA do uses dependency hierarchy I guess it makes sense to include it. It might help to enrich the current UI if that info is also available in `/api/v2/sbom/{id}/packages`

docs/adrs/00007-purl-report.md
ruromero: In the Trustification UI use case, it doesn't make sense to generate the HTML report in the server. DA does that because there is no web client as such.
DA will not need to send the SBOM to trustification, it will only request the vulnerabilities by purl as it already does.

ruromero: I was just agreeing with you on the fact that for trustify UI the HTML should be generated in the client.
The DA backend generates a unified HTML because there's no UI to execute in the client. It goes directly to a file that is persisted in the file system to be displayed by the IDE.

I wanted to clarify these 2 differences between the DA and the UI. No big deal :)

ruromero: As I mentioned before, I don't think DA should move this functionality to TPA, it is an unnecessary overhead for a functionality that is already in place and is not expected to change.

carlosthe19916 (approved):
LGTM

PR #1655, opened 2025-05-21
chore: update otel doc with diagrams

helio-frota:
Rendered version: https://github.com/helio-frota/trustify/blob/issue-1648/docs/otel.md

### DIFF
--- a/docs/otel.md
+++ b/docs/otel.md
@@ -1,97 +1,213 @@
 # OpenTelemetry
 
-[OpenTelemetry](https://opentelemetry.io/) is an observability framework for collecting, generating,
+[OpenTelemetry](https://opentelemetry.io/) (OTEL) is an observability framework for collecting, generating,
 and exporting telemetry data (traces, metrics, and logs, aka [signals](https://opentelemetry.io/docs/concepts/signals/))
 to improve system monitoring and performance. It is a [CNCF project](https://www.cncf.io/projects/opentelemetry/),
 ensuring vendor-neutral and standardized observability solutions for cloud-native applications.
 
-[Jaeger](https://www.jaegertracing.io/), [Grafana Tempo](https://grafana.com/oss/tempo/) and [Prometheus](https://prometheus.io/) are also CNCF projects for observability.
-Jaeger and Grafana Tempo focuses on distributed tracing, while Prometheus handles metrics and monitoring.
+[Grafana Tempo](https://grafana.com/oss/tempo/) (Tempo) and [Prometheus](https://prometheus.io/) are also CNCF projects for observability.
+Grafana Tempo focuses on distributed tracing, while Prometheus handles metrics and monitoring.
 
 The OpenTelemetry Collector (aka OTELCOL or OTEL collector) is a vendor-neutral service for
-processing and exporting telemetry data, that also acts as a facade for both Jaeger and Prometheus,
+processing and exporting telemetry data, that also acts as a facade for both Tempo and Prometheus,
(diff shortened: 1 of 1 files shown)


docs/otel.md
```diff
@@ -95,3 +95,90 @@ cargo run --bin trustd api --db-password trustify --auth-disabled --metrics enab
...
+
+The diagrams below reflect the current state of both Trustify and [Trustify Helm Charts](https://github.com/trustification/trustify-helm-charts)
+
+### Scenarios
+
+#### Development
+
+> Based on **etc/telemetry/compose.yaml** and **etc/telemetry/compose-tempo.yaml**
+
+```
```
ctron, line 109:
```suggestion
```mermaid
```

helio-frota: Thanks for catching this : )
First time I'm using Markdown Preview plugin for Neovim, and the plugin just forgot to complain about this

helio-frota:
> it's documenting the status quo

yes that's correct.

> , so maybe we need to change that too.

Yeah, we can adjust anything to match our needs :+1: 

> The compose setup with tempo seems to be only about "traces", I don't see "metrics" in there. Is that the case?

Fair question. Since those compose files are related to the development environment, they exist mainly to:

* Validate and review pull requests
* Run local experiments to ensure the otel-rust libraries still work after dependency or API updates without need to deploy on minikube
* Run local experiments by instrumenting all functions with #[instrument]
* Investigate performance issues in slow functions

And with that we already have `metrics` on the other `compose.yaml`, but yeah we can change that as well.

> The compose setup also seems to have Jaeger and Temp. Are both required? Isn't Tempo a replacement for Jaeger? If so, can we drop Jaeger?

We have [compose.yaml](https://github.com/trustification/trustify/blob/main/etc/telemetry/compose.yaml) with `Jaeger` and `Prometheus` and we have other [compose-tempo.yaml](https://github.com/trustification/trustify/blob/main/etc/telemetry/compose-tempo.yaml) with `Tempo`.

* Yeah it was part of the transition (Jaeger -> Tempo). Probably the ideal is to keep `Tempo` only. 
* It helps to validate Trustify side (otel-rust libraries and etc) without need to deploy on Minikube.

> If not, then wouldn't it make sense to combine these two compose files? I think we should have only one.

:+1:  let's do it

> The Helm charts says "production", but having "in-memory" Jaeger, it seems more like an example setup. If that's the case, that's fine, but we should name it as such.

Yeah I was expecting comments about this part as well :smiley:  probably from "Production" -> to "Deployment" ?

> Judging from this, it looks like what we need on the trustify side only the "collector address".

Exactly you got the point :+1:  

* No more hard dependencies on specific backends
* No more exposing `/metrics` in Trustify. Prometheus scraping it is no longer required

We should talk only with the OTEL collector via OpenTelemetry protocol (OTLP).

`Devs <---> Collector`

Example: https://github.com/trustification/trustify/pull/1635 I have nothing against changing the OTELCOL processor for development purposes. But I'm `-1`, _as developer_, changing that for production/deployment.

> Is any other information required? Like "protocol"?

I'll add this information on Trustify part of diagram thanks pointing it as we are using grpc-tonic feature enabled on Cargo.toml and yeah that's requires `Tokio` I'll add Tokio on the diagram as well :+1:

helio-frota:
Issue created for the other task https://github.com/trustification/trustify/issues/1661 :+1:

helio-frota:
Diagram updated with the new crate name for actix-web auto-instrumentation https://github.com/trustification/trustify/pull/1689

PR #1659, opened 2025-05-22
tests for sbom/advisory label api methods

m-brophy:


bxf12315 (changes requested):
```rust
`if let Some(advisory) = fetched_again {
        assert_eq!(
            advisory.head.labels.0,
            HashMap::from([
                ("label_1".into(), "First Label".into()),
                ("label_2".into(), "Second Label".into())
            ]),
            "Labels were not set correctly"
        );
    } else {
        panic!("Can not find the advisory");
    }`
```

m-brophy:
> I suggest making some changes here.
> 
> ```rust
> `if let Some(advisory) = fetched_again {
>         assert_eq!(
>             advisory.head.labels.0,
>             HashMap::from([
>                 ("label_1".into(), "First Label".into()),
>                 ("label_2".into(), "Second Label".into())
>             ]),
>             "Labels were not set correctly"
>         );
>     } else {
>         panic!("Can not find the advisory");
>     }`
> ```

Changes made as per review suggestion

bxf12315:
> > I suggest making some changes here.
> > ```rust
> > `if let Some(advisory) = fetched_again {
> >         assert_eq!(
> >             advisory.head.labels.0,
> >             HashMap::from([
> >                 ("label_1".into(), "First Label".into()),
> >                 ("label_2".into(), "Second Label".into())
> >             ]),
> >             "Labels were not set correctly"
> >         );
> >     } else {
> >         panic!("Can not find the advisory");
> >     }`
> > ```
> 
> Changes made as per review suggestion

let advisory = fetched_again.unwrap_or_else(|| panic!("The advisory does not exist."));
I suggest you use this, because this approach is very common in our codebase.

PR #1664, opened 2025-05-27
chore: removes unused duplicated status

helio-frota:
The tables `purl_status` and `product_status` both reference status.id as a foreign key and they are empty during migration init.


dejanb:
I looked at this a bit more. It seems that duplication was there from the start

https://github.com/trustification/trustify/commit/8ab2b877e073ff22471dc4df5d8cf343f1aee57d

As for the fix:

1. This needs to be done through the new migration, we cannot just delete it from the inital one
2. Strangely, when I import ds3 both statuses are used, although `c0273e43-2b0c-4dae-a3b3-c4f9733fbfa7` much less. So, migration should run an update to fix this before deleting the status.

helio-frota:
Great investigation, thanks!
Let's close this PR :+1:

Issue #1637, opened 2025-05-16
We need an endpoint for fetching Labels


carlosthe19916:
Requirement coming from :

https://issues.redhat.com/browse/TC-2497
https://issues.redhat.com/browse/TC-2498

Requirement:
- Have a way to fetch all Labels of all SBOMs
- Have a way to fetch all Labels of all Advisories

To be defined:
- Do we need the Label endpoints to return ALL instances or return only a subset of them and allow the result to be filtered. E.g. `GET /api/v2/sbom-labels?q=some_label`?

Why do we need it:
- https://issues.redhat.com/browse/TC-2497 and https://issues.redhat.com/browse/TC-2498 describe the new Feature: "Labels" that will be available for SBOMs and Advisories. We should be able to filter SBOMs/Advisories by Labels. From the UI point of view, there was an agreement coming from UX and PM that we could adopt the OCP way of filtering by Label.

Having the endpoints for Labels will allow the UI to populate the Label Dropdown Filter with all available Labels and let the user choose one of them to start the filtering process. 

![Image](https://github.com/user-attachments/assets/6dd252d1-8ef9-46ba-b6c1-45d89cbfc158)

There might be things we did still not discovered and we can use this issue to track any difficulties we find in the way and ask for feedback to the UX team if we need to change anything.

ctron:
Just to clarify what you are asking for: labels right now means "key and value". Do you want an endpoint, which returns all key/value combinations for all SBOMs (or other document types)?

carlosthe19916:
@ctron I don't think we need "all combinations" of key/value pairs. I think the objective is to simulate what OCP does (image of the issue description). What OCP does is to list all unique  "key/value" pairs. E.g.

- `SBOM1` has labels `key1:value1`, `samekey:differentvalue1`, `samekey:samevalue`, `tag1`, `sametag`
- `SBOM2` has labels `key2:value2`, `samekey:differentvalue2`, `samekey:samevalue`, `tag2`, `sametag`

The output of listing labels would be:
-  `key1:value1`
- `key2:value2`
- `samekey:differentvalue1`
- `samekey:differentvalue2`
- `samekey:samevalue`
- `tag1`
- `tag2`
- `sametag`

Which means, the user will see those elements in the Dropdown (labels selector of the image of the issue description)

WDYT?

ctron:
Ok, that's what I meant by saying "all key/value combinations".

[...]

PR #1667, opened 2025-05-28
Add endpoint to fetch all labels from SBOM|Advisory

carlosthe19916:
fixes: https://github.com/trustification/trustify/issues/1637

This PR creates 2 new endpoints:
-  /api/v2/sbom-labels .- list all unique key/value labels of the sbom table.
- /api/v2/advisory-labels .- list all unique key/value labels of the advisory table.

Both endpoints have pagination in place (offset, limit) and a Query Parameter named "filter_text" which is an input to filter out the results.

Use Case:
- The UI will have a dropdown of labels that a user can select in order to filter SBOM and Advisories.


modules/fundamental/src/advisory/endpoints/label.rs
```diff
@@ -1,9 +1,52 @@
...
+struct LabelQuery {
+    #[serde(default)]
+    filter_text: String,
+}
+
+#[utoipa::path(
+    tag = "advisory",
+    operation_id = "listAdvisoryLabels",
+    params(
+        Query,
```
jcrossley3, line 25:
```suggestion
        LabelQuery,
```

modules/fundamental/src/advisory/endpoints/label.rs
```diff
@@ -1,9 +1,52 @@
...
+        Paginated,
+    ),
+    responses(
+        (status = 200, description = "List all Labels from all Advisories", body = Vec<Value>),
+    ),
+)]
+#[get("/v2/advisory-labels")]
```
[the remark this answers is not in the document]
carlosthe19916: Should it be `#[get("/v2/advisoryLabels")]` ?

modules/fundamental/src/advisory/model/raw_sql.rs
```diff
@@ -0,0 +1,29 @@
...
+        FROM advisory,
+            LATERAL jsonb_each_text(labels) AS kv
+        WHERE
+            CASE 
+                WHEN kv.value IS NULL THEN kv.key
+                ELSE kv.key || '=' || kv.value
+            END ILIKE $1
+        ORDER BY
+            kv.key, kv.value
+        OFFSET {offset} {}
```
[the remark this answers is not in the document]
carlosthe19916: Thanks, I made a change and now the SQL is not generated dynamically.

modules/fundamental/src/advisory/service/label.rs
```diff
@@ -0,0 +1,29 @@
...
+    pub async fn fetch_labels<C: ConnectionTrait>(
+        &self,
+        filter_text: &str,
+        limit: u64,
+        connection: &C,
+    ) -> Result<Vec<serde_json::Value>, Error> {
+        let statement = Statement::from_sql_and_values(
+            DbBackend::Postgres,
+            SEARCH_LABELS_SQL,
+            [format!("%{}%", filter_text).into()],
```
ctron, line 17: That looks like a candidate for an SQL inject attack. Please use SQL parameters and maybe `CONCAT`.

modules/fundamental/src/sbom/service/label.rs
```diff
@@ -1,13 +1,40 @@
...
+    pub async fn fetch_labels<C: ConnectionTrait>(
+        &self,
+        filter_text: &str,
+        limit: u64,
+        connection: &C,
+    ) -> Result<Vec<serde_json::Value>, Error> {
+        let statement = Statement::from_sql_and_values(
+            DbBackend::Postgres,
+            SEARCH_LABELS_SQL,
+            [format!("%{}%", filter_text).into()],
```
ctron, line 25: That looks like a candidate for an SQL inject attack. Please use SQL parameters and maybe CONCAT.

ctron (changes requested):
I think there two candidates for an SQL injection attack.

Also, the default seems to be to return "all" entries. Which is way too much. I think it makes more sense to have a default of e.g. 10.

As codecov points out, that PR is missing test coverage. I think we should have at least some basic tests for it.

modules/fundamental/src/common/service.rs
```diff
@@ -0,0 +1,61 @@
+use crate::Error;
+use sea_orm::{ConnectionTrait, DbBackend, FromQueryResult, PaginatorTrait, Statement};
+
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub enum DocumentType {
+    Advisory,
+    Sbom,
+}
+
+/// Fetch all unique key/value labels of all Advisories
```
mrizzi, line 10:
```suggestion
/// Fetch all unique key/value labels matching the `filter_text` for all the `r#type` entities, i.e. `DocumentType::Advisory` or `DocumentType::Sbom`
```

mrizzi (commented):
Just a tiny detail on the doc for the `fetch_labels` function.

ctron:
Good catch. I think @carlosthe19916 fixed this. Thanks!

PR #1669, opened 2025-05-29
Quay importer

### LINKED ISSUE #1633: Quay Importer
Downstream issue: https://issues.redhat.com/browse/TC-2289

GOAL: Enable walker for Quay registry.

## Story 

As Devsecops I want see TPA analysis of SBOMs stored in Quay

### Acceptance Criteria

I expect to configure TPA to access and walk a Quay registry for existing SBOMs
I expect the Quay SBOMs to be imported into TPA and analyzed like other imported SBOMs|

jcrossley3:
Fixes #1633 

May be configured with a registry source, a namespace, an API token, and a size limit. The only namespace I've found any SBOM's in is 'redhat-user-workloads' which contains ~150K SBOM's within ~8K repos.

Quay API requests are concurrently fetched, up to 32 at a time, but ingestions occur synchronously, so as not to overwhelm the DB.

As of this time, only SBOM attachments are supported, as that's how most SBOM's are published on quay.io. SBOM attestations will be a future enhancement as more become available.


modules/importer/src/runner/quay/walker.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+impl<C: RunContext> QuayWalker<C> {
+    pub fn new(
+        importer: QuayImporter,
+        ingestor: IngestorService,
+        report: Arc<Mutex<ReportBuilder>>,
+        context: C,
+    ) -> Result<Self, Error> {
+        let client = match std::env::var(QUAY_API_TOKEN) {
```
ctron, line 39: Importers to have a configuration section. In cases we of system configuration, we use `clap`. So I am wondering why this token is fetched directly from an environment variable?

Also, having only one token would restrict us to only one quay instance, with only one quay "user" (token). What would happen if a user has two quay instances? What if there are two users/tokens with two different sets of permissions?

jcrossley3: Good points. I'll add the token to the importer config.

Issue #1665, opened 2025-05-28
Implement "extract PURLs" endpoint from ADR 00007 / #1654


ctron:
ADR 00007 defines a new endpoint for extracting PURLs from an SBOM, without storing it. This needs to be implemented.

Also see: https://github.com/trustification/trustify/pull/1654

PR #1673, opened 2025-06-02
feat: implement PURL extraction endpoint

ctron:
Closes: #1665


modules/ui/src/endpoints.rs
```diff
@@ -32,6 +60,67 @@ impl UiResources {
...
+    tag = "ui",
+    operation_id = "extractSbomPurls",
+    request_body = inline(BinaryData),
+    params(
+        ExtractSbomPurls,
+    ),
+    responses(
+        (
+            status = 200,
+            description = "User preference stored under this key",
```
dejanb, line 86: Are these descriptions correct? They look out of context.

PR #1690, opened 2025-06-03
chore: database and migration on otel compose

helio-frota:
* For convenience, avoiding 2 extra terminals opened during development experiments.

### DIFF
--- a/etc/telemetry/compose.yaml
+++ b/etc/telemetry/compose.yaml
@@ -31,7 +31,7 @@ services:
     restart: unless-stopped
 
   collector:
-    image: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector:0.119.0
+    image: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector:0.127.0
     command: ["--config=/otel-collector-config.yaml"]
     volumes:
       - './config-collector.yaml:/otel-collector-config.yaml:z'


helio-frota:
TODO: Update https://github.com/trustification/trustify/pull/1655 in case we merge this PR.

etc/telemetry/compose.yaml
```diff
@@ -38,3 +38,41 @@ services:
...
+
+  # NOTE: Database part to avoid opening two more terminals for extra commands.
+  # Doesn't need to be in the diagrams, it's just for convenience.
+  migration:
+    image: ghcr.io/trustification/trustd:latest
+    # network_mode to avoid:
+    # migration-1   | Error: Connection Error: pool timed out while waiting for an open connection
+    # migration-1   | Caused by:
+    # migration-1   |   pool timed out while waiting for an open connection
+    network_mode: "host"
```
ctron, line 50: Host `host` mode brings some disadvantages. It's better to run with the normal. The hostname would be `postgres` in this case, as that's the name of the service later on in the file.

helio-frota: Thanks it worked!

```
postgres-1    | 2025-06-04 09:55:10.732 UTC [1] LOG:  database system is ready to accept connections
migration-1 exited with code 0
```

ctron (approved):

Issue #1674, opened 2025-06-02
trustify-pm binary pg-embedded fails to start due to github rate limit


queria:
trustd binary obtained from Github Releases page,
can fail to start due to progresql_embedded needs to fetch pg binaries and utilizes github api for it which is rate limited
(and may be already exhausted for shared IP).

In case of development mode like following Readme and using `cargo run/build` approach,
it wouldn't be surprising to me.

But in case of downloading pre-built binary release it is bit surprising.

Seems it may be same or related to as in #501.

E.g. it be reproduced utilizing network namespace for blocking network access:
```
#!/bin/bash

# switch to temp workdir
mkdir -p /tmp/trustd-pm
cd /tmp/trustd-pm

# download and extract released binary
URL="https://github.com/trustification/trustify/releases/download/v0.2.19/trustd-pm-0.2.19-x86_64-unknown-linux-gnu.tar.gz"
TGZ="trustd-pm-0.2.19-x86_64-unknown-linux-gnu.tar.gz"

if [[ ! -f "${TGZ}" ]]; then curl --location --fail --retry 3 -o "./${TGZ}__tmp" "${URL}" && mv "./${TGZ}__tmp" "./${TGZ}"; fi
tar xf "${TGZ}"
cd ./trustd-pm-0.2.19-x86_64-unknown-linux-gnu/

# ensure we have no db initialized yet, neither pg binaries cached
rm -rf ./.trustify

# start trust (in network namespace)
unshare -r -n ./trustd-pm
```

PR #1691, opened 2025-06-04
fix: align the embedded postgres version with the CI

ctron:
In PM mode, the postgresql binary will be embedded. This requires the build to know which version to download. However, this didn't align with the code requesting a database.

This change prefers the use of the CI version, but falls back to the locally (code) provided value. This ensures that if the CI builds with a provided version, it will be used during runtime.

Closes: #1674

### DIFF
--- a/.github/workflows/build-binary.yaml
+++ b/.github/workflows/build-binary.yaml
@@ -123,7 +123,7 @@ jobs:
       - name: Build | Build
         shell: bash
         env:
-          POSTGRESQL_VERSION: 16
+          POSTGRESQL_VERSION: "=17.2"
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # required for retrieving postgres
         run: |
           set -x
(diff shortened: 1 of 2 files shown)


ctron:
@queria @helio-frota as you both mentioned to experience, maybe you can give this PR a try.

Compile it once with `cargo build`, and then run the binary in the `target` folder. Building should download the DB archive. But running should not.

ctron:
> Thanks for this, will check carefully to understand better. I stopped using the dump on github because I need to
> 
>     1. run trustify-pm
> 
>     2. populate with the dump
> 
>     3. kill trustify-pm
> 
>     4. run trustify-pm again to apply extra missing migrations
> 
> 
> and on the step 4) ^ the CI complains about github rate limit downloading pg-embedded

With this PR, you should be able to run PM mode as often as you like. Without it fetching any postgres binary. Except for the build time.

helio-frota:
>  Except for the build time.

Awesome I'm running cargo build https://github.com/helio-frota/trustify-api-hurl-tests/blob/main/.github/workflows/ci.yaml#L38 then it will work : )

Should we merge first and check later or are you waiting for a concrete feedback ? ( in this case I'll need a bit of time to change my configs and report here later )

helio-frota (approved):

PR #1692, opened 2025-06-04
chore: add 'async-graphql' feature flag to prevent unused_imports warnings

mrizzi:
Running tests locally while developing, I experienced these warnings that confused me:
```
warning: unused import: `std::sync::Arc`
 --> entity/src/advisory.rs:4:5
  |
4 | use std::sync::Arc;
  |     ^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused import: `trustify_common::db`
 --> entity/src/advisory.rs:7:5
  |
7 | use trustify_common::db;
  |  
```
so I've added the `"async-graphql` feature flag to properly manage these `use` statements.


ctron (approved):
:heart: There's another way to deal with this: in cases like that use the fully qualified type. The result is the same, just a style thing. And just that you know.

PR #1700, opened 2025-06-04
TC-2564 : Add versions column to sbom to allow version filtering

gildub:
Along with https://github.com/trustification/trustify/pull/1749, this PR is partially fixing https://issues.redhat.com/browse/TC-2564 which is about using full text search to find information about version.

To be able to search versions data requires to have a `versions` column added to sbom table.
That's the main purpose of this PR. 
This will also allow to specifically search `versions` which is different than full text search across the sbom.

## Summary by Sourcery

Add a new "versions" field to SBOM entities and enable version-specific filtering

New Features:
- Persist a versions column in the SBOM database table with a migration
- Expose SBOM.versions in GraphQL queries and vulnerability advisory summaries

Enhancements:
- Populate the new versions field when ingesting SPDX, CycloneDX, and ClearlyDefined SBOMs
- Extend the SBOM search translator to support exact and substring matches on versions

Build:
- Add migration m0001080_sbom_add_versions to add and remove the versions column

Tests:
- Add integration tests for querying SBOMs by version filters

### DIFF
--- a/migration/src/m0001080_sbom_add_versions.rs
+++ b/migration/src/m0001080_sbom_add_versions.rs
@@ -1,0 +1,43 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(Sbom::Table)
+                    .add_column(
+                        ColumnDef::new(Sbom::Versions)
+                            .array(ColumnType::Text)
+                            .not_null()
+                            .default("{}"),
+                    )
+                    .to_owned(),
+            )
+            .await?;
+
+        Ok(())
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(Sbom::Table)
+                    .drop_column(Sbom::Versions)
+                    .to_owned(),
+            )
+            .await?;
+        Ok(())
+    }
+}
+
+#[derive(DeriveIden)]
+enum Sbom {
+    Table,
+    Versions,
+}
--- a/modules/fundamental/src/sbom/endpoints/test.rs
+++ b/modules/fundamental/src/sbom/endpoints/test.rs
@@ -1287,6 +1287,7 @@ async fn query_sboms_by_array_values(ctx: &TrustifyContext) -> Result<(), anyhow
     ctx.ingest_documents([
         "quarkus-bom-2.13.8.Final-redhat-00004.json",
         "spdx/rhelai1_binary.json",
+        "cyclonedx/container.json",
     ])
     .await?;
 
@@ -1309,6 +1310,25 @@ async fn query_sboms_by_array_values(ctx: &TrustifyContext) -> Result<(), anyhow
     query(2, "authors>ZZZ").await;
     query(2, "organization").await;
     query(1, "tool: syft").await;
+    query(1, "versions=2.13.8.Final-redhat-00004").await;
+    query(1, "versions~2.13.8").await;
+    query(1, "version~2.13.8").await;
+    query(1, "2.13.8.Final-redhat-00004").await;
+    query(1, ".8.Final-redhat").await;
+    query(
+        1,
+        "versions=sha256:fbf470d8b5b84606f797d78775b9de88e14fdb43cc47b2db6ff3747b46df323e",
+    )
+    .await;
+    query(1, "versions~fbf470d8b5").await;
+    query(1, "version~fbf470d8b5").await;
+    query(
+        1,
+        "sha256:fbf470d8b5b84606f797d78775b9de88e14fdb43cc47b2db6ff3747b46df323e",
+    )
+    .await;
+    query(1, "fbf470d8b5").await;
+    query(3, "").await;
 
     Ok(())
 }
--- a/modules/fundamental/src/sbom/service/sbom.rs
+++ b/modules/fundamental/src/sbom/service/sbom.rs
@@ -141,6 +141,10 @@ impl SbomService {
                     .alias("sbom_node", "r0")
                     .translator(|f, op, v| match f.split_once(':') {
                         Some(("label", key)) => Some(format!("labels:{key}{op}{v}")),
+                        None => match f {
+                            "version" => Some(format!("versions{op}{v}")),
+                            _ => None,
+                        },
                         _ => None,
                     }),
             )?
--- a/modules/ingestor/src/graph/sbom/spdx.rs
+++ b/modules/ingestor/src/graph/sbom/spdx.rs
@@ -72,6 +72,23 @@ fn suppliers(sbom: &SPDX) -> Vec<String> {
     Vec::from_iter(result)
 }
 
+/// Extract versions for a SPDX SBOM by collecting versions of describing packages
+fn versions(sbom: &SPDX) -> Vec<String> {
+    // packages describing the SBOM
+    let describing = describing_packages(sbom);
+
+    // collect versions for matching packages
+    let mut result = HashSet::new();
+    for p in &sbom.package_information {
+        if !describing.contains(p.package_spdx_identifier.as_str()) {
+            continue;
+        }
+        result.extend(p.package_version.clone());
+    }
+
+    Vec::from_iter(result)
+}
+
 impl<'a> From<Information<'a>> for SbomInformation {
     fn from(value: Information<'a>) -> Self {
         let sbom = value.0;
@@ -96,6 +113,7 @@ impl<'a> From<Information<'a>> for SbomInformation {
                 .clone(),
             suppliers: suppliers(sbom),
             data_licenses: vec![value.0.document_creation_information.data_license.clone()],
+            versions: versions(sbom),
         }
     }
 }
--- a/entity/src/sbom.rs
+++ b/entity/src/sbom.rs
@@ -21,6 +21,8 @@ pub struct Model {
 
     pub source_document_id: Option<Uuid>,
 
+    pub versions: Vec<String>,
+
     #[cfg_attr(
         feature = "async-graphql",
         graphql(derived(owned, into = "HashMap<String,String>", with = "Labels::from"))
--- a/etc/test-data/cyclonedx/container.json
+++ b/etc/test-data/cyclonedx/container.json
@@ -1,0 +1,28034 @@
+{
+  "$schema": "http://cyclonedx.org/schema/bom-1.4.schema.json",
+  "bomFormat": "CycloneDX",
+  "specVersion": "1.4",
+  "serialNumber": "urn:uuid:6972bd6d-919f-483a-9bfc-97b879bc60f2",
+  "version": 1,
+  "metadata": {
+    "timestamp": "2025-05-29T02:38:28Z",
+    "tools": [
+      {
+        "vendor": "anchore",
+        "name": "syft",
+        "version": "1.19.0"
+      }
+    ],
+    "component": {
+      "bom-ref": "8b3cf52115dba09d",
+      "type": "container",
+      "name": "/tmp/files/image",
+      "version": "sha256:fbf470d8b5b84606f797d78775b9de88e14fdb43cc47b2db6ff3747b46df323e"
+    },
+    "properties": [
+      {
+        "name": "syft:image:labels:architecture",
+        "value": "x86_64"
+      },
+      {
+        "name": "syft:image:labels:build-date",
+        "value": "2025-05-14T18:14:25"
+      },
+      {
+        "name": "syft:image:labels:com.redhat.component",
+        "value": "openjdk-21-runtime-ubi9-container"
+      },
+      {
+        "name": "syft:image:labels:com.redhat.license_terms",
+        "value": "https://www.redhat.com/en/about/red-hat-end-user-license-agreements#UBI"
+      },
+      {
+        "name": "syft:image:labels:description",
+        "value": "Image for Red Hat OpenShift providing OpenJDK 21 runtime"
+      },
+      {
+        "name": "syft:image:labels:distribution-scope",
+        "value": "public"
+      },
+      {
+        "name": "syft:image:labels:io.buildah.version",
+        "value": "1.37.6"
+      },
+      {
+        "name": "syft:image:labels:io.cekit.version",
+        "value": "4.13.0.dev0"
+      },
+      {
+        "name": "syft:image:labels:io.k8s.description",
+        "value": "Platform for running plain Java applications (fat-jar and flat classpath)"
+      },
+      {
+        "name": "syft:image:labels:io.k8s.display-name",
+        "value": "Java Applications"
+      },
+      {
+        "name": "syft:image:labels:io.openshift.tags",
+        "value": "java"
+      },
+      {
+        "name": "syft:image:labels:maintainer",
+        "value": "Red Hat OpenJDK \u003Copenjdk@redhat.com\u003E"
+      },
+      {
+        "name": "syft:image:labels:name",
+        "value": "ubi9/openjdk-21-runtime"
+      },
+      {
+        "name": "syft:image:labels:org.jboss.product",
+        "value": "openjdk"
+      },
+      {
+        "name": "syft:image:labels:org.jboss.product.openjdk.version",
+        "value": "21"
+      },
+      {
+        "name": "syft:image:labels:org.jboss.product.version",
+        "value": "21"
+      },
+      {
+        "name": "syft:image:labels:org.opencontainers.image.documentation",
+        "value": "https://rh-openjdk.github.io/redhat-openjdk-containers/"
+      },
+      {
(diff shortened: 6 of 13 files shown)


gildub:
@jcrossley3 , is that test error expected ?
`2025-06-04T16:12:49.457923Z ERROR trustify_common::db::query::columns::tests: Query syntax error: 'missing' is an invalid field. Try [authors, id, location, name, published, purl, score, severity, title, type, version] `

jcrossley3:
> @jcrossley3 , is that test error expected ? `2025-06-04T16:12:49.457923Z ERROR trustify_common::db::query::columns::tests: Query syntax error: 'missing' is an invalid field. Try [authors, id, location, name, published, purl, score, severity, title, type, version] `

Yes. Per our chat, many of our unit tests invoke code branches that may result in error messages logged to stdout. It's the failures, rather than the errors, you want to focus on.

ctron:
To my understanding, this will join and expand searching for SBOMs with all packages. Which can be huge! Just for the case when the user might want to search for this information. And not even all information, just for the packages describing the SBOM, right?

I think we should only add a join like this in cases where the user actually wants to search for this information.

ctron:
To me it looks like a hit on the sbom search performance. We should investigate this.

gildub:
@ctron, to be able to search on the version we have to join with sbom_package, I can't see another way. 
And regarding the performance hit, a view seems to be the appropriate answer. Views can be managed  using raw SQL during migration with `manager.exec_stmt()`.

jcrossley3:
I don't think a view PoC is a good use of resources, so I'm -1 on that.

I question the current structure of the response. What does `described_by` even mean? And what user story requires that a single SBOM should have anything other than one and only one version?

I suggest we add a `version: String` field to our SBOM table, or if we must, a `version: Vec<String>` and address the issue at ingestion.

gildub:
@ctron . @jcrossley3, the test are all [GREEN] now.

I initially used an Inner Join instead of a LeftJoin, the latter preserves all the records of the left table (sbom) even when there are no matches with the right table (sbom_package).

Also I had to change the test for `query_sboms_by_ingested_time` because the order of the items was changes due to the necessary.
Altogether, a `distinct_on()` followed by a `order_by_desc()` seems to be less efficient than using group_by. 
I'm working out to have a stabilized version with the latter.

gildub:
@ctron. @jcrossley3, following-up some of your earlier comment.

Effectively a more efficient approach would be a dedicated search only when columns of related tables are involved.
But the search "anything" is not helping identify fields from related tables to trigger a tailored search.

For me this goes back to the discussion about removing the search "anything"  feature in the GUI and have only search some fields explicitly.

BTW the scale test seems to be [GREEN]

ctron:
> BTW the scale test seems to be [GREEN]

I see a massive decrease in SBOM search performance.

gildub:
@ctron, 
> I see a massive decrease in SBOM search performance.

It leaves us with either :
- Dedicated search requests only to be used when filtering is required by user so we have a clear demarcation line when fetching default SBOM (or CVEs, etc).
- Use DBMS Views

ctron:
> Let me add a patch using that latter approach...

Please ensure that there's a migration as part of the patch filling the entry for existing entries.

ctron (changes requested):

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -139,6 +139,7 @@ impl SbomService {
                     .alias("sbom_node", "r0")
                     .translator(|f, op, v| match f.split_once(':') {
                         Some(("label", key)) => Some(format!("labels:{key}{op}{v}")),
+                        Some(("version", _)) => Some(format!("versions @> ARRAY[{v}]")),
```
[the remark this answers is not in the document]
jcrossley3: Confirmed: #1746

gildub:
@jcrossley3, with latest commit the version search works only for a full text (the entire version string) but it doesn't work for a sub-string of the version (the "like" case).

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -139,7 +139,16 @@ impl SbomService {
...
+                        Some(("version", _key)) => match op {
+                            // Exact match for a version in the array
+                            "eq" => Some(format!("versions @> ARRAY[{}]::text[]", v)),
+                            // Substring search within any version in the array
+                            "like" => Some(format!(
+                                "EXISTS (SELECT 1 FROM unnest(versions) AS v WHERE v ILIKE '%' || {} || '%')",
+                                v
+                            )),
+                            _ => None,
+                        },
```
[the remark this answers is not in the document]
gildub: @jcrossley3, I understand but I tried various solutions and so far that's the only one that worked for me.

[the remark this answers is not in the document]
jcrossley3: That branch isn't affecting your results. You can delete it and get the same thing you're getting.

jcrossley3:
> In the meantime I get a test example here is working example the curl output using the `_tmp.json` file from TC-2564:
> ```
> $  curl -s "http://localhost:8080/api/v2/sbom?limit=10&offset=0&q=versions=fbf470d8b5b84606f797d78775b9de88e14fdb43cc47b2db6ff3747b46df323e&sort=name:asc" | jq
> ```
You're using the wrong operator in your query. With `=`, the value has to exactly match the field. But `~` will find partial matches, i.e. "contains" or "like". Try `q=versions~fbf47`

> ```
> $  curl -s "http://localhost:8080/api/v2/sbom?limit=10&offset=0&q=versions=sha256:fbf470d8b5b84606f797d78775b9de88e14fdb43cc47b2db6ff3747b46df323e&sort=name:asc" | jq
> ```
This works because you're using `=` and the value is an exact match.

Neither of these address TC-2564, of course, which would be `q=fbf47` but you'll need to merge #1749 for that to work.

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -933,4 +936,43 @@ mod test {
...
+        log::debug!("Version substring search results: {:#?}", fetched.items);
+        assert_eq!(1, fetched.total);
+
+        // Test all SBOMs without version filter
+        let fetched = service
+            .fetch_sboms(Query::default(), Paginated::default(), (), &ctx.db)
+            .await?;
+        assert_eq!(2, fetched.total);
+
+        Ok(())
```
jcrossley3, line 976: I would encourage you to reorganize this test to look more like [this test](https://github.com/trustification/trustify/blob/60ff46581c36ec09814f1b64067ab0d92d3fce39/modules/fundamental/src/sbom/endpoints/test.rs#L664-L688). Specifically, use `ctx.ingest_documents`, create a `query` closure that can take various "q" values, and then call it for all the queries you need to support, including a full text search, e.g. `q=fbf470d`.

jcrossley3 (changes requested):

gildub:
@jcrossley3, oh now I understand why you mean by _full text search_ allowing to search any text in all available columns of the query. 

I see, the versions column _and_ #1746 are both needed to address TC-2564.

modules/fundamental/src/sbom/service/sbom.rs
jcrossley3: @gildub You may as well just add more `query` calls to `query_sboms_by_array_values` cited above. You can ingest a 3rd doc (the `_tmp.json` one, but maybe rename it something like `etc/test-data/cyclonedx/tc-2564_tmp.json`), and all the existing query expectations should still work, I think.

jcrossley3: `versions` becomes just another array field to test in there with `suppliers` and `authors`

modules/fundamental/src/sbom/endpoints/test.rs
```diff
@@ -684,6 +685,15 @@ async fn query_sboms_by_array_values(ctx: &TrustifyContext) -> Result<(), anyhow
...
     query(2, "authors>ZZZ").await;
+    query(1, "versions=2.13.8.Final-redhat-00004").await;
+    query(1, "versions~2.13.8").await;
+    query(
+        1,
+        "versions=sha256:fbf470d8b5b84606f797d78775b9de88e14fdb43cc47b2db6ff3747b46df323e",
+    )
+    .await;
+    query(1, "versions~fbf470d8").await;
+    query(3, "").await;
```
jcrossley3, line 696: nice, thanks! :smile:

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -139,6 +139,9 @@ impl SbomService {
                     .alias("sbom_node", "r0")
                     .translator(|f, op, v| match f.split_once(':') {
                         Some(("label", key)) => Some(format!("labels:{key}{op}{v}")),
+                        Some(("version", _key)) => {
+                            Some(format!("versions @> ARRAY[{}]::text[]", v))
+                        }
```
jcrossley3, line 144: Whatever you think this is doing, it's not. :stuck_out_tongue_winking_eye: 

If you want to alias the field "version" to the actual field "versions", you could do...

```suggestion
                        None => match f {
                            "version" => Some(format!("versions{op}{v}")),
                            _ => None,
                        }
```

jcrossley3 (commented):
almost there!

modules/fundamental/src/sbom/service/sbom.rs
gildub: @jcrossley3, 
Oh my, and I thought I was doing something useful with that ! 
I still have a split brain between the versions search and the full text search.
Thanks for the suggestion, just added it.

gildub:
@jcrossley3, after rebase following https://github.com/trustification/trustify/pull/1749 and full text search added.

I believe another scale-test to be done once everything is [GREEN].

jcrossley3:
Excellent! I might squash all those commits, too.

jcrossley3 (approved):

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -139,6 +139,10 @@ impl SbomService {
                     .alias("sbom_node", "r0")
                     .translator(|f, op, v| match f.split_once(':') {
                         Some(("label", key)) => Some(format!("labels:{key}{op}{v}")),
+                        None => match f {
```
ctron, line 142: I think it should be possible adding a test for this as well.

modules/ingestor/src/graph/sbom/spdx.rs
```diff
@@ -72,6 +72,25 @@ fn suppliers(sbom: &SPDX) -> Vec<String> {
     Vec::from_iter(result)
 }
 
+/// Extract versions for a SPDX SBOM by collecting versions of describing packages
+fn versions(sbom: &SPDX) -> Vec<String> {
+    // packages describing the SBOM
+    let describing = describing_packages(sbom);
```
ctron, line 78: Initially I'd prefer to have packages (or their versions) returned right away. Taking a look at the implementation of `describing_packages` that's however not really possible, as one source of this information only holds the string, not the full package information.

I think the compromise (searching through everything once again) it ok, but we should document this here.

modules/ingestor/src/graph/sbom/spdx.rs
```diff
@@ -72,6 +72,25 @@ fn suppliers(sbom: &SPDX) -> Vec<String> {
...
+
+    // collect versions for matching packages
+    let mut result = HashSet::new();
+    for p in &sbom.package_information {
+        if !describing.contains(p.package_spdx_identifier.as_str()) {
+            continue;
+        }
+        if let Some(version) = &p.package_version {
```
ctron, line 86: It's ok as it is. There's another way this should work, just that you know, and maybe you already do:

```rust
result.extend(p.package_version.clone());
```

`Option<T>` is an iterable (zero or one item), and `extend` can accept any iterable.

ctron (approved):
LGTM. I'd appreciate a test for covering one more case. But not a blocker.

modules/fundamental/src/sbom/service/sbom.rs
gildub: @ctron, well the following might be simpler :

```suggestion
                    .translator(|f, op, v| match f.split_once(':') {
                        Some(("label", key)) => Some(format!("labels:{key}{op}{v}")),
                        Some(("version", _key)) => Some(format!("versions{op}{v}")),
                        _ => None,
                    }),
```

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -139,10 +139,7 @@ impl SbomService {
                     .alias("sbom_node", "r0")
                     .translator(|f, op, v| match f.split_once(':') {
                         Some(("label", key)) => Some(format!("labels:{key}{op}{v}")),
-                        None => match f {
-                            "version" => Some(format!("versions{op}{v}")),
-                            _ => None,
-                        },
+                        Some(("version", _key)) => Some(format!("versions{op}{v}")),
```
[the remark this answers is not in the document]
jcrossley3: And PLEASE squash all these commits! :smile:

migration/src/m0000010_init_up.sql
```diff
@@ -1896,7 +1896,8 @@ CREATE TABLE public.sbom (
     authors character varying[],
     labels jsonb NOT NULL,
     source_document_id uuid,
-    data_licenses text[] DEFAULT ARRAY[]::text[] NOT NULL
+    data_licenses text[] DEFAULT ARRAY[]::text[] NOT NULL,
+    versions text[] DEFAULT ARRAY[]::text[] NOT NULL
```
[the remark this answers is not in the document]
mrizzi: @gildub please follow @ctron's comment and remove this from the initial script.

mrizzi (changes requested):
The DB migration isn't executed and it requires a further enhancement to really fix the issue and let the user search by version for the already ingested SBOMs.

migration/src/m0001080_sbom_add_versions.rs
```diff
@@ -0,0 +1,42 @@
...
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(Sbom::Table)
+                    .add_column(
+                        ColumnDef::new(Sbom::Versions)
+                            .array(ColumnType::Text)
+                            .not_null(),
```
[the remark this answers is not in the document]
gildub: @mrizzi, thanks, I missed the `lib.rs` reference.

modules/fundamental/src/sbom/service/sbom.rs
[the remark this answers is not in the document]
gildub: @jcrossley3, sure but Github does that usually for me which is much faster and less error prone.

modules/fundamental/src/sbom/endpoints/test.rs
```diff
@@ -1309,6 +1311,24 @@ async fn query_sboms_by_array_values(ctx: &TrustifyContext) -> Result<(), anyhow
     query(2, "authors>ZZZ").await;
     query(2, "organization").await;
     query(1, "tool: syft").await;
+    query(1, "versions=2.13.8.Final-redhat-00004").await;
+    query(1, "versions~2.13.8").await;
```
jcrossley3, line 1315: Since you have the logic in the translator that aliases `versions` to `version`, you can add a test for that, too. Hopefully, that'll make the coverage report a little happier...
```suggestion
    query(1, "versions~2.13.8").await;
    query(1, "version~2.13.8").await;
```

jcrossley3 (commented):
Let's add one more tiny test, please.

modules/fundamental/src/sbom/endpoints/test.rs
gildub: @jcrossley3, done.

migration/src/m0001080_sbom_add_versions.rs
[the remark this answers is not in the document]
mrizzi: @gildub sorry but this issue hasn't be addressed. The already ingested SBOMs will have no value for the `versions` column so `Search it with "78775b9de88el4fdb43cc47b2db6ff3747b46df323e"` will still return nothing after this PR will be merged and released because the SBOM expected to be in the result list won't have any value in the `versions` column to match the query.

There must be a filling of data for the new `versions` column for the existing entries in the `sbom` table to really solve the issue, that's what the second part of my comment above (and @ctron as well) was referring to.

mrizzi (changes requested):
There's still a need for a data filling for already ingested rows into the `sbom` table.

mrizzi:
@sourcery-ai summary

mrizzi:
@sourcery-ai review

gildub:
@ctron, could you please provide an update or the link regarding the migration work issue ?

ctron:
> @ctron, could you please provide an update or the link regarding the migration work issue ?

Sure: https://github.com/guacsec/trustify/pull/1913

It might be a good idea adding your use case the approach in this PR. Just to see it would cover this use case.

gildub:
@mrizzi, a fully supported migration with re-processing of documents is going to take time if accepted.

In the meantime this PR and more importantly the bug behind is not make progress.

We need to decide that until we have that re-processing feature in place then we must accept that there will be a breaking version at some point.

Could we please consider that ?

PR #1707, opened 2025-06-05
feat: start exctracting vulnerability scores from cvss v4 and v2

dejanb:
Currently we only extract base score and severity values. There's need to be more work done to properly save all scores of all versions for vulnerabilities and advisories. The base score used for the vulnerability will be picked from the highest cvss version available. Since base severity was not standardized in cvss v2.0 it is recalculated from the score.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2393,14 +2393,13 @@ dependencies = [
 
 [[package]]
 name = "cve"
-version = "0.3.2"
+version = "0.4.0"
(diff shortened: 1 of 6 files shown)


dejanb:
This PR depends on releasing cve crate with https://github.com/scm-rs/cve/pull/6 in

dejanb:
@ctron @mrizzi with cve crate with the fix now released, this is ready for review

ctron (approved):

PR #1725, opened 2025-06-09
ci: set non-blocking codecov status checks

jcrossley3:
We get the benefit of the report without the annoyance of a failed build.


helio-frota (approved):

ctron:
> We get the benefit of the report without the annoyance of a failed build.

The build was **never** failed by this. Nothing was blocked.

PR #1732, opened 2025-06-10
feat(analysis): log graph cache size during startup

ctron:


mrizzi (approved):
This is useful, thanks @ctron 👍

Issue #1726, opened 2025-06-09
Would be nice to delegate the tests to the coverage-strategy we are using


helio-frota:
To avoid avoid running the tests `2x`:

* The tests
* The tests with coverage data

update:

literally run the tests only in `codecov.yaml` and remove from `ci.yaml`

jcrossley3:
Can we just separate the concerns? Run the codecov workflow daily or maybe on every push to main, not on every PR push.

helio-frota:
> Run the codecov workflow daily or maybe on every push to main, not on every PR push

Honestly I have no idea about the strategy, I was not following the JIRAs so, similar to snyk which I know is part of a JIRA goal/strategy/task , in the case of coverage-feedback-strategy I have no clue on what is happening 👍

helio-frota:
> From my point of view, those two concerns currently are separated

👍 I can live with that.

For me this is the most important part:

> For example, only block if coverage drops by more than Y%. Or, we could configure (if possible) codecov to block the PR if the coverage falls below a base number, let's say, 80% as example. So if a PR causes coverage to drop to 79%, it would be automatically blocked -> by codecov

ctron:
> > That's the information a reviewer can use in approving or rejecting a PR
> 
> ➖ I'm -1 here because this could create friction for both reviewers and contributors. It would be better to delegate this decision to codecov using a defined threshold. For example, only block if coverage drops by more than `Y%`. Or, we could configure (if possible) codecov to block the PR if the coverage falls below a base number, let's say, `80%` as example. So if a PR causes coverage to drop to `79%`, it would be automatically blocked -> by codecov

Right now nothing is blocked. The decision is left to reviewers. There may be good reasons for accepting a bad test coverage in a PR. That depends on the situation.

PR #1735, opened 2025-06-11
Revert "ci: set non-blocking codecov status checks (#1725)"

ctron:
This reverts commit 1bb18f5e712c9eb7d56ee901b5e28d1eebd39489.

As discussed in today's handover call with @PhilipCattanach, this will be reverted. The check was never blocking anything, just reporting. Furthermore, such PRs should be discussed first.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,8 +1,0 @@
-coverage:
-  status:
-    project:
-      default:
-        informational: true
-    patch:
-      default:
-        informational: true


helio-frota (approved):

mrizzi (approved):
Checking the `informational` documentation at https://docs.codecov.com/docs/commit-status#informational, I learnt `If true is specified the resulting status will pass no matter what the coverage is or what other settings are specified` but I think the merit in having codecov is to have a clear understanding of the effect of a PR on the current coverage.
As reported having the codecov builds green is not required for GH to allow moving forward with merging but having a clear feedback on the effect of the PR on the coverage is useful.

helio-frota:
Another confusion on my part about why I merged it:

I thought all the other PRs would have to wait 40+ minutes to be good for merge after clicking "re-run" on CI, which I later found out isn't true

jcrossley3:
I think we're all in agreement that a missed coverage threshold shouldn't block the ability to merge a PR.

The only contention is the "red X" that github uses to visually indicate a failed build, because github doesn't have icons to distinguish between "failed" and "blocked".

The "red X" negatively affects my development workflow, because I use it to determine which PR's are ready for review.

If you don't use it, what's the harm? Why revert the PR?

helio-frota:
> I think we're all in agreement that a missed coverage threshold shouldn't block the ability to merge a PR.

Yes, that's correct :+1:  To be more specific:

* The threshold wasn't blocking anything before.
* As @i386x mentioned in the meeting, there are still some issues in how coverage is collected and processed, and this is from rustc compiler not the tool we are using.
* The fact that it's running in a separate workflow.yaml file is meant to serve as a kind of helper for the reviewer similar to scale-tests. This strategy is consistent with the point Jiri mentioned earlier, that some PRs can be accepted even with a certain percentage of coverage drop, depending on the context.
* Since the strategy does not rely on a fixed minimum base number (e.g. 60%), and considering the previous point raised by Jiri, it doesn't make sense to have a blocking mechanism as I had suggested [here](https://github.com/trustification/trustify/issues/1726#issuecomment-2958798307)

PR #1739, opened 2025-06-11
Use license expressions as the data source for the license filter

bxf12315:


modules/fundamental/src/sbom/endpoints/test.rs
```diff
@@ -30,29 +30,86 @@ async fn fetch_unique_licenses(ctx: &TrustifyContext) -> Result<(), anyhow::Erro
...
+        "MIT AND WTFPL".to_string(),
         "MPL-2.0".to_string(),
         "NOASSERTION".to_string(),
         "Python-2.0".to_string(),
-        "Sendmail".to_string(),
-        "Unlicense".to_string(),
-        "WTFPL".to_string(),
         "Zlib".to_string(),
+        "Zlib AND LicenseRef-Boost".to_string(),
+        "Zlib AND Sendmail AND LicenseRef-8".to_string(),
```
jcrossley3, line 112: I think you can safely remove all the `.to_string()` calls in here, and your assertion will still work.

modules/fundamental/src/sbom/endpoints/test.rs
```diff
@@ -76,16 +133,20 @@ async fn fetch_unique_licenses(ctx: &TrustifyContext) -> Result<(), anyhow::Erro
...
+        "EPL 1.0".to_string(),
         "EPL-1.0".to_string(),
         "EPL-2.0".to_string(),
+        "GNU Lesser General Public License".to_string(),
         "GPL-2.0-with-classpath-exception".to_string(),
         "LGPL-2.1-only".to_string(),
         "MIT".to_string(),
         "MPL-2.0".to_string(),
         "Openfont-1.1".to_string(),
+        "The GNU General Public License, v2 with Universal FOSS Exception, v1.0".to_string(),
```
jcrossley3, line 145: these, too

modules/fundamental/src/sbom/endpoints/test.rs
```diff
@@ -76,16 +133,20 @@ async fn fetch_unique_licenses(ctx: &TrustifyContext) -> Result<(), anyhow::Erro
...
     ];
     match response {
         Value::Array(ref list) => {
+            println!("{:?}", list.clone());
```
jcrossley3, line 149: log::debug?

modules/fundamental/src/sbom/endpoints/test.rs
```diff
@@ -30,29 +30,86 @@ async fn fetch_unique_licenses(ctx: &TrustifyContext) -> Result<(), anyhow::Erro
     let req = TestRequest::get().uri(&uri).to_request();
     let response: Value = app.call_and_read_body_json(req).await;
     let expected_result = &vec![
+        "(LicenseRef-0 OR LicenseRef-AFL) AND LicenseRef-0".to_string(),
+        "(LicenseRef-10 OR LicenseRef-0) AND LicenseRef-6".to_string(),
+        "(LicenseRef-4 OR LicenseRef-Artistic) AND LicenseRef-5 AND LicenseRef-UCD".to_string(),
```
mrizzi, line 35: I wasn't expecting the `LicenseRef` to show up and I'm sure we don't want this as evaluated meaningless to the user but it should rather show the license name (ref. `If the license stored against the package is an internal licenseId then it will contain the corresponding name` in [TC-1902](https://issues.redhat.com/browse/TC-1902?focusedId=27095913&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27095913)).
Furthermore, checking the `LicenseRef-Artistic` for example, I can not find anyway its name, i.e. `Artistic`, in the previous test so it looks like, in a way or another, there's a bug in the code in the current `main` branch.

bxf12315: First, it is very difficult to determine whether a license expression is a license reference before parsing it.
[1]  https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/
Second, how should we handle the cases("MIT AND LicenseRef-2 AND LicenseRef-CC-BY AND LicenseRef-GPLv3") where a license expression includes both an SPDX license ID (such as MIT) and a license reference?

modules/fundamental/src/sbom/endpoints/test.rs
bxf12315: Done

modules/fundamental/src/sbom/endpoints/test.rs
bxf12315: done

modules/fundamental/src/sbom/endpoints/test.rs
bxf12315: done

carlosthe19916:
@bxf12315 thanks for the PR. Seems like @mrizzi is looking into this. So far I see this PR is consistent with what I see in the CSV file report.

PR #1751, opened 2025-06-13
refactor(tests): Refactor organization tests

helio-frota:
Group assertions, reuse code, and speeding up tests

Before:
boefore

After:

Related to https://github.com/trustification/trustify/discussions/1742

### DIFF
--- a/modules/fundamental/src/organization/endpoints/test.rs
+++ b/modules/fundamental/src/organization/endpoints/test.rs
@@ -8,56 +8,53 @@ use test_log::test;
 use trustify_common::db::query::Query;
 use trustify_common::hashing::Digests;
 use trustify_common::model::Paginated;
-use trustify_module_ingestor::graph::advisory::AdvisoryInformation;
+use trustify_module_ingestor::graph::{
+    Outcome,
+    advisory::{AdvisoryContext, AdvisoryInformation},
+};
 use trustify_test_context::{TrustifyContext, call::CallService};
+
+async fn ingest_advisory<'ctx>(
+    ctx: &'ctx TrustifyContext,
+    id: &str,
+    issuer: &str,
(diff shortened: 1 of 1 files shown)


modules/fundamental/src/organization/endpoints/test.rs
```diff
@@ -68,33 +65,11 @@ async fn all_organizations(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
...
     );
 
-    Ok(())
-}
-
-#[test_context(TrustifyContext)]
-#[test(actix_web::test)]
-async fn one_organization(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
-    let app = caller(ctx).await?;
+    db.refresh().await?;
```
ctron, line 68: I'm against re-using database instances. We re-create DB instances for a good reason.

modules/fundamental/src/organization/endpoints/test.rs
```diff
@@ -34,30 +38,23 @@ async fn all_organizations(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
...
-            &ctx.db,
-        )
-        .await?;
+    Ok(advisory)
+}
 
-    let uri = "/api/v2/organization?sort=name";
+#[test_context(TrustifyContext)]
+#[test(actix_web::test)]
+async fn organization_tests(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
```
ctron, line 46: I am against merging all tests into one. That makes it very hard to debug single failed tests.

ctron (changes requested):

modules/fundamental/src/organization/endpoints/test.rs
helio-frota: I agree with this part here :+1:

PR #1755, opened 2025-06-13
fix(analysis): don't load all SBOMs when checking by name

ctron:


ctron:
@JimFuller-RedHat changed the tests, I'd appreciate another review from you.

PR #1759, opened 2025-06-16
fix: prevent dumping massive amounts of log info

ctron:
Having the `graphs` variable logged dumps out an enormous amount of data, not adding any value. Blocking the log file, having an impact on performance.

I believe this has an impact on: https://issues.redhat.com/browse/TC-2567


JimFuller-RedHat (approved):
LGTM

PR #1761, opened 2025-06-16
Compile dependencies with no debuginfo

i386x:
See https://github.com/trustification/guac-rs/commit/265fba20fb1f359b311029a348e9be73f406a266

This should save us a few gigabytes of disk space on GitHub runners, moving `ldd: no space left on device` error a bit further.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -184,6 +184,11 @@ postgresql_archive = { version = "0.18.0", default-features = false, features = 
 postgresql_embedded = { version = "0.18.0", default-features = false, features = ["theseus", "rustls"] }
 postgresql_commands = { version = "0.18.0", default-features = false, features = ["tokio"] }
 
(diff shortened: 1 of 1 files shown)


helio-frota:
I think we should try this +1

ctron:
Is there a way we can do this by using just CLI parameters? So we could enable this in the CI workflow?

i386x:
So `tarpaulin` failed after 24 minutes with the same error as previously: https://github.com/trustification/trustify/actions/runs/15678399958/job/44163935816?pr=1761#step:9:817

It seems that this change has no effect on how `tarpaulin` compiles the project. Maybe `tarpaulin` is overriding the settings in `Cargo.toml` using its own environment?

i386x:
Closing this as this PR does not solve the problem with space.

PR #1764, opened 2025-06-16
fix: mark advisory average scores and severites deprecated

dejanb:
According to adr https://github.com/trustification/trustify/blob/main/docs/adrs/00004-advisory-scores.md
we don't want to calculate average scores for advisories anymore.
The values like aggregated severity will be extracted from appropriate advisories
and stored as labels.
This commit deprecates average score and severity fields in models and modifies the query to
stop calculating these values.


dejanb:
@jcrossley3 there are no real scores/severities on advisory level, only what advisory says about vulnerabilities it describes. The "average" scores that we had are artificial and not something that is used in practice, hence the removal.
I thought about removing the fields immediately instead of marking them deprecated for now. I was thinking to do one PR with braking API changes as there will be a few more in the same area.

PR #1777, opened 2025-06-19
fix: ensure latest cpe query returns latest roots

JimFuller-RedHat:
There were a few issues here to fix:

- [x]  loosen the requirement to have duplicate bom-refs in cdx (which is invalid cdx!), specifically on the metadata.component
- [x]  ensure cpe queries actually returned all top level roots
- [x]  added test for TC-2606
- [x] review/fix existing cdx tests
- [x] review/fix existing spdx tests

(fixes TC-2606)

### DIFF
--- a/entity/src/sbom_node.rs
+++ b/entity/src/sbom_node.rs
@@ -20,6 +20,12 @@ pub enum Relation {
     )]
     Package,
     #[sea_orm(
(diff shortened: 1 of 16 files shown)


dejanb (approved):
Looks good. Thanks

Issue #1778, opened 2025-06-19
fetch_unique_licenses test isn't deterministic


jcrossley3:
https://github.com/trustification/trustify/blob/880f425f3c00fd9a251d511a9cf2deea8abf9966/modules/fundamental/src/sbom/endpoints/test.rs#L250-L255

The above assertion fails locally for me. The results in the response are in a different order than expected. Here's what I got:
```
[
  {
    "license_id": "LicenseRef-AFL",
    "license_name": "AFL"
  },
  {
    "license_id": "Apache-2.0",
    "license_name": "Apache-2.0"
  },
  {
    "license_id": "LicenseRef-Artistic",
    "license_name": "Artistic"
  },
  {
    "license_id": "LicenseRef-5",
    "license_name": "Artistic 2.0"
  },
  {
    "license_id": "LicenseRef-2",
    "license_name": "ASL 2.0"
  },
  {
    "license_id": "LicenseRef-Boost",
    "license_name": "Boost"
  },
  {
    "license_id": "LicenseRef-BSD",
    "license_name": "BSD"
  },
  {
    "license_id": "BSD-2-Clause",
    "license_name": "BSD-2-Clause"
  },
  {
    "license_id": "BSD-2-Clause-Views",
    "license_name": "BSD-2-Clause-Views"
  },
  {
    "license_id": "BSD-3-Clause",
    "license_name": "BSD-3-Clause"
  },
  {
    "license_id": "BSD-3-Clause-Clear",
    "license_name": "BSD-3-Clause-Clear"
  },
  {
    "license_id": "CC0-1.0",
    "license_name": "CC0-1.0"
  },
  {
    "license_id": "LicenseRef-CC-BY",
    "license_name": "CC-BY"
  },
  {
    "license_id": "CC-BY-SA-4.0",
    "license_name": "CC-BY-SA-4.0"
  },
  {
    "license_id": "CDDL-1.0",
    "license_name": "CDDL-1.0"
  },
  {
    "license_id": "CDDL-1.1",
    "license_name": "CDDL-1.1"
  },
  {
    "license_id": "LicenseRef-7",
    "license_name": "Copyright only"
  },
  {
    "license_id": "EPL-1.0",
    "license_name": "EPL-1.0"
  },
  {
    "license_id": "LicenseRef-14",
    "license_name": "EPL-2.0 OR GNU General Public License, version 2 with the GNU Classpath Exception"
  },
  {
    "license_id": "LicenseRef-13",
[...]

PR #1779, opened 2025-06-19
Fix: Avoid errors caused by different orderings of ouput.

bxf12315:
fixs: https://github.com/trustification/trustify/issues/1778


jcrossley3 (approved):
nice, thanks! confirmed the fix locally.

PR #1783, opened 2025-06-20
ci: using the same ubuntu version across workflows

helio-frota:


helio-frota:
> For others it causes problems, especially when libraries get updated like libc

+1

I'll update the PR

ctron (approved):

PR #1801, opened 2025-06-24
ci: Add a filter for docs avoiding unnecessary run

helio-frota:
Run `npm ci` and `make all` only if `docs/book` was changed


helio-frota:
ok let's close this :+1:

Issue #1684, opened 2025-06-03
Vulnerabilities not reported for go package


ruromero:
# Description

When using the `/api/v2/vulnerability/analyze` endpoint for some packages, nothing is reported. I have used the same purl against the old `/api/v1/analyze` endpoint and I get the expected results.

## Reproduce

In the staging environment where the DA Exhort backend is integrated with (server-trustify-latest.apps.cluster.trustification.rocks)

Package "pkg:golang/golang.org/x/text@v0.3.6" has 2 vulnerabilities cve-2022-32149 and cve-2021-38561 but the following body returns an empty response

Request:

```
{
  "purls": [
    "pkg:golang/golang.org/x/text@v0.3.6"
  ]
}
```

Response 200 OK:

```
{}
```

## Additional information

- When I query other packages like `pkg:npm/lodash@4.17.21` I get the expected vulnerability
- The golang vulnerabilities are accessible through the UI https://server-trustify-latest.apps.cluster.trustification.rocks/vulnerabilities/CVE-2022-32149

dejanb:
Fixed with #1821

PR #1821, opened 2025-07-02
fix: add support for golang versions with 'v' prefix

dejanb:
Go ecosystem often uses 'v' prefix before semver formatted versions

https://go.dev/ref/mod#versions

On the other hand, OSV and other adivsories are reporting pure semver versions. Newly introduced golang_version_matches pgsql function, strips the prefix if it is present and then continues with semver match.


dejanb:
Fixes #1684

PR #1822, opened 2025-07-02
refactor: optimise graph loading

JimFuller-RedHat:
Turns out we were getting duplicates due to processing external sbom references. This PR avoids  extra loops as well as return distinct set of sbom analysis graphs.


JimFuller-RedHat:
closing this as I have another PR which will supersede

PR #1824, opened 2025-07-03
chore: prepares for future rust update

helio-frota:


ctron:
I'd want to put a hold on this one. It seems like Rust versions are problematic for downstream right now. And I'd not want to disrupt releases that much at the current phase of things.

helio-frota:
Should I revert only the rust version part and keep the string interpolation ?

ctron (approved):

PR #1827, opened 2025-07-03
chore: removes unused dependencies

helio-frota:
Based on https://gist.github.com/helio-frota/fd06f1b67168b22cc649504729b9eeec
But I had to check one by one, unfortunately `cargo build --timings` showed almost nothing speed gains from the root dir. I noticed only ~5 seconds improvements when building inside `xtask` dir for example.


ctron (approved):
Valid. But doesn't really change the dependency list itself.

PR #1860, opened 2025-07-10
ci: give cargo llvm-cov a chance

ctron:
## Summary by Sourcery

Update the Codecov GitHub Actions workflow to switch from cargo-tarpaulin to cargo-llvm-cov and install required LLVM tools

CI:
- Use taiki-e/install-action@v2 to install cargo-llvm-cov
- Add llvm-tools-preview via rustup
- Replace cargo tarpaulin coverage step with cargo llvm-cov generating codecov JSON output

### DIFF
--- a/.github/workflows/codecov.yaml
+++ b/.github/workflows/codecov.yaml
@@ -54,20 +54,17 @@ jobs:
         run: |
           docker compose -f etc/deploy/compose/compose-minio.yaml up -d --wait
 
-      - name: Setup cargo-binstall (Linux)
-        if: runner.os != 'Windows'
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-llvm-cov
 
-      - name: Install cargo-tarpaulin
-        run:
-          cargo binstall cargo-tarpaulin -y --force
+      - name: Install llvm-tools-preview
+        run: |
+          rustup component add llvm-tools-preview
 
       - name: Test
         # use only one job, trying to reduce memory usage
-        run: cargo tarpaulin --jobs 1 --timeout 300 --out xml --skip-clean --features _test-s3
+        run: cargo llvm-cov --codecov --jobs 1 --features _test-s3 --output-path codecov.json
         env:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}" # for embedded postgresql
 


.github/workflows/codecov.yaml
```diff
@@ -54,20 +54,11 @@ jobs:
...
-        if: runner.os != 'Windows'
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
-
-      - name: Install cargo-tarpaulin
-        run:
-          cargo binstall cargo-tarpaulin -y --force
+      - uses: taiki-e/install-action@cargo-llvm-cov
```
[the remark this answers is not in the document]
ctron: I general this is correct, but in this specific case, you will also need to add the tool name as a parameter.

helio-frota:
@ctron in this PR it was not required to install llvm-tools and it worked fine https://github.com/trustification/trustify/pull/1834/files

.github/workflows/codecov.yaml
```diff
@@ -54,20 +54,17 @@ jobs:
...
-        run:
-          cargo binstall cargo-tarpaulin -y --force
+      - name: Install llvm-tools-preview
+        run: |
+          rustup component add llvm-tools-preview
 
       - name: Test
         # use only one job, trying to reduce memory usage
-        run: cargo tarpaulin --jobs 1 --timeout 300 --out xml --skip-clean --features _test-s3
+        run: cargo llvm-cov --codecov --jobs 1 --features _test-s3 --output-path codecov.json
```
helio-frota, line 67: Do we really need this `jobs` restriction ?
this worked fine https://github.com/trustification/trustify/pull/1834/files#diff-d968391b3a7edaed2de53a8c60aa2a164c40c9bf62511aee0d95657672245637R68 without causing `OOM`

ctron: Yes, because otherwise it may run into out of memory issues when linking all binaries at the same time.

helio-frota:
this PR says `Successful in 35m` even with `--jobs 1` :+1: 

Compared with an PR from other day `55 m` 

462202484-e5ed88b0-660b-4d05-89ee-a735ed99905a

helio-frota:
I don't know exactly why you are using this:
`cargo llvm-cov --codecov --jobs 1 --features _test-s3 --output-path codecov.json`

in my previous comparison I used this: 
`cargo llvm-cov --jobs 1 --features _test-s3 --no-clean --cobertura --output-path cobertura.xml`

Because tarpaulin is using `--out xml`:
`cargo tarpaulin --jobs 1 --timeout 300 --out xml --skip-clean --features _test-s3`

Docs:

ctron:
`--codecov` is the switch for generating codecov.io compatible data. what's the benefit of using the other format?

helio-frota:
cool then, it is still as a draft pr :+1:

Issue #1864, opened 2025-07-10
Add a delete method to the StorageBackend trait


jcrossley3:
It's yet to be determined how/when it will be invoked -- possibly triggered by a similar mechanism for #1851 -- but we know we'll need implementations for our backends when that time comes.

PR #1871, opened 2025-07-14
feat: add delete method to Storage trait

jcrossley3:
fixes: #1864

## Summary by Sourcery

Implement an idempotent delete operation across all storage backends, refactor filesystem retrieval logic with a locate helper, and extend tests to cover deletion behavior

New Features:
- Add delete method to StorageBackend trait
- Implement delete for FileSystemBackend, S3Backend, and DispatchBackend

Enhancements:
- Extract locate helper in FileSystemBackend to unify file lookup across retrieve and delete
- Refactor FileSystemBackend retrieve method to use the new locate helper

Tests:
- Introduce test_store_read_and_delete to verify delete functionality and idempotency
- Update existing storage backend tests to use the new delete-aware test helper

### DIFF
--- a/modules/storage/src/service/fs.rs
+++ b/modules/storage/src/service/fs.rs
@@ -6,7 +6,7 @@ use bytes::Bytes;
 use futures::Stream;
 use std::{
     fmt::Debug,
-    io::{Error as IoError, ErrorKind, Result as IoResult},
+    io::{ErrorKind, Result as IoResult},
     path::{Path, PathBuf},
 };
 use strum::IntoEnumIterator;
@@ -165,10 +165,7 @@ impl StorageBackend for FileSystemBackend {
     async fn delete(&self, key: StorageKey) -> Result<(), Self::Error> {
         match self.locate(key).await? {
             Some((path, _)) => remove_file(path).await,
-            None => Err(IoError::new(
-                ErrorKind::NotFound,
-                "No document found in storage matching key",
-            )),
+            None => Ok(()),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


jcrossley3:
@sourcery-ai review

modules/storage/src/service/fs.rs
```diff
@@ -93,6 +93,17 @@ impl FileSystemBackend {
             .await
             .map(|result| (result, dir))
     }
+
+    fn locate(&self, StorageKey(hash): StorageKey) -> IoResult<Option<(PathBuf, Compression)>> {
```
ctron, line 97: As file system call may block, this should be `async` all the way.

ctron:
What would happen if one request reads the document, while the other one deletes it?

ctron:
How would it manifest on the REST API if we'd delete a non-existing document? I think that should be documented too.

jcrossley3:
> How would it manifest on the REST API if we'd delete a non-existing document? I think that should be documented too.

Perhaps, but the REST API is beyond the scope of this PR.

jcrossley3:
> how could the REST API distinguish between a non-existent file and another error.

The error returned will be `ErrorKind::NotFound` for a non-existent file.

jcrossley3:
> Same for the S3 implementation? And the "dispatch"?

S3 returns its own "NotFound" variant. And, as you well know, dispatch returns an anyhow::Error. We could introduce a new error variant to allow callers to distinguish "notfoundness" or we could just make delete idempotent, as the AI suggests.

As we've yet to build the caller, I'm not sure how much the caller will care about the error, so I'm inclined to go the simpler route and make the delete idempotent. wfy?

ctron:
@sourcery-ai review

PR #1883, opened 2025-07-17
fix: properly filter cvss scores for the advisory

dejanb:
Currently the logic incorrectly filters CVSS scores based on the vulnerability ID instead of the advisory ID. This change ensures that only CVSS scores related to the specific advisory are included in the advisory summary.

## Summary by Sourcery

Fix CVSS filtering logic for advisory summaries to use advisory_id instead of vulnerability_id, update related invocation, and add validation tests

Bug Fixes:
- Filter CVSS scores by advisory_id in VulnerabilityAdvisoryHead instead of vulnerability_id
- Correct arguments passed to VulnerabilityAdvisoryHead::from_entities in vulnerability summary

Enhancements:
- Remove unused vulnerability parameter from VulnerabilityAdvisoryHead::from_entities signature

Tests:
- Add test verifying only advisory-specific CVSS scores are included in fetched vulnerabilities

### DIFF
--- a/modules/fundamental/src/vulnerability/model/details/vulnerability_advisory.rs
+++ b/modules/fundamental/src/vulnerability/model/details/vulnerability_advisory.rs
@@ -85,7 +85,6 @@ impl VulnerabilityAdvisoryHead {
         }
     }
     pub async fn from_entities<C: ConnectionTrait>(
-        vulnerability: &vulnerability::Model,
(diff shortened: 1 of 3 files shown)


helio-frota (approved):
approving the rust part -- no idea about the business

ctron (approved):

PR #1886, opened 2025-07-21
fix: prevent calculating the total if we don't need it

ctron:
In some cases we calculate a "total number of items" although we don't need it. This is due to the fact that we support pagination, and for this we need to execute the SQL statement twice: once for getting the page of items, and once for getting the total number of items.

In some cases however, we just drop the total, without ever using it. In such cases executing the count() statement is a wast of time and resources.

This change allows the caller of a function to influence the output type and with that trigger an executing of the count() variant, or skip it.


JimFuller-RedHat (approved):
LGTM ... this may have an impact on query plans - running load tests should catch any perf regression ... it might also be that we can drop some indexes at some point as well but would wait a few releases to do that.

Issue #1898, opened 2025-07-28
Can only see results for "percentage of search request response times above 1s" when reducing bucket granularity


helio-frota:
Currently, I'm only able to see the results for the percentage of search request response times above 1 second when I reduce the bucket granularity.
With higher-level granularity, the data either doesn’t show up or is too sparse to be useful. Would be good to understand whether this is expected behavior or if there's a way to improve visibility without lowering the bucket resolution too much.

update:
This is related to this blank panel [here](https://github.com/trustification/trustify/pull/1899#issuecomment-3126416318)

helio-frota:
Ok, @ctron @mrrajan I’ve figured out what’s going on here, It’s not just a matter of the search operations (blank panel in previous comment ^) being fast, or because I was previously testing with an empty database, as I had mentioned internally before. I’ve outlined a step-by-step explanation for confirmation and also a reproducible example locally:

* Decisions were made to use [seconds instead of milliseconds](https://github.com/open-telemetry/opentelemetry-specification/issues/2977)
* Here's a [commit](https://github.com/OutThereLabs/actix-web-opentelemetry/commit/be46aff899fd26e9d9965116bd1b800a13eb4fb9#diff-d36878526203bd26015871a44e04601167ef4f5186b284530041e68ec1f75f0dR26) in actix-web-opentelemetry that transitions to using lower granularity buckets
* Why [views](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view) are used to customize aggregations
* The [default values](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#explicit-bucket-histogram-aggregation) (in seconds)
* With that, our `http_server_duration_seconds_bucket` will fall into something bellow `5s` bucket for the search, meaning that 
this PromQL `http_server_duration_seconds_bucket{http_route=~"/api/v2/sbom", job = "trustify", le="1"}` will show nothing
* Why did I only notice this now? Well, `v1` was using a different crate for this: https://github.com/nlopes/actix-web-prom, which talks directly to Prometheus. Now we’re talking with OTELCOL, so if we don’t define a custom view with more granular buckets, we’ll have to adapt the PromQL to work with the default bucket configuration, which ends up only exposing values (for that blank panel ^) above 5 seconds. That doesn’t make sense, since it’s essentially a different metric, and we also lose a lot of precision in the process.

helio-frota:
Closing, we need to use the `view` for this

PR #1908, opened 2025-08-01
chore: using custom bucket boundaries for http metrics

helio-frota:
* This aligns with https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpserverrequestduration
* This is related to https://github.com/trustification/trustify/issues/1898

## Summary by Sourcery

Configure custom histogram bucket boundaries for HTTP server duration metrics per OTEL conventions and enable metrics views support in the SDK.

Enhancements:
- Add a view for http.server.duration to use explicit bucket histogram boundaries and record min/max.
- Enable the spec_unstable_metrics_views feature in opentelemetry_sdk to support custom metric views.

### DIFF
--- a/common/infrastructure/Cargo.toml
+++ b/common/infrastructure/Cargo.toml
@@ -21,7 +21,7 @@ mime = { workspace = true }
 openssl = { workspace = true }
 opentelemetry = { workspace = true }
 opentelemetry-otlp = { workspace = true, features = ["grpc-tonic"] }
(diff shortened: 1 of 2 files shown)


ctron:
I think we agreed an the last metrics call to not perform this change, but wait until someone specifically asks for this. Or there's an alternative.

helio-frota:
yeah ok let's close the PR :+1:

PR #1911, opened 2025-08-04
perf(analysis): Parallelize graph query execution

dejanb:
This commit significantly improves the performance of graph analysis by applying parallelization techniques to the run_graph_query and collect_graph functions.

  Previously, the collection of ancestors and descendants for a given node was performed sequentially. By refactoring run_graph_query to use futures::join!, we now process both directions concurrently, reducing the overall
  execution time.

  Building on this, the collect_graph function was optimized to process all discovered nodes in parallel using join_all. This ensures that we leverage available resources more efficiently when analyzing multiple entry points in
  the graph.

Assisted-by: Gemini

## Summary by Sourcery

Optimize graph analysis performance by refactoring query and collection routines to execute traversals and node processing in parallel using futures::join! and join_all.

Enhancements:
- Parallelize run_graph_query to concurrently collect ancestors and descendants using futures::join!
- Process all candidate nodes in collect_graph concurrently across graphs with join_all instead of sequential iteration
- Refactor DiscoveredTracker to use tokio::sync::Mutex and make visit asynchronous for non-blocking locking
- Modify collector.collect_graph to traverse edges in parallel with join_all and concurrent futures
- Update collect_graph signature to accept async closure factories for node creation

### DIFF
--- a/modules/analysis/src/service/collector.rs
+++ b/modules/analysis/src/service/collector.rs
@@ -1,4 +1,5 @@
 use super::*;
+use futures::stream::{self, StreamExt};
 use parking_lot::Mutex;
 use std::{collections::HashMap, sync::Arc};
 
@@ -33,6 +34,7 @@ pub struct Collector<'a, C: ConnectionTrait> {
     discovered: DiscoveredTracker,
     relationships: &'a HashSet<Relationship>,
     connection: &'a C,
+    concurrency: usize,
 }
 
 impl<'a, C: ConnectionTrait> Collector<'a, C> {
@@ -47,6 +49,7 @@ impl<'a, C: ConnectionTrait> Collector<'a, C> {
             discovered: self.discovered.clone(),
             relationships: self.relationships,
             connection: self.connection,
+            concurrency: self.concurrency,
         }
     }
 
@@ -61,6 +64,7 @@ impl<'a, C: ConnectionTrait> Collector<'a, C> {
         depth: u64,
         relationships: &'a HashSet<Relationship>,
         connection: &'a C,
+        concurrency: usize,
     ) -> Self {
         Self {
             graph_cache,
@@ -72,6 +76,7 @@ impl<'a, C: ConnectionTrait> Collector<'a, C> {
             discovered: Default::default(),
             relationships,
             connection,
+            concurrency,
         }
     }
 
@@ -100,6 +105,7 @@ impl<'a, C: ConnectionTrait> Collector<'a, C> {
             discovered: self.discovered.clone(),
             relationships: self.relationships,
             connection: self.connection,
+            concurrency: self.concurrency,
         }
     }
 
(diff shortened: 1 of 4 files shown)


modules/analysis/src/service/collector.rs
```diff
@@ -240,44 +240,54 @@ impl<'a, C: ConnectionTrait> Collector<'a, C> {
...
     pub async fn collect_graph(&self) -> Vec<Node> {
-        let mut result = vec![];
         log::debug!("Collecting graph for {:?}", self.node);
-        for edge in self.graph.edges_directed(self.node, self.direction) {
+        let edges: Vec<_> = self
+            .graph
+            .edges_directed(self.node, self.direction)
+            .collect();
+
+        let futures = edges.iter().map(|edge| {
```
ctron, line 249: I'm wondering of the `.collect` is required? Or if it can be used as the iterator in the next step?

dejanb:
The changes in this commit improved performance of the analysis graph tenfold for the large sboms.

In particular this call on the load tests dataset wen from

```
time curl "http://localhost:8080/api/v2/analysis/component/cpe%3A%2Fa%3Aredhat%3Aopenstack%3A13%3A%3Ael7?descendants=10&limit=20&offset=0"
```
from
```
Executed in  671.87 secs - sbom not in the cache
Executed in  651.17 secs - sbom cached
```
to
```
Executed in   49.15 secs - sbom not in the cache
Executed in   23.03 secs - sbom cached
```

For smaller sboms, the difference is not that significant when the sbom is not in the cache, but it's also noticeable when it's it already ion the memory.

```
time curl "http://localhost:8080/api/v2/analysis/latest/component/cpe%3A%2Fa%3Aredhat%3Aquarkus%3A2.13%3A%3Ael8?descendants=10&limit=20&offset=0"
```
from
```
Executed in    3.78 secs
Executed in  476.32 millis
```
to
```
Executed in    3.37 secs
Executed in    2.72 secs
```

dejanb:
I'll go next and examine sourcery (and ctron's :) ) reviews

ctron:
My concern with this change (I know I always have concerns) is that there's an unbounded fork of tasks. Sure that will improve performance, but it might also spike resource consumption. Question is, can we somehow limit this? And make this limit configurable?

dejanb:
The reworked implementation now uses `buffer_unordered()` instead of `join_all()` and bound the parallelisation. I did some manual tests with different values to find a sweet spot for performance/peak memory usage.

The test with 

`time curl "http://localhost:8080/api/v2/analysis/latest/component/cpe%3A%2Fa%3Aredhat%3Aopenstack%3A13%3A%3Ael7?descendants=10&limit=20&offset=0"` 

against load testing dataset.

 | Concurrency | Time (s) |  Peak Memory (MB) |
 |:------------|---------:|-----------------:|
 | `before`      |   638.96 |            413.2 |
 | `1`         |   705.16 |            370.7 |
 | `2`         |   351.76 |           407.5 |
 | `5`         |   139.47 |             424.1 |
 | `10`        |   101.33 |             444.2 |
 | `100`       |    92.81 |             673.3 |
 | `1000`      |    62.72 |         1,173.1 |
 | `10000`     |    52.41 |          1,390.1 |

So, I picked `10` a default value as it improves performances significantly without blowing up the memory. The value is configurable thought environment variable if anyone need more performances and have resources to support it. There's a few more things we could improve in the future, but it's not critical to do it now.

It'd be great if we could periodically run memory profiling during load/scaling tests in the future.

dejanb:
@ctron Thanks for the suggestions. Flattening the stream definitely makes sense and I think additionally impacts the memory usage in a positive way.

I also implemented the concurrency config with NonZeroUsize. I'm not super happy with the usage pattern, but I think it's good for now. There are also a few more places in the code where this could be applied.

Can you give it another quick look before merging?

ctron (approved):

PR #1930, opened 2025-08-20
fix(compose): /init-sso/init.sh: Permission denied

helio-frota:


ctron (changes requested):
`SELINUX_VOLUME_OPTIONS` is actually exactly for this. As not all platforms have SElinux.

PR #1958, opened 2025-09-05
feat: enhance correlation for SBOMs without CPE

mrizzi:
https://issues.redhat.com/browse/TC-2805


dejanb (approved):
Looks good. I did manual scale test for the endpoint. No regressions. The issue to improve this for the future is https://github.com/guacsec/trustify-scale-test-runs/issues/17

PR #1974, opened 2025-09-12
TC-2826	Packages List API - Add licenses array

bxf12315:
## Summary by Sourcery

Enhance the PURL endpoints to return a detailed licenses array and SPDX reference mappings, centralize license models, update the OpenAPI spec, and add an integration test to validate the new functionality.

New Features:
- Add `licenses` and `licenses_ref_mapping` arrays to PURL details response to include declared and concluded license information and SPDX LicenseRef mappings

Enhancements:
- Introduce `LicenseInfo` and `LicenseRefMapping` models in the common module and remove duplicates
- Implement `extract_license_ref_mappings` to parse SPDX expressions and populate reference mappings
- Expose `get_licensing_infos` service method publicly for retrieving SBOM license metadata

Documentation:
- Update OpenAPI spec to reference the new `LicenseInfo` and `LicenseRefMapping` schemas in the PurlDetails component

Tests:
- Add integration test for the PURL license details endpoint to verify licenses and reference mappings are returned correctly

### DIFF
--- a/modules/fundamental/src/purl/model/details/purl.rs
+++ b/modules/fundamental/src/purl/model/details/purl.rs
@@ -41,7 +41,7 @@ pub struct PurlDetails {
     pub base: BasePurlHead,
     pub advisories: Vec<PurlAdvisory>,
     pub licenses: Vec<LicenseInfo>,
-    pub license_ref_mapping: Vec<LicenseRefMapping>,
+    pub licenses_ref_mapping: Vec<LicenseRefMapping>,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, ToSchema, FromQueryResult)]
@@ -149,7 +149,7 @@ impl PurlDetails {
             base: BasePurlHead::from_entity(&package),
             advisories: PurlAdvisory::from_entities(purl_statuses, product_statuses, tx).await?,
             licenses: purl_license_info,
-            license_ref_mapping,
+            licenses_ref_mapping: license_ref_mapping,
         })
     }
 }
--- a/entity/src/sbom_package_license.rs
+++ b/entity/src/sbom_package_license.rs
@@ -68,21 +68,3 @@ impl Related<super::license::Entity> for Entity {
 }
 
 impl ActiveModelBehavior for ActiveModel {}
-
-impl LicenseCategory {
-    /// Returns the string representation of the license category
-    ///
-    /// # Examples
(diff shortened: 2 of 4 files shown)


modules/fundamental/src/purl/model/details/purl.rs
```diff
@@ -36,7 +40,15 @@ pub struct PurlDetails {
     pub version: VersionedPurlHead,
     pub base: BasePurlHead,
     pub advisories: Vec<PurlAdvisory>,
-    pub licenses: Vec<PurlLicenseSummary>,
+    pub licenses: Vec<LicenseInfo>,
+    pub license_ref_mapping: Vec<LicenseRefMapping>,
```
mrizzi, line 44: Please rename this to `licenses_ref_mapping` to be exactly the same as the one current returned from the SBOM packages endpoint, ref. https://github.com/guacsec/trustify/blob/3818788ca7f893d3d9bfa03bd71b03216017a857/modules/fundamental/src/sbom/model/mod.rs#L135

mrizzi (changes requested):
Rename a field to give consistency with the same array already provided from another endpoint. Remember to accordingly update the test.

modules/fundamental/src/purl/model/details/purl.rs
mrizzi: You did partially because, as I wrote in the review above, also the `test_purl_license_details` test must be changed accordingly and, in fact, that test fails with the last commit.
Please always remember to run affected tests locally before pushing the code.

bxf12315: ok

mrizzi:
@sourcery-ai summary

mrizzi:
It looks like the only affected scale test is the `search_purls` which, checking the last days executions in https://guacsec.github.io/trustify-scale-test-runs/, looks pretty unstable considering that, at the time of writing, the last commit on Trustify has been on Sept. 11th but the performance kept changing a lot every day.

carlosthe19916:
@mrizzi @bxf12315 I have reviewed this PR and the endpoint `/api/v2/purl/{id}` does contains new fields called `licenses` and `licenses_ref_mapping` which are exactly the same fields we use in the SBOM Details Page. So we are being consistent. Good job. LGTM!

mrizzi (approved):
@bxf12315 thank you 👍

PR #1979, opened 2025-09-16
remove vulnerability delete api method

m-brophy:


ctron (approved):

ctron:
I am wondering if this should target 0.3.x, or better `main` first.

ctron:
> > I am wondering if this should target 0.3.x, or better `main` first.
> 
> Could we backport it to main? Or is it better doing it the other way around?

I'd say the other way round. New features (unless critical bug fixes) should to go main and then get backported.

Issue #1977, opened 2025-09-15
`gc_purls.sql` may delete packages that are only referenced by an advisory


helio-frota:
I can see the packages when turning off with `TRUSTD_GC_FREQ=0s`

<img width="589" height="89" alt="Image" src="https://github.com/user-attachments/assets/ae697640-c0e6-4c09-8959-374fe378f643" />

jcrossley3:
I'm guessing there may be multiple files referring to the same base package identified by a purl. See if you can find 2 CSAF files each referring to the same packages that demonstrates the issue. That would at least give us steps to reproduce.

helio-frota:
Steps to reproduce:

1) Delete `.trustify`
2) Run trustify with GC disabled `TRUSTD_GC_FREQ=0s AUTH_DISABLED=true cargo run --bin trustd`
3) Go to http://localhost:8080/advisories/upload page and upload all `/etc/datasets/ds1/csaf/2023` json files
4) Then go to http://localhost:8080/packages and we can see this:

<img width="914" height="135" alt="Image" src="https://github.com/user-attachments/assets/d08cec71-1939-4307-a24b-3db4c880545f" />

5) Stop the server and delete `.trustify`
6) Run trustify with GC 10 seconds frequency `TRUSTD_GC_FREQ=10s AUTH_DISABLED=true cargo run --bin trustd`
3) Go to http://localhost:8080/advisories/upload page and upload all `/etc/datasets/ds1/csaf/2023` json files
4) Then go to http://localhost:8080/packages and we can see this:

<img width="995" height="168" alt="Image" src="https://github.com/user-attachments/assets/619993c8-b791-49fa-bef2-5b3a527202cf" />

helio-frota:
Now that we understand what is going, on this issue title needs to be changed to 

_`gc_purls.sql` was not designed to run from time to time, only when we delete sbom or advisory._

and I think that makes no sense to keep this issue opened as this is basically informative.

helio-frota:
NOTE: The packages deletion is happening on `advisories` only

PR #1980, opened 2025-09-16
Add a test to reproduce #1977

jcrossley3:
Also deleted some superfluous code, i.e. no advisories ingested so none need to be deleted in the now-renamed `gc_purls_from_sbom` test.

### DIFF
--- a/modules/fundamental/src/purl/service/test.rs
+++ b/modules/fundamental/src/purl/service/test.rs
@@ -1,5 +1,4 @@
 use crate::{
-    advisory::service::AdvisoryService,
     purl::{model::details::purl::StatusContext, service::PurlService},
     sbom::service::SbomService,
 };
@@ -702,7 +701,7 @@ async fn contextual_status(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
 
 #[test_context(TrustifyContext)]
 #[test(actix_web::test)]
-async fn gc_purls(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
+async fn gc_purls_from_sbom(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
     let purl_service = PurlService::new();
     assert_eq!(
         0,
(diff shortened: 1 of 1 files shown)


helio-frota:
This change makes the test pass:



but unfortunately we can not merge , we need a better way to say that packages from advisory still "alive", because we have cardinality impact with big databases.

modules/fundamental/src/purl/service/test.rs
```diff
@@ -801,6 +791,37 @@ async fn gc_purls(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
...
+    assert_eq!(
+        6,
+        purl_service
+            .purls(Query::default(), Paginated::default(), &ctx.db)
+            .await?
+            .items
+            .len()
+    );
+
+    // should any packages be garbage collected? I honestly don't know.
```
jcrossley3, line 819: Can you remove this comment, please?

ctron:
> > trustify-scale-testing calls trustify endpoints, I'm not sure if I understood well, for this we would need to create a new endpoint in trustify just to call `PurlService::gc_purls()` ?
> 
> Good idea. Let's add an endpoint, `/api/v2/purl/gc`, under the "purl" tag. I'm sure QE would appreciate that now that the default period is longer. That might even obviate exposing TRUSTD_GC_FREQ in the helm chart.

That might actually be dangerous. As we know it has side effects. This might be good way for a DoS.

helio-frota:
`That might actually be dangerous. As we know it has side effects. This might be good way for a DoS.`

:heavy_plus_sign: 

`Then may end up with an illusion that the performance good or acceptable`

let's revert then?

ctron:
> `That might actually be dangerous. As we know it has side effects. This might be good way for a DoS.`
> 
> ➕
> 
> `Then may end up with an illusion that the performance good or acceptable`
> 
> let's revert then?

Having tests is fine. Exposing the GC on an HTTP endpoint I think is not.

jcrossley3:
I think this is obsolete now

PR #1988, opened 2025-09-23
feat(fundamental): return all scores in addition for sbom advisories

ctron:
## Summary by Sourcery

Consolidate CVSS scoring into a shared Score model and include full score lists in SBOM and advisory analyses

New Features:
- Expose a vector of all advisory scores in AnalysisAdvisory responses
- Include a vector of all CVSS scores in SbomStatus

Enhancements:
- Extract ScoreType and Score definitions into a common model to eliminate duplication
- Refactor vulnerability service and SbomStatus to use the shared Score::try_from conversion
- Remove legacy score schema and conversion code from individual modules

Tests:
- Add a schema validation test for ScoreType enum variants in the common model

### DIFF
--- a/modules/fundamental/src/common/mod.rs
+++ b/modules/fundamental/src/common/mod.rs
@@ -6,6 +6,7 @@ use trustify_entity::sbom_package_license::LicenseCategory;
 use utoipa::ToSchema;
 
 pub mod license_filtering;
+pub mod model;
 pub mod service;
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, ToSchema, FromQueryResult)]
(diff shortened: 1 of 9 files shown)


mrizzi (commented):
@ctron the changes looks good to me. I've opened https://github.com/ctron/trustify/pull/2 to add a test to check for the new `scores` array to speed up this PR completion.

dejanb (approved):
Looks good. A few things:

* Maybe we should mark average values as deprecated? If we are doing UI change along with this one, we can also stop calculating them?

* I opened a separate issue https://github.com/guacsec/trustify-scale-test-runs/issues/17 to include this endpoint (with some reasonable sbom) into scale tests, just to have some visibility in the performance changes (until large sboms are properly fixed)

mrizzi (approved):
@ctron thank you 👍

PR #1996, opened 2025-09-26
feat: adding gc endpoint

helio-frota:
Related to https://github.com/guacsec/trustify/pull/1980#issuecomment-3338643165


modules/fundamental/src/purl/endpoints/mod.rs
```diff
@@ -81,5 +82,21 @@ pub async fn all(
...
+    tag = "purl",
+    responses(
+        (status = 200, description = "Performs garbage collection for orphaned packages", body = String),
+    ),
+)]
+#[get("/v2/purl/gc")]
+pub async fn gc(
+    service: web::Data<PurlService>,
+    db: web::Data<Database>,
+    _: Require<ReadSbom>,
```
jcrossley3, line 96: To stay consistent with how it previously worked, this should probably require delete privs.
```suggestion
    _: Require<DeleteSbom>,
```

jcrossley3 (changes requested):
Looks good. Just one minor nit.

modules/fundamental/src/purl/endpoints/mod.rs
helio-frota: thanks, pr updated :+1:

modules/fundamental/src/purl/endpoints/test.rs
```diff
@@ -310,3 +311,22 @@ async fn test_purl_license_details(ctx: &TrustifyContext) -> Result<(), anyhow::
...
+
+    let uri = "/api/v2/purl/gc";
+    let request = TestRequest::get().uri(uri).to_request();
+    let response = app.call_service(request).await;
+
+    assert_eq!(response.status(), StatusCode::OK);
+
+    let body_bytes = actix_web::test::read_body(response).await;
+    let body_str = std::str::from_utf8(&body_bytes)?;
+    assert_eq!(body_str, "8");
```
jcrossley3, line 329: This is currently affirming the bug that #1980 fixes, correct? Because I don't see anything being deleted that would generate any "garbage" to be collected.

I suppose we can merge this, and then rebase #1980 to change this to 0?

helio-frota: https://github.com/guacsec/trustify/pull/1980 already affirms the fix :+1: 

> Because I don't see anything being deleted that would generate any "garbage" to be collected.

Yeah my focus was on the endpoint-call-only as the 1980 already is the actual test.

But you right, that is confusing. I'm reusing the `setup` function which ingests only packages with no parent sbom or advisory, so that they are orphans by default.

> I suppose we can merge this, and then rebase https://github.com/guacsec/trustify/pull/1980 to change this to 0?

Not exactly because we are not ingesting sbom or advisory in this test, just calling the endpoint :+1:

modules/fundamental/src/purl/endpoints/mod.rs
```diff
@@ -81,5 +82,21 @@ pub async fn all(
...
 
+#[utoipa::path(
+    operation_id = "garbageCollect",
+    tag = "purl",
+    responses(
+        (status = 200, description = "Performs garbage collection for orphaned packages", body = String),
+    ),
+)]
+#[get("/v2/purl/gc")]
+pub async fn gc(
```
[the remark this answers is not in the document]
helio-frota: PR updated :+1:

modules/fundamental/src/purl/endpoints/test.rs
jcrossley3: Ah, I was wrongly assuming they weren't orphans. Since they are, the test makes sense.

jcrossley3 (approved):

PR #2002, opened 2025-09-29
feat(api): recommendations api

Strum355:
## Summary by Sourcery

Introduce a new recommendations API endpoint for PURLs to suggest updated package versions and related vulnerability remediations.

New Features:
- Add POST /api/v2/purl/recommend endpoint with RecommendRequest and RecommendResponse schemas.

Enhancements:
- Implement PurlService::recommend_purls to filter by regex, compare semver, and assemble recommendation entries.
- Define RecommendEntry and VulnerabilityStatus models plus VexStatus and VexJustification enums.
- Extend OpenAPI spec and Actix routing with Utoipa annotations for the new endpoint.

Build:
- Add packageurl, semver, regex, and lenient_semver dependencies for version parsing and pattern matching.

Tests:
- Add integration test for the recommendations endpoint using sample CVE fixtures.


modules/fundamental/src/vulnerability/endpoints/mod.rs
```diff
@@ -121,3 +132,104 @@ pub async fn analyze(
...
+        parameters.push(("ty", purl.ty()));
+        parameters.push(("name", purl.name()));
+
+        query.push_str(
+            &parameters
+                .iter()
+                .map(|pair| format!("{}={}", pair.0, pair.1))
+                .collect::<Vec<_>>()
+                .join("&"),
+        );
```
Strum355, line 175: Would be nice if we could pass properly-typed structs [similarly to the guac client originally used in trustification v1](https://github.com/trustification/trustification/blob/1d65f2c1cce303a89f1f9ca1e8cd6285f1d23de0/exhort/api/src/server.rs#L178-L189) instead of having to construct the query string manually. The [example query in the ADR](https://github.com/guacsec/trustify/blob/main/docs/adrs/00008-purls-recommendation.md#decision) isnt correct, so I had to do some digging to see what the actual query format is 😅

Strum355:
@dejanb unless Im forgetting something, the missing items are:
1. user-provided pattern for filtering to trusted purls (currently still using the v1 redhat pattern)
2. `justification` field present in v1 currently isnt filled in, I couldnt see that we process the `vulnerabilities.flags` section in csaf vex files

modules/fundamental/src/vulnerability/endpoints/mod.rs
```diff
@@ -121,3 +133,143 @@ pub async fn analyze(
...
+    tag = "vulnerability",
+    request_body = AnalysisRequest,
+    responses(
+        (status = 200, description = "Get recommendations and remediations for provided purls", body = RecommendResponse)
+    )
+)]
+#[post("/v2/vulnerability/recommend")]
+pub async fn recommend(
+    purl_service: web::Data<PurlService>,
+    db: web::Data<Database>,
```
jcrossley3, line 148: This is a sign that this endpoint belongs not in the vulnerability module but in the purl module. There is generally a 1-to-1 relationship between endpoints and their services.

I have some specific feedback relating to the implementation of this function, but before we discuss that, I think we should put these functions in the correct modules first. Notice how the `analyze` function above this one delegates most of its behavior to `service.analyze_purls`. In general, the endpoint functions should be relatively small, and the "meat" should live in the corresponding service functions. So you should have a `recommend` endpoint and a `recommend_purls` service function analogous to the `analyze` endpoint, but since you're relying on a `PurlService` and not a `VulnerabilityService`, I believe your two new functions belong beneath `src/purl/endpoints` and `src/purl/service` instead of in here. Does that make sense?

Once you have them in the right place, then I can review the logic of the service function itself.

Strum355: > In general, the endpoint functions should be relatively small, and the "meat" should live in the corresponding service functions.

Thanks for bringing this up as clearly as you did, had a suspicion this might be the case so thats good to have confirmed :+1: I've moved everything over to the purl module, Ill bump the test coverage again based on the codecov MR failing while I await your review the function logic. Thanks! @jcrossley3

modules/fundamental/src/vulnerability/endpoints/mod.rs
jcrossley3: > When run through `Query.parse` ([implementation here](https://github.com/guacsec/trustify/blob/143a5ae/common/src/db/query.rs#L85-L122)), it sets the `field` attribute of each Constraint to e.g. `purl:namespace`/`purl:version` etc which wasnt matching the column names (and giving me 0 results)

That is the correct syntax. The colon delimits the name of the json column ("purl") in the table and the key of the object ("namespace"). When I change your (now obsolete) code to the following, your test still passes:
```rust
        if let Some(namespace) = purl.namespace() {
            parameters.push(("purl:namespace", namespace));
        }

        parameters.push(("purl:ty", purl.ty()));
        parameters.push(("purl:name", purl.name()));
```
There are (perhaps too) many ways a user might specify the fields of a query for purls. But I see you're using `purl~...` now, so that should be fine.

Strum355: >  When I change your (now obsolete) code to the following, your test still passes:

maybe I had something else wrong at the time :smile:

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -339,6 +344,100 @@ impl PurlService {
...
+                    Err(e) => {
+                        return Err(Error::Internal(format!(
+                            "{} in similar purl {} is not valid semver: {}",
+                            version, summary.head.purl, e
+                        )));
+                    }
+                };
+                // From the original version in trustification v1:
+                // remove the RHT stupid renaming because semver thinks it means pre-release
+                // and that breaks stupid comparisions.
```
jcrossley3, line 407: This confused me at first. I might include a link instead...

```suggestion
                // See https://github.com/trustification/trustification/blob/1d65f2c1cce303a89f1f9ca1e8cd6285f1d23de0/exhort/api/src/server.rs#L204-L205
```

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -339,6 +344,100 @@ impl PurlService {
...
+                similar_version.pre = Prerelease::EMPTY;
+
+                if similar_version >= input_version {
```
jcrossley3, line 410: This is a bit of a style nit, but just to be consistent with your other "continues", I might do 

```suggestion
            if similar_version < input_version {
                continue;
            }
```

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -339,6 +344,100 @@ impl PurlService {
...
+                            .flat_map(|advisory| {
+                                advisory.status.iter().map(|status| VulnerabilityStatus {
+                                    id: status.vulnerability.identifier.clone(),
+                                    status: Some(status.into()),
+                                    justification: None,
+                                })
+                            })
+                            .collect(),
+                    });
+                }
```
jcrossley3, line 435: I think it makes more sense to follow the v1 logic here, mostly to account for the possibility of a user passing the same purl more than once in the input list. I think that will result in duplicate entries being returned. Try putting the same purl in your test twice to see what I mean. I would recommend removing the `recommendations_for_purl` logic entirely and just add the vec to the result map if it's non-empty.

```suggestion
            if let Some(purl_details) = purl_service
                .versioned_purl_by_uuid(&summary.head.purl.version_uuid(), db.as_ref())
                .await?
            {
                recommended_purls.push(RecommendEntry {
                    package: summary.head.purl.to_string(),
                    vulnerabilities: purl_details
                        .advisories
                        .iter()
                        .flat_map(|advisory| {
                            advisory.status.iter().map(|status| VulnerabilityStatus {
                                id: status.vulnerability.identifier.clone(),
                                status: Some(status.into()),
                                justification: None,
                            })
                        })
                        .collect(),
                });
            }
        }
        if !recommended_purls.is_empty() {
            recommendations.insert(purl.to_string(), recommended_purls);
        }
 ```

**Review by jcrossley3 — CHANGES_REQUESTED** [review]
Great work! Thanks for indulging me 😄

File: modules/fundamental/src/purl/service/mod.rs [hunk 6]
```diff
@@ -339,6 +344,100 @@ impl PurlService {
...
+            };
+            let input_version = match lenient_semver::parse(input_version_str) {
+                Ok(v) => v,
+                Err(e) => {
+                    return Err(Error::BadRequest(format!(
+                        "version {} in purl {} is not valid semver: {}",
+                        input_version_str, purl, e
+                    )));
+                }
+            };
```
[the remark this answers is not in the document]
dejanb: If I'm reading the v1 logic correctly, for these kind of purls no recommendation is returned

https://github.com/trustification/trustification/blob/1d65f2c1cce303a89f1f9ca1e8cd6285f1d23de0/exhort/api/src/server.rs#L221

So, let's do that for now?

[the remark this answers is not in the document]
dejanb: Sounds good to me ... On that note, I think we should stick with the original pattern matching for now and develop all further changes in following prs and tests with dependency analytics.

jcrossley3 (approved):
Looks good. All my comments are nits, so do with them what you will.

PR #2011, opened 2025-10-03
feat: license filter consistent for SBOM packages tab

mrizzi:
Add SBOM package license filtering with expanded SPDX LicenseRef resolution to achieve consistency with license filtering approach used in SBOM and Packages lists.

The SBOM package license filter BEFORE this change looks like this when filtering licenses for `gpl+`

Screenshot 2025-10-03 at 10 18 09

The SBOM package license filter WITH this change looks like this when filtering licenses for `gpl+`

Screenshot 2025-10-03 at 10 17 53

Changes
  - Enhanced license filtering and  license expansion for SBOM packages:
    - Enables consistent license management across SBOM and package list endpoints, pretty clear from changes to current test
  - Deprecated licenses_ref_mapping field across all license endpoints because not useful anymore:
    - Now returns always an empty array
    - Updated OpenAPI specification to reflect deprecation
    - Updated test expectations to expect empty licenses_ref_mapping
  - Unified using shared PLSQL `case_license_text_sbom_id` function adding reusable get_case_license_text_sbom_id() helper function
  - Replaced raw SQL UNION queries in license service with SeaORM-based implementation 
  - Updated test data to reflect expanded license expressions:
    - Increased recursion limit to 256 to parse entensive JSON expected results in test `fetch_unique_licenses`

### DIFF
--- a/modules/fundamental/src/common/license_filtering.rs
+++ b/modules/fundamental/src/common/license_filtering.rs
@@ -7,7 +7,7 @@ use sea_query::{
     extension::postgres::PgExpr,
 };
 use trustify_common::db::{
-    ExpandLicenseExpression,
+    CaseLicenseTextSbomId, ExpandLicenseExpression,
     query::{Columns, Filtering, IntoColumns, Query, q},
 };
 use trustify_entity::{license, sbom_package, sbom_package_license, sbom_package_purl_ref};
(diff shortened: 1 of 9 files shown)


PhilipCattanach:
@jcrossley3 - Jim this is the last piece of the puzzle regarding the License Search changes. We had to make some compromises with the as we didn't want to change the data model in this release and introduce data migration challenges. Instead Marco did the best he could with the existing, flawed, data model (it should be SBOM specification agnostic, but isn't). So with that said, is there anything preventing this PR from getting merged?

jcrossley3:
> @jcrossley3 - Jim this is the last piece of the puzzle regarding the License Search changes. We had to make some compromises with the as we didn't want to change the data model in this release and introduce data migration challenges. Instead Marco did the best he could with the existing, flawed, data model (it should be SBOM specification agnostic, but isn't). So with that said, is there anything preventing this PR from getting merged?

Not to my knowledge. Merge away!

PR #2042, opened 2025-10-16
perf: license filtering with CTEs (TC-3006)

mrizzi:
Some considerations based upon evaluation against the largest testing dataset, i.e. the scale test one.

## Optimization at high level

- CTE 1: Prepare license mappings
- CTE 2: Deduplicate input data
- CTE 3: Perform expensive operation on minimum set
- Final: Filter and return results

Each CTE ([Common Table Expressions](https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH)) has a single, well-defined purpose.

### Reduction in Expensive Operations
  - Before: 766,498 total rows → 255,000+ function calls to EXPAND_LICENSE_EXPRESSION
  - After: Deduplicate to 31,459 unique combinations → only 31,459 function calls
  - Result: 24x reduction in data volume, 8x reduction in function calls
  - Impact: ~4x performance improvement (25.5s → 6s)

### Exploit Data Invariants
  - Key insight: Same license text in same SBOM always expands identically
  - CTE benefit: DISTINCT in CTE enforces this invariant at SQL level
  - Avoids: Redundant computation of identical expansions

 ### Single Query Execution vs N+1 Pattern
  - Without CTE: Query → fetch → loop → call function N times → aggregate
  - With CTE: Single query executes everything in database engine
  - Benefit: Eliminates round-trips between application and database

### Database-Level Optimization
  - PostgreSQL can optimize each CTE independently
  - Query planner chooses optimal join order and indexes
  - Intermediate results can be materialized or inlined based on cost
  - Better use of database caching and buffer pools

## Further changes
- new way for expressing custom functions for SQL in `common/src/db/func.rs` (follow-up PR to migrate the existing ones once agreed upon applying it)
- added some more tests

Fixes [TC-3006](https://issues.redhat.com/browse/TC-3006)

## Summary by Sourcery

Streamline SPDX license filtering by moving deduplication and expression expansion into three database CTEs and a new PL/pgSQL function, updating service layers to use a single-query approach for significant performance gains, and adding the necessary migration and endpoint tests

New Features:
- Introduce CTE-based license filtering pipeline in license, SBOM, and PURL services to deduplicate and expand SPDX expressions entirely in SQL
- Add a new PL/pgSQL function `expand_license_expression_with_mappings` with a composite `license_mapping` type via migration m0001180
- Expose the new expand function in Rust through a `CustomFunc::ExpandLicenseExpressionWithMappings` enum variant

Enhancements:
- Refactor service queries to use `build_license_filtering_with_clause`, reducing redundant expansion calls and eliminating N+1 patterns
- Leverage distinct CTEs for mapping aggregation, unique license/SBOM pairs, and expression expansion, yielding ~4× performance improvement
- Integrate full-text search and sorting translations into the new CTE-based filtering logic

Build:
- Add migration m0001180 to create `license_mapping` composite type and optimized expansion function

Tests:
- Add endpoint tests to verify license search filtering and sorting order for the updated license API

### DIFF
--- a/common/src/db/func.rs
+++ b/common/src/db/func.rs
@@ -98,3 +98,9 @@ impl Iden for CaseLicenseTextSbomId {
         write!(s, "case_license_text_sbom_id").unwrap()
     }
 }
+
+#[derive(Iden)]
+pub enum CustomFunc {
+    #[iden = "expand_license_expression_with_mappings"]
+    ExpandLicenseExpressionWithMappings,
+}
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -29,6 +29,7 @@ mod m0001140_expand_spdx_licenses_function;
 mod m0001150_case_license_text_sbom_id_function;
 mod m0001160_improve_expand_spdx_licenses_function;
 mod m0001170_non_null_source_document_id;
+mod m0001180_expand_spdx_licenses_with_mappings_function;
 
 pub struct Migrator;
 
@@ -65,6 +66,7 @@ impl MigratorTrait for Migrator {
             Box::new(m0001150_case_license_text_sbom_id_function::Migration),
             Box::new(m0001160_improve_expand_spdx_licenses_function::Migration),
             Box::new(m0001170_non_null_source_document_id::Migration),
+            Box::new(m0001180_expand_spdx_licenses_with_mappings_function::Migration),
         ]
     }
 }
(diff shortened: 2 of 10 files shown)


mrizzi:
As agreed, added scale tests to cover the changed endpoints in

- https://github.com/guacsec/trustify-scale-testing/pull/70

dejanb (approved):
Looks good. I like the idea of using CTEs for making queries more performant (but also more readable). We use it in correlation logic a bit, but we should investigate if it could help more.

In general I think using raw sql for these complex queries is OK (to me even makes it more readable), but you already did the hard part.

PR #2049, opened 2025-10-21
[TC-3060] gensbom: Support TPA with certs

i386x:
~~Add `-k` flag to `curl` in case TPA is configured with SSL certificates.~~
Detect provided trust anchors and ~~install it on the container~~ pass them to `curl`.

## Summary by Sourcery

Add the insecure (-k) flag to curl invocations in the gensbom.sh script to allow interactions with TPA services using custom or self-signed SSL certificates

Enhancements:
- Add the -k option to curl download check to skip SSL verification
- Add the -k option to curl POST operations to skip SSL verification

### DIFF
--- a/etc/gensbom/gensbom.sh
+++ b/etc/gensbom/gensbom.sh
@@ -71,6 +71,11 @@ function usage() {
 	    A file with the valid container registry credentials following the
 	    Docker format. This file must be in the current working directory
 
+	  ${_VIOLET}trust.crt${_RESET}
+	    Optional custom trust anchors, needed to be installed on the container
+	    to make the TPA instance accessible from it, in case the TPA instance
+	    uses a certificate signed by these trust anchors
+
 	  ${_VIOLET}TPA_AUTH_TOKEN${_RESET}
 	    Authorization token for TPA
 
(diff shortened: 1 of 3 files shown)


ctron (changes requested):
Please never just use `-k`. It turns off all TLS validation. This is dangerous!

Instead, allow the user to provide a CA cert bundle and keep TLS active.

Also see: https://curl.se/docs/sslcerts.html

etc/gensbom/gensbom.sh
```diff
@@ -71,6 +71,13 @@ function usage() {
 	    A file with the valid container registry credentials following the
 	    Docker format. This file must be in the current working directory
 
+	  ${_VIOLET}ca-bundle.crt${_RESET}
+	    SSL/TLS certificates bundle. If the TPA instance uses SSL/TLS, this
```
ctron: This is only required if trustify uses a certificate signed by a custom trust anchor. Not in general when using TLS.

ctron:
What's wrong with having support for `--cacert`?

ctron:
I think if the user provides a wrong/mismatching certificate, it's ok to get an error.

ctron (approved):

PR #2055, opened 2025-10-22
test: implement test using a previous DB export

ctron:
## Summary by Sourcery

Provide end-to-end migration testing by downloading previous commit database and storage dumps, importing them into an embedded database, and running migrations. Restructure the test context architecture, extend the embedded DB helper for dump imports, and update CI and dependencies to support these capabilities.

New Features:
- Introduce TrustifyMigrationContext that downloads and imports a previous database and storage dump for migration tests
- Add Migration manager to fetch, verify, and provide branch-specific migration dumps from S3

Enhancements:
- Restructure test-context crate by renaming TrustifyContext to TrustifyTestContext and splitting contexts into default, migration, and read-only modules
- Extend common/db embedded creation to support importing from SQL/XZ dumps via new Source and Options enums and an AsyncRead-based import pipeline
- Update existing tests to borrow the database reference (`&ctx.db`) instead of consuming it

Build:
- Add new dependencies (git2, reqwest, sha2, tar, tempfile, async-compression, postgresql_commands, tokio full features, fs4) across Cargo.toml files to support migration and import functionality

CI:
- Set TRUSTIFY_S3_AWS_REGION and TRUSTIFY_S3_AWS_BUCKET environment variables in CI workflow for migration dumps

Tests:
- Add a new migration test (`migration/tests/previous.rs`) to verify importing and migrating a dump from the previous commit

### DIFF
--- a/.github/workflows/migration-upload.yaml
+++ b/.github/workflows/migration-upload.yaml
@@ -29,13 +29,18 @@ jobs:
           # we don't need to compress the storage tar file, as individual files are already compressed
 
       - name: Upload to S3
-        if: ${{ secrets.TRUSTIFY_S3_UPLOAD_ACCESS_KEY_ID }}
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.TRUSTIFY_S3_UPLOAD_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.TRUSTIFY_S3_UPLOAD_SECRET_ACCESS_KEY }}
           AWS_REGION: eu-west-1
           AWS_BUCKET: guacsec-migration-dumps
         run: |
+          # we can't check in an `if` condition, as secrets are not available there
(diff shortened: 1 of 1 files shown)


common/db/src/lib.rs
```diff
@@ -60,8 +63,58 @@ impl<'a> Database<'a> {
...
+
+            Ok::<_, anyhow::Error>(())
+        })
+        .await??;
+
+        let _ = child.wait_with_output().await?;
+
+        // some maintenance after import
+
+        db.execute_unprepared(r#"VACUUM FULL ANALYZE"#).await?;
```
dejanb: Should we do 

`REINDEX database trustify`

here as well?

dejanb (approved):
Looks good. Two minor questions inlined. Also, a bit of documentation on how the whole process works would help other folks maintain it in the future.

common/db/src/lib.rs
ctron: Done

PR #2064, opened 2025-10-24
fix: make GET endpoint working with repeatable read transaction (TC-3085)

mrizzi:
While working on adding the scale test for SBOM delete endpoint (ref. https://github.com/guacsec/trustify/pull/2052#issuecomment-3431423132), I ran locally the scale tests and it turned out that if the delete SBOM endpoint is invoked, the `list_sboms` scale test (which just does a GET call to `/api/v2/sbom` endpoint) starts to report `404` errors and some `200` response with empty body.
The screenshot shows that running 83 successful deletes caused 26 `404` errors on the `list_sboms` test
Screenshot 2025-10-24 at 14 39 12

and an extract of the log that shows how the `GET /api/v2/sbom` call returns a `404` and then `200` again  
```bash
2025-10-23T13:15:09.249692Z  WARN ThreadId(24) 01973121-ffba-7432-86fa-5ec42dc4749a describes 1 packages
2025-10-23T13:15:09.250751Z  WARN ThreadId(24) 01973121-ffba-7432-86fa-5ec42dc4749a has source document
2025-10-23T13:15:13.022515Z  INFO ThreadId(24) "DELETE /api/v2/sbom/urn:uuid:01973121-ffba-7432-86fa-5ec42dc4749a HTTP/1.1" 200 1302 "-" "goose/0.18.0" 0.105058
2025-10-23T13:15:13.120979Z  WARN ThreadId(19) 01973121-ffba-7432-86fa-5ec42dc4749a describes 0 packages
2025-10-23T13:15:13.122135Z  INFO ThreadId(19) error=Not found: Missing source document sbom=Model {sbom_id: 01973121-ffba-7432-86fa-5ec42dc4749a
2025-10-23T13:15:13.122369Z  INFO ThreadId(19) "GET /api/v2/sbom HTTP/1.1" 404 76 "-" "goose/0.18.0" 0.650402
2025-10-23T13:15:14.075891Z  INFO ThreadId(24) "GET /api/v2/sbom HTTP/1.1" 200 16487 "-" "goose/0.18.0" 0.991112
```

Investigating this issue it turned out the `404` is generated by reading the related `source_document` in https://github.com/guacsec/trustify/blob/5bd7c2cabc17b50bfd84a3e79d047b6e0f3856e5/modules/fundamental/src/sbom/model/mod.rs#L96-L100

Hence it was clear that the **all** the GET endpoints collecting data from multiple tables using the default `READ COMMITTED` database connection would be exposed to such read inconsistent data due to concurrent writes during request processing.

The solution is based upon using a `REPEATABLE READ` transaction in all the GET endpoints to prevent phantom reads and non-repeatable reads.
PostgreSQL has _Multiversion Concurrency Control_ (MVCC) that makes this approach affordable because no locks are acquired (MVCC handles it), i.e. "_The main advantage of using the MVCC model of concurrency control rather than locking is that in MVCC locks acquired for querying (reading) data do not conflict with locks acquired for writing data, and so reading never blocks writing and writing never blocks reading_" from https://www.postgresql.org/docs/17/mvcc-intro.html

With this changes applied, the scale test execution reports no further errors for the `list_sboms` test
Screenshot 2025-10-24 at 14 40 22

I've also checked that transactions are properly dropped after the end of the sale test execution monitoring the DB sessions for them to come back idle just like reported in the next dashboard
Screenshot 2025-10-24 at 14 42 17

## Summary by Sourcery

Use PostgreSQL’s MVCC by running all GET endpoints in REPEATABLE READ read-only transactions and refactoring data fetching to bulk prefetch, eliminating phantom and non-repeatable reads and preventing inconsistent GET errors during concurrent SBOM deletes.

New Features:
- Add DatabaseExt trait with begin_read method to start REPEATABLE READ read-only transactions.

Bug Fixes:
- Fix intermittent 404s and empty GET /api/v2/sbom responses under concurrent deletes by using repeatable-read isolation.

Enhancements:
- Wrap all GET endpoints in begin_read transactions and update service/model layers to use bulk prefetch (.all/query_all) instead of streaming for consistent snapshots.
- Refactor SBOM details and vulnerability analysis services to collect IDs, pre-fetch related entities in bulk, and build in-memory maps to work seamlessly within repeatable-read transactions.

Build:
- Add tracing-log dependency and use EnvFilter for global tracing setup.

Tests:
- Adjust GraphQL test indexing to align with updated result ordering.

### DIFF
--- a/modules/fundamental/src/sbom/model/details.rs
+++ b/modules/fundamental/src/sbom/model/details.rs
@@ -492,8 +492,7 @@ impl SbomAdvisory {
                         .get(&(each.advisory.id, each.vulnerability.id.clone()))
                         .cloned()
                         .unwrap_or_default(),
-                )
-                .await?;
+                )?;
                 advisory.status.push(status);
                 if let Some(status) = advisory.status.last_mut() {
                     status
@@ -558,7 +557,7 @@ impl SbomStatus {
         ),
         err(level=tracing::Level::INFO)
     )]
-    pub async fn new(
+    pub fn new(
         advisory_vulnerability: &advisory_vulnerability::Model,
         vulnerability: &vulnerability::Model,
         status: String,


ctron:
I am not a fan of passing in two database connections as the same time.

I wonder if we can't have the `DatabaseExt` as a blanket implementation on `ConnectionTrait`?

mrizzi:
> I am not a fan of passing in two database connections as the same time.
> 
> I wonder if we can't have the `DatabaseExt` as a blanket implementation on `ConnectionTrait`?

What if, for example, I would add the `db` field to the `VulnerabilityService` struct just like we already do in `SbomService`?
https://github.com/guacsec/trustify/blob/98727dd48649bfd80a97bccc6b72b276ef63a6d3/modules/fundamental/src/sbom/service/mod.rs#L10-L12
In this way the `Database` will be provided by the `VulnerabilityService`, removing the need to have the `db: &trustify_common::db::Database` param in `analyze_purls`.

mrizzi:
We've been discussing this in the Trustify community Matrix channel, i.e. https://matrix.to/#/!heUmIqxVhZejiSjzVu:matrix.org/$FYsKPlKVM_OhmzRDB10cZheIqqsdj5SSIZYEE7ETi3M?via=matrix.org&via=dentrassi.de and tried to get rid of streaming in order to just use a single transaction in each GET endpoint.
Commit e7f7da3 reflects this effort.

It has been successfully tested running https://github.com/guacsec/trustify/blob/1681c18a4f3a4cd7a908bd32e46383886587d2bb/modules/fundamental/tests/issues/oom.rs against the (currently used in scale tests) DB dump (requires hash to be `720e4451b0c13f028e55e48b9d522fee8a20124b41379c99a9396562471644471`).

ctron:
I used the following SBOM: `sha256:a08f4d8723d3f2e1e12ba4a8961c6ebccfd603577d784b24576c09be8925af40`

ctron (approved):
As to my understanding from the call we just had, there have been additional optimizations, bringing the memory usage down to the baseline on main.

PR #2069, opened 2025-10-29
test: fix up oom test

ctron:
First we need to ensure `teardown` is being run. This can be done by dropping the `skip_teardown` attribute.

Second, we need to refresh the ID. Adding the SQL statement used for future reference.

Third, add some more explanation about the expectation towards the test.

## Summary by Sourcery

Refine the OOM test by restoring teardown execution, migrating to test_log's test attribute, updating the SBOM ID, and enhancing documentation with an SQL snippet and usage explanation.

Bug Fixes:
- Restore test teardown by removing the skip_teardown attribute

Enhancements:
- Switch from tokio::test to test_log::test for better logging integration
- Remove obsolete global subscriber setup
- Change the test signature to accept a reference to TrustifyContext

Tests:
- Refresh the SBOM ID used in the test
- Embed the SQL query used to derive the SBOM ID for future reference
- Add a doc comment explaining memory usage expectations under --test-threads=1

### DIFF
--- a/etc/deploy/compose/compose.yaml
+++ b/etc/deploy/compose/compose.yaml
@@ -10,3 +10,5 @@ services:
     shm_size: '1g'
     command: >
       postgres -c random_page_cost=1.1 -c max_parallel_workers_per_gather=4 -c shared_preload_libraries='pg_stat_statements'
+#    volumes:
+#      - <base>/dump.sql.gz:/docker-entrypoint-initdb.d/dump.sql.gz:Z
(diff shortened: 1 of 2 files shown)


etc/deploy/compose/compose.yaml
```diff
@@ -10,3 +10,5 @@ services:
     shm_size: '1g'
     command: >
       postgres -c random_page_cost=1.1 -c max_parallel_workers_per_gather=4 -c shared_preload_libraries='pg_stat_statements'
+#    volumes:
```
mrizzi: what is this for?

ctron: For starting the container with a database dump.

mrizzi: And is it fine to keep it commented or should it be executed?

mrizzi (approved):

PR #2070, opened 2025-10-29
chore: rename adr another adr with same number was added

dejanb:
## Summary by Sourcery

Rename ADR and mark it as approved

Documentation:
- Rename ADR file and internal heading from 00009 to 00010 to avoid duplicate numbering
- Update ADR status from PROPOSED to APPROVED


dejanb:
Let's use this PR to test out scale test changes https://github.com/guacsec/trustify-scale-testing/pull/72

PR #2076, opened 2025-10-31
perf: fetch_sbom_details optimize CPE-based query (TC-3139)

mrizzi:
### Optimize CPE-based vulnerability query with CTEs and index

Refactor the `product_advisory_info_sql()` function to use Common Table Expressions (CTEs) and split OR conditions into UNION queries for performance improvement.

The original query had a performance issue caused by an OR condition in the JOIN clause:

```sql
  FROM sbom_package_purl_ref spr
  JOIN qualified_purl qp ON spr.qualified_purl_id = qp.id
  JOIN versioned_purl vp ON qp.versioned_purl_id = vp.id
  JOIN base_purl bp ON vp.base_purl_id = bp.id
  JOIN product_status ps ON ps.package = bp.name
                         OR ps.package = CONCAT(bp.namespace, '/', bp.name)
```
The OR condition in a JOIN generates a big Cartesian product intermediate result set whose cardinality slows down the performance.

The change here is splitting the OR into UNION, i.e. creating two separate queries and combines them with UNION.
Using CTEs also allowed for some pre-filtering, e.g. in `sbom_purls` the packages are pre-filtered to be only the ones with the SBOM.

Also added the index `product_status_package_idx` (on product_status.package) to support some join conditions, e.g. `ps.package = sp.name`.

The final result is the query's execution time move from ~3 minutes to ~0.5 sec.

Relates to https://issues.redhat.com/browse/TC-3139

## Summary by Sourcery

Optimize the CPE-based vulnerability query by refactoring it with Common Table Expressions and splitting OR conditions into UNIONs, and support the changes by adding a new index to speed up package lookups

Enhancements:
- Refactor product_advisory_info_sql to use CTEs for pre-filtering and union-based joins to eliminate OR conditions
- Split OR join into two match queries combined via UNION to leverage indexes and reduce Cartesian products
- Add sbom_purls CTE to pre-filter packages by SBOM and avoid repeated scans

Build:
- Add migration to create and drop product_status_package_idx on product_status.package

### DIFF
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -30,6 +30,7 @@ mod m0001150_case_license_text_sbom_id_function;
 mod m0001160_improve_expand_spdx_licenses_function;
 mod m0001170_non_null_source_document_id;
 mod m0001180_expand_spdx_licenses_with_mappings_function;
+mod m0001190_optimize_product_advisory_query;
 
 pub struct Migrator;
 
@@ -67,6 +68,7 @@ impl MigratorTrait for Migrator {
             Box::new(m0001160_improve_expand_spdx_licenses_function::Migration),
             Box::new(m0001170_non_null_source_document_id::Migration),
             Box::new(m0001180_expand_spdx_licenses_with_mappings_function::Migration),
+            Box::new(m0001190_optimize_product_advisory_query::Migration),
         ]
     }
 }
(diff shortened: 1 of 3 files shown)


dejanb:
Yeah, are we able to uncomment them now?

mrizzi:
> Yeah, are we able to uncomment them now?

Done in https://github.com/guacsec/trustify-scale-test-runs/pull/28

PR #2077, opened 2025-11-03
fix: Exclude labels with key 'file' from fetch_labels query

helio-frota:
Signed-off-by: Helio Frota <00hf11@gmail.com>

Issue described by @mrrajan:

```
Advisories added with default "file" label, making label recommendation
and filtering drop downs less effective.

For each Advisories, the label "file" added with unique value which
listed out under label recommendation and filtering condition.
```

## Summary by Sourcery

Filter out labels with the key 'file' in the fetch_labels SQL query and update its documentation accordingly

Bug Fixes:
- Exclude labels with key 'file' from the fetch_labels query results

Enhancements:
- Update the fetch_labels function documentation to mention exclusion of 'file' labels


helio-frota:
No problem :+1: . I was thinking on making `filter_text` using negative form but then I noticed that it also contains incremental search, which will invalidate something I was thinking about (to add a default value of `file` in UI and negate the `ILIKE`)...

PR #2079, opened 2025-11-04
feat: add version range to purl status

Strum355:
PR for https://github.com/guacsec/trustify/pull/2056

## Summary by Sourcery

Add support for version ranges in PurlStatus by introducing a VersionRange type, updating the vulnerability analysis flow and API schema, and adjusting tests to validate the new field.

New Features:
- Include version_range in PurlStatus and return it in analysis responses
- Introduce VersionRange enum and schema to represent package version bounds

Enhancements:
- Refactor vulnerability analysis to replace status map with a flat list of PurlStatus entries
- Simplify CVSS3 score handling by using Cvss3Base and streamline advisories aggregation

Documentation:
- Extend OpenAPI schema with PurlStatus changes and VersionRange definition

Tests:
- Update tests to verify version_range and new purl_statuses structure

## Summary by Sourcery

Add support for version ranges in PurlStatus and refactor vulnerability analysis to return a flat list of statuses with version_range, updating models, API schema, and tests accordingly.

New Features:
- Include version_range in PurlStatus and return it in analysis responses
- Introduce VersionRange enum and schema to represent package version bounds

Enhancements:
- Refactor vulnerability analysis to produce a flat list of PurlStatus entries instead of a status-to-advisory map
- Simplify CVSS3 score processing by consolidating on Cvss3Base and using Score.from_iter

Documentation:
- Extend OpenAPI schema to replace status map with purl_statuses array and define VersionRange

Tests:
- Update tests to validate version_range field and new purl_statuses structure in vulnerability analysis and advisory reingestion

### DIFF
--- a/modules/fundamental/src/purl/model/details/version_range.rs
+++ b/modules/fundamental/src/purl/model/details/version_range.rs
@@ -24,6 +24,7 @@ pub enum VersionRange {
         high_version: String,
         high_inclusive: bool,
     },
+    Unbounded,
 }
 
 impl VersionRange {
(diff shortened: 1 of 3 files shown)


dejanb:
Looking at the database, the issue looks like that `low_version` is `null` and `low_inclusive` is `false`. That breaks the original logic. You can ingest the dataset with this data like https://docs.guac.sh/trustify/getting-started#5-ingest-a-dataset
Anyhow, we need to add more tests for the possible combinations. I would assume that the value is `false` if not present for `inclusive` values.

PR #2092, opened 2025-11-06
fix: remove check for previously ingested SBOM from gensbom script

jcrossley3:
downstream issue: https://issues.redhat.com/browse/TC-3160

There's no reason to do this. SBOM ingestion is designed to be idempotent.

## Summary by Sourcery

Remove the redundant SBOM existence check in the gensbom script to rely on the idempotent ingestion process

Enhancements:
- Remove SHA512-based GET request and related logging in gensbom.sh
- Eliminate skip logic for already ingested SBOM files


i386x:
LGTM

PR #2100, opened 2025-11-12
fix: ensure content is fully written

ctron:
## Summary by Sourcery

Ensure output streams are properly shut down after writing to flush all buffered data.

Bug Fixes:
- Add shutdown calls after io::copy in Compression::write for both None and Zstd variants to guarantee full content is written
- Invoke shutdown on the destination writer in download_artifacts to finalize the write process

Enhancements:
- Introduce a copy_shutdown helper in the compression module to unify io::copy and shutdown logic

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2055,10 +2055,11 @@ dependencies = [
 [[package]]
 name = "csaf"
 version = "0.5.0"
-source = "git+https://github.com/trustification/csaf-rs#4d65f7c791e16b7f80e0a3945bde9b91820f8f5b"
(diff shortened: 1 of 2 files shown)


dejanb:
@ctron I only see Cargo.lock, but not Cargo.toml committed, is this on purpose?

dejanb (approved):
Looks good

I've looked at a few more places where this can be relevant.

https://github.com/guacsec/trustify/blob/main/common/db/src/lib.rs#L107 - maybe we should handle the error here?

https://github.com/guacsec/trustify/blob/main/modules/storage/src/service/fs.rs#L150 - maybe we should use the same pattern here?

This can all be handled in a separate PR (if we decide it is needed)

PR #2103, opened 2025-11-14
fix: SBOM details filters out the deprecated advisories (TC-3191)

mrizzi:
Starting with the investigation to define the expected numbers of vulnerabilities in https://github.com/guacsec/trustify-ui/pull/820, it turned out that there's no deprecated advisories filtering applied when retrieving the SBOM details.

A test for reproducing this has been added.

Related to https://issues.redhat.com/browse/TC-3191

## Summary by Sourcery

Filter out deprecated advisories when fetching SBOM details to prevent obsolete entries from appearing

Bug Fixes:
- Exclude deprecated advisories in SBOM details queries (ORM and raw SQL)

Tests:
- Add integration test to verify deprecated advisories are filtered out from the SBOM advisory endpoint

### DIFF
--- a/etc/test-data/cve/CVE-2024-26308-updated.json
+++ b/etc/test-data/cve/CVE-2024-26308-updated.json
@@ -1,0 +1,181 @@
+{
+    "dataType": "CVE_RECORD",
+    "dataVersion": "5.1",
+    "cveMetadata": {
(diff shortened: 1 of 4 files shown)


dejanb (approved):
Looks great, thanks for looking into it

PR #2132, opened 2025-11-24
Revert b051a77 and 8848e2b, actually fixing TC-2717

ctron:
This reverts the following commits:

* b051a77
* 8848e2b

As they have been extending the scope of the API in a way it never should have. While this might be considered a breaking change, the original definition of the API (which might not have been documented properly) never intended for a "key" to support "fuzzy matching".

Also the commits didn't completely implement the fuzzy matching, as the implementation of the DB loading phase, latest and non-latest are not aligned. Creating an inconsistent behavior.

It also fixed the actually reported TC-2717 issue (500 in certain cases).

## Summary by Sourcery

Revert unintended fuzzy-matching behavior in component lookup APIs and restore exact-match semantics while aligning query building across lookup types to fix TC-2717 errors.

Bug Fixes:
- Fix TC-2717 by preventing 500 errors during component lookup when using exact identifier queries.

Enhancements:
- Unify and generalize ranked query construction for component lookups across different entity types to ensure consistent latest/non-latest behavior.
- Tighten in-memory graph matching for pURLs to require exact matches instead of prefix-based matching.

Documentation:
- Clarify API and OpenAPI documentation to state that component lookups use exact matching for name, pURL, and CPE keys.

Tests:
- Extend latest-filters endpoint tests to cover exact vs fuzzy matching for IDs and latest/non-latest scenarios, validating the restored semantics.

### DIFF
--- a/entity/src/sbom_node_checksum.rs
+++ b/entity/src/sbom_node_checksum.rs
@@ -26,6 +26,12 @@ pub enum Relation {
     Sbom,
     #[sea_orm(has_many = "super::sbom_node::Entity")]
     Node,
+    #[sea_orm(
+        belongs_to = "Entity",
+        from = "(Column::Value, Column::Type)",
+        to = "(Column::Value, Column::Type)"
+    )]
+    SelfRef,
 }
 
 impl Related<super::sbom::Entity> for Entity {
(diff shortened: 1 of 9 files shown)


ctron:
@sourcery-ai review

PR #2143, opened 2025-11-26
chore: refactor sort_by_id into sort_by_key

mrizzi:
Addressing comment https://github.com/guacsec/trustify/pull/2138#pullrequestreview-3510103919

## Summary by Sourcery

Enhancements:
- Remove the generic sort_by_id macro and inline explicit ID-based sort_by_key calls for all relevant model collections prior to batch inserts.

### DIFF
--- a/modules/ingestor/src/service/advisory/csaf/creator.rs
+++ b/modules/ingestor/src/service/advisory/csaf/creator.rs
@@ -150,6 +150,7 @@ impl<'a> StatusCreator<'a> {
 
             let product_id = ProductInformation::create_uuid(org_id, product.product.clone());
 
+            // Warn: id must be Set(), required for sorting
             let product_entity = product::ActiveModel {
                 id: Set(product_id),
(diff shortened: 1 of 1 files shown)


ctron:
I removed the auto-merge so you can address those two comments, or ignore them and submit it to merging.

mrizzi:
Good catch, thanks.

>     * There are a bunch of more `.clone()` calls which could be removed
Done in 5ff7048

>     * `.as_ref()` for `ActiveModel` may panic in case the model value is `NotSet`. This doesn't happen in this case I think. But it migth be worth a note.
I'm not very confident on the idiomatic way to solve this so I asked for some clarification in https://matrix.to/#/!heUmIqxVhZejiSjzVu:matrix.org/$Nq8xBX6Wo_RZoYEgntXfBraqLiXKP5hm8DGNOGcJY-4?via=matrix.org&via=dentrassi.de

PR #2151, opened 2025-11-27
Add expression index for sargable queries

gildub:
Add expression index built on the result of expression applied to the package column to make JOIN or WHERE queries more efficient.

The current join:
```sql
JOIN base_purl ON 
    product_status.package = base_purl.name 
    OR product_status.package LIKE CONCAT(base_purl.namespace, '/', base_purl.name)
```
The problem is that even if with an index on product_status.package, PostgreSQL cannot use it because:
- LIKE with a pattern that starts with a variable (CONCAT(...)) prevents index usage
- The OR condition makes it worse
- It's not sargable (able to use indexes -> Search ARGument  ABLE)

This expression index pre-computes the split (see the slash in the CONCAT) so PostgreSQL doesn't have to do it for every row during the join:
- Without index: PostgreSQL scans 100k rows and calls split_part() 100k times
- With index: PostgreSQL looks up the pre-computed values in O(log n) B-tree time

The migration adds `package_namespace` and `package_name` as GENERATED STORED columns
derived from the existing `package` field, along with a composite index to optimize lookups.

# Compatibility Guarantees
1. **Source of Truth**: `package` remains the authoritative field. The generated columns
   are automatically maintained by PostgreSQL and cannot diverge from their source.

2. **Existing Code Paths**: All existing code that reads/writes `package` continues to work:
   - Insertions only need to set `package` (generated columns auto-populate)
   - Updates to `package` automatically recalculate the generated columns
   - Queries using `product_status.package` remain valid (not replaced by this migration)

3. **Query Pattern Equivalence**:
   OLD: `package = 'name' OR package = 'namespace/name'`
   NEW: `(package_namespace IS NULL AND package_name = 'name') OR
         (package_namespace = 'namespace' AND package_name = 'name')`
   Both patterns return identical results; the new pattern can use the composite index.

4. **Index Usage**: The index on (package_namespace, package_name) efficiently handles:
   - Lookups with NULL namespace (packages without '/')
   - Lookups with non-NULL namespace (scoped packages like npm, maven)
   - B-tree indexes naturally support NULL values in PostgreSQL

# Migration Safety
- GENERATED STORED columns are computed at write-time, not query-time
- Backfill UPDATE ensures all existing rows have populated generated columns
- CONCURRENTLY option prevents write blocking during index creation
- Columns added with IF NOT EXISTS for migration idempotency

# Affected Code Locations
- entity/src/product_status.rs: Entity updated with new columns
- modules/fundamental/src/sbom/model/raw_sql.rs: Can use new columns for optimized joins
- modules/fundamental/src/purl/model/details/purl.rs: Can migrate to indexed lookups
- modules/fundamental/src/vulnerability/model/details/vulnerability_advisory.rs: Already uses new columns

See https://github.com/guacsec/trustify/issues/2039
And https://issues.redhat.com/browse/TC-3106

## Summary by Sourcery

Introduce a database migration to optimize product status lookups and update advisory queries to use the new indexed columns.

Enhancements:
- Register a new migration that improves product_status performance via additional schema/index changes.
- Update vulnerability advisory summary queries to join base_purl using package_namespace and package_name instead of non-sargable LIKE/OR conditions.

### DIFF
--- a/migration/src/m0001220_improve_product_status.rs
+++ b/migration/src/m0001220_improve_product_status.rs
@@ -48,11 +48,10 @@ impl MigrationTrait for Migration {
             )
             .await?;
 
-        // CONCURRENTLY (not supported by SeaORM) to avoid blocking writes
         manager
             .get_connection()
             .execute_unprepared(
-                "CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_product_status_package_lookup \
+                "CREATE INDEX IF NOT EXISTS idx_product_status_package_lookup \
                  ON product_status (package_namespace, package_name)",
             )
             .await?;


gildub:
@sourcery-ai dismiss

gildub:
@sourcery-ai dismiss

gildub:
@sourcery-ai resolve

migration/src/m0001220_improve_product_status.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .get_connection()
+            .execute_unprepared("DROP INDEX IF EXISTS idx_product_status_package_lookup")
+            .await?;
+
+        manager
+            .get_connection()
+            .execute_unprepared("ALTER TABLE product_status DROP COLUMN IF EXISTS package_name")
```
ctron: This can be done in seaorm. Like the others.

PR #2156, opened 2025-12-01
fix: ensure %2F in purl names is handled properly

### LINKED ISSUE #2146: PURL's with %2F in their name aren't ingested properly
https://github.com/guacsec/trustify/blame/9aaf962b917d0f8d267e2e1e58a89c57f57b90d6/modules/fundamental/src/sbom/endpoints/test.rs#L1672-L1685

Note that the %2F characters that appear in the original SBOM aren't preserved in our API results. We add a namespace to the PURL's that doesn't exist in the SBOM.

I've created an [upstream issue](https://github.com/scm-rs/packageurl.rs/issues/28) for this.

jcrossley3:
This fixes #2146 and therefore indirectly resolves the downstream issue, https://issues.redhat.com/browse/TC-3090

It relies on unpublished forks of two interdependent crates: csaf-rs and packageurl.rs

Ultimately, it'd be swell if we could instead depend on something less dead and more supported like https://github.com/csaf-rs/csaf

## Summary by Sourcery

Handle percent-encoded slashes in package URLs and align SBOM and advisory translation logic with updated packageurl behavior.

Bug Fixes:
- Preserve and correctly round-trip percent-encoded slashes (%2F) in purl names when formatting and parsing package URLs.
- Ensure SBOM lookup returns consistent results for equivalent huggingface and generic package URLs with encoded namespace separators.

Enhancements:
- Update purl construction and advisory translation helpers to propagate errors from namespace and version setters instead of silently ignoring failures.

Build:
- Bump packageurl to a newer unreleased revision and switch csaf to a branch with purl name encoding fixes, wiring these forks into Cargo dependencies.

Tests:
- Extend SBOM endpoint and purl unit tests to cover purls containing percent-encoded slashes and their equivalence across ecosystems.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2090,7 +2090,7 @@ dependencies = [
 [[package]]
 name = "csaf"
 version = "0.5.0"
-source = "git+https://github.com/trustification/csaf-rs?branch=purl-name-encoding-fix#9f4e0d24795a1d01eef1d805ab434c28d6f4df8a"
+source = "git+https://github.com/trustification/csaf-rs?branch=main#63ac9e19d881cbf1808de38b6849635cda19931d"
 dependencies = [
  "chrono",
  "cpe",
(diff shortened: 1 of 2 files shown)


dejanb (approved):
This looks OK to me ... Let's look next how to properly maintain csaf library or use the alternative

PR #2174, opened 2025-12-09
Complete the logic for removing the advisory.

bxf12315:
## Summary by Sourcery

Ensure deleting an advisory also removes all associated records and is covered by tests.

Bug Fixes:
- Fix advisory deletion to also clean up related CVSS, package status, vulnerability description, and advisory-vulnerability records.

Enhancements:
- Refine advisory deletion logic to return a boolean when no advisory is found and to update deprecated advisory metadata and source documents after cleanup.

Tests:
- Add an integration test that ingests an advisory with related records, deletes it via the service, and verifies all associated rows are removed.

### DIFF
--- a/modules/fundamental/src/advisory/service/mod.rs
+++ b/modules/fundamental/src/advisory/service/mod.rs
@@ -1,11 +1,7 @@
-use crate::{
-    Error,
-    advisory::model::{AdvisoryDetails, AdvisorySummary},
-};
 use sea_orm::{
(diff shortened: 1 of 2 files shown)


bxf12315:
> @bxf12315 I think this should be working as expected, as pr #1569 added `CASCADE DELETE` for cvss tables. I think the test should pass even without changes.

Yes, that’s absolutely correct.

PR #2183, opened 2025-12-12
perf: Improves a bit the recommend endpoint

helio-frota:
Assisted-by: Claude Code

Before:

After:

Before:
https://gist.github.com/helio-frota/25af13c01066bde8c506bee4a7b32b9e

After:
https://gist.github.com/helio-frota/0d17454a2ef37c5f3cbe41f64aec6073

## Summary by Sourcery

Optimize advisory recommendation query by batching related entity loading and reusing pre-fetched status and organization data.

Enhancements:
- Batch-load advisory organizations and statuses to reduce database queries and improve performance of the recommend endpoint.
- Refactor VersionedPurlStatus construction to accept pre-fetched status models instead of querying by ID per record.


mrizzi (approved):
@helio-frota thanks for the performance improvement 👏

PR #2186, opened 2025-12-17
doc(adr): add ADR 00012 SBOM Pruning CLI Tool

dejanb:
Propose new CLI for manually pruning SBOMs

## Summary by Sourcery

Documentation:
- Add ADR describing a proposed `trustify` CLI with a `prune` subcommand for bulk SBOM deletion, including requirements, design, and future phases for automated lifecycle management.

### DIFF
--- a/docs/adrs/00012-sbom-pruning-cli.md
+++ b/docs/adrs/00012-sbom-pruning-cli.md
@@ -1,0 +1,243 @@
+# 00010. SBOM Pruning CLI Tool
+
+## Status
+
+PROPOSED
+
+## Context
+
+### Problem Statement
+
+Production Trustify deployments contain millions of SBOMs, causing significant storage and database cost concerns. Currently, Trustify lacks:
+- Bulk deletion capabilities
+- Automated pruning mechanisms
+- Tools for lifecycle management of SBOMs
+
+### Current State
+
+- **Delete API**: `DELETE /v2/sbom/{id}` supports single SBOM deletion only
+- **Search API**: Supports date filtering (`ingested<30 days ago`) and label filtering
+- **Gaps**: No batch operations, no CLI tooling, no automation
+
+### Requirements
+
+**Immediate need**: Manual tool for bulk SBOM deletion to reduce storage costs
+
+**Future needs** (subsequent ADRs):
+- Automated/scheduled pruning (Phase 2)
+- Soft-delete and restore capabilities (Phase 3+)
+- Policy-based lifecycle management (Phase 4+)
+
+## Decision
+
+### Create Separate CLI Binary: `trustify`
+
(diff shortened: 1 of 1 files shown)


docs/adrs/00012-sbom-pruning-cli.md
```diff
@@ -0,0 +1,243 @@
...
+
+    /// Max concurrent deletions
+    #[arg(long, default_value = "5")]
+    max_concurrent: usize,
+}
+```
+
+### Core Algorithm
+
+1. Authenticate with Trustify API using OIDC token
```
ruromero: Having users retrieve and provide OIDC tokens given the short life span (10 mins) they have will be frustrating to users and not practical for unassisted tasks (example Jobs).

For unassisted tasks like Jobs or CI I guess we can either provide an offline_token (retrieved from the trustify UI, but that requires a specific OIDC client with offline_access) or client_id + client_secret that will interact directly with the IdP to retrieve a valid token.
For standard cli usage like `oc` or `podman` users can use the Device Auth Flow (the token will be exposed by the UI) or even involve a browser and do Authentication Code Flow (I don't think that's worth it)

ctron:
I like the approach. What might get tricky is the amount of connection information required (endpoint, issuer URL, client ID and secret), … I think it would make sense to store this in a configuration, with a profile (multiple instances). Then use `--profile` or `TRUSTIFY_PROFILE` (env-var), with the option to have a default one.

This file could be created manually at first. But later on, we could have `trustify login` as a command, performing those steps and manage the creation of this file.

bxf12315:
Regarding this ADR, I have three questions.​
First, the deletion here should be a physical deletion rather than a soft deletion, right?​
Second, is it sufficient to record deletions only through logs, without using a history table to capture who deleted the historical data, why these data were deleted, and related information, in order to form a complete, auditable deletion trail?​
Third, do we need to consider re‑ingest the deleted records?​

dejanb:
Thinking a bit more about it, maybe a better approach would be to create an 'admin' API instead of the CLI. The functionality would be completely the same, we would just avoid thinking about all these auth topics and publishing/distributing the binary. Also, users wouldn't need to install anything and could just use the feature using curl/httpie.

So, the example usage could be something like

`http POST https://trustify.example.com/v1/admin/sbom/prune filter=="ingested<90 days ago&label:env=staging" batch-size ==500 dry-run==true`

@ruromero @PhilipCattanach @ctron wdyt?
@PhilipCattanach Automated process is the end goal, but it would take much more time to develop. Creating manual cli/endpoint is the step we need to do anyway (it contains the actual pruning logic) and it will provide users with something useful much sooner. Additionally, having a tool for manual management is useful on its own for cleaning up one-off incidents and testing automated configurations.

PR #2197, opened 2026-01-12
doc(adr): SBOM Prune API

dejanb:
## Summary by Sourcery

Documentation:
- Add ADR describing the SBOM prune REST endpoint, including API contract, safety features, logging requirements, and future scheduled pruning service.


bxf12315:
Do we need to consider re‑ingest the deleted records?​

dejanb:
I'll close this one for now as the predominant opinion is that we should not go this route. We can re-open it if needed later on

i386x:
@dejanb This can help with TC-3934.

PR #2222, opened 2026-01-29
Add Mermaid system, components and container diagrams

gildub:
The design/README.md articulates together the design information of the Trustify project using Mermaid providing diagram as code with the following :

- The domain model, which purpose is to have a common domain knowledge for all participate to communicate efficiently.
- The architecture (C4 diagrams): the system context, the container's diagram, 2 components views (API service, Ingestor service)
- The data model represented with erDiagram (Entity Relation)
- An how to install and use `mermaid-cli` tool in order to regenerate images (PNG, SVG or PDF) along with the default theme and scale parameters.


ctron:
Wouldn't an ER diagram be a better fit for this?

Also, why does it have another diagram as PNG? And how do we ensure this stays in sync? Which tool was used to create this? Can the resolution be enhanced.

ctron:
Thanks for explaining this. I still don't see the benefit of having the (quite small) PNG, compared to the complexity it adds. And the problems when things go out of alignment.

I think it's easy enough to view the diagram on GitHub, and at least in IntelliJ too. I'd assume VScode has something similar.

So which case would require having this PNG?

gildub:
@ctron, 

The PNG (or SVG, JPEG) are definitely optional and I understand everyone might have different needs to use them or not. I removed the images and let's keep only the instruction for anyone to generate them if needed.

gildub:
@ctron, I believe the ADR documents approach was put in place afterward the start of the project, therefore there is no "initial" ADR which paved the design document. The intent of this PR is to fill the gap and give us that initial design document. Following ADRs will have the responsibility to amend the main design document to reflect what they bring to the table.

That said, when adding ADR we could discover that some part of the main design would be better if dropped or broken down to avoid excessive work, then we'll adapt.

docs/design/data-model.md
```diff
@@ -0,0 +1,436 @@
...
+    versioned_purl ||--o{ purl_license_assertion : "licensed_as"
+
+    qualified_purl ||--o{ sbom_package_purl_ref : "referenced_in"
+
+    %% Relationships - SBOM
+    sbom ||--o{ source_document : "stored_in"
+    sbom ||--o{ sbom_node : "contains"
+    sbom ||--o{ sbom_package : "has_packages"
+    sbom ||--o{ sbom_file : "has_files"
+    sbom ||--o{ sbom_node_checksum : "has_checksums"
```
ctron: I don't think that's correct. The `sbom_node_checksum` related to the `sbom_node`.

docs/design/data-model.md
```diff
@@ -0,0 +1,436 @@
...
+    sbom ||--o{ sbom_external_node : "references_external"
+    sbom ||--o{ sbom_external_node : "referenced_by"
+    sbom ||--o{ package_relates_to_package : "defines_relationships"
+    sbom ||--o{ purl_license_assertion : "asserts_purl_license"
```
ctron: Not exactly sure how that relates, but pretty sure it does not relate to the `sbom`.

ctron (changes requested):

rh-jfuller:
@gildub Is it possible this work is now out of date ... maybe we should consider closing this PR and create a new one ? Or maybe rebase this one and get it landed ?

ctron:
Closing this for now.

PR #2233, opened 2026-02-04
feat: support querying fields containing the literal string, "null"

jcrossley3:
Relates #2230

BREAKING-CHANGE: Querying for NULL fields is now achieved using an ASCII NUL value, percent-encoded as %00, instead of the literal string "null".

## Summary by Sourcery

Document and implement a new convention for querying unset (NULL) fields using an ASCII NUL value instead of the literal string "null", and update tests accordingly.

Enhancements:
- Change query parsing so that an ASCII NUL value (\x00) in [Not]Equal filters represents NULL/NOT NULL checks instead of the literal string "null".

Documentation:
- Update OpenAPI documentation to describe using ASCII NUL percent-encoded as %00 for querying fields that are NULL, replacing the previous 'null' sentinel examples.

Tests:
- Adjust database and endpoint tests to use the ASCII NUL (\x00) sentinel in query parameters when asserting NULL/NOT NULL filtering semantics.

### DIFF
--- a/common/src/db/query/columns.rs
+++ b/common/src/db/query/columns.rs
@@ -161,18 +161,30 @@ impl Columns {
         value: &str,
     ) -> Result<SimpleExpr, Error> {
         self.for_field(field).and_then(|(lhs, ct)| {
-            match (value.to_lowercase().as_str(), operator, &ct) {
-                ("\x00", Operator::Equal, _) => Ok(lhs.is_null()),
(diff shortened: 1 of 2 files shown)


carlosthe19916 (approved):
Thanks for the heads up guys. I ran some tests and verify the current UI code. After this PR is merged I will need to patch the UI through this PR https://github.com/guacsec/trustify-ui/pull/919 so everything fine from the UI side.

PR #2242, opened 2026-02-11
perf(license): add indexes and refactor 'NOT EXIST' query (TC-3591)

mrizzi:
- HASH index on `license(text)` for exact-match lookups; HASH is used instead of B-tree because license expressions can exceed the B-tree maximum row size of 2704 bytes                                                                                                                                               
  - B-tree composite index on `sbom_package_license(license_id, sbom_id)` to accelerate the join between `sbom_package_license` and `license` in the `unique_license_sbom` CTE                                                                                                                                                 
  - B-tree composite index on `licensing_infos(sbom_id, license_id, name)` as a covering index for the `licensing_infos_mappings` CTE aggregation, eliminating heap fetches                                                                                                                                                
  - `expand_license_expression_with_mappings()` PL/pgSQL function replaced with optimized version that adds an early-exit check (`license_text !~ 'LicenseRef-'`) before entering the loop, skipping invocations where no `LicenseRef` pattern is present                                                          
  - License listing query (`licenses()` in `license/service/mod.rs`) rewritten to use `NOT EXISTS` instead of `LEFT JOIN ... IS NULL` for finding licenses without SBOMs, enabling a Nested Loop Anti Join with index-only scan instead of a full table scan on `sbom_package_license`

Refers to https://issues.redhat.com/browse/TC-3591

## Summary by Sourcery

Improve license query performance and correctness for LicenseRef expansions while adding supporting database indexes and tests.

Bug Fixes:
- Prevent partial LicenseRef matches from corrupting expanded license names and ensure no raw LicenseRef tokens remain in license listings.

Enhancements:
- Rewrite the license listing query to use a NOT EXISTS anti-join instead of a LEFT JOIN with NULL filter when selecting licenses without SBOMs.
- Optimize the expand_license_expression_with_mappings() function to short-circuit when no LicenseRef pattern is present and to avoid partial identifier matches in replacements.

Build:
- Add a new database migration that introduces indexes on license text, sbom_package_license, and licensing_infos tables to speed up license-related queries and updates the license expansion function definition.

Tests:
- Add an endpoint test covering overlapping LicenseRef identifiers to verify correct expansion behavior and absence of partial matches.

### DIFF
--- a/etc/test-data/spdx/license-ref-overlap.json
+++ b/etc/test-data/spdx/license-ref-overlap.json
@@ -1,0 +1,88 @@
+{
+  "spdxVersion": "SPDX-2.2",
+  "dataLicense": "CC0-1.0",
+  "SPDXID": "SPDXRef-DOCUMENT",
+  "name": "license-ref-overlap-test",
+  "documentNamespace": "https://example.org/test/license-ref-overlap",
+  "creationInfo": {
+    "created": "2024-01-01T00:00:00Z",
+    "creators": [
+      "Tool: test"
+    ],
+    "licenseListVersion": "3.19"
+  },
+  "hasExtractedLicensingInfos": [
+    {
+      "licenseId": "LicenseRef-BSD",
+      "extractedText": "BSD License text",
+      "name": "BSD License"
+    },
+    {
+      "licenseId": "LicenseRef-BSD-with-advertising",
+      "extractedText": "BSD with advertising License text",
+      "name": "BSD with advertising License"
+    },
+    {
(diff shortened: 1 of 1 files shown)


ctron:
@JimFuller-RedHat maybe you could take a look at this

mrrajan:
@JimFuller-RedHat I am in the mid of setting up an environment to compare the performance improvement. Will share the results once the environment is ready.

JimFuller-RedHat (approved):
this PR does not make things better or worst

PR #2247, opened 2026-02-13
fix(sbom): accept SHA-256 identifiers in packages and related endpoints

jangel97:
The `/packages` and `/related` SBOM endpoints declared their path parameter as `web::Path<Uuid>`, which caused actix-web to reject non-UUID identifiers (e.g. `sha256:...`) before the handler was executed.

All other SBOM endpoints already accept identifiers as `String` and parse them via `Id::from_str()`, allowing both UUIDs and SHA-256 hashes.

Changes:
- Updated `/packages` and `/related` endpoints to use `web::Path<String>`
- Parse identifiers using `Id::from_str()`, matching the pattern used by other SBOM endpoints
- Fixes 404 errors when accessing SBOMs by SHA-256 hash

## Summary by Sourcery

Allow SBOM package and related endpoints to resolve SBOMs by either UUID or SHA-256 identifier and return 404 when the SBOM is not found.

Bug Fixes:
- Fix SBOM /packages and /related endpoints rejecting non-UUID identifiers by supporting SHA-256-based lookups.

Enhancements:
- Align SBOM package and related endpoints with other SBOM APIs by resolving a generic Id and using the canonical SBOM head ID internally.
- Document 404 responses for SBOM package and related listing endpoints in the OpenAPI metadata.

Tests:
- Add integration tests verifying /packages and /related endpoints work with SHA-256 IDs, match UUID-based results, and return 404 for unknown hashes.

### DIFF
--- a/modules/fundamental/src/sbom/endpoints/mod.rs
+++ b/modules/fundamental/src/sbom/endpoints/mod.rs
@@ -26,7 +26,7 @@ use crate::{
 use actix_web::{HttpResponse, Responder, delete, get, http::header, post, web};
 use config::Config;
 use futures_util::TryStreamExt;
-use sea_orm::{TransactionTrait, prelude::Uuid};
+use sea_orm::TransactionTrait;
 use serde_qs::actix::QsQuery;
 use std::str::FromStr;
 use trustify_auth::{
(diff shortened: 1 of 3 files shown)


helio-frota:
thanks! it works fine

PR #2250, opened 2026-02-17
Fixed matching vulnerabilities to Rust crates

waldemar-kindler:
## Summary

  **Problem:** OSV advisories from crates.io use `ECOSYSTEM` as their range type (rather than `SEMVER`). The OSV loader
   had no explicit match for `(RangeType::Ecosystem, Ecosystem::CratesIO)`, so these advisories fell through to the
  catch-all handler which calls `build_package_status_versions`. That handler creates exact-match version entries
  instead of proper semver ranges — meaning affected version ranges like `[1.2.1, 1.11.1)` were not being captured, and
   version matching against installed packages would fail silently.

  **Approach:**
  - Add an explicit match arm for `(RangeType::Ecosystem, Ecosystem::CratesIO)` in the OSV loader that routes through
  `build_package_status` with a new `Cargo` version scheme
  - Add `Cargo` as a version scheme variant, backed by semver matching (since Cargo enforces semver)
  - Add a DB migration to insert the `cargo` version scheme and update the `version_matches` SQL function to handle it
  - Add a test using a real crates.io advisory (`GHSA-434x-w66g-qw3r`) that verifies purl status entries are created
  with proper semver version ranges

  ## Test plan
  - [x] New test `loader_crates_io` verifies that ingesting a crates.io advisory produces `purl_status` entries with
  `Cargo` version scheme and correct semver range bounds

## Summary by Sourcery

Handle crates.io OSV advisories using a Cargo-specific version scheme so Rust crate vulnerabilities get correct semver ranges and match installed package versions.

Bug Fixes:
- Ensure crates.io advisories with ECOSYSTEM ranges are ingested as proper semver ranges instead of generic exact-match versions, restoring correct vulnerability matching for Rust crates.

Enhancements:
- Introduce a Cargo version scheme mapped to semver for version range handling across the system.

Build:
- Add a database migration to register the Cargo version scheme and extend the version_matches SQL function to support it.

Tests:
- Add an integration test using a real crates.io advisory to verify Cargo version scheme usage and correct affected version bounds in generated purl_status records.

### DIFF
--- a/modules/ingestor/src/service/advisory/osv/loader.rs
+++ b/modules/ingestor/src/service/advisory/osv/loader.rs
@@ -587,7 +587,9 @@ mod test {
     use sea_orm::{ColumnTrait, EntityTrait, QueryFilter};
     use test_context::test_context;
     use test_log::test;
-    use trustify_entity::{advisory_vulnerability_score, purl_status, version_range, version_scheme};
+    use trustify_entity::{
+        advisory_vulnerability_score, purl_status, version_range, version_scheme,
+    };
     use trustify_test_context::{TrustifyContext, document};
 
     #[test_context(TrustifyContext)]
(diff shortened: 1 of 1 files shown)


waldemar-kindler:
The migration is basically a copy from migration/src/m0001130_gover_cmp_fns with cargo added.

waldemar-kindler:
Added 'ON CONFLICT DO NOTHING' to allow rerunning migration, resolved merge conflict and ran pre-commit hooks

PR #2320, opened 2026-04-10
perf(migration): add FK indexes on advisory child tables

### LINKED ISSUE #2319: Add missing FK indexes on advisory child tables for CASCADE delete performance
## Problem

Three advisory child tables lack indexes on their `advisory_id` foreign key columns, causing PostgreSQL to perform sequential scans during CASCADE deletes and advisory ingestion queries. PostgreSQL does **not** automatically create indexes on FK columns — only on primary keys and unique constraints.

This impacts:
- **Advisory ingestion speed** — `ScoreCreator::create()` unconditionally DELETEs all scores by `advisory_id` on every advisory ingestion (wipe-and-replace pattern)
- **Advisory deletion time** — CASCADE deletes must scan child tables without index support

The performance degradation is linear with table size.

## Benchmark Results (RPi4, 24GB database)

| Table | Rows | Without Index | With Index | Speedup |
|---|---|---|---|---|
| `advisory_vulnerability_score` | ~345K | 75.7ms (seq scan) | 0.4ms (index scan) | **187x** |
| `purl_status` | ~8M | 4.1s (seq scan) | 0.4s (bitmap scan) | **10x** |
| `product_status` | ~124K | 43ms (seq scan) | 5ms (bitmap scan) | **8.5x** |

**End-to-end advisory deletion** (all CASCADEs combined): 68.4s → 22.0s (**3.1x faster**)

## Proposed Solution

Add a SeaORM migration creating 3 single-column `(advisory_id)` indexes:

1. `advisory_vulnerability_score(advisory_id)` — serves CASCADE deletes and ScoreCreator's wipe-and-replace DELETE on every advisory ingestion
2. `purl_status(advisory_id)` — serves CASCADE deletes
3. `product_status(advisory_id)` — serves CASCADE deletes

> **Note:** `purl_status` and `product_status` use single-column indexes rather than composites. The `ingest_package_status()` query is already covered by `package_status_idx(base_purl_id, advisory_id, status_id)`, and the vulnerability detail query filters by `vulnerability_id` only — not `advisory_id`.

All indexes use `IF NOT EXISTS` for idempotency. No storage overhead concern — the 24GB database size remained unchanged after adding all indexes.

## Scale Impact

- `purl_status` is already at 8M rows (3.6GB) and grows with every advisory ingested (CSAF, CVE, and OSV formats all write to it)
- `advisory_vulnerability_score` is at 345K rows and is wiped/rewritten on every advisory re-ingestion
- `product_status` is at 124K rows and grows with each advisory containing product trees

Jira: [TC-4064](https://redhat.atlassian.net/browse/TC-4064)

[TC-4064]: https://redhat.atlassian.net/browse/TC-4064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ruromero:
## Summary

- Add SeaORM migration `m0002180_advisory_fk_indexes` creating 3 missing FK indexes on advisory child tables
- Eliminates sequential scans during CASCADE deletes and advisory ingestion queries
- Measured **3.1x faster** end-to-end advisory deletion (68.4s → 22.0s on 24GB database)

### Indexes

| Table | Index Columns | Use Cases |
|---|---|---|
| `advisory_vulnerability_score` | `(advisory_id)` | ScoreCreator wipe-and-replace DELETE on every ingestion, CASCADE delete |
| `purl_status` | `(advisory_id)` | CASCADE delete |
| `product_status` | `(advisory_id)` | CASCADE delete |

All three indexes cover the `advisory_id` FK column that PostgreSQL does not auto-index. The primary use case is the CASCADE delete path:

```
DELETE /v2/advisory/{key}
  → DELETE FROM advisory WHERE id = $1
    → CASCADE to advisory_vulnerability_score, purl_status, product_status
```

The `advisory_vulnerability_score` index additionally serves `ScoreCreator::create()` which does a direct `DELETE ... WHERE advisory_id = $1` on every advisory ingestion (CSAF, CVE, OSV loaders).

> **Note:** `purl_status` and `product_status` use single-column `(advisory_id)` indexes rather than composites. The `ingest_package_status()` query is already covered by `package_status_idx(base_purl_id, advisory_id, status_id)`, and the vulnerability detail query filters by `vulnerability_id` only — not `advisory_id`.

## Test plan

- [x] `cargo check -p trustify-migration` — compiles
- [x] `cargo clippy -p trustify-migration --all-targets` — no warnings
- [x] `cargo fmt --check` — formatted
- [x] `cargo test -p trustify-migration` — migration up/down passes

Implements [TC-4064](https://redhat.atlassian.net/browse/TC-4064)
Closes trustification/trustify#2319

[TC-4064]: https://redhat.atlassian.net/browse/TC-4064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/migration/src/m0002180_advisory_fk_indexes.rs
+++ b/migration/src/m0002180_advisory_fk_indexes.rs
@@ -22,34 +22,30 @@ impl MigrationTrait for Migration {
             )
             .await?;
 
-        // Composite index on purl_status(advisory_id, status_id)
+        // Index on purl_status(advisory_id)
         //
-        // Required for CASCADE deletes on advisory deletion. The composite also
-        // serves ingest_package_status() which filters by (advisory_id, status_id).
+        // Required for CASCADE deletes on advisory deletion.
         manager
             .create_index(
                 Index::create()
                     .if_not_exists()
                     .table(PurlStatus::Table)
-                    .name(Indexes::IdxPurlStatusAdvisoryIdStatusId.to_string())
+                    .name(Indexes::IdxPurlStatusAdvisoryId.to_string())
                     .col(PurlStatus::AdvisoryId)
-                    .col(PurlStatus::StatusId)
                     .to_owned(),
             )
             .await?;
 
-        // Composite index on product_status(advisory_id, vulnerability_id)
+        // Index on product_status(advisory_id)
         //
-        // Required for CASCADE deletes on advisory deletion. The composite also
-        // serves vulnerability detail view JOINs on (advisory_id, vulnerability_id).
+        // Required for CASCADE deletes on advisory deletion.
         manager
             .create_index(
                 Index::create()
                     .if_not_exists()
                     .table(ProductStatus::Table)
-                    .name(Indexes::IdxProductStatusAdvisoryIdVulnId.to_string())
+                    .name(Indexes::IdxProductStatusAdvisoryId.to_string())
                     .col(ProductStatus::AdvisoryId)
-                    .col(ProductStatus::VulnerabilityId)
                     .to_owned(),
             )
             .await?;
@@ -63,7 +59,7 @@ impl MigrationTrait for Migration {
                 Index::drop()
                     .if_exists()
                     .table(ProductStatus::Table)
-                    .name(Indexes::IdxProductStatusAdvisoryIdVulnId.to_string())
+                    .name(Indexes::IdxProductStatusAdvisoryId.to_string())
                     .to_owned(),
             )
             .await?;
@@ -73,7 +69,7 @@ impl MigrationTrait for Migration {
                 Index::drop()
                     .if_exists()
                     .table(PurlStatus::Table)
-                    .name(Indexes::IdxPurlStatusAdvisoryIdStatusId.to_string())
+                    .name(Indexes::IdxPurlStatusAdvisoryId.to_string())
                     .to_owned(),
             )
             .await?;
@@ -96,8 +92,8 @@ impl MigrationTrait for Migration {
 #[derive(DeriveIden)]
 enum Indexes {
     IdxAdvVulnScoreAdvisoryId,
-    IdxPurlStatusAdvisoryIdStatusId,
-    IdxProductStatusAdvisoryIdVulnId,
+    IdxPurlStatusAdvisoryId,
+    IdxProductStatusAdvisoryId,
 }
 
 #[derive(DeriveIden)]
@@ -110,12 +106,10 @@ enum AdvisoryVulnerabilityScore {
 enum PurlStatus {
     Table,
     AdvisoryId,
-    StatusId,
 }
 
 #[derive(DeriveIden)]
 enum ProductStatus {
     Table,
     AdvisoryId,
-    VulnerabilityId,
 }
--- a/migration/src/m0002180_advisory_fk_indexes.rs
+++ b/migration/src/m0002180_advisory_fk_indexes.rs
@@ -1,0 +1,121 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+#[allow(deprecated)]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        // Index on advisory_vulnerability_score(advisory_id)
+        //
+        // Required for CASCADE deletes and ScoreCreator's wipe-and-replace DELETE
+        // that runs on every advisory ingestion. Also used by UI CVSS score loading.
+        manager
+            .create_index(
+                Index::create()
+                    .if_not_exists()
+                    .table(AdvisoryVulnerabilityScore::Table)
+                    .name(Indexes::IdxAdvVulnScoreAdvisoryId.to_string())
+                    .col(AdvisoryVulnerabilityScore::AdvisoryId)
+                    .to_owned(),
+            )
+            .await?;
+
+        // Composite index on purl_status(advisory_id, status_id)
+        //
+        // Required for CASCADE deletes on advisory deletion. The composite also
+        // serves ingest_package_status() which filters by (advisory_id, status_id).
+        manager
+            .create_index(
+                Index::create()
+                    .if_not_exists()
+                    .table(PurlStatus::Table)
+                    .name(Indexes::IdxPurlStatusAdvisoryIdStatusId.to_string())
+                    .col(PurlStatus::AdvisoryId)
+                    .col(PurlStatus::StatusId)
+                    .to_owned(),
+            )
+            .await?;
+
+        // Composite index on product_status(advisory_id, vulnerability_id)
+        //
+        // Required for CASCADE deletes on advisory deletion. The composite also
+        // serves vulnerability detail view JOINs on (advisory_id, vulnerability_id).
+        manager
+            .create_index(
+                Index::create()
+                    .if_not_exists()
+                    .table(ProductStatus::Table)
+                    .name(Indexes::IdxProductStatusAdvisoryIdVulnId.to_string())
+                    .col(ProductStatus::AdvisoryId)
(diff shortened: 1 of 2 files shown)


Strum355:
Would it be possible to get specific before & after query plans for each of the affected queries? Itd be good to confirm these results concretely rather than trusting the LLM :smile: 

An example of my "doubts" (for need of a better word) is in the table where it mentions that `purl_status.(advisory_id, status_id)` index helps with queries in `ingest_package_status`. In that function, we have a query on `purl_status` filtering on `base_purl_id`, `advisory_id` and `status_id`, for which there is already an index `"package_status_idx" btree (base_purl_id, advisory_id, status_id)`, so Im unclear on how these changes help in that function or if the LLM is hallucinating.

ruromero:
> Would it be possible to get specific before & after query plans for each of the affected queries? Itd be good to confirm these results concretely rather than trusting the LLM 😄
> 
> An example of my "doubts" (for need of a better word) is in the table where it mentions that `purl_status.(advisory_id, status_id)` index helps with queries in `ingest_package_status`. In that function, we have a query on `purl_status` filtering on `base_purl_id`, `advisory_id` and `status_id`, for which there is already an index `"package_status_idx" btree (base_purl_id, advisory_id, status_id)`, so Im unclear on how these changes help in that function or if the LLM is hallucinating.

These numbers are not generated by the LLM but coming from actual benchmarks ran against my local environment. 
5 executions are run for each test case and I have the execution with the `EXPLAIN ANALYZE` output for each run.

Example:
```
TEST: adv_vuln_score_advisory_id

  Table:     advisory_vulnerability_score
  Column:    advisory_id
  Use case:  DELETE during every advisory ingestion (ScoreCreator::create) + CASCADE. Also benefits: UI advisory detail CVSS loading
  Test ID:   advisory_id = 037fc28e-de30-4998-8414-21475dba5b19
  Query:     DELETE FROM advisory_vulnerability_score WHERE advisory_id = '037fc28e-de30-4998-8414-21475dba5b19';
  Index DDL: CREATE INDEX idx_bench_avs_advisory_id ON advisory_vulnerability_score (advisory_id);
  Runs:      5

  WITHOUT index:
    Median: 75.667 ms  (min=74.777, max=214.205)
    Plan:   ->  Seq Scan on advisory_vulnerability_score  (cost=0.00..10334.15 rows=2 width=6) (actual time=0.021..213.864 rows=26 loops=1)

  Creating index...
  WITH index:
    Median: 0.405 ms  (min=0.387, max=0.664)
    Plan:   ->  Index Scan using idx_bench_avs_advisory_id on advisory_vulnerability_score  (cost=0.42..3.76 rows=2 width=6) (actual time=0.126..0.142 rows=26 loops=1)

  RESULT: 75.667 ms → 0.405 ms  (186.8x faster, median of 5 runs)
  ```
  
  The explain for run 5 before:
  ```
BEGIN
                                                            QUERY PLAN                                                            
----------------------------------------------------------------------------------------------------------------------------------
 Delete on advisory_vulnerability_score  (cost=0.00..10334.15 rows=0 width=0) (actual time=75.303..75.304 rows=0 loops=1)
   Buffers: shared hit=6047
   ->  Seq Scan on advisory_vulnerability_score  (cost=0.00..10334.15 rows=2 width=6) (actual time=0.052..75.208 rows=26 loops=1)
         Filter: (advisory_id = '037fc28e-de30-4998-8414-21475dba5b19'::uuid)
         Rows Removed by Filter: 345026
         Buffers: shared hit=6021
 Planning:
   Buffers: shared hit=91
 Planning Time: 1.027 ms
 Execution Time: 75.506 ms
(10 rows)

ROLLBACK
  ```
  
  The explain for run 5 after:
```
BEGIN
                                                                          QUERY PLAN                                                                           
---------------------------------------------------------------------------------------------------------------------------------------------------------------
 Delete on advisory_vulnerability_score  (cost=0.42..3.76 rows=0 width=0) (actual time=0.211..0.212 rows=0 loops=1)
   Buffers: shared hit=30
   ->  Index Scan using idx_bench_avs_advisory_id on advisory_vulnerability_score  (cost=0.42..3.76 rows=2 width=6) (actual time=0.097..0.113 rows=26 loops=1)
         Index Cond: (advisory_id = '037fc28e-de30-4998-8414-21475dba5b19'::uuid)
         Buffers: shared hit=4
 Planning:
   Buffers: shared hit=107
 Planning Time: 1.275 ms
 Execution Time: 0.402 ms
(9 rows)

ROLLBACK
```

File: migration/src/m0002180_advisory_fk_indexes.rs [hunk 1]
```diff
@@ -0,0 +1,121 @@
...
+                    .col(AdvisoryVulnerabilityScore::AdvisoryId)
+                    .to_owned(),
+            )
+            .await?;
+
+        // Composite index on purl_status(advisory_id, status_id)
+        //
+        // Required for CASCADE deletes on advisory deletion. The composite also
+        // serves ingest_package_status() which filters by (advisory_id, status_id).
+        manager
```
Strum355: Agreed on the CASCADE part, I don't agree on the `ingest_package_status` section. Unless Im mistaken (correct me if Im wrong), this references the following query:
```rs
let package_status = purl_status::Entity::find()
            .filter(purl_status::Column::BasePurlId.eq(package.base_purl.id))
            .filter(purl_status::Column::AdvisoryId.eq(self.advisory.advisory.id))
            .filter(purl_status::Column::StatusId.eq(status.id))
            .left_join(version_range::Entity)
            .filter(info.clone().into_condition())
            .one(connection)
            .await?;
```
This query should already be covered by the following index:
```
Indexes:
    ...
    "package_status_idx" btree (base_purl_id, advisory_id, status_id)
```
while yes, postgres can use multiple indexes (using a bitmap join), the index being proposed would potentially be better off being an index _only_ on `advisory_id`. Unless theres a specific query that uses `(advisory_id, status_id)` filter _without_ `base_purl_id`

ruromero: @Strum355 You're right. Thanks for the careful review.                                                                                                                                      

I traced the end-to-end paths for each index and confirmed your finding:                                                                                                                                
                                                                                                                                                                                                           
   **purl_status**: The `ingest_package_status()` query filters by `(base_purl_id, advisory_id, status_id)` and is already fully covered by the existing `package_status_idx btree (base_purl_id,          
   advisory_id, status_id)`. The composite `(advisory_id, status_id)` was redundant.                                                                                                                       
                                                                                                                                                                                                           
   **product_status**: Similarly, the vulnerability detail query (`vulnerability_advisory.rs`) filters by `vulnerability_id` only — not `advisory_id` — so the composite `(advisory_id,                    
   vulnerability_id)` didn't serve that use case either.

   Both have been simplified to single-column `(advisory_id)` indexes. The only real use case for all three indexes is the CASCADE delete path:

   ```
   DELETE /v2/advisory/{key}
     → DELETE FROM advisory WHERE id = $1
       → CASCADE to advisory_vulnerability_score, purl_status, product_status
   ```

   The `advisory_vulnerability_score(advisory_id)` index additionally serves `ScoreCreator::create()` which does a direct `DELETE ... WHERE advisory_id = $1` on every advisory ingestion.

   Updated in 86910e96.

Strum355: Fantastic, thanks for confirming! I'm good with these changes, indexes for cascades is very important

Strum355 (approved):

ctron:
Is there an estimate on what that image will cost in storage?

ruromero:
> Is there an estimate on what that image will cost in storage?

If calculations are correct, for 8M rows given that the advisory_id is of type UUID (16bytes) + 8 bytes of header it'll translate into a 240MB.

PR #2325, opened 2026-04-14
feat: add endpoint for returning all AIBOM's, optionally filtered

### LINKED ISSUE #2324: Complete work on AI endpoint
This is mostly a placeholder for the downstream issue, https://redhat.atlassian.net/browse/TC-3198, which was partially fulfilled by #2241 and #2255 

To complete the work, we need an endpoint that returns all the AI models in the system, optionally filtered with the `q=` param.

jcrossley3:
Fixes #2324

## Summary by Sourcery

Add a new API endpoint to list all AI SBOM models with flexible querying and pagination, alongside adapting existing model retrieval logic to support both per-SBOM and global queries.

New Features:
- Expose a /api/v2/sbom/models endpoint to search across all AI models with query, sorting, offset, and limit parameters.

Enhancements:
- Generalize SbomService model-fetching to accept an optional SBOM id so it can be reused for both single-SBOM and global model searches.
- Register the new all-models handler in the SBOM endpoints configuration and document it in the OpenAPI/utoipa specs.

Tests:
- Add integration tests covering various positive and negative query patterns for the new all-models AI SBOM models endpoint.

### DIFF
--- a/modules/fundamental/src/sbom/endpoints/mod.rs
+++ b/modules/fundamental/src/sbom/endpoints/mod.rs
@@ -64,6 +64,7 @@ pub fn configure(
         .service(v3::all)
         .service(all_related)
         .service(count_related)
+        .service(all_models)
         .service(get)
         .service(get_sbom_advisories)
         .service(delete)
@@ -478,9 +479,35 @@ pub async fn models(
 ) -> actix_web::Result<impl Responder> {
     let tx = db.begin_read().await?;
     let result = fetch
-        .fetch_sbom_models(id.into_inner(), search, paginated, &tx)
+        .fetch_sbom_models(Some(id.into_inner()), search, paginated, &tx)
         .await?;
+    Ok(HttpResponse::Ok().json(result))
+}
 
+/// Search for all AI models
+#[utoipa::path(
+    tag = "sbom",
+    operation_id = "listAllModels",
+    params(
+        Query,
+        Paginated,
+    ),
+    responses(
+        (status = 200, description = "AI Models", body = PaginatedResults<SbomModel>),
+    ),
+)]
+#[get("/v2/sbom/models")]
+pub async fn all_models(
+    fetch: web::Data<SbomService>,
+    db: web::Data<Database>,
+    web::Query(search): web::Query<Query>,
+    web::Query(paginated): web::Query<Paginated>,
+    _: Require<ReadSbom>,
+) -> actix_web::Result<impl Responder> {
+    let tx = db.begin_read().await?;
+    let result = fetch
+        .fetch_sbom_models(None, search, paginated, &tx)
+        .await?;
     Ok(HttpResponse::Ok().json(result))
 }
 
--- a/modules/fundamental/src/sbom/endpoints/test.rs
+++ b/modules/fundamental/src/sbom/endpoints/test.rs
@@ -1860,6 +1860,54 @@ async fn query_aibom_models(
 
 #[test_context(TrustifyContext)]
 #[rstest]
+#[case("hugging", 2)]
+#[case("granite", 1)]
+#[case("pkg:huggingface/ibm-granite", 1)]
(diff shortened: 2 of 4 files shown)


ctron (commented):
I think the AI is right. So I think it makes sense:

* add a test variant without `q`
* not only test for the number of results, but for the actual models returned (which implies the number then)
* also: add some tests to the scale-testing repository using this new endpoint. Not only with one case, but using a few with `q` (and without)

jcrossley3:
> I think the AI is right. So I think it makes sense:
> 
> * add a test variant without `q`
> * not only test for the number of results, but for the actual models returned (which implies the number then)
> * also: add some tests to the scale-testing repository using this new endpoint. Not only with one case, but using a few with `q` (and without)

done

ctron:
> > I think the AI is right. So I think it makes sense:
> > 
> > * add a test variant without `q`
> > * not only test for the number of results, but for the actual models returned (which implies the number then)
> > * also: add some tests to the scale-testing repository using this new endpoint. Not only with one case, but using a few with `q` (and without)
> 
> done

Sorry I can't find the new tests in the scale-testing repository. Could you please point me towards them.

Also (not a blocker), I'm wondering how the user can get a reference on the SBOM with the search results of the models? There seems to be no "sbom ID" as part of the response.

carlosthe19916 (commented):
@jcrossley3 sorry for my late review. Looks good; only one thing missing

The mockups (image below) contains the column SBOMs which is supposed to count the number of SBOMs that contain a particular AI Model. How can I filter SBOMs based on a Model?

Image

jcrossley3:
> Sorry I can't find the new tests in the scale-testing repository. Could you please point me towards them.

I created an issue in that repo: https://github.com/guacsec/trustify-scale-testing/issues/99

This would have to me merged before work can proceed on that, right?

jcrossley3:
> The mockups (image below) contains the column SBOMs which is supposed to count the number of SBOMs that contain a particular AI Model. How can I filter SBOMs based on a Model?

Hmmm. I totally missed that. I assume your question is because that's what's required in response to a user clicking the "sbom count" link?

carlosthe19916:
> > The mockups (image below) contains the column SBOMs which is supposed to count the number of SBOMs that contain a particular AI Model. How can I filter SBOMs based on a Model?
> 
> Hmmm. I totally missed that. I assume your question is because that's what's required in response to a user clicking the "sbom count" link?

Following the Licenses Page Pattern, if the user clicks on the "SBOM Count" then he would be redirected to the SBOM List page with a pre-defined filter that renders only those SBOMs.

Watch the video below of the licenses.

In the UI side:
- The SBOM List page needs a new filter: Filter by AI Model
- The Model List page needs a column with "sbom count"


https://github.com/user-attachments/assets/63c0e10b-5b84-40cf-a91b-fd19305c2f6b

carlosthe19916:
@jcrossley3 If it help I think we can tackle the "sbom count" in a separate PR so the UI takes this one and move things forward. If you want to do everything here it is also fine, whatever generates less problems for you

jcrossley3:
> @jcrossley3 If it help I think we can tackle the "sbom count" in a separate PR so the UI takes this one and move things forward. If you want to do everything here it is also fine, whatever generates less problems for you

Yes, please. It's not a trivial change, so deserving of its own PR, I think.

carlosthe19916 (approved):
LGTM, I tested it locally and there is this PR on the UI side that renders the models https://github.com/guacsec/trustify-ui/pull/995

As agreed with @jcrossley3 the column SBOM Count can be done in a separate PR

ctron:
> > Sorry I can't find the new tests in the scale-testing repository. Could you please point me towards them.
> 
> I created an issue in that repo: [guacsec/trustify-scale-testing#99](https://github.com/guacsec/trustify-scale-testing/issues/99)
> 
> This would have to me merged before work can proceed on that, right?

I'd say it can be run in parallel. And you should be able to run a test locally, just to show some numbers.

PR #2327, opened 2026-04-14
test: add a reproducer for assigning groups concurrently

ctron:
Also see TC-4012

There actually is not race condition. I believe that claude is wrong. There is however an mismatch in expectations.

The "update" call "updates" the groups assignments. Not "amending" them. So the result of assigning "group A" to "SBOM 1" is that "SBOM 1 is assigned to group A only".

In the case of a concurrent update, only one must win.

## Summary by Sourcery

Add a regression test to verify SBOM group assignment replace semantics under concurrent updates.

Enhancements:
- Introduce a reusable helper for assigning SBOM groups in tests and a utility to resolve multiple group IDs by name paths.

Tests:
- Add a test covering sequential and parallel SBOM group assignments to ensure only one group remains after concurrent updates.

### DIFF
--- a/modules/fundamental/src/common/test.rs
+++ b/modules/fundamental/src/common/test.rs
@@ -333,6 +333,16 @@ pub fn locate_id(
         .clone()
 }
 
+/// Look up multiple group IDs by their name paths at once.
+pub fn locate_ids<I, P, S>(ids: &HashMap<Vec<String>, String>, paths: I) -> Vec<String>
+where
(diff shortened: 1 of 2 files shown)


mrrajan:
@ctron Thanks for the clarification and I Agree on having differences in the expectation. I was under the assumption that the group assignment would append/amend to the existing assignments, But it is replacing. I will check Michelle before closing the JIRA.

mrrajan (approved):
LGTM

PR #2329, opened 2026-04-15
feat(server): add OIDC_LOAD_USER configuration support

mrrajan:
Implements TC-4108

## Summary by Sourcery

New Features:
- Introduce an OIDC_LOAD_USER configuration option to toggle loading of OIDC user information.


ctron (changes requested):
I think the AI is right, this should be a boolean. Also, I think the name should be aligned with the other parameters in this struct. 

And, to my understanding, this is a field coming from the `trustify_ui` crate. If it's not possible to change that there as well, then at least here we should use a proper type.

Please also update the env-var list: https://github.com/guacsec/trustify/blob/main/docs/env-vars.md

ctron (approved):

PR #2337, opened 2026-04-20
Add a read-only switch

ctron:
Also see: https://redhat.atlassian.net/browse/TC-4162
Also see: https://github.com/guacsec/trustify-ui/pull/1008

## Summary by Sourcery

Documentation:
- Add an ADR describing the environment flag, middleware guard, importer suspension, and status endpoint changes for read-only mode.

## Summary by Sourcery

Introduce a configurable read-only mode that blocks mutating operations while keeping read operations and status introspection available.

New Features:
- Add a TRUSTD_READ_ONLY / --read-only flag to API and importer processes to enable read-only mode across the system.
- Expose instance read-only status via the /.well-known/trustify endpoint with a new readOnly field for UI and client adaptation.
- Introduce shared middleware and standard app wiring to guard mutating HTTP methods when read-only mode is active.

Enhancements:
- Suspend importer job execution while preserving process liveness when running in read-only mode.
- Refactor test app helpers and extend integration tests to cover read-only guarding behavior and the status endpoint.
- Adjust existing ADRs and add a new ADR documenting the design and behavior of the read-only switch.
- Tidy documentation examples and image paths in existing ADRs for consistency.

Documentation:
- Document the TRUSTD_READ_ONLY environment variable and its behavior in the environment variables reference.
- Add an ADR describing the read-only mode switch, middleware guarding, importer suspension, and status endpoint.

Tests:
- Add integration tests to verify that read-only mode blocks mutating API methods, preserves read operations, and surfaces mode in the well-known endpoint.

## Summary by Sourcery

Introduce a configurable read-only mode that prevents mutating operations while keeping read operations and status introspection available across the API and importer services.

New Features:
- Add a TRUSTD_READ_ONLY / --read-only flag for API and importer processes to enable read-only mode.
- Expose instance read-only status via the /.well-known/trustify endpoint through a new readOnly field.
- Introduce shared middleware and standard app wiring to guard mutating HTTP methods when read-only mode is active.

Enhancements:
- Suspend importer job execution while preserving process liveness when running in read-only mode.
- Refine shared test helpers and add coverage for read-only guarding behavior and the status endpoint.
- Update the OpenAPI schema and infrastructure wiring to include read-only state in application configuration.
- Adjust ADRs and documentation assets for consistency, including image paths and JSON examples.

Documentation:
- Add an ADR documenting the design and behavior of the read-only mode switch, middleware guarding, importer suspension, and status endpoint.
- Document the TRUSTD_READ_ONLY environment variable and its behavior in the environment variable reference.

Tests:
- Add integration tests to verify that read-only mode blocks mutating API methods, preserves read operations, and surfaces the mode via the well-known endpoint.


ctron:
@sourcery-ai review

ctron:
@sourcery-ai review

ctron:
> * The `ReadOnlyState` extractor silently defaulting to `false` when no app data is registered can easily mask misconfiguration; consider either failing loudly (e.g. panic/log error in debug) or requiring explicit registration so the guard is never accidentally disabled in new apps/tests.

This falls back to the old behavior. And in the tests we do test it. The application itself sets it too. And this is not considered a library. So I think it's ok.

> * The `read_only_guard` currently allows any method other than `GET`/`HEAD`/`OPTIONS` when not in read-only mode and blocks everything else when enabled; it may be clearer and safer to explicitly guard only well-known mutating methods (`POST`, `PUT`, `PATCH`, `DELETE`) so that unexpected methods like `TRACE` or `CONNECT` don't change behavior in the future.

I think it's better to actively allow new HTTP methods, which wouldn't happen that often anyway. Rather than accidentally accepting them.

helio-frota:
Will this be propagated to Helm Charts?

ctron:
> Will this be propagated to Helm Charts?

It should be.

ctron:
This also has to be backported, as we would require this switch in a 2.x to 3.0 migration to be present in the 2.x instance.

jcrossley3 (approved):
LGTM, but I defer to whoever asked for this feature.

PhilipCattanach:
@ctron - I can't see any indication in the comments that that the backport has been applied to the 0.4.Z

PR #2343, opened 2026-04-29
fix: improve performance of advisory list

ctron:
The left join can be replaced with an inner join, as the source_document_id is never null and always matches a record.

This allows to speed up postgres doing the most frequent searches, especially the initial dashboard search.

This allows postgres to optimize the call, speeding it up from ~3.2 seconds to ~0.6 seconds on DS4.

## Summary by Sourcery

Enhancements:
- Replace left joins with inner/inner-typed joins for advisory-source_document and sbom-source_document queries to improve query performance in common paths.

### DIFF
--- a/modules/fundamental/src/advisory/service/mod.rs
+++ b/modules/fundamental/src/advisory/service/mod.rs
@@ -42,7 +42,7 @@ impl AdvisoryService {
     ) -> Result<PaginatedResults<AdvisorySummary>, Error> {
         let limiter = advisory::Entity::find()
             .with_deprecation(deprecation)
-            .left_join(source_document::Entity)
+            .inner_join(source_document::Entity)
             .join(JoinType::LeftJoin, advisory::Relation::Issuer.def())
             .filtering_with(
                 search,
@@ -78,7 +78,7 @@ impl AdvisoryService {
         connection: &C,
     ) -> Result<Option<AdvisoryDetails>, Error> {
         let results = advisory::Entity::find()
-            .left_join(source_document::Entity)
+            .inner_join(source_document::Entity)
             .join(JoinType::LeftJoin, advisory::Relation::Issuer.def())
             .try_filter(id)?
             .try_into_multi_model::<AdvisoryCatcher>()?
--- a/modules/fundamental/src/license/service/mod.rs
+++ b/modules/fundamental/src/license/service/mod.rs
@@ -249,7 +249,7 @@ impl LicenseService {
     ) -> Result<Option<Vec<LicenseRefMapping>>, Error> {
         // check the SBOM exists searching by the provided Id
         let sbom = sbom::Entity::find()
-            .join(JoinType::LeftJoin, sbom::Relation::SourceDocument.def())
+            .join(JoinType::InnerJoin, sbom::Relation::SourceDocument.def())
             .try_filter(id)?
             .one(connection)
             .await?;
--- a/modules/ingestor/src/graph/sbom/mod.rs
+++ b/modules/ingestor/src/graph/sbom/mod.rs
@@ -79,7 +79,7 @@ impl Graph {
         connection: &C,
     ) -> Result<Option<SbomContext>, Error> {
         Ok(sbom::Entity::find()
-            .join(JoinType::LeftJoin, sbom::Relation::SourceDocument.def())
+            .join(JoinType::InnerJoin, sbom::Relation::SourceDocument.def())
             .filter(
                 Condition::any()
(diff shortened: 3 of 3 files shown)


helio-frota:
I observed a performance degradation in a basic test.
I am using `pm-mode` (without the `--release` flag) with the load tests dump.
Here are the results:

ctron:
> I observed a performance degradation in a basic test. I am using `pm-mode` (without the `--release` flag) with the load tests dump. Here are the results:
> 

Interesting, as scale-tests show the same behavior. What is `advisory-list.js` doing?

ctron:
So, checking this on Linux (was MacOS) I can see it's slower too.

ctron:
Ok, it seems that the inner join improves the the actual select, but degrades the "select count" before.

ctron:
After digging into this, I got misled by AI. … The inner join may (or may not) help the actual statement. The optimizer can actually gather statistics and use the same query plan for both inner and left join.

However, with the "count", the information from the join is not required and the "left join" can't reduce the number of entries. So, the join can be removed and Postgres can do this. However, Postgres cannot do this with inner joins, as it would need to prove that the entry would always be there, which Postgres doesn't support.

So yea, the original problem is "fixed", but by creating a new, worse one.

helio-frota:
> What is advisory-list.js doing?

@ctron The script is accessing the endpoint in a specific way where we can configure users and duration:

we can use this for others endpoints with `GET` just changing this `const URL = "http://localhost:8080/api/v3/advisory";`

```js
const USERS = parseInt(process.argv[2], 10) || 20;
const DURATION_SECONDS = parseInt(process.argv[3], 10) || 30;

const URL = "http://localhost:8080/api/v3/advisory";

const END_TIME = Date.now() + DURATION_SECONDS * 1000;

const MAX_RETRIES = 3;
const BASE_BACKOFF_MS = 200;

const responseTimes = [];

function median(arr) {
    if (arr.length === 0) return 0;
    const sorted = [...arr].sort((a, b) => a - b);
    const mid = Math.floor(sorted.length / 2);
    return sorted.length % 2 !== 0
        ? sorted[mid]
        : (sorted[mid - 1] + sorted[mid]) / 2;
}

function sleep(ms) {
    return new Promise((resolve) => setTimeout(resolve, ms));
}

async function callApi(userId, requestId) {
    let attempt = 0;

    while (attempt <= MAX_RETRIES) {
        attempt++;

        try {
            const start = Date.now();
            const res = await fetch(URL);
            const duration = Date.now() - start;

            if (res.status === 408) {
                if (attempt <= MAX_RETRIES) {
                    const backoff = BASE_BACKOFF_MS * attempt;
                    console.warn(
                        `[User ${userId}] request ${requestId} timeout (${res.status}), retry ${attempt}/${MAX_RETRIES} after ${backoff}ms (${duration}ms)`,
                    );
                    await sleep(backoff);
                    continue;
                }

                console.error(
                    `[User ${userId}] request ${requestId} failed after retries (${res.status}) (${duration}ms)`,
                );
                responseTimes.push(duration);
                return;
            }

            if (!res.ok) {
                console.error(
                    `[User ${userId}] request ${requestId} failed with status ${res.status} (${duration}ms)`,
                );
                responseTimes.push(duration);
                return;
            }

            // console.log(
            //   `[User ${userId}] request ${requestId} succeeded (${duration}ms)`
            // );
            responseTimes.push(duration);
            return;
        } catch (err) {
            if (attempt <= MAX_RETRIES) {
                const backoff = BASE_BACKOFF_MS * attempt;
                console.warn(
                    `[User ${userId}] request ${requestId} network error, retry ${attempt}/${MAX_RETRIES} after ${backoff}ms`,
                );
                await sleep(backoff);
            } else {
                console.error(
                    `[User ${userId}] request ${requestId} failed after retries: ${err.message}`,
                );
                return;
            }
        }
    }
}

async function userLoop(userId) {
    let requestId = 0;

    while (Date.now() < END_TIME) {
        requestId++;
        await callApi(userId, requestId);
    }

    //console.log(`[User ${userId}] finished`);
}

async function main() {
    console.log("Starting load test");
    console.log(`Concurrent users: ${USERS}`);
    console.log(`Duration: ${DURATION_SECONDS} seconds`);
    console.log("");

    const users = [];

    for (let i = 1; i <= USERS; i++) {
        users.push(userLoop(i));
    }

    await Promise.all(users);

    const medianTime = median(responseTimes);
    console.log("");
    console.log("finished");
    console.log(`Total requests: ${responseTimes.length}`);
    console.log(`Median response time: ${medianTime}ms`);
}

main();
```

PR #2344, opened 2026-05-01
chore: workaround to make some tests to pass

helio-frota:
* when using linux + btrfs without long_running feature or --all-features enabled.
* Related to https://github.com/guacsec/trustify/issues/2340

## Summary by Sourcery

Add a configuration flag to optionally disable Btrfs-based snapshots in test migrations and fall back to temporary directories to ensure tests pass in non-long_running environments.

New Features:
- Allow tests to use a non-Btrfs TempDir-based snapshot mechanism when Btrfs snapshots are not requested.

Enhancements:
- Wire the new snapshot configuration flag through the migration test context to control whether Btrfs or TempDir is used.

### DIFF
--- a/test-context/src/ctx/migration/mod.rs
+++ b/test-context/src/ctx/migration/mod.rs
@@ -167,6 +167,7 @@ impl<ID: DumpId> TrustifyMigrationContext<ID> {
                     snapshot_file: None,
                     strip: 0,
                     fix_zstd: false,
+                    use_btrfs: false,
                 }
             }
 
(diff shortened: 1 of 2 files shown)


ctron:
As mentioned in the chat, it looks like the issue is that we bail out of all btrfs processing for all requirements except one, the mount flag. If that's the case, then we should add a check for this too, and just bail out in that case.

helio-frota:
closing in favor of https://github.com/guacsec/trustify/pull/2345

PR #2348, opened 2026-05-05
docs: refresh documentation

ctron:
## Summary by Sourcery

Update design documentation diagrams to reflect new CPE and PURL status modeling and expanded SBOM node capabilities.

Documentation:
- Document CPE as a first-class entity and introduce separate ProductStatus and PurlStatus relationships in the product design diagram.
- Clarify that SBOM nodes can represent additional SBOM sub-parts and adjust SBOM naming relationships from packages to generic nodes.

### DIFF
--- a/docs/design/sbom.md
+++ b/docs/design/sbom.md
@@ -16,7 +16,7 @@ An SBOM contains a list of nodes, which are connected by relationships (like A d
 (e.g. SPDX), this may include references to the document itself (A describes SBOM), or to external documents
 (SBOM-A amends SBOM-B) and nodes in external documents (A depends on SBOM-B/C).
 
-A node can be either the SBOM itself, packages (aka components), or files.
+A node can be either the SBOM itself, packages (aka components), files, or other parts of the SBOM.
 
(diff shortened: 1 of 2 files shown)


docs/design/sbom.md
```diff
@@ -16,7 +16,7 @@ An SBOM contains a list of nodes, which are connected by relationships (like A d
 (e.g. SPDX), this may include references to the document itself (A describes SBOM), or to external documents
 (SBOM-A amends SBOM-B) and nodes in external documents (A depends on SBOM-B/C).
 
-A node can be either the SBOM itself, packages (aka components), or files.
+A node can be the SBOM itself, a package (aka component), a file, or another part of the SBOM.
```
jcrossley3, line 19: Terminology is hard. I feel like if we had focused on CycloneDX rather than SPDX when first designing our data model, we would've used the term "component" rather than "package". CDX has 13 different types of component, including `file` and at least 3 that could be considered a "package", e.g. `application`, `library`, `framework`, and yet SPDX has only 2: `package` and `file`.

I haven't looked in depth yet, but I'm betting the new "profiles" in SPDX 3 were introduced to address that.

jcrossley3 (approved):

PR #2349, opened 2026-05-05
Add a count of sboms to the models response

### LINKED ISSUE #2332: Provide SBOM counts on the AI models summary page
@jcrossley3 sorry for my late review. Looks good; only one thing missing

The mockups (image below) contains the column SBOMs which is supposed to count the number of SBOMs that contain a particular AI Model. How can I filter SBOMs based on a Model?

Image

_Originally posted by @carlosthe19916 in https://github.com/guacsec/trustify/pull/2325#pullrequestreview-4112616218_

jcrossley3:
Fixes: #2332

## Summary by Sourcery

Add an SBOM usage count to model responses and propagate it through the API and tests.

New Features:
- Include the number of SBOMs containing each model in the SBOM models API response.

Enhancements:
- Batch compute SBOM counts for model PURLs to avoid per-item queries.

Documentation:
- Document the sbom_count field in the OpenAPI schema for SBOM model responses.

Tests:
- Extend SBOM endpoint tests to assert the new sbom_count field in model responses.

### DIFF
--- a/modules/fundamental/src/sbom/service/sbom.rs
+++ b/modules/fundamental/src/sbom/service/sbom.rs
@@ -478,9 +478,63 @@ impl SbomService {
             &self.cache,
         );
 
-        let LimitedResult { items, total } = limiter.fetch().await?;
-        let total = total.requested(paginated.total).await?;
-        let items = items.into_iter().map(SbomModel::from_row).collect();
+        let total = limiter.total().await?;
+        let model_catchers = limiter.fetch().await?;
+
+        // Parse PURLs once per model and collect all unique PURL UUIDs
+        let parsed: Vec<(String, String, serde_json::Value, Vec<PurlSummary>)> = model_catchers
+            .into_iter()
+            .map(|c| {
+                let purls = PurlSummary::from_values(c.purls);
+                (c.id, c.name, c.properties, purls)
+            })
+            .collect();
+
+        let mut all_purl_uuids = Vec::new();
+        for (.., purls) in &parsed {
+            all_purl_uuids.extend(purls.iter().map(|p| p.head.uuid));
+        }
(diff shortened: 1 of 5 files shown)


modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -478,11 +478,61 @@ impl SbomService {
...
+        } else {
+            sbom_node_purl_ref::Entity::find()
+                .select_only()
+                .column(sbom_node_purl_ref::Column::QualifiedPurlId)
+                .column_as(sbom_node_purl_ref::Column::SbomId.count(), "count")
+                .filter(sbom_node_purl_ref::Column::QualifiedPurlId.is_in(all_purl_uuids))
+                .group_by(sbom_node_purl_ref::Column::QualifiedPurlId)
+                .into_tuple::<(Uuid, i64)>()
+                .all(connection)
+                .await?
```
ctron, line 515: Please add instrumentation.

ctron (changes requested):
Instead of adding this information, which to my understanding causes additional table joins, to all responses, we are turning towards a model which makes such information optional, and have the caller request it to not slow down operations for users which are not interested in it.

There also is support for fields like this: https://github.com/guacsec/trustify/blob/main/common/src/requested_field.rs

modules/fundamental/src/sbom/service/sbom.rs
jcrossley3: done

PhilipCattanach:
@jcrossley3 - Hi Jim, 
Can you resolve the conflicts so we can get this change cleanly merged please?

ctron (approved):
Looks good to me. Thanks. Maybe AI is right and we can increase test coverage. Not a blocker for me.

PR #2356, opened 2026-05-08
feat: implement the read-only database connection

ctron:
Implement a second DB connection which is only used for read-only operations. Defaulting that connection to the primary (read-write) one.

The idea is to configure the system to use read-only replicas, scaling up more pods. But using the R/W connection when required.

Also see: https://redhat.atlassian.net/browse/TC-3439

## Summary by Sourcery

Introduce separate read-write and read-only database connection wrappers and wire the application to use them across modules, enabling optional use of a dedicated read-only database while enforcing read-only access at the transaction level.

New Features:
- Add ReadWrite and ReadOnly database connection wrapper types exposing appropriate SeaORM traits for mutation and read-only use cases.
- Introduce separate read-only database configuration (CLI flags and environment variables) that can override or fall back to the primary database settings.
- Provide module- and endpoint-level wiring to inject read-only or read-write connections explicitly via actix-web data, including for analysis, importer, ingestor, user, and fundamental modules.

Enhancements:
- Enforce read-only access by opening read-only transactions from ReadOnly connections and mapping database errors into module error types with appropriate HTTP responses.
- Refactor ingestion, SBOM, advisory, product, SBOM group, and dataset write paths to use explicit read-write transactions with manual begin/commit instead of generic transaction helpers.
- Adjust OpenAPI generation and tests to construct and use both read-write and read-only database wrappers for better parity with runtime configuration.

Documentation:
- Add ADR 00018 documenting the design and rationale for dual database connections (read-write primary plus optional read-only replica) and their configuration.

Tests:
- Update module and integration tests to use the new ReadWrite and ReadOnly wrappers instead of the raw Database where appropriate.

Chores:
- Propagate new database wrappers through server startup configuration and module configure() signatures without changing the external HTTP API surface.


queria:
I think You may want also update the guidance in CONVENTIONS.md:
```
## Endpoint Patterns
...
- Read operations acquire a read transaction: `let tx = db.begin_read().await?;`
```
and maybe hint there when ReadOnly vs ReadWrite should be used (although that may be obvious to people or AI enough).

queria:
Also seems to me that with this, the `begin_read` and so `DatabaseExt` may be unecessary?

docs/adrs/00018-dual-database-connections.md
```diff
@@ -0,0 +1,207 @@
...
+
+The two features compose naturally:
+
+| Deployment | R/W connection | R/O connection | `--read-only` |
+|------------|---------------|---------------|---------------|
+| Single instance, full mode | Primary | Primary (fallback) | `false` |
+| Single instance, read-only | Not needed | Primary | `true` |
+| Multi-instance, mixed | Primary | Replica | `false` |
+| Multi-instance, read-only replica | Not needed | Replica | `true` |
+
```
queria, line 139: Is the NO-RW mode actually supported, i may not be understanding it right, but seems to me that ReadWrite db is currently not optional, is it?

server/src/profile/api.rs
queria, line 254: Database instance is not conditional - at least not conditional like `db_ro  = database_ro.is_configured() { ... }` below.

server/src/profile/api.rs
```diff
@@ -252,6 +257,15 @@ impl InitData {
...
         }
 
+        let db_ro = if run.database_ro.is_configured() {
+            let ro_config = run.database_ro.to_database_config(&run.database);
+            let ro_db = db::Database::new(&ro_config).await?;
+            db::ReadOnly::new(ro_db)
+        } else {
+            db::ReadOnly::new(db.clone())
+        };
+        let db_rw = db::ReadWrite::new(db.clone());
```
queria, line 262: And it's then always cloned and below passed to InitData.

queria (commented):
I am not sure if the mode without R/W DB configuration, as mentioned in ADR, is possible currently?

helio-frota:
> Also seems to me that with this, the begin_read and so DatabaseExt may be unecessary?

yeah seems unused:

```
➜  trustify git:(feature/ro_instance_1) rg DatabaseExt
modules/fundamental/src/db.rs
5:pub trait DatabaseExt {
17:impl<T> DatabaseExt for T
```

when runnin it with pm-mode, I'm not 100% following why I'm logged as 'Marvin' and it shows extra parts that I can see only when using SSO




**UPDATE**: nvm ^^^ I forgot auth-disabled env var ...

docs/adrs/00018-dual-database-connections.md
```diff
@@ -0,0 +1,207 @@
...
+Each module's `configure()` function signature changes to accept the connection type it needs:
+
+* **Read-only modules** (fundamental queries, analysis, user preferences reads) receive `ReadOnly`.
+* **Read-write modules** (ingestor, importer) receive `ReadWrite`.
+* **Mixed modules** (e.g. fundamental, which has both query and mutation endpoints) receive both.
+
+The `server/src/profile/api.rs` `configure()` function constructs both wrapper types and passes them
+to each module accordingly.
+
+## Alternatives considered
```
jcrossley3, line 161: Was [PGPool-II ](https://www.pgpool.net/) considered?

jcrossley3 (commented):
Why is application-level DB routing required? What existing problem is it solving? And is the configuration complexity trade-off worth it? Whoever will be responsible for that configuration should review this PR.

docs/adrs/00018-dual-database-connections.md
helio-frota: I'm not understanding this table part as well :+1: 
I checked this behavior and it worked:

helio-frota:
other than git conflict, overall lgtm

docs/adrs/00018-dual-database-connections.md
ctron: No, as we would need to build, productize and support it. It would also be required to incorporated into deployments like RDS.

RDS on the other side offers read-only replicas out of the box.

docs/adrs/00018-dual-database-connections.md
ctron: I agree, that table is confusing. I'll re-iterate over it.

server/src/profile/api.rs
ctron: It is used later. But, passing it on to the Analysis service, that should be the readonly connection.

ctron:
> Why is application-level DB routing required? What existing problem is it solving?

I updated the ADR to explain this better.

> And is the configuration complexity trade-off worth it?

Yes. As all we're talking about is adding a new set of DB connection parameters *only if* this should be enabled.

helio-frota:
it was cool to see working!

```
| Primary + replica, full mode | Primary | Replica | `false` | Reads go to replica, writes to primary |
```

I ran the first time using the env-var pointing to the primary and then changed later to the replica and refreshed the dashboard N times:

primary:



replica:

helio-frota (approved):

PR #2358, opened 2026-05-14
Added ADR for architectural standards

mrrajan:
### Summary
- Add ADR-00018 proposing a CONVENTIONS.md file and documenting pattern standards for the Trustify project
- Catalogs recurring anti-patterns found across the codebase (N+1 queries, unbounded collections, sequential independent queries, missing database indexes, uncontrolled pagination, redundant deserialization) with specific file locations and severity ratings
- Presents convention options for each anti-pattern category for the team to evaluate during PR review

Rendered version https://github.com/mrrajan/trustify/blob/TC-4289/docs/adrs/00018-conventions-file.md

## Summary by Sourcery

Introduce an ADR proposing a centralized CONVENTIONS.md file and cataloging key performance and coding anti‑patterns in the Trustify codebase, along with candidate standards for addressing them.

Documentation:
- Add ADR-00017 documenting the rationale, scope, and maintenance process for a repository-wide CONVENTIONS.md conventions file used by contributors and AI tools.
- Document a detailed analysis of recurring performance and coding anti-patterns across the codebase, including their locations and proposed convention options for remediation.


PhilipCattanach:
Thanks @mrrajan Very interesting reading
I would suggest choosing the appropriate convention for some of the performance issue will probably need tests at scale. I'm concerned that some of the potential solutions might make things worse at scale.

I'd like us to come up with a way of prioritizing our next steps. As I would like to think there is some non contentious conventions we can adopt that will address some of these things. 

And we might need some Spikes to evaluate the best convention for some of the issues highlighted.

docs/adrs/00019-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| **Migration Patterns** | Database migration conventions | Idempotency guards, naming, raw SQL loading |
+| **Rust Idioms** | Preferred Rust patterns | Type inference, iterator ownership, `.zip()`, capacity |
+| **SeaORM Query Patterns** | ORM query conventions | `.is_in()`, chunking, `IntoIterator` parameters |
+| **Observability** | Tracing and instrumentation | `#[instrument]` usage, span conventions, error levels |
+| **Shared Table Patterns** | Concurrent insert handling | Nested transaction duplicate-key and `.on_conflict` patterns |
+
+### Content Principles
+
+1. **Prescriptive, not descriptive**: each convention states what to do and what to avoid, with concrete code examples
+2. **Derived from existing code**: conventions are extracted from established patterns in the codebase, not invented
```
mrizzi, line 59: Maybe the initial version but not once stable because, once fully adopted, it should be other way around, i.e. `code should be derived from existing conventions`

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+
+- **Better AI output**: AI tools generate code that matches project style from the first attempt
+- **Onboarding**: new contributors can read one file to understand "how we write code here"
+- **Consistency**: single source of truth reduces convention drift across modules and contributors
+- **Faster reviews**: reviewers can reference specific convention sections instead of explaining patterns from scratch
+- **Accountability**: convention changes are tracked in git history with review
+
+### Trade-offs
+
+- **Maintenance burden**: the file must be kept in sync with evolving practices — stale conventions are worse than no conventions
```
mrizzi: I don't think this exists because, as described in the above `Maintenance Process`, when a convention has to change, then an ADR arrives. Once merged, the impacted code has to refactored to be aligned with the new code convention. I don't see how "stale conventions" could happen other than implementing code not aligned with the conventions, which should be considered a mistake to prevent.

mrizzi (commented):
A first review of the methodology before getting into the anti-patterns analysis with the maintainers.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+codified in `CONVENTIONS.md` and enforced through code review and AI-assisted development
+workflows.
+
+Each category below follows this structure:
+- **What it is** — description of the anti-pattern
+- **Why it matters** — performance impact
+- **Occurrences found** — specific locations in the codebase
+- **Convention options** — choices for the team to decide on (to be resolved via PR review)
+
+> **How to use this section**: During PR review, maintainers should mark their preferred
```
mrizzi: @jcrossley3 @ctron @ruromero @queria please review this

ctron:
I'm not sure an ADR is the right approach to this. We already have the conventions file, and this ADR would need to be copied over to that file in a second step.

Why no simply make the modifications to the conventions file and discuss this on a PR? Because what would we review on that second PR (moving content over to the conventions file)? Feels like doppelmoppel.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+
+Date: 2026-05-14
+
+## Status
+
+PROPOSED
+
+## Context
+
+Trustify uses AI-assisted development workflows (Claude Code, Copilot, and similar tools). These tools perform best when they have access to explicit, machine-readable project conventions — coding patterns, naming rules, error-handling idioms, testing practices, and architectural norms.
```
ctron: As we do have some documentation (for less then we should have) for humans. I believe that this file should be usable by humans as well. Replacing the existing documents. Which means, that this file should be structured and written to serve both.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| **Error Handling** | Error type design and propagation | `thiserror` enums, `ResponseError` mapping, `From<DbErr>` |
+| **Testing Conventions** | Test infrastructure and patterns | `TrustifyContext`, test placement, assertion style |
+| **Commit Messages** | Commit format and trailers | Conventional Commits, Jira references |
+| **Pre-commit Workflow** | CI-equivalent local checks | `cargo xtask precommit` steps |
+| **Dependencies** | Dependency management policy | Workspace-level pinning, key crate choices |
+| **Endpoint Patterns** | HTTP endpoint conventions | `configure()`, authorization, transactions, OpenAPI |
+| **Entity Model Patterns** | ORM model conventions | `DeriveEntityModel`, relations, `Linked` structs |
+| **Migration Patterns** | Database migration conventions | Idempotency guards, naming, raw SQL loading |
+| **Rust Idioms** | Preferred Rust patterns | Type inference, iterator ownership, `.zip()`, capacity |
+| **SeaORM Query Patterns** | ORM query conventions | `.is_in()`, chunking, `IntoIterator` parameters |
```
ctron: `IntoIterator` wouldn't be a SeaORM thing, but a Rust thing.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| **Observability** | Tracing and instrumentation | `#[instrument]` usage, span conventions, error levels |
+| **Shared Table Patterns** | Concurrent insert handling | Nested transaction duplicate-key and `.on_conflict` patterns |
```
ctron: Not sure I understand the scope of this section. I'd prefer to have some "database" section maybe. Defining how database resources/entities are created, column types, tables, enums, indexes, ...

docs/adrs/00019-conventions-file.md
ctron: Question is, how do we deal with current code that doesn't align to those newly defined patterns?

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+
+### Content Principles
+
+1. **Prescriptive, not descriptive**: each convention states what to do and what to avoid, with concrete code examples
+2. **Derived from existing code**: conventions are extracted from established patterns in the codebase, not invented
+3. **Minimal and actionable**: each entry should be short enough that a contributor (or AI tool) can apply it without reading surrounding prose
+4. **Reference implementations**: point to specific files in the codebase as canonical examples
```
ctron: I think that may be tricky. As those code locations will change over time. And that would mean updating the conventions file, or dead links.

docs/adrs/00019-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+- **Deprecation**: when a convention is superseded, update the section rather than appending contradictory guidance
+- **Scope creep guard**: the file documents *conventions* (how to write code), not *architecture* (why the system is designed this way) — architecture belongs in ADRs
+
+### AI Tool Integration
+
+The conventions file is designed to be consumed by AI coding assistants:
+- Claude Code automatically loads `CONVENTIONS.md` from the repository root as part of its project context
+- The file uses markdown with code blocks, making it parseable by any LLM
+- Conventions are structured as clear rules with examples, optimizing for AI instruction-following
+- When a `CLAUDE.md` file is present (for tool-specific configuration), `CONVENTIONS.md` complements it — `CONVENTIONS.md` focuses on language and framework patterns that apply regardless of the tool
```
ctron, line 77: I wonder how that aligns with review tools? like sourcery.ai?

docs/adrs/00018-conventions-file.md
ctron: To my understanding, feedback from reviews might turn in PR changing the conventions file? If that's the case, then this scenario may happen. Unless of course, a change in the conventions file is accompanied by reworking the code base.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+- **Maintenance burden**: the file must be kept in sync with evolving practices — stale conventions are worse than no conventions
+- **Completeness tension**: too few conventions and the file is unhelpful; too many and it becomes noise that contributors (and AI tools) ignore
+- **Convention vs. enforcement gap**: not all conventions can be enforced by CI — some rely on review discipline
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|------|------------|
+| File becomes outdated | Treat convention violations in review as a signal to update the file |
+| File grows too large for AI context windows | Keep entries concise; split into linked files if needed |
```
ctron: I think that's not only a concern, but for humans alike.

docs/adrs/00019-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|------|------------|
+| File becomes outdated | Treat convention violations in review as a signal to update the file |
+| File grows too large for AI context windows | Keep entries concise; split into linked files if needed |
+| Conventions conflict with each other | PR review process catches contradictions before merge |
+| Over-prescription stifles judgment calls | Focus on patterns with clear consensus; leave room for discretion |
+
```
ctron, line 103: One risk I see is the lack of possibility to deviate from the rules when it's beneficial to the system. Maybe that's what the last point meant.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| Conventions conflict with each other | PR review process catches contradictions before merge |
+| Over-prescription stifles judgment calls | Focus on patterns with clear consensus; leave room for discretion |
+
+## Anti-Pattern Analysis
+
+Analysis of the Trustify codebase has identified recurring anti-patterns across
+multiple modules. This section catalogs each category, lists the occurrences found, and
```
ctron: I see the same problem as before. If we "point" to those locations, and those locations will be gone over time (which should be the case with anti-patterns) we have a document of dead links.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+presents convention options for the team to evaluate. The selected conventions will be
+codified in `CONVENTIONS.md` and enforced through code review and AI-assisted development
+workflows.
+
+Each category below follows this structure:
+- **What it is** — description of the anti-pattern
+- **Why it matters** — performance impact
```
ctron: I assume "performance impact" is an example. But in general, "why" can be all kind of reasons.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+- **Occurrences found** — specific locations in the codebase
```
ctron: See my comment about links above.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+> added to `CONVENTIONS.md` and this ADR's status will change from PROPOSED to ACCEPTED.
+
+---
+
+### AP-1: N+1 Query Anti-pattern
+
+**What it is**: Database queries executed inside loops — fetching related data one entity
+at a time instead of loading all related data in a single batch query.
+
+**Why it matters**: For a list of N entities, this generates N additional queries instead of
```
ctron: On the "pro" side, it may reduce code complexity. In some cases dramatically. And when it is known that "N" will always be a small number, and the operation itself is fast. Then keeping the N+1 pattern would be a good thing IMHO. It depends on the case.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+1 batch query. On API read paths serving collections, this scales linearly with result size
+and dominates response latency.
+
+**Occurrences found (5)**:
```
ctron: I think there are a lot more in, especially in the "analysis" module.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+
+| # | File | Line(s) | Description | Severity |
```
ctron: How is "severity" evaluated here?

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| 4 | `modules/fundamental/src/purl/model/details/purl.rs` | ~279 | Up to 3 DB queries per iteration via `PurlStatus::from_entity` (status, CPE, version_range lookups by ID), plus additional queries from `PurlStatus::new` (`VulnerabilityHead` description, `AdvisoryHead`); conditional organization query in the parent loop | High |
+| 5 | `modules/fundamental/src/purl/model/details/versioned_purl.rs` | ~134-145 | `VersionedPurlStatus::from_entity()` per iteration calls `VulnerabilityHead::from_vulnerability_entity(..., Memo::NotProvided, tx)`, which issues a per-iteration `SELECT` for `vulnerability_description` (lang = "en"). Bulk loading for other relations is already done outside the loop, but this read-side N+1 remains. | Medium |
+
+> **Note**: Some loop-based queries (e.g., `purl/service/mod.rs` lines 575, 730, 755) are
+> intentional chunking for PostgreSQL's 65535 bind-parameter limit. These are valid patterns
+> and are not N+1 issues.
+
+**Convention options**:
+
+- **Option A — Batch with JOINs/IN clauses**: All collection data access MUST use batch
```
ctron: Maybe we should consider this: https://pganalyze.com/blog/5mins-postgres-performance-in-vs-any

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+  ```
+
+- **Option B — Batch for read paths only**: N+1 is prohibited on API read paths (GET
+  endpoints). Allowed on write/delete paths where entity counts are typically small and
+  transactional isolation may require per-entity operations.
+
+- **Option C — Keep as-is**: No convention. Performance is addressed case-by-case when
+  bottlenecks are observed. N+1 patterns are acceptable if the loop iteration count is
+  expected to be small.
+
```
ctron: I would say "Option D": prefer option A, but fall back to "N+1" if there are good reasons.

"C" doesn't seem to cover this, as it says "no convention". I think the should try, and try hard to remediate N+1 patterns. But not at all cost.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+`LIMIT`, pagination, or any bound on result size.
+
+**Why it matters**: As data grows, unbounded queries consume increasing memory and network
+bandwidth. A single unbounded query on a large table can cause OOM or timeout.
+
+**Occurrences found (3)**:
+
+| # | File | Line(s) | Description |
+|---|------|---------|-------------|
+| 1 | `modules/importer/src/service.rs` | ~133 | `importer::Entity::find().all()` fetches all importers, sorts in memory |
```
ctron: Which I don't see as a problem, as the number of importers is considered small (below 50). Sure it may be an improvement. But in some cases it's not. Having pagination on the HTTP API would be beneficial.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+
+### AP-2: Unbounded Queries
+
+**What it is**: Queries that fetch all rows from a table or relation without applying
+`LIMIT`, pagination, or any bound on result size.
+
+**Why it matters**: As data grows, unbounded queries consume increasing memory and network
+bandwidth. A single unbounded query on a large table can cause OOM or timeout.
+
+**Occurrences found (3)**:
```
ctron: Again, there is more. Especially in the analysis module.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+
+- **Option C — Keep as-is**: No convention. Performance is addressed case-by-case when
+  bottlenecks are observed. N+1 patterns are acceptable if the loop iteration count is
+  expected to be small.
+
+---
+
+### AP-2: Unbounded Queries
+
+**What it is**: Queries that fetch all rows from a table or relation without applying
```
ctron: In some cases we load all data, and then keep in memory to process it. Especially as mentioned above to do things like batch loading, pre-loading, ... 

In such cases it doesn't seem to make sense to use limits, as we load and store the data in-memory anyway. In fact, it would be counter productive (even batch loading). Batch loading only makes sense to limit the number of postgres parameters. Having to page through data would mean to execute the query multiple times. Compared to one single query.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| 3 | `modules/fundamental/src/vulnerability/model/details/mod.rs` | ~46-57 | `.all(tx)` fetching all advisory-vulnerabilities and all scores **filtered by `vulnerability_id`** — not a full table scan, but lacks `LIMIT`; risk is unbounded per-entity fan-out if a single vulnerability has many advisories/scores |
+
+**Convention options**:
+
+- **Option A — Require pagination or limits on all queries**: All list/collection endpoints
+  MUST use the `Paginated` wrapper. Internal service queries on potentially-large tables
+  must use `.limit()` or chunked iteration. Exceptions only for tables with a known small
+  upper bound (e.g., importers, which are admin-configured).
+
+- **Option B — Limit only on public API endpoints**: Public API list endpoints must be
```
ctron: Option B 👍

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+
+**Why it matters**: Transfers unnecessary data over the network, consumes application memory,
+and bypasses database index optimizations. The performance gap widens as the unfiltered
+dataset grows.
+
+**Occurrences found (2)**:
+
+| # | File | Line(s) | Description |
+|---|------|---------|-------------|
+| 1 | `modules/fundamental/src/vulnerability/model/details/mod.rs` | ~54-68 | All scores fetched via `.all(tx)`, then filtered with `.filter(\|s\| s.advisory_id == ...)` |
```
ctron: I'd say that's a false positive. As the code does re-use that data. But eliminates a second query by re-using that data and manually filtering out information for the first use case of the data.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| 2 | `modules/fundamental/src/advisory/model/summary.rs` | ~40-67 | Bulk scores fetched, then O(advisories x vulns x scores) in-memory iteration |
```
ctron: Again, looks like a false positive to me. The data is re-used in a way to actually present an N+1 selection issue.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+
+**Convention options**:
```
ctron: The two wrong examples show me, that AI did get it wrong here. In general this is correct. But the provided examples are a valid deviation from the norm.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+were loaded from the database **only** to obtain a count — materializing a large result
+set that is not otherwise needed.
+
+**Why it matters**: Transfers and deserializes every row just to count them. For paginated
+endpoints, this means fetching the entire table to compute `total` instead of a single
+aggregate query.
+
+**Occurrences found (0)**:
+
+No confirmed occurrences were found. The previously cited case
```
ctron: I'm confused by this section. The example is, that we never had this problem?

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+  let total = items.len();
+  ```
+
+- **Option B — SQL COUNT for paginated endpoints only**: Paginated endpoints must use SQL
+  `COUNT()` for the `total` field. Internal service code may use `.len()` if the full
+  dataset is already loaded for processing.
+
+- **Option C — Keep as-is**: No convention. `.len()` on query results is acceptable for
+  simplicity.
+
```
ctron: I guess again Option D: prefer Option A, unless there are reasons not to. But have this as a convention rather than no convention at all. Which we currently seem to follow.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+---
+
+### AP-5: Sequential DB Calls (Missing Parallelization)
+
+**What it is**: Multiple independent database queries awaited sequentially when they could
+be executed concurrently using `tokio::join!` or `futures::try_join!`.
+
+**Why it matters**: Sequential independent queries add their latencies together. With
```
ctron: We have to be careful here. That may actually be a thing that has more negatives than positives. And I'd say that's the reason why we avoid it.

Running multiple queries without a transaction (which isn't a good idea) will actually use multiple database connections. And thus also lead to inconsistent results.

While running queries with the same transaction/DB connection would not lead to increased performance, as PSQL only handles one query per connection at a time: https://www.postgresql.org/message-id/4A8B56E2.6050704%40postnewspapers.com.au

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+  let scores = advisory_vulnerability_score::Entity::find()...all(tx).await?;
+  ```
+
+- **Option B — Parallelize on hot paths only**: Parallelization is recommended but not
+  required. It MUST be applied on API-facing read paths with measurable latency. Internal
+  batch/ingestion code may remain sequential for readability.
+
+- **Option C — Keep as-is**: No convention. Sequential queries are acceptable. Parallelization
+  is a micro-optimization to apply when profiling indicates a bottleneck.
+
```
ctron: Option D: never parallelize DB queries.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+---
+
+### AP-5: Sequential DB Calls (Missing Parallelization)
+
+**What it is**: Multiple independent database queries awaited sequentially when they could
+be executed concurrently using `tokio::join!` or `futures::try_join!`.
+
+**Why it matters**: Sequential independent queries add their latencies together. With
+`try_join!`, the wall-clock time is the maximum of the individual queries, not their sum.
```
ctron: Additionally, it would only benefit small number of users. Scaling up more users, we can process more queries in parallel. Parallelizing in a single API operation would exhaust the DB connection pool faster, and this reduce the API throughput. Having less parallelization in the end but more memory consumption.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+  }
+  ```
+
+- **Option B — Bulk for large collections only**: Bulk operations are required when the
+  collection size can exceed ~10 items. Small fixed-size loops (e.g., deleting 1-3 related
+  entities) may use individual operations.
+
+- **Option C — Keep as-is**: No convention. Individual operations in loops are acceptable
+  for clarity. Batch operations are a performance optimization to apply when needed.
+
```
ctron: I'd say: it depends. If there are multiple entries to delete, then do it all at once. If there is only one present, e.g. because the API call is a single delete, then just do a single delete.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+**Why it matters**: Dependency trees in SBOMs can be deeply nested. Without depth limits,
+a single malformed or deeply nested document can cause stack overflow or excessive
+computation time.
+
+**Occurrences found (2)**:
+
+| # | File | Line(s) | Description |
+|---|------|---------|-------------|
+| 1 | `modules/analysis/src/model/roots.rs` | ~24-35, 68-88 | Two `roots_into` functions — recursive DFS with ancestor vector cloning per call |
```
ctron: I'm not really seeing a problem here. We are working on a bounded data set. One that we already have in-memory anyways.

Now the cloning is a different story, using a non-recursive approach could eliminate that. But it also wouldn't fix the (infinite) depth recursion, as it would trade it for an infinite loop problem.

So the depth bound doesn't seem to be the problem here, but the cloning required by the recursive calls.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| 2 | `modules/ingestor/src/service/advisory/csaf/util.rs` | ~50-64 | `walk_product_branches_ref` — recursive DFS with no depth limit |
```
ctron: Again, we do have a depth limit because we had been able to load the file. And we are not allocating anything here, as we are working with references.

If you end up at this code spot, a lot of recursion has already happened, parsing the document. So solving anything here doesn't seem like the right place. How would you want to handle a depth limit here? Fail processing? That would introduce a lot of code what I would expect to be never executed (as the failure would happen way before).

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+  recursive functions to avoid stack overflow. Recursive implementations must include a
+  depth counter parameter.
+
+  ```rust
+  // Approved: iterative with depth limit
+  fn walk_ancestors(start: &Node, max_depth: usize) -> Vec<Node> {
+      let mut stack = vec![(start, 0)];
+      let mut result = Vec::new();
+      while let Some((node, depth)) = stack.pop() {
+          if depth >= max_depth { continue; }
```
ctron: That would actually lead to wrong data. Wouldn't it?

Just skipping content like that introduced bugs before. That would actually be an anti-pattern to me.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+  }
+  ```
+
+- **Option B — Depth limits only**: Recursive traversal is acceptable but MUST include a
+  maximum depth parameter. No requirement to convert to iterative. Reasonable default
+  depth: 256.
+
+- **Option C — Keep as-is**: No convention. Trust that input data is well-formed. Address
+  traversal limits only if stack overflow is observed in production.
+
```
ctron: Option D: Ensure we are able to handle the data when receiving it. Then assuming we can process it. But don't just "skip" data, raise an error.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+---
+
+### AP-8: Extra DB round-trips vs JOINs / single-query loading
+
+**What it is**: After an initial query, related data is loaded via **multiple sequential**
+SeaORM `load_*` calls (each call is typically **one batched SQL statement** for the whole
+set, not per-row N+1) instead of fewer queries using JOINs or combined selects such as
+`find_also_related()` / `find_with_related()`.
+
+**Why it matters**: Each extra round-trip adds latency (often ~sum of sequential query
```
ctron: Careful with that. That might actually. be an anti-pattern too. Aggressive joining of table has negative impacts as well.

Joining means that all data has the returned in all rows. If for example, you're joining a lot of "packages to purls", you might end up with the same PURL over and over again. That would lead to returning the same data over and over again. Same with rows expanding, assuming a relationship of A -> B -> C. If there are many Bs for an A, then that would mean that data of B and C would be repeated, although not being used (reduced) later. That would mean a lot more traffic on the postgres connection, for not good reason. Harming performance.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+times vs ~max with parallelization, or ~one round-trip with a well-shaped JOIN). Distinct
+from “missing batch load”: `load_one` on a `Vec<Model>` is already the standard SeaORM batch
+pattern.
+
+**Occurrences found (1)**:
+
+| # | File | Line(s) | Description |
+|---|------|---------|-------------|
+| 1 | `modules/fundamental/src/purl/service/mod.rs` | ~602-626 | Four sequential `load_one` / `load_many_to_many` calls after one `all_statuses` query — four extra batched queries vs potentially one heavier JOIN (trade-off: clarity vs round-trips) |
```
ctron: This was actually refactored by @ruromero away from this pattern, IIRC. Specifically adding the additional loads to improve performance. Which was proven by providing measurements.

Maybe there are cases where joins help. But you have to be really careful with that, as there are other cases where it might hurt massively.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+  a shaped JOIN over **many sequential** batched `load_*` calls on the same parent set.
+  Each `load_one`/`load_many` on a `Vec<Model>` is typically **one** SQL statement for the
+  whole set (not per-row N+1); the trade-off is often **several batched queries vs one
+  heavier JOIN**.
+
+- **Option B — Optimize hot paths only**: Fewer round-trips / JOINs are recommended on
+  API-facing read paths with measurable latency. Internal code may keep sequential batched
+  `load_*` for clarity; profile before changing.
+
+- **Option C — Keep as-is**: No convention. Sequential batched `load_one`/`load_many` is
```
ctron: C: As it depends. But better not to require anything then heading in the wrong direction.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+
+---
+
+### AP-9: Separate queries for validation vs follow-on work (Optional Improvement)
+
+**What it is**: One query establishes **404 / existence** and supplies an identifier (e.g.
+`sbom_id`); a **second** query implements different logic (packages, related packages) and
+does **not** re-load the same SBOM tuple as the first call.
+
+**Why it matters**: This is two round-trips for two **separate concerns** (guard + domain
```
ctron: Not sure I understand the idea here.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+query), not duplicate fetching of the same row for the same purpose. This is a common and
+defensible pattern for separation of concerns — not a true anti-pattern. Listed here as an
+optional improvement opportunity only.
+
+**Occurrences noted (1)**:
+
+| # | File | Line(s) | Description |
+|---|------|---------|-------------|
+| 1 | `modules/fundamental/src/sbom/endpoints/mod.rs` | ~447, 555 | `fetch_sbom(id)` for 404 + `sbom_id`, then `fetch_sbom_packages` / `fetch_related_packages(sbom_id, …)` — second path builds different SQL; room to reduce to one round-trip if desired |
```
ctron: This is actually just a side effect here. The use for the is call is a lookup of the SBOM from a different table (source_document).

I'd not recommend doing an initial lookup all the time to prove existence. That can be done in one call (as all the other optimizations are suggesting) But in this case, it's a lookup from one ID to another.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+
+**Convention options**:
+
+- **Option A — Pass fetched entities downstream**: When an entity is already loaded for
```
ctron: Yes, option A. Unless the use case is a different one and this is a side effect.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+  acceptable for clarity unless profiling shows a need to merge.
+
+---
+
+### AP-10: Missing Database Indexes
+
+**What it is**: Frequently-queried columns that lack database indexes, causing full table
+scans on filter and sort operations.
+
+**Why it matters**: Without indexes, the database must scan every row for each query that
```
ctron: As tables grow, so do indexes. That's the downside of them. As we allow for user provided query and sorting operations, we cannot just create an index for every combination. That's just too costly.

But for cases where we do know the queries upfront, and where the performance improvement is significiant, we should definitely have them.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+filters or sorts on these columns. As tables grow, query latency increases linearly.
+
+**Occurrences found (2 tables)**:
+
+| # | Table | Column(s) | Queried From | Impact |
+|---|-------|-----------|-------------|--------|
+| 1 | `product` | `name` | `ingestor/graph/product/mod.rs` — lookup by name during ingestion | Medium |
```
ctron: I guess that's a red herring. Yes, that function does exist (`get_product_by_name`) but it's dead code. Never called by anyone. So instead of adding an index, we should drop that code.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+  -- Example: B-tree index for product name lookups
+  CREATE INDEX IF NOT EXISTS idx_product_name
+      ON product (name);
+  ```
+
+- **Option B — Index high-traffic columns only**: Indexes are required for columns used
+  in public API endpoints that serve list/search results. Columns used only in low-traffic
+  internal operations do not require indexes.
+
+- **Option C — Keep as-is**: No convention. Index creation is decided per-query based on
```
ctron: Option C: it depends. There should definitely be a pros/cons list for making a decision

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+
+In addition to performance anti-patterns, analysis of the codebase has identified recurring
+coding anti-patterns that affect maintainability, correctness, and observability. This
+section follows the same structure as the performance analysis above.
+
+---
+
+### CA-1: Swallowed Errors
+
+**What it is**: Error values silently discarded via `.ok()`, `let _ =`, or `if let Ok(...)`
```
ctron: I general yes. But something being an error, doesn't mean something "when wrong". An error may every well be an expected outcome. Logging all cases of "expected errors" would lead to a massive log. Which would create problems on it own.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+without logging or handling the error case. The caller has no signal that something went
+wrong.
+
+**Why it matters**: Silent error swallowing hides bugs, makes debugging difficult, and can
+lead to data loss or inconsistent state. When a deserialization or DB write fails silently,
+the system continues with missing or stale data.
+
+**Occurrences found (9)**:
+
+| # | File | Line(s) | Pattern | Severity |
```
ctron: What's the definition of "severity" here?

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+|---|------|---------|---------|----------|
+| 1 | `modules/importer/src/model/mod.rs` | ~331 | `serde_json::from_value(report).ok()` — importer run report deserialization silently dropped | High |
```
ctron: In this case, this provides additional diagnostic information, if possible. If the serialization of that fails, that would just not be possible. Logging would be fine, but throwing an error not.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| 2 | `modules/importer/src/server/progress.rs` | ~48, 77 | `let _ = self.service.set_progress_message(...)` — DB write failures for progress silently discarded | High |
```
ctron: In this case it would actually be beneficial to add this to the instrumentation. Rather than an explicit log call.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| 3 | `modules/importer/src/server/mod.rs` | ~186 | `serde_json::to_value(report).ok()` — importer run report serialization silently dropped on `update_finish`; report data lost from database | High |
+| 4 | `modules/importer/src/model/mod.rs` | ~65 | `OffsetDateTime::from_unix_timestamp_nanos(t).ok()` — heartbeat timestamp errors silently dropped | Medium |
```
ctron: Here it might actually be beneficial to remove the root cause of this. Using a proper timestamp in the database, if possible.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| 5 | `modules/ingestor/src/service/advisory/csaf/loader.rs` | ~44, 48 | `published`/`modified` timestamp conversions silently become `None` | Medium |
```
ctron: In this case it would be beneficial to report this back to the user, as we do have a mechanism for reporting such things as "warning" during the upload. This would also end up in the report. Logging this, or failing, wouldn't help.

ctron: Same for the other CSAF related ones.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+  serde_json::from_value(report).ok()
+  ```
+
+- **Option B — Log on high-severity paths only**: Silent error discarding is prohibited
+  on paths that affect data integrity (DB writes, state serialization, lock acquisition).
+  Allowed on best-effort paths (timestamp formatting, display rendering) with a comment.
+
+- **Option C — Keep as-is**: No convention. Developers judge whether an error is worth
+  logging. `.ok()` is acceptable for non-critical conversions.
+
```
ctron: Option D:

* if we have instrumentation wrapping it, that's ok, no need for extra processing
* if we have a way to report this case back to the user, leverage this, but don't extra log it
* if that case prevents us from processing the data, and the data is essential, propagate the error
* if running into an error is an acceptable and expected case, that's fine, do nothing
* if we can ignore the error, but have no other way to handle it, log it, but try to be compact doing it, e.g. like `.inspect_err()` rather than a full blown `match`.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| 2 | `modules/fundamental/src/purl/endpoints/base.rs`, `modules/fundamental/src/purl/endpoints/mod.rs` | ~39, ~60 | Inconsistent prefix check: `"pkg:"` vs `"pkg"` (missing colon) for pURL detection | High |
+| 3 | `modules/fundamental/src/vulnerability/service/mod.rs` | ~79 | Sort field and direction compared as raw strings: `field == "id" && order == "asc"` — **Note**: constrained by the `Columns::translator` callback API (`fn(&str, &str, &str) -> Option<String>`); not fixable at the application level without framework changes | Medium |
+| 4 | `modules/fundamental/src/sbom/endpoints/mod.rs` | ~372 | Hardcoded `vec!["affected".to_string()]` instead of enum | Low |
+| 5 | `modules/fundamental/src/sbom/model/details.rs`, `modules/fundamental/src/vulnerability/service/mod.rs`, `modules/ingestor/src/graph/purl/status_creator.rs` | ~552, ~703, ~19 | `VexStatus` enum exists but multiple structs carry `status: String` | Medium |
+
+**Convention options**:
+
+- **Option A — Enums for fixed value sets**: Any value drawn from a fixed set (statuses,
```
ctron: Option A

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+|---|-------|-------------|----------|
+| 1 | `modules/fundamental/src/advisory/service/mod.rs` ~132-197 vs `modules/fundamental/src/sbom/service/label.rs` ~16-77 | Nearly identical `set_labels`/`update_labels` with FOR UPDATE locking — differs only by entity type | High |
+| 2 | `modules/fundamental/src/advisory/endpoints/label.rs` (~119 lines) vs `modules/fundamental/src/sbom/endpoints/label.rs` (~120 lines) | Identical label endpoint handlers — same `LabelQuery`, same `all`/`set`/`update` functions | High |
+| 3 | `modules/fundamental/src/sbom/service/sbom.rs` ~222-268, ~338-390 and `modules/fundamental/src/purl/service/mod.rs` ~414-473 | Same SPDX+CycloneDX license filtering subquery pattern copied three times | Medium |
+| 4 | `modules/fundamental/src/vulnerability/service/mod.rs` ~335-454, ~668-896 | `format_response`/`format_response_v2` share ~90% logic; `row_to_vuln_v3`/`v2` share ~40-50% (v3 has richer `AdvisoryEntry` with version_range, remediations, context_cpe) — significant duplication in the coordinators, moderate in the row mappers | Medium |
+
+**Convention options**:
+
+- **Option A — Extract shared logic into generics or traits**: When two or more modules
```
ctron: Option A

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+  (e.g., `Deprecation`, `AdvisoryHead`, `VulnerabilityHead`) MUST live in `common/` or
+  `entity/` crates, not in the module that happens to define them. Domain modules may only
+  import from `common/`, `entity/`, and `query/` — never from sibling domain modules.
+
+- **Option B — Allow read-model imports, prohibit layer violations**: Domain modules may
+  import read-only model types (`*Head`, `*Summary`) from sibling modules for API response
+  composition. However, the `fundamental` layer MUST NOT depend on `ingestor` — shared
+  types like `Deprecation` must be moved to `common/`.
+
+- **Option C — Keep as-is**: No convention. Cross-module imports are acceptable within
```
ctron: Option C

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| 5 | `vulnerability/service/mod.rs` | `row_to_vuln_v3` | ~130 | Inline struct definitions, JSONB deserialization, BTreeMap construction, advisory iteration |
+
+**Convention options**:
+
+- **Option A — Maximum function length**: Functions MUST NOT exceed 100 lines (excluding
+  blank lines and comments). When a function has multiple distinct phases, extract each
+  phase into a private helper method with a descriptive name. The parent function becomes
+  a coordinator that calls the helpers in sequence.
+
+- **Option B — Maximum cyclomatic complexity**: No hard line limit, but functions with
```
ctron: Option B. And I'd say it is ok to use inner strucs, fns, ... for doing this.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+  handlers MUST also be instrumented. Follow the conventions already in `CONVENTIONS.md`
+  for `skip`, `skip_all`, and `err(level = ...)`.
+
+- **Option B — Instrument service methods only**: All public service methods must have
+  `#[instrument]`. Endpoint handlers are exempt because Actix-web middleware provides
+  request-level tracing. Focus instrumentation on the service layer.
+
+- **Option C — Keep as-is**: No convention. Instrumentation is added when needed for
+  debugging. Partial coverage is acceptable.
+
```
ctron: Option D: Option A, but with a twist. I think this is already well described in the tracing/logging markdown file. In a nutshell, everything that is calling out to external things (DB, storage, ... ) should be wrapped in instrumentation. Everything that does not, as long as it is considered trivial, must not be wrapped. Even if things cannot be wrapped (multiple sections of loading where loading functions don't have instrumentation) use `.instrument` or other constructs amending this. But don't do it twice (inner vs outer). Not limited to `pub`, but based on logic blocks and the fact that external factors are in play.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| 2 | `vulnerability/service/mod.rs` | `VulnerabilityService` struct, all 5 methods |
+| 3 | `weakness/service/mod.rs` | `WeaknessService` struct, both methods |
+| 4 | `organization/service/mod.rs` | `OrganizationService` struct, both methods |
+| 5 | `product/service/mod.rs` | `ProductService` struct, all 3 methods |
+| 6 | `purl/service/mod.rs` | `PurlService` struct, most methods |
+| 7 | `storage/src/service/mod.rs` | `StorageKey`, `StorageResult`, `StoreError` |
+
+**Convention options**:
+
+- **Option A — Document all public items**: Every public struct, enum, trait, function,
```
ctron: Option A.

In addition, also add non-verbose comments inside functions.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+
+---
+
+### CA-8: Mixed Logging Frameworks (`log::` vs `tracing::`)
+
+**What it is**: The same file uses both `log::debug!`/`log::warn!`/`log::info!` macros
+and `tracing::instrument`/`tracing::Instrument` attributes, creating inconsistent
+observability output.
+
+**Why it matters**: `tracing` and `log` have different span-awareness behavior. `log::*`
```
ctron: Not sure this is true, as `log` actually forwards to `tracing`.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+| 1 | `vulnerability/service/mod.rs` | `log::debug!` (6 calls) mixed with `tracing::instrument` imports |
+| 2 | `sbom/service/sbom.rs` | `log::debug!`, `log::warn!` mixed with `tracing::instrument` + `tracing::Instrument` |
+| 3 | `purl/service/mod.rs` | `log::debug!` (2 calls) mixed with `tracing::{instrument, Instrument, info_span}` |
+| 4 | `analysis/src/service/mod.rs` | `log::info!`, `log::warn!`, `log::debug!`, `log::trace!` (11 calls) mixed with `tracing::instrument` |
+| 5 | `advisory/endpoints/mod.rs` | `log::warn!`, `log::info!` only (no `tracing::` in this file) — module-level inconsistency: advisory *service* uses `tracing::instrument` while advisory *endpoints* uses `log::` only |
+| 6 | `ingestor/src/service/mod.rs` | `log::debug!`, `log::warn!` mixed with `tracing::instrument` |
+
+**Convention options**:
+
+- **Option A — Use `tracing::` exclusively**: All logging MUST use `tracing::` macros
```
ctron: Option A sounds reasonable.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+  format!("AND relationship = {RELATIONSHIP_DESCRIBES}")
+  ```
+
+- **Option B — Named constants for domain values only**: Domain-specific magic values
+  (relationship types, status codes, heuristic thresholds) must use named constants.
+  Infrastructure values (timeouts, poll intervals) may be inline if used only once.
+
+- **Option C — Keep as-is**: No convention. Inline values are acceptable with explanatory
+  comments.
+
```
ctron: Option D: It depends. In general we should have Option A. However, there are cases where it causes confusion or has other side effects.

Assume the example from above. Having 13 replaced by a const value, would require to render the string every time used. Rather than having a const struct. This seems counter productive. It would be better to keep that 13 in the string, but document around it "why" there's a 13 in there.

Same for structs:

```rust
        ImporterConfiguration::Cve(CveImporter {
            common: CommonImporter {
                disabled: true,
                period: Duration::from_secs(300),
                description: Some(description.into()),
                labels: Default::default(),
            },
            source: DEFAULT_SOURCE_CVEPROJECT.into(),
            years: HashSet::default(),
            start_year,
        }),
```

Having `DEFAULT_SOURCE_CVEPROJECT` here, imported from another module. Only being used once. Doesn't make much sense to me. It would be clearer if that would be in the same file. Unless, it would indeed be re-used. Then it could be refactored.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+**Why it matters**: Inlined parameters prevent the database from reusing query plans and
+remove a layer of SQL injection protection. While the current code parameterizes sub-queries
+individually, the final concatenation undoes this.
+
+**Occurrences found (1)**:
+
+| # | File | Line(s) | Description |
+|---|------|---------|-------------|
+| 1 | `modules/fundamental/src/vulnerability/service/mod.rs` | ~455-657 | `build_vulnerabilities_query_string` (~455-521) uses `format!()` to interpolate SQL fragments; `build_query` (~523-657) constructs sub-queries with `Statement::from_sql_and_values`, then `.to_string()`s them and concatenates with `UNION ALL` into a `Statement::from_string` |
```
ctron: This example might not be ideal. But it seems like a decent trade off handling the complexity. Having that replaced by something else, the question would be if the alternative is actually better,

There is no SQL injection in play here, as that's gets sorted out before.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1027 @@
...
+
+**Convention options**:
+
+- **Option A — Prohibit `Statement::from_string` with concatenated queries**: All SQL
+  queries MUST use parameterized statements end-to-end. If `UNION ALL` is needed, use
+  SeaORM's `UnionType` builder or a single parameterized CTE. `Statement::from_string`
+  is only acceptable for static SQL with no dynamic values.
+
+- **Option B — Allow with review gate**: `Statement::from_string` with dynamic content
```
ctron: Option B: There should be a good reason to not use Option A, and a discussion around that. But to me the example shows a valid reason to deviate from the norm.

docs/adrs/00018-conventions-file.md
mrrajan: Agree - Removed

docs/adrs/00018-conventions-file.md
mrrajan: Agree! We should have a dedicated database section.

docs/adrs/00018-conventions-file.md
mrrajan: Agree! We should have just the examples not the references to the specific files. Rephrased to `share canonical examples` on the conventions file.

docs/adrs/00019-conventions-file.md
mrrajan: Sourcery.ai applies generic best practices and CONVENTIONS.md captures project specific ones. So, project specific practices might called out by the review tools. We have to explore how to align review tools to use CONVENTIONS.md file as reference.

docs/adrs/00018-conventions-file.md
mrrajan: Agree!

docs/adrs/00019-conventions-file.md
mrrajan: Yes! Conventions are defaults, but should not be enforced.

docs/adrs/00018-conventions-file.md
mrrajan: Agree! The references should use function/ symbol names or example snippets in the CONVENTIONS.md file

docs/adrs/00018-conventions-file.md
mrrajan: Agree! rephrased to `performance impact, maintainability, consistency`

docs/adrs/00018-conventions-file.md
mrrajan: Agree! This should be references not specific locations in codebase.

docs/adrs/00018-conventions-file.md
mrrajan: Severity is based on,
 - Whether the N is bounded or not
 - Code path is API or internal

docs/adrs/00018-conventions-file.md
mrrajan: Sounds good! Added option D
```
- **Option D — Prefer batch, allow exceptions**: Default to Option A — batch loading on
  collection paths and remediate existing N+1 patterns as tech debt. Per-entity DB calls
  inside loops are permitted only when batching is impractical or costly. 
```

docs/adrs/00018-conventions-file.md
mrrajan: Agree! It is a false positive. I will remove this example section - But as a convention for this issue, shall we suggest `option B`?

docs/adrs/00018-conventions-file.md
mrrajan: Yes, we don't have any occurrences for this issue. But, we could keep this on our CONVENTIONS.md file for future reference.

docs/adrs/00018-conventions-file.md
mrrajan: Agree!

docs/adrs/00018-conventions-file.md
mrrajan: In that case, I will remove this section.

docs/adrs/00018-conventions-file.md
mrrajan: Agree!

docs/adrs/00018-conventions-file.md
mrrajan: Agree!

docs/adrs/00018-conventions-file.md
mrrajan: Agree!

docs/adrs/00018-conventions-file.md
mrrajan: Sounds good!

docs/adrs/00018-conventions-file.md
mrrajan: It is misunderstanding on the existing workflow - I hope this section can be removed.

docs/adrs/00018-conventions-file.md
mrrajan: Ah, Agree!

docs/adrs/00018-conventions-file.md
mrrajan: Sounds good!

docs/adrs/00018-conventions-file.md
mrrajan: Same as above, Severity is based on
- Whether the N is bounded or not
- Code path is API or internal

docs/adrs/00018-conventions-file.md
mrrajan: Agree! It is additional noise

docs/adrs/00018-conventions-file.md
mrrajan: Agree!

docs/adrs/00018-conventions-file.md
mrrajan: Agree! we should fix the root cause instead!

docs/adrs/00018-conventions-file.md
mrrajan: Sounds good!

docs/adrs/00018-conventions-file.md
mrrajan: Sounds good!

docs/adrs/00018-conventions-file.md
mrrajan: Sounds good!

docs/adrs/00018-conventions-file.md
mrrajan: Agree! Rephrased to,
```
Trustify needs a single, explicit reference for coding patterns, naming rules, error-handling idioms, testing practices, and architectural norms. Contributors and reviewers use it during implementation and review; the same clarity also helps AI-assisted workflows (Claude Code, Copilot, and similar tools) when those tools load project context.
```

docs/adrs/00018-conventions-file.md
mrrajan: Agree! lets limit only for the public API endpoints

docs/adrs/00018-conventions-file.md
mrrajan: Agree!

docs/adrs/00018-conventions-file.md
mrrajan: Referring to https://docs.rs/tracing-log/latest/tracing_log/, we need logTracer::init() to have logs and tracing connected. But in our project only the test file, `test-context/src/flame.rs` has this configuration for `setup_global_subscriber`.

mrrajan (commented):
Thanks @ctron for your review, I have added a new section for database and addressed the review comments. I have added a new section `Preferred option:` under each category based on the review.

docs/adrs/00019-conventions-file.md
helio-frota: Good question 👍 

Sending random and sometimes partial PRs without JIRAs?

side note: Yesterday I asked AI to read the [instrumentation](https://github.com/guacsec/trustify/blob/main/docs/design/log_tracing.md#instrumentation) section and to produce a [checklist](https://gist.github.com/helio-frota/c4f9f3ace876d481025da22d5221eed4) for me, not changed the code yet as some parts are subjective (example -> the size of returned struct data) + still need to validate if all the checklist is 100% correct. Imagining all this checklist is correct, all the instrumentation-related-code from all files should be updated via a single PR? Should this initiative be separated by our current workspace-crates to avoid git conflicts and long PR reviews?

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1454 @@
...
+| 2 | `modules/fundamental/src/advisory/endpoints/label.rs` (~119 lines) vs `modules/fundamental/src/sbom/endpoints/label.rs` (~120 lines) | Identical label endpoint handlers — same `LabelQuery`, same `all`/`set`/`update` functions | High |
+| 3 | `modules/fundamental/src/sbom/service/sbom.rs` ~222-268, ~338-390 and `modules/fundamental/src/purl/service/mod.rs` ~414-473 | Same SPDX+CycloneDX license filtering subquery pattern copied three times | Medium |
+| 4 | `modules/fundamental/src/vulnerability/service/mod.rs` ~335-454, ~668-896 | `format_response`/`format_response_v2` share ~90% logic; `row_to_vuln_v3`/`v2` share ~40-50% (v3 has richer `AdvisoryEntry` with version_range, remediations, context_cpe) — significant duplication in the coordinators, moderate in the row mappers | Medium |
+
+**Convention options**:
+
+- **Option A — Extract shared logic into generics or traits**: When two or more modules
+  implement the same logic pattern (e.g., label CRUD, license filtering), extract it into
+  a shared generic function, trait, or macro parameterized by the entity type. Duplicated
+  logic blocks longer than ~20 lines are prohibited.
```
helio-frota: informational only:

I quickly modified my personal duplicate code segment tool to focus on 20+ lines and found only these: https://gist.github.com/helio-frota/87f1a9ab12fe84480f2574bc7b92e8a7 
( I'm skipping `tests,test,entity,migration,benches` )

docs/adrs/00018-conventions-file.md
helio-frota: we are using a tracing-log feature here: https://github.com/guacsec/trustify/blob/main/common/infrastructure/Cargo.toml#L33

> log actually forwards to tracing

this explains why even using log::info here https://gist.github.com/helio-frota/171ac4ac8d362827e22b66ce165f32cf?permalink_comment_id=6121576#gistcomment-6121576, I was able to see the correlation with traces using [opentelemetry-appender-tracing](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/) instead of [opentelemetry-appender-log](https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-appender-log) :+1:

docs/adrs/00019-conventions-file.md
mrizzi: > Question is, how do we deal with current code that doesn't align to those newly defined patterns?

It will mean having tech debt initiatives to enforce consistency in current code not yet aligned with expected patterns and code conventions.

docs/adrs/00018-conventions-file.md
mrizzi: Changes from PRs are meant to be additive to improve code conventions for identified lacks. Changing them to "adhere" to the code in a PR should be blocked otherwise code will drift from conventions.

docs/adrs/00018-conventions-file.md
```diff
@@ -0,0 +1,1454 @@
+# 00018. Conventions File and Performance Anti-Pattern Standards
+
+Date: 2026-05-14
+
+## Status
+
+PROPOSED
```
ctron: This should be `ACCEPTED` as it would be accepted when the PR is merged.

ctron (changes requested):
Looks good to me. With the exception that I still see the actual content (AP, …) My understanding was, that his would be removed from this ADR, but then added in another PR to the conventions file.

docs/adrs/00018-conventions-file.md
mrrajan: Thanks for the clarification @helio-frota! updated this section.

ctron (approved):

PR #2364, opened 2026-05-19
Add `DELETE /api/v3/sbom` endpoint for batch SBOMs deletion

i386x:
Supports [Batch/Bulk operations conventions](https://github.com/guacsec/trustify/blob/main/CONVENTIONS.md#missing-batchbulk-operations).

Related Jiras: TC-4003, TC-3934

## Summary by Sourcery

Introduce server-side SBOM pruning APIs and CLI integration for bulk deletion with preview support.

New Features:
- Add /api/v3/sbom/prune GET and DELETE endpoints to list and prune SBOMs based on existing query filters with pagination-like parameters.
- Expose SBOM prune context and states (fetched, deleted, skipped, failed) via new API schemas and models, including storage keys for follow-up cleanup.
- Extend the CLI to call the new prune endpoints, supporting dry-run and actual prune operations while mapping results into existing delete reporting.

Enhancements:
- Refactor SBOM deletion logic to support bulk deletion, including coordinated cleanup of related entities and storage artifacts, and reuse this path for single-SBOM delete.
- Augment storage service keys and related types with serialization and OpenAPI schema support for use over the API.
- Document the SBOM pruning API design and behavior in a new ADR.

Build:
- Update Cargo.toml dependencies to wire fundamental and storage modules into the CLI and enable schema/serde support.

Documentation:
- Add an ADR describing the SBOM prune API endpoints, usage, logging, safety features, and future scheduling plans.

### DIFF
--- a/modules/fundamental/src/sbom/endpoints/mod.rs
+++ b/modules/fundamental/src/sbom/endpoints/mod.rs
@@ -10,7 +10,7 @@ use uuid::Uuid;
 
 use crate::{
     Error,
-    common::{LicenseRefMapping, service::delete_doc},
+    common::LicenseRefMapping,
     license::{
         get_sanitize_filename,
         service::{LicenseService, license_export::LicenseExporter},
@@ -23,7 +23,6 @@ use crate::{
         service::{SbomService, sbom::FetchOptions},
     },
     sbom_group::service::SbomGroupService,
-    source_document::model::SourceDocument,
 };
 use actix_web::{HttpResponse, Responder, delete, get, http::header, post, web};
 use config::Config;
@@ -47,7 +46,7 @@ use trustify_module_ingestor::{
     model::IngestResult,
     service::{Cache, Format, IngestorService},
 };
-use trustify_module_storage::service::StorageBackend;
+use trustify_module_storage::service::{StorageBackend, StorageKey};
 
 #[derive(Clone, Debug, PartialEq, Eq, Default, serde::Deserialize, IntoParams)]
 pub struct ModelGetParams {
@@ -78,6 +77,7 @@ pub fn configure(
         .service(get)
         .service(get_sbom_advisories)
         .service(delete)
+        .service(delete_many)
         .service(packages)
         .service(models)
         .service(related)
@@ -388,6 +388,20 @@ pub async fn get_sbom_advisories(
 
 all!(GetSbomAdvisories -> ReadSbom, ReadAdvisory);
 
+async fn delete_blobs<T: StorageBackend>(digests: &[String], storage: &T) {
+    if let Err(e) = storage
+        .delete_many(
+            &digests
+                .iter()
+                .map(|k| StorageKey::from_sha256(k))
+                .collect::<Vec<_>>(),
+        )
+        .await
+    {
+        log::error!("Failed to remove SBOMs from the storage: {e:#?}");
+    }
+}
+
 /// Delete an SBOM
 #[utoipa::path(
     tag = "sbom",
@@ -410,16 +424,51 @@ pub async fn delete(
     let tx = db.begin().await?;
 
     let id = Id::from_str(&id)?;
-    if let Some((v, _, source_document)) = service.fetch_sbom(id, &tx).await?
-        && service.delete_sbom(v.sbom_id, &tx).await?
+    if let Some((v, _, _)) = service.fetch_sbom(id, &tx).await?
+        && let digests = service.delete_sboms(vec![v.sbom_id], &tx).await?
+        && !digests.is_empty()
     {
         tx.commit().await?;
-        if let Err(e) =
-            delete_doc(&SourceDocument::from_entity(&source_document), i.storage()).await
-        {
-            log::warn!("Ignoring {e}");
-        }
+        delete_blobs(&digests, i.storage()).await;
     }
+    Ok(HttpResponse::NoContent().finish())
+}
+
+/// Delete multiple SBOMs
+#[utoipa::path(
+    tag = "sbom",
+    operation_id = "deleteSboms",
+    request_body(
+        content = Vec<String>,
+        description = "List of ids of SBOMs to be deleted",
+        content_type = "application/json",
+    ),
+    responses(
+        (status = 204, description = "Requested SBOMs were deleted or did not exist"),
+    ),
+)]
+#[delete("/v3/sbom")]
+pub async fn delete_many(
+    i: web::Data<IngestorService>,
+    service: web::Data<SbomService>,
+    db: web::Data<db::ReadWrite>,
+    web::Json(body): web::Json<Vec<String>>,
+    _: Require<DeleteSbom>,
+) -> actix_web::Result<impl Responder, Error> {
+    let tx = db.begin().await?;
+
+    let ids = body
+        .into_iter()
+        .filter_map(|x| Uuid::try_parse(&x).ok())
+        .collect();
+
+    let digests = service.delete_sboms(ids, &tx).await?;
+
+    if !digests.is_empty() {
+        tx.commit().await?;
+        delete_blobs(&digests, i.storage()).await;
+    }
+
     Ok(HttpResponse::NoContent().finish())
 }
 
--- a/modules/fundamental/src/sbom/endpoints/test.rs
+++ b/modules/fundamental/src/sbom/endpoints/test.rs
@@ -3,15 +3,18 @@ use crate::{
         Group, GroupRef, UpdateAssignments, create_groups, locate_id, read_assignments,
         resolve_group_refs,
     },
+    purl::model::summary::purl::PurlSummary,
     sbom::model::{SbomPackage, SbomSummary},
     test::{caller, label::Api},
 };
 use actix_http::StatusCode;
 use actix_web::{
     body::MessageBody,
+    dev::ServiceResponse,
     test::{TestRequest, read_body},
 };
 use flate2::bufread::GzDecoder;
+use futures::future::join_all;
 use rstest::rstest;
 use serde_json::{Value, json};
 use std::{collections::HashMap, io::Read, str::FromStr};
@@ -21,9 +24,10 @@ use trustify_common::{id::Id, model::PaginatedResults};
 use trustify_module_ingestor::{model::IngestResult, service::Format};
 use trustify_module_storage::service::{StorageBackend, StorageKey};
 use trustify_test_context::{
-    TrustifyContext, call::CallService, document_bytes, subset::ContainsSubset,
+    IngestionResult, TrustifyContext, call::CallService, document_bytes, subset::ContainsSubset,
 };
 use urlencoding::encode;
+use uuid::Uuid;
 
 #[test_context(TrustifyContext)]
 #[test(actix_web::test)]
@@ -1237,6 +1241,133 @@ async fn delete_sbom(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
     log::debug!("Code: {}", response.status());
     assert_eq!(response.status(), StatusCode::NO_CONTENT);
     assert!(response.into_body().try_into_bytes().unwrap().is_empty());
+
+    Ok(())
+}
+
+/// Test deleting sboms in a batch
+#[test_context(TrustifyContext)]
+#[test(actix_web::test)]
+async fn delete_sboms(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
+    const TASKS: usize = 20;
+    let app = caller(ctx).await?;
+    let storage = &ctx.storage;
+    let [_, id1, id2, id3] = ctx
+        .ingest_documents([
+            // this advisory refers to many packages in both the Quarkus SBOMs
+            "csaf/rhsa-2024-2705.json",
+            "spdx/quarkus-bom-3.2.11.Final-redhat-00001.json",
+            "spdx/quarkus-bom-3.2.12.Final-redhat-00002.json",
+            // this SBOM is totally unrelated with the previous documents
+            "ubi9-9.2-755.1697625012.json",
+        ])
+        .await?
+        .into_uuid();
+
+    async fn get_storage_key(
+        app: &impl CallService,
+        id: &Uuid,
+    ) -> Result<StorageKey, anyhow::Error> {
+        let uri = format!("/api/v3/sbom/urn:uuid:{id}");
+        let req = TestRequest::get().uri(&uri).to_request();
+        let sbom = app.call_and_read_body_json::<SbomSummary>(req).await;
+
+        Ok(sbom.source_document.try_into()?)
+    }
+
+    let key1 = get_storage_key(&app, &id1).await?;
+    assert!(storage.retrieve(key1.clone()).await?.is_some());
+    let key2 = get_storage_key(&app, &id2).await?;
+    assert!(storage.retrieve(key2.clone()).await?.is_some());
+    let key3 = get_storage_key(&app, &id3).await?;
+    assert!(storage.retrieve(key3.clone()).await?.is_some());
+
+    async fn count_purls(app: &impl CallService) -> Option<u64> {
+        let req = TestRequest::get()
+            .uri("/api/v3/purl?total=true")
+            .to_request();
+
+        app.call_and_read_body_json::<PaginatedResults<PurlSummary>>(req)
+            .await
+            .total
+    }
+
+    async fn delete<const N: usize>(app: &impl CallService, ids: [&Uuid; N]) -> ServiceResponse {
+        let req = TestRequest::delete()
+            .uri("/api/v3/sbom")
+            .set_json(ids.into_iter().map(|x| Id::Uuid(*x)).collect::<Vec<_>>())
+            .to_request();
+
+        app.call_service(req).await
+    }
+
+    async fn delete_concurrently<const N: usize>(
+        app: &impl CallService,
+        ids: [&Uuid; N],
+        count: usize,
+    ) -> Vec<ServiceResponse> {
+        let mut tasks = vec![];
+
+        for _ in 0..count {
+            let req = TestRequest::delete()
+                .uri("/api/v3/sbom")
+                .set_json(ids.into_iter().map(|x| Id::Uuid(*x)).collect::<Vec<_>>())
+                .to_request();
+            let app_ref = &app;
+
+            tasks.push(async move { app_ref.call_service(req).await });
+        }
+        join_all(tasks).await
+    }
+
+    assert_eq!(count_purls(&app).await, Some(2087));
+
+    // Delete with no input provided
+    let response = delete(&app, []).await;
+    assert_eq!(response.status(), StatusCode::NO_CONTENT);
+    assert!(response.into_body().try_into_bytes().unwrap().is_empty());
+    // - no purl should be GC'ed (garbage collected)
+    assert_eq!(count_purls(&app).await, Some(2087));
+    // - check the storage
+    assert!(storage.retrieve(key1.clone()).await?.is_some());
+    assert!(storage.retrieve(key2.clone()).await?.is_some());
+    assert!(storage.retrieve(key3.clone()).await?.is_some());
+
+    // Delete quarkus #1
+    let response = delete(&app, [&id1]).await;
+    assert_eq!(response.status(), StatusCode::NO_CONTENT);
+    assert!(response.into_body().try_into_bytes().unwrap().is_empty());
+    // - 4 purls should be GC'ed
+    assert_eq!(count_purls(&app).await, Some(2083));
+    // - check the storage
+    assert!(storage.retrieve(key1.clone()).await?.is_none());
+    assert!(storage.retrieve(key2.clone()).await?.is_some());
+    assert!(storage.retrieve(key3.clone()).await?.is_some());
+
+    // Delete ubi9 and quarkus #1 again, do it concurrently
+    let responses = delete_concurrently(&app, [&id1, &id3], TASKS).await;
+    assert_eq!(responses.len(), TASKS);
+    for response in responses {
+        assert_eq!(response.status(), StatusCode::NO_CONTENT);
+        assert!(response.into_body().try_into_bytes().unwrap().is_empty());
+    }
+    // - 610 purls should be GC'ed
+    assert_eq!(count_purls(&app).await, Some(1473));
+    // - check the storage
+    assert!(storage.retrieve(key1.clone()).await?.is_none());
+    assert!(storage.retrieve(key2.clone()).await?.is_some());
+    assert!(storage.retrieve(key3.clone()).await?.is_none());
+
+    // Delete quarkus #2, and ubi9 and quarkus #1 again
+    let response = delete(&app, [&id2, &id1, &id3]).await;
+    assert_eq!(response.status(), StatusCode::NO_CONTENT);
+    assert!(response.into_body().try_into_bytes().unwrap().is_empty());
+    // - 1 purl should be GC'ed
+    assert_eq!(count_purls(&app).await, Some(1472));
+    // - check the storage
+    assert!(storage.retrieve(key1.clone()).await?.is_none());
+    assert!(storage.retrieve(key2.clone()).await?.is_none());
+    assert!(storage.retrieve(key3.clone()).await?.is_none());
 
     Ok(())
 }
--- a/modules/fundamental/src/sbom/service/sbom.rs
+++ b/modules/fundamental/src/sbom/service/sbom.rs
@@ -17,7 +17,7 @@ use sea_orm::{
 use sea_query::{ColumnType, Expr, JoinType, UnionType, extension::postgres::PgExpr};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
-use std::{collections::HashMap, fmt::Debug, sync::Arc};
+use std::{collections::HashMap, fmt::Debug, sync::Arc, vec::Vec};
 use tracing::{Instrument, info_span, instrument};
 use trustify_common::{
     cpe::Cpe,
@@ -114,80 +114,81 @@ impl SbomService {
         })
     }
 
-    /// delete one sbom
+    /// delete multiple sboms
     #[instrument(skip(self, connection), err(level=tracing::Level::INFO))]
-    pub async fn delete_sbom<C: ConnectionTrait>(
+    pub async fn delete_sboms<C: ConnectionTrait>(
         &self,
-        id: Uuid,
+        ids: Vec<Uuid>,
         connection: &C,
-    ) -> Result<bool, Error> {
+    ) -> Result<Vec<String>, Error> {
+        if ids.is_empty() {
+            return Ok(vec![]);
+        }
+
         // IMPORTANT: Capture qualified_purl IDs before CASCADE deletion.
-        // After SBOM deletion, CASCADE removes sbom_node_purl_ref entries,
+        // After SBOMs deletion, CASCADE removes sbom_node_purl_ref entries,
         // then GC uses the captured IDs to clean up orphaned PURLs.
         let qualified_purl_ids: Vec<Uuid> = sbom_node_purl_ref::Entity::find()
             .select_only()
             .column(sbom_node_purl_ref::Column::QualifiedPurlId)
-            .filter(sbom_node_purl_ref::Column::SbomId.eq(id))
+            .filter(sbom_node_purl_ref::Column::SbomId.is_in(ids.clone()))
             .into_tuple()
             .all(connection)
             .await?;
 
         log::debug!(
-            "Captured {} qualified_purl IDs from SBOM {} for cleanup",
+            "Captured {} qualified_purl IDs from SBOMs {:?} for cleanup",
             qualified_purl_ids.len(),
-            id
+            ids
         );
 
-        // Delete the SBOM - CASCADE will properly delete sbom_package and sbom_node_purl_ref
+        // Delete SBOMs - CASCADE will properly delete sbom_package and sbom_node_purl_ref
         let stmt = Statement::from_sql_and_values(
             connection.get_database_backend(),
-            r#"DELETE FROM sbom WHERE sbom_id=$1 RETURNING source_document_id"#,
-            [id.into()],
+            r#"DELETE FROM sbom WHERE sbom_id = ANY($1) RETURNING source_document_id"#,
+            vec![ids.clone().into()],
         );
 
         let result = connection.query_all(stmt).await?;
-        if result.len() > 1 {
-            return Err(Error::Data(format!("Too many rows deleted for {id}")));
-        }
 
-        for row in &result {
-            let source_document = row.try_get_by_index::<Option<Uuid>>(0)?;
-            if let Some(doc) = source_document {
-                source_document::Entity::delete_by_id(doc)
-                    .exec(connection)
-                    .await?;
-            }
-        }
+        let source_document_ids: Vec<_> = result
+            .iter()
+            .map(|r| r.try_get_by_index::<Uuid>(0))
+            .collect::<Result<_, _>>()?;
+
+        let digests = match source_document_ids.is_empty() {
+            true => vec![],
+            false => source_document::Entity::delete_many()
+                .filter(source_document::Column::Id.is_in(source_document_ids))
+                .exec_with_returning(connection)
+                .await?
+                .iter()
+                .map(|x| x.sha256.clone())
+                .collect(),
+        };
 
         // Cleanup orphaned PURLs if deletion succeeded and we had PURLs to check
-        if !qualified_purl_ids.is_empty() && result.len() == 1 {
-            // Build array parameter for the GC query
-            let array_param = sea_query::Value::Array(
-                sea_query::ArrayType::Uuid,
-                Some(Box::new(
-                    qualified_purl_ids
-                        .iter()
-                        .map(|&id| sea_query::Value::Uuid(Some(Box::new(id))))
-                        .collect(),
-                )),
-            );
-
+        if !qualified_purl_ids.is_empty() {
             let gc_stmt = Statement::from_sql_and_values(
                 connection.get_database_backend(),
                 // it looks much more readable in an SQL file
                 include_str!("gc_purls_after_sbom_deletion.sql"),
-                vec![array_param],
+                vec![qualified_purl_ids.into()],
             );
 
-            let gc_result = connection.execute(gc_stmt).await?;
+            let gc_result = connection
+                .execute(gc_stmt)
+                .instrument(info_span!("delete_sboms::gc"))
+                .await?;
             log::debug!(
-                "Cleaned up {} orphaned purl records after SBOM {} deletion",
+                "Cleaned up {} orphaned purl records after SBOMs {:?} deletion",
                 gc_result.rows_affected(),
-                id
+                ids,
             );
         }
 
-        Ok(result.len() == 1)
+        // Return a list of keys of blobs to be removed from the storage
+        Ok(digests)
     }
 
     /// fetch all SBOMs
@@ -1334,8 +1335,20 @@ mod test {
 
         let service = SbomService::new(PaginationCache::for_test());
 
-        assert!(service.delete_sbom(sbom_v1.sbom.sbom_id, &ctx.db).await?);
-        assert!(!service.delete_sbom(sbom_v1.sbom.sbom_id, &ctx.db).await?);
+        assert!(
+            // A digest is expected
+            !service
+                .delete_sboms(vec![sbom_v1.sbom.sbom_id], &ctx.db)
+                .await?
+                .is_empty()
+        );
+        assert!(
+            // No SBOM, no digest
+            service
+                .delete_sboms(vec![sbom_v1.sbom.sbom_id], &ctx.db)
+                .await?
+                .is_empty()
+        );
 
         Ok(())
     }
--- a/modules/storage/src/service/mod.rs
+++ b/modules/storage/src/service/mod.rs
@@ -13,6 +13,7 @@ use crate::service::fs::FileSystemBackend;
 use bytes::Bytes;
 use futures::Stream;
 use hex::ToHex;
+use std::collections::HashMap;
 use std::fmt::{Debug, Display, Formatter};
 use std::future::Future;
 use tokio::io::AsyncRead;
@@ -27,8 +28,28 @@ pub enum StoreError<B: Debug> {
     Backend(#[source] B),
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum DeleteManyError<E> {
+    /// The whole operation failed
+    #[error("{0}")]
+    Generic(E),
+    /// Individual delete errors
+    #[error("individual delete errors: {0}")]
+    Individual(HashMap<StorageKey, E>),
+}
+
 #[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct StorageKey(String);
+
+impl StorageKey {
+    /// Create a storage key from the digest.
+    ///
+    /// The digest must be SHA256 (without "sha256:" prefix) and user must
+    /// ensure this (for simplicity we omitted checks here).
+    pub fn from_sha256(digest: &str) -> Self {
+        Self(digest.into())
+    }
+}
 
 impl Display for StorageKey {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
@@ -112,4 +133,29 @@ pub trait StorageBackend {
     /// (i.e., return `Ok(())`) and not result in an error. This ensures consistent
     /// behavior across all backends.
     fn delete(&self, key: StorageKey) -> impl Future<Output = Result<(), Self::Error>>;
+
+    /// Batch variant of `delete`, with the same requirements.
+    fn delete_many(
+        &self,
+        keys: &[StorageKey],
+    ) -> impl Future<Output = Result<(), DeleteManyError<Self::Error>>> {
+        async move {
+            // Default implementation: delete blobs one by one (can be overridden in
+            // implementations for particular backends).
+            let mut whats_wrong = HashMap::new();
+
+            for key in keys {
+                // Do not stop on the first error---instead, accumulate all errors
+                // into one and return it when we are done (we do not want one faulty
+                // request to stop the entire deletion process)
+                if let Err(e) = self.delete(key.clone()).await {
+                    whats_wrong.insert(key.clone(), e);
+                }
+            }
+            match whats_wrong.is_empty() {
+                true => Ok(()),
+                false => Err(DeleteManyError::Individual(whats_wrong)),
+            }
+        }
+    }
 }
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2749,6 +2749,23 @@ paths:
                 $ref: '#/components/schemas/IngestResult'
         '400':
           description: One or more group IDs are invalid or do not exist
+    delete:
+      tags:
+      - sbom
+      summary: Delete multiple SBOMs
+      operationId: deleteSboms
+      requestBody:
+        description: List of ids of SBOMs to be deleted
+        content:
+          application/json:
+            schema:
+              type: array
+              items:
+                type: string
+        required: true
+      responses:
+        '204':
+          description: Requested SBOMs were deleted or did not exist
   /api/v3/sbom-labels:
     get:
       tags:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8645,6 +8645,7 @@ dependencies = [
  "csv",
  "cve",
  "flate2",
+ "futures",
  "futures-util",
  "hex",
  "humantime",
--- a/cli/src/api/client.rs
+++ b/cli/src/api/client.rs
@@ -2,6 +2,7 @@ use std::{sync::Arc, time::Duration};
 
 use indicatif::style::TemplateError;
 use reqwest::{Client, RequestBuilder, StatusCode};
+use serde::Serialize;
 use thiserror::Error;
 use tokio::{sync::RwLock, time::sleep};
 
@@ -157,6 +158,21 @@ impl ApiClient {
         self.execute_with_retry(|| async {
             let url = self.url(path);
             let request = self.client.delete(&url);
+            let response = self.authorize(request).await.send().await?;
+            self.handle_response(response).await
+        })
+        .await
+    }
+
+    /// Perform a DELETE request with JSON and retry logic
+    pub async fn delete_with_json<T: Serialize + ?Sized>(
+        &self,
+        path: &str,
+        json: &T,
+    ) -> Result<String, ApiError> {
+        self.execute_with_retry(|| async {
+            let url = self.url(path);
+            let request = self.client.delete(&url).json(json);
             let response = self.authorize(request).await.send().await?;
             self.handle_response(response).await
         })
--- a/cli/src/api/sbom.rs
+++ b/cli/src/api/sbom.rs
@@ -15,8 +15,8 @@ use tokio::sync::Mutex;
 
 use super::client::{ApiClient, ApiError};
 use crate::common::{
-    DeleteEntry, DeleteResult, ListParams, PruneParams, build_prune_query, delete_entries,
-    new_delete_result,
+    DeleteEntry, DeleteResult, DeletedResult, FailedResult, ListParams, PruneParams,
+    build_prune_query, delete_entries, new_delete_result,
 };
 
 const SBOM_PATH: &str = "/v3/sbom";
@@ -300,6 +300,11 @@ pub async fn delete(client: &ApiClient, id: &str) -> Result<(), ApiError> {
     Ok(())
 }
 
+/// Delete multiple SBOMs by their IDs in one batch
+pub async fn delete_many(client: &ApiClient, ids: &Vec<String>) -> Result<String, ApiError> {
+    client.delete_with_json(SBOM_PATH, ids).await
+}
+
 /// Delete all SBOMs by listing and deleting each one
 pub async fn delete_by_query(
     client: &ApiClient,
@@ -398,8 +403,37 @@ pub async fn prune(client: &ApiClient, params: &PruneParams) -> Result<DeleteRes
         return Ok(new_delete_result(total));
     }
 
+    let ids = entries.iter().map(|x| x.id.clone()).collect();
+    let mut deleted = vec![];
+    let mut failed = vec![];
+
     // Perform the actual deletion
-    delete_list(client, entries, params.concurrency).await
+    match delete_many(client, &ids).await {
+        Ok(_) => deleted.extend(entries.iter().map(|x| DeletedResult {
+            id: x.id.clone(),
+            identifier: x.identifier.clone(),
+        })),
+        Err(e) => failed.extend(entries.iter().map(|x| FailedResult {
+            id: x.id.clone(),
+            identifier: x.identifier.clone(),
+            error: e.to_string(),
+        })),
+    }
+
+    let deleted_total = deleted.len() as u32;
+    let skipped = vec![];
+    let skipped_total = 0;
+    let failed_total = failed.len() as u32;
+
+    Ok(DeleteResult {
+        deleted,
+        deleted_total,
+        skipped,
+        skipped_total,
+        failed,
+        failed_total,
+        total,
+    })
 }
 
 /// Read delete entries from a file
--- a/modules/fundamental/Cargo.toml
+++ b/modules/fundamental/Cargo.toml
@@ -62,6 +62,7 @@ csaf = { workspace = true }
 csv = { workspace = true }
 cve = { workspace = true }
 flate2 = { workspace = true }
+futures = { workspace = true }
 hex = { workspace = true }
 humantime = { workspace = true }
 jsonpath-rust = { workspace = true }
--- a/modules/fundamental/src/sbom/service/test.rs
+++ b/modules/fundamental/src/sbom/service/test.rs
@@ -547,7 +547,13 @@ async fn delete_sbom_orphaned_purl_test(ctx: &TrustifyContext) -> Result<(), any
     let tx = ctx.db.begin().await?;
     let sbom_service = SbomService::new(PaginationCache::for_test());
     // delete the UBI SBOM
-    assert!(sbom_service.delete_sbom(ubi9_sbom.id.parse()?, &tx).await?);
+    assert!(
+        // Digest is expected
+        !sbom_service
+            .delete_sboms(vec![ubi9_sbom.id.parse()?], &tx)
+            .await?
+            .is_empty()
+    );
     tx.commit().await?;
 
     // it should not leave behind orphaned purls
@@ -558,9 +564,11 @@ async fn delete_sbom_orphaned_purl_test(ctx: &TrustifyContext) -> Result<(), any
     // delete the quarkus sbom....
     let tx = ctx.db.begin().await?;
     assert!(
-        sbom_service
-            .delete_sbom(quarkus_sbom.id.parse()?, &tx)
+        // Digest is expected
+        !sbom_service
+            .delete_sboms(vec![quarkus_sbom.id.parse()?], &tx)
             .await?
+            .is_empty(),
     );
     tx.commit().await?;
 
@@ -620,7 +628,8 @@ async fn delete_sbom_preserves_advisory_referenced_packages(
     let sbom_uuid = results[1].id.parse().expect("SBOM should have a UUID");
     let tx = ctx.db.begin().await?;
     assert!(
-        service.delete_sbom(sbom_uuid, &tx).await?,
+        // Digest is expected
+        !service.delete_sboms(vec![sbom_uuid], &tx).await?.is_empty(),
         "SBOM should be deleted"
     );
     tx.commit().await?;
@@ -659,7 +668,8 @@ async fn delete_sbom_preserves_advisory_referenced_packages(
     let sbom_uuid = results[2].id.parse().expect("SBOM should have a UUID");
     let tx = ctx.db.begin().await?;
     assert!(
-        service.delete_sbom(sbom_uuid, &tx).await?,
+        // Digest is expected
+        !service.delete_sboms(vec![sbom_uuid], &tx).await?.is_empty(),
         "SBOM should be deleted"
     );
     tx.commit().await?;
@@ -682,7 +692,11 @@ async fn delete_sbom_preserves_advisory_referenced_packages(
     let ubi_sbom_uuid = results[3].id.parse().expect("SBOM should have a UUID");
     let tx = ctx.db.begin().await?;
     assert!(
-        service.delete_sbom(ubi_sbom_uuid, &tx).await?,
+        // Digest is expected
+        !service
+            .delete_sboms(vec![ubi_sbom_uuid], &tx)
+            .await?
+            .is_empty(),
         "SBOM should be deleted"
     );
     tx.commit().await?;
--- a/modules/storage/src/service/s3.rs
+++ b/modules/storage/src/service/s3.rs
@@ -1,8 +1,8 @@
 use crate::{
     config::S3Config,
     service::{
-        StorageBackend, StorageKey, StorageResult, StoreError, compression::Compression,
-        temp::TempFile,
+        DeleteManyError, StorageBackend, StorageKey, StorageResult, StoreError,
+        compression::Compression, temp::TempFile,
     },
 };
 use anyhow::{Context, anyhow, bail};
@@ -15,6 +15,7 @@ use aws_sdk_s3::{
     },
     operation::get_object::GetObjectError,
     primitives::FsBuilder,
+    types::{Delete, ObjectIdentifier},
 };
 use aws_smithy_http_client::tls::{Provider, TlsContext, TrustStore, rustls_provider::CryptoMode};
 use aws_smithy_types::endpoint::Endpoint;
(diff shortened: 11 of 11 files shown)


ruromero (dismissed):
## Verification Report

I built and ran the server from this branch (`AUTH_DISABLED=true cargo run --bin trustd api --devmode`), ingested 2 test SBOMs (`etc/test-data/quay/sbom.json` and `etc/test-data/cyclonedx/openssl-3.0.7-18.el9_2.cdx_1.6.sbom.json`), and exercised the new endpoints via curl.

### What works

- **GET /api/v3/sbom/prune** returns the correct response shape with `fetched` count and SBOM list
- **Limit and offset** work correctly — `limit=1` caps results, `offset=1` skips the first item
- **DELETE /api/v3/sbom/prune** actually deletes SBOMs from the database (confirmed via `GET /api/v2/sbom`)
- **DELETE with limit=0** correctly prevents any deletion (safety mechanism)
- **Date filtering** works (`q=ingested>1 day ago` matches expected SBOMs)
- **Invalid field names** in `q` return a helpful 400 error listing valid fields
- **Refactored single-delete** (`DELETE /api/v2/sbom/{id}`) returns 204 and removes the SBOM correctly

### Bug: SBOM state never updates in the response

After `DELETE /api/v3/sbom/prune?offset=0&limit=1`, the response is:

```json
{
  "fetched": 2,
  "deleted": 1,
  "sboms": [
    {
      "sbom_id": "urn:uuid:019e6d7d-da8c-7b23-9094-a207496ab2e8",
      "state": "fetched"
    }
  ]
}
```

The `deleted` counter is correct, but each SBOM's `state` remains `"fetched"` instead of `{"deleted": "sha512:..."}` as the ADR specifies.

**Root cause:** `commit_deleted()` at `modules/fundamental/src/sbom/model/mod.rs:347` iterates over `self.sboms.clone()`:

```rust
for sbom in &mut self.sboms.clone().unwrap_or(vec![]) {
    sbom.state = DeletableSbomState::Deleted(id2key.get(&doc).cloned());
    self.deleted += 1;
}
```

`.clone()` creates a throwaway copy — state mutations apply to the clone and are never written back to `self.sboms`. The same pattern exists in `commit_failure()` at line 377. The counters work because they write to `self` directly.

This means callers cannot determine which specific SBOMs were deleted vs skipped vs failed, and the `StorageKey` intended for storage cleanup is lost from the response.

### Bug: CLI path mismatch

The CLI defines `PRUNE_PATH = "/v3/prune"` at `cli/src/api/sbom.rs:24`, but the server registers the endpoint at `/v3/sbom/prune` (`modules/fundamental/src/sbom/endpoints/mod.rs:448,474`). Confirmed: `curl /api/v3/prune?offset=0` returns 404.

### Minor: `offset` is required despite ADR default

`DeleteParams.offset` has no `#[serde(default)]`, making it required. The ADR says the default should be 0. Omitting `offset` returns 400: `"missing field 'offset'"`.

i386x:
Thanks @ruromero for the report. Will look closer one the issues.
EDIT: Requested changes should be now addressed.

modules/fundamental/src/sbom/model/mod.rs
```diff
@@ -106,6 +173,249 @@ impl<P: IntoPackage> SbomSummary<P> {
...
+
+        let id2key: HashMap<Uuid, StorageKey> = keys
+            .iter()
+            .filter_map(|key| key.clone().try_into().map_or(None, |x| Some((key.id, x))))
+            .collect();
+
+        self.skipped = 0;
+        self.deleted = 0;
+
+        for sbom in &mut self.sboms.clone().unwrap_or(vec![]) {
```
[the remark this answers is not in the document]
i386x: Done.

modules/fundamental/src/sbom/model/mod.rs
```diff
@@ -106,6 +173,249 @@ impl<P: IntoPackage> SbomSummary<P> {
...
+            skipped: 0,
+            deleted: 0,
+            failed: 0,
+            limit: DeleteParams::default().limit,
+            sboms: None,
+        }
+    }
+}
+
+impl DeleteContext {
```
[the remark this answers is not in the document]
i386x: Done.

modules/fundamental/src/sbom/model/mod.rs
```diff
@@ -9,16 +9,88 @@ use crate::{
...
+        value: T,
+        connection: &C,
+    ) -> impl Future<Output = Result<Self, Self::Error>> + Send
+    where
+        C: ConnectionTrait;
+}
+
+#[derive(Serialize, Deserialize, Debug, Copy, Clone, IntoParams)]
+#[into_params(parameter_in = Query)]
+pub struct DeleteParams {
```
ctron: This smells like the existing `Paginated` type. Why do we need a new one?

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -308,6 +356,50 @@ impl SbomService {
         Ok(PaginatedResults { total, items })
     }
 
+    /// Fetch SBOMs selected for pruning
+    #[instrument(skip(self, connection), err(level=tracing::Level::INFO))]
+    pub async fn fetch_prunable_sboms<C: ConnectionTrait>(
```
ctron: I'm not sure why we can't re-use the existing list function?

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -43,6 +45,22 @@ use trustify_entity::{
     versioned_purl, vulnerability,
 };
 
+struct SeaQueryValueWrapper(sea_query::Value);
```
ctron: Why do we need this?

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -428,6 +433,74 @@ pub async fn delete(
...
+    responses(
+        (status = 200, description = "Candidate SBOMs to be pruned", body = DeleteContext),
+    ),
+)]
+#[get("/v3/sbom/prune")]
+pub async fn prunable(
+    service: web::Data<SbomService>,
+    db: web::Data<Database>,
+    web::Query(search): web::Query<Query>,
+    web::Query(params): web::Query<DeleteParams>,
```
ctron: Why can't we use `Paginated`?

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -428,6 +433,74 @@ pub async fn delete(
...
+    Ok(HttpResponse::Ok().json(ctx))
+}
+
+/// Prune SBOMs
+#[utoipa::path(
+    tag = "sbom",
+    operation_id = "pruneSboms",
+    params(
+        Query,
+        DeleteParams,
```
ctron: Why can't we use `Paginated`?

modules/fundamental/src/sbom/model/mod.rs
```diff
@@ -106,6 +178,251 @@ impl<P: IntoPackage> SbomSummary<P> {
...
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, ToSchema, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum DeletableSbomState {
+    /// An SBOM has been fetched, its `id`, `document_id`, and `source_document_id`
+    /// are present
+    #[default]
+    Fetched,
+    /// An SBOM has been successfully removed from the database, use the provided
+    /// storage key to remove its leftovers from the storage
+    Deleted(Option<StorageKey>),
```
ctron: The `StorageKey` type should not be leaked. IDs are strings.

modules/fundamental/src/sbom/model/mod.rs
```diff
@@ -106,6 +178,251 @@ impl<P: IntoPackage> SbomSummary<P> {
...
+/// the id of associated document
+#[derive(Serialize, Deserialize, Debug, Clone, ToSchema, Default)]
+pub struct DeletableSbom {
+    #[serde(flatten)]
+    pub head: DeletableSbomHead,
+    /// The deletion status of the SBOM
+    pub state: DeletableSbomState,
+}
+
+/// Source document storage keys for later deletion from the storage
```
ctron: I'm not sure why we need this. It seems we're extracting only the id anyway.

modules/fundamental/src/sbom/model/mod.rs
```diff
@@ -106,6 +178,251 @@ impl<P: IntoPackage> SbomSummary<P> {
...
+            return Ok(key);
+        }
+
+        Err(IdError::MissingPrefix)
+    }
+}
+
+/// The context of SBOM prune/delete operation
+#[derive(Serialize, Deserialize, Debug, Clone, ToSchema)]
+pub struct DeleteContext {
```
ctron: I don't understand why we need this?

ctron (changes requested):
Maybe I'm missing something. But this seems overly complex. Or I don't understand the idea behind it.

For fetching SBOMs we already have an endpoint.

For deleting SBOMs, we would need a "multi-delete" endpoint. Ideally one that accepts the query request as the list endpoint does.

So we could just have a `delete_sboms` service method. Not sure if we can bulk delete dependent entities. But we could try that. Internally, we should delete by ID only. Not the SHAs.

i386x:
Why there are [`Database` not found in this scope](https://github.com/guacsec/trustify/actions/runs/26821525026/job/79077135135?pr=2364#step:8:1448) errors when `Database` is clearly imported and used in previous endpoints?

Also it compiles successfully on my machine.

modules/fundamental/src/sbom/model/mod.rs
i386x: Yes, that's true and I am aware of this, however I struggled how to override the default value from 25 to 100 in the prune API endpoint and left it 25 for the rest of API endpoints so I introduced a new type.

i386x: However, I have an idea to make `Paginated` a generic type with the default limit as a parameter. This could help to get rid off duplicates.

modules/fundamental/src/sbom/service/sbom.rs
i386x: You mean `fetch_sboms`? This is the point of the whole effort. `fetch_sboms` executes a lot of queries and subqueries and pulls items that are not needed for deletion operation. That was point out by @bxf12315 in TC-3934.

i386x:
> Maybe I'm missing something. But this seems overly complex. Or I don't understand the idea behind it.

Well, the idea:
- Optimize fetching SBOMs for deletion operation only (the current `fetch_sboms` executes 6 too long queries). As you mentioned, only `sbom_id`s can suffice. However, it can be also nice to have `document_id`s fetched for better user experience (user can deduce from it some basic info about SBOM). And while we are fetching ids, why not to fetch also `source_document_id`s and SHAs so we do not need to fetch them again when they become useful?
- Preserve `trustify-cli` UX compatibility when doing `prune`. That is, I need to fetch at least `sbom_id` and `document_id` since these are returned to a CLI user in form of `DeletedItem`, `SkippedItem`, or `FailedItem`, which additionally contains also an error (this become a challenge when doing multi-deletion since for a bunch of SBOMs only one error can be returned).

The design:
- Provide `GET` endpoint listing only information needed for deletion. This endpoint is used when user pass `--dry-run` to the `trustify-cli prune` (you have a good point about only ids since at this point user do not need `source_document_id`s&mdash;a boolean parameter to `fetch_prunable_sboms` could help to refine what is really needed).
- Provide `POST` endpoint doing the deletion. This first call `fetch_prunable_sboms` and store fetched data to the `DeletionContext`. Then the deletion is done over the context. Why we have `DeletionContext`? To have something which is easily convertible to `trustify-cli`'s `DeleteResult`.

Can this be done more easily? Yes, with loss of `trustify-cli prune` UX compatibility:
- `GET` returns only `sbom_id`s
- `POST` takes a list of ids from `GET` and performs delete, returning the list of fetched ids and possible error if the deletion failed (note that this does not fulfill the ADR, which requires detailed info about each deleted SBOM).

> 
> For fetching SBOMs we already have an endpoint.

Unoptimized one (according to @bxf12315 analyzis).

> For deleting SBOMs, we would need a "multi-delete" endpoint. Ideally one that accepts the query request as the list endpoint does.

`POST /api/v3/sbom/prune` is trying to do this.

> So we could just have a `delete_sboms` service method. Not sure if we can bulk delete dependent entities. But we could try that. Internally, we should delete by ID only. Not the SHAs.

Arent SHAs used to locate and delete source documents on storage server?

modules/fundamental/src/sbom/model/mod.rs
i386x: To save the running the very similar query again when SHAs will be needed for the source document deletion from the storage server. This comes from joined `sbom` and `source_document` tables. `id` is extracted to use it to delete a record from the `source_document` table. SHAs are used later to delete the source document from the storage.

modules/fundamental/src/sbom/endpoints/mod.rs
i386x: I think `Paginated` can be used here.

modules/fundamental/src/sbom/endpoints/mod.rs
i386x: I think `Paginated` can be used here.

modules/fundamental/src/sbom/model/mod.rs
i386x: Ok, will work on it.

modules/fundamental/src/sbom/service/sbom.rs
i386x: To convert `Vec<Uuid>` to `sea_query::Value`. The conversion is done on two places, so I factored the code out to prevent duplicating it. Unfortunately, Rust do not allow me to write `imp From<Vec<Uuid>> for sea_query::Value` since neither `Uuid` nor `sea_query::Value` are defined in this module. Hence the wrapper.

modules/fundamental/src/sbom/model/mod.rs
i386x: I tried to explain it in the reply.

modules/fundamental/src/sbom/service/sbom.rs
ctron: Right, but since then, thing have changed. The query time is much more reasonable. And if there's a slight overhead, that's ok, compared to the overhead of maintaining a lot of custom code for a single use case.

modules/fundamental/src/sbom/service/sbom.rs
ctron: Why not just use what's already there?

```rust
fn test() {
    use uuid::Uuid;
    let ids = vec![Uuid::new_v4(), Uuid::new_v4(), Uuid::new_v4()];
    let _ids: sea_orm::Value = ids.into();
}
```

modules/fundamental/src/sbom/model/mod.rs
ctron: I think having a default of 25 is just fine. If you want more, it can be provided. Having a duplicated type just because of the expectation of a different default for a single operations feels like overkill. Please just use `Paginated` as is.

modules/fundamental/src/sbom/model/mod.rs
ctron: Ok I see. The fact that we implictly use the SHA256 as ID threw me off. We still have too many IDs.

So what we need is the SHA256 (for storage), and the SBOM ID (for referencing on the API). If that's the case, then a simple lookup of ID -> (ID, SHA256) would be sufficient, right? Returning the ID in the result to be able to create a hashmap for lookup.

However, I'm not sure if we need all this complexity. What about having a simple approach of:

* Perform ID -> SHA256 lookup
* Delete all SBOMs by ID
* If ok, delete all BLOBs by SHA256

If the delete (or commit) fails, don't delete (as they are still in the database).

modules/fundamental/src/sbom/model/mod.rs
```diff
@@ -9,16 +9,88 @@ use crate::{
...
+    requested_field::RequestedField,
+};
 use trustify_entity::{
     labels::Labels, relationship::Relationship, sbom, sbom_node, sbom_package, source_document,
 };
-use utoipa::ToSchema;
+use trustify_module_storage::service::StorageKey;
+use utoipa::{IntoParams, ToSchema};
+
+pub trait TryFromDb<T, C>: Sized {
```
ctron: It seems we're only using this once.

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -428,6 +433,74 @@ pub async fn delete(
...
+    let mut ctx = service.fetch_prunable_sboms(search, params, &tx).await?;
+
+    service.delete_sboms(&mut ctx, &tx).await?;
+    if ctx.deleted > 0 {
+        tx.commit().await?;
+
+        let storage = i.storage();
+        for sbom in ctx.sboms.clone().unwrap_or(vec![]) {
+            if let DeletableSbomState::Deleted(Some(key)) = &sbom.state
+                && let Err(e) = storage.delete(key.clone()).await
```
ctron: I don't think this logic should be in the endpoint. But in the service.

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -428,6 +433,74 @@ pub async fn delete(
...
+            {
+                log::warn!("Ignoring {e}");
```
ctron: There's no real context information attached to it. Not sure this is helpful.

docs/adrs/00012-sbom-pruning-api.md
```diff
@@ -0,0 +1,223 @@
...
+      "state": "fetched"
+    }
+  ]
+}
+```
+
+Again, `GET` will only return what will be deleted, limited to first N items
+(`limit` field).
+
+`DELETE` endpoint should return the following (example):
```
ctron: What HTTP error codes are there to expect? What happens when one SBOM deletes, but another doesn't?

modules/fundamental/src/sbom/model/mod.rs
```diff
@@ -106,6 +178,251 @@ impl<P: IntoPackage> SbomSummary<P> {
...
+        self.skipped = 0;
+        self.deleted = 0;
+
+        if let Some(sboms) = self.sboms.as_mut() {
+            for sbom in sboms.iter_mut() {
+                if sbom.state != DeletableSbomState::Fetched {
+                    continue;
+                }
+                let doc = sbom.head.source_document_id;
+                match docs.contains(&doc) {
```
ctron: Instead of using `contains` + `get`, why not use `entry`?

ctron: Sorry, just noticed there are two different maps.

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -121,73 +139,103 @@ impl SbomService {
...
+        &self,
+        ctx: &mut DeleteContext,
+        connection: &C,
+    ) -> Result<(), Error> {
+        let ids = ctx.get_fetched_sbom_ids();
+
+        if ids.is_empty() {
+            return Ok(());
+        }
+        let nids: u64 = ids.len() as u64;
```
ctron: Why do we need to cast this `u64`?

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -121,73 +139,103 @@ impl SbomService {
...
+
         // IMPORTANT: Capture qualified_purl IDs before CASCADE deletion.
-        // After SBOM deletion, CASCADE removes sbom_node_purl_ref entries,
+        // After SBOMs deletion, CASCADE removes sbom_node_purl_ref entries,
         // then GC uses the captured IDs to clean up orphaned PURLs.
         let qualified_purl_ids: Vec<Uuid> = sbom_node_purl_ref::Entity::find()
             .select_only()
             .column(sbom_node_purl_ref::Column::QualifiedPurlId)
-            .filter(sbom_node_purl_ref::Column::SbomId.eq(id))
+            .filter(match nids {
```
ctron: Can't we just pass in a array of 1?

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -121,73 +139,103 @@ impl SbomService {
...
         );
 
-        let result = connection.query_all(stmt).await?;
-        if result.len() > 1 {
-            return Err(Error::Data(format!("Too many rows deleted for {id}")));
+        let result = connection
+            .query_all(stmt)
+            .await
+            .inspect_err(|e| ctx.commit_failure(e.to_string()))?;
+        let nres: u64 = result.len() as u64;
```
ctron: Why do we need to cast this to `u64`?

modules/fundamental/src/sbom/model/mod.rs
ctron: This feels like building some kind of transaction layer. Having this in the `model` module feels wrong. Either because it is internal logic (which should not be leaked), or because it's not really needed in the model.

cli/src/common/mod.rs
```diff
@@ -108,6 +109,56 @@ pub struct FailedResult {
...
+                        id,
+                        identifier,
+                        error: error.clone(),
+                    });
+                }
+                _ => (),
+            }
+        }
+
+        let deleted_total = deleted.len() as u32;
```
ctron: Why do we need to cast them to `u32`?

cli/src/common/mod.rs
```diff
@@ -108,6 +109,56 @@ pub struct FailedResult {
...
 
+impl From<DeleteContext> for DeleteResult {
+    fn from(value: DeleteContext) -> Self {
+        let sboms = value.sboms.unwrap_or(vec![]);
+
+        let mut deleted = vec![];
+        let mut skipped = vec![];
+        let mut failed = vec![];
+
+        for sbom in &sboms {
```
ctron: Can we consume the entries? And skip cloning later?

cli/Cargo.toml
```diff
@@ -11,6 +11,8 @@ name = "trustify"
 path = "src/main.rs"
 
 [dependencies]
+trustify-module-fundamental = { workspace = true }
```
ctron: This worries me a bit. As we're pulling in a backend internal dependency into the CLI. I understand why, as we need/want to re-use the API model definition. However, the CLI is a change in architecture.

I'm not sure we have to solve this in this PR. However, we should solve it. I could imagine having a dedicated `client` or `model` crate for this.

Please create a new issue JIRA and GitHub to address this in the future.

docs/adrs/00012-sbom-pruning-api.md
```diff
@@ -0,0 +1,223 @@
...
+- **Search API:** Supports date filtering (`ingested<30 days ago`) and label
+  filtering
+- **Gap:** No server-side bulk prune operation with tracking
+
+### Requirements
+
+- Server-side API endpoint for bulk SBOM pruning
+- Reuse existing filter query syntax
+- Safety (requires proper permission to perform pruning operations)
+- Audit logging for compliance
```
ctron: We should define what that means. To me "audit logging for compliance" is not logging things to the console.

docs/adrs/00012-sbom-pruning-api.md
```diff
@@ -0,0 +1,223 @@
...
+- Foundation for future scheduled pruning service
+
+## Decision
```
ctron: Based on the document so far, I don't see how we arrived at this decision.

I don't think we need a new `GET` endpoint. I think, as a good first step, we could add a "delete many SBOMs` endpoint. e.g. `DELETE /api/v3/sbom` with a JSON payload, containing the IDs.

That would be a good first step. I am sure, we could later extend this with a query (q style vs providing IDs). But looking at the current implementation, we don't gain much (one round-trip, same queries) if we start with a two step approach.

docs/adrs/00012-sbom-pruning-api.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Responses
+
+`GET` endpoint should return the following (example):
+
+```json
+{
+  "fetched": 2,
+  "skipped": 0,
+  "deleted": 0,
```
ctron: Why would the `GET` endpoint return `deleted`? Or re-return the limit?

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -121,73 +139,103 @@ impl SbomService {
...
 
-        for row in &result {
-            let source_document = row.try_get_by_index::<Option<Uuid>>(0)?;
-            if let Some(doc) = source_document {
-                source_document::Entity::delete_by_id(doc)
-                    .exec(connection)
-                    .await?;
+        let source_document_ids = ctx.commit_deleted(&result, connection).await?;
+
+        if !source_document_ids.is_empty() {
```
ctron: Instead of having an outer `if`, could this be added at "match 0"?

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -121,73 +139,103 @@ impl SbomService {
...
     ) -> Result<bool, Error> {
+        let mut ctx = DeleteContext::try_from_db(id, connection).await?;
+        self.delete_sboms(&mut ctx, connection).await?;
+
+        Ok(ctx.deleted == 1)
+    }
+
+    /// delete multiple sboms
+    #[instrument(skip(self, connection), err(level=tracing::Level::INFO))]
+    pub async fn delete_sboms<C: ConnectionTrait>(
```
ctron, line 119: Trying to understand what's going on here. I think this can be simplified. Why not just pass in an array, skip the "if 1" checks.

modules/fundamental/src/sbom/model/mod.rs
i386x: Ok.

modules/fundamental/src/sbom/service/sbom.rs
i386x: I see. That means that we should focus on batch/bulk deletion then, which also supports https://github.com/guacsec/trustify/blob/main/CONVENTIONS.md#missing-batchbulk-operations

modules/fundamental/src/sbom/service/sbom.rs
i386x: It seems that this works, which means that
```rust
            // Build array parameter for the GC query
            let array_param = sea_query::Value::Array(
                sea_query::ArrayType::Uuid,
                Some(Box::new(
                    qualified_purl_ids
                        .iter()
                        .map(|&id| sea_query::Value::Uuid(Some(Box::new(id))))
                        .collect(),
                )),
            );
```
from https://github.com/guacsec/trustify/blob/main/modules/fundamental/src/sbom/service/sbom.rs#L164 is probably redundant.

modules/fundamental/src/sbom/model/mod.rs
i386x: I like the idea. But are we sure that SHA256 will be already available? Why we have then another two SHAs, SHA384 and SHA512, in the database record?

ctron: Because we can? :grimacing: I guess the problem is, and this may lead to confusion, that we implicitly use the SHA256 returned by the storage as an ID for linking to/from the source document (which links to the SBOM).

Ideally the `StorageService` returns a `StorageKey` in addition, which then gets used as "storage key/id". But that would mean some refactoring and would be outside of this ADR/PR.

ctron:
> Why there are [`Database` not found in this scope](https://github.com/guacsec/trustify/actions/runs/26821525026/job/79077135135?pr=2364#step:8:1448) errors when `Database` is clearly imported and used in previous endpoints?
> 
> Also it compiles successfully on my machine.

Maybe you need to rebase?

modules/fundamental/src/sbom/model/mod.rs
i386x: I will try to get rid off it and redesign `delete_sboms`.

ctron:
> > Maybe I'm missing something. But this seems overly complex. Or I don't understand the idea behind it.
> 
> Well, the idea:
> 
>     * Optimize fetching SBOMs for deletion operation only (the current `fetch_sboms` executes 6 too long queries). As you mentioned, only `sbom_id`s can suffice. However, it can be also nice to have `document_id`s fetched for better user experience (user can deduce from it some basic info about SBOM). And while we are fetching ids, why not to fetch also `source_document_id`s and SHAs so we do not need to fetch them again when they become useful?
> 

IIRC, this has already been optimized. I am sure, there are other cases which would benefit from a father SBOM query, and which don't need additional data. So optimizing the original API endpoint would serve those as well. Compared to introducing a very opinionated and specific endpoint. Which deviates from the existing patterns a lot. I don't think that's worth the complexity.

>     * Preserve `trustify-cli` UX compatibility when doing `prune`. That is, I need to fetch at least `sbom_id` and `document_id` since these are returned to a CLI user in form of `DeletedItem`, `SkippedItem`, or `FailedItem`, which additionally contains also an error (this become a challenge when doing multi-deletion since for a bunch of SBOMs only one error can be returned).

Not sure we ever released the CLI. Or consider it stable. I think the CLI should be easy to use, and easy to understand. When I ask it to delete a batch of items, I would expect them to be deleted. Or the operation failed. But not some deleted, some skipped, and some failed. I do understand there is the transaction gap between the S3 storage and the DB. However, from a user perspective, the DB either deletes all documents, or none. Any mishap with the S3 storage should be logged and flagged for an administrator. Not for the user.

> 
> 
> The design:
> 
>     * Provide `GET` endpoint listing only information needed for deletion. This endpoint is used when user pass `--dry-run` to the `trustify-cli prune` (you have a good point about only ids since at this point user do not need `source_document_id`s—a boolean parameter to `fetch_prunable_sboms` could help to refine what is really needed).
> 

I understand the idea of "dry run". But that could just be "search for SBOMs". Returning their IDs. Which could be the input for the DELETE operation.

>     * Provide `POST` endpoint doing the deletion. This first call `fetch_prunable_sboms` and store fetched data to the `DeletionContext`. Then the deletion is done over the context. Why we have `DeletionContext`? To have something which is easily convertible to `trustify-cli`'s `DeleteResult`.

If you'd pass a list of SBOM IDs, and perform the delete operation, that should be sufficient. I'm not sure you need to return anything else. Other than "ok", "not ok". Because either the operation succeeds, or fails. As described above. Anything else is for the operator of the instance, not for the user. The user can't deal with this.

> 
> 
> Can this be done more easily? Yes, with loss of `trustify-cli prune` UX compatibility:
> 
>     * `GET` returns only `sbom_id`s
> 
>     * `POST` takes a list of ids from `GET` and performs delete, returning the list of fetched ids and possible error if the deletion failed (note that this does not fulfill the ADR, which requires detailed info about each deleted SBOM).
> 

As the ADR is part of this PR, that should be easy to change. I think we never released any CLI tool that we now need to keep stable. We also have a breaking release upcoming.

> 
> > For fetching SBOMs we already have an endpoint.
> 
> Unoptimized one (according to @bxf12315 analyzis).

Again, things have changed since then. And if we need further optimizations, those would benefit everyone else too.

> 
> > For deleting SBOMs, we would need a "multi-delete" endpoint. Ideally one that accepts the query request as the list endpoint does.

Ideally one that accepts an array of IDs. `DELETE /api/v3/sbom`.

> 
> `POST /api/v3/sbom/prune` is trying to do this.
> 
> > So we could just have a `delete_sboms` service method. Not sure if we can bulk delete dependent entities. But we could try that. Internally, we should delete by ID only. Not the SHAs.
> 
> Arent SHAs used to locate and delete source documents on storage server?

Yes, they are. So the approach would be:

* `DELETE /api/v3/sbom` -> body: `{"id": []}`
* Lookup SHA254 via IDs, one query
* Delete all SBOMs by IDs
* Run GC by IDs

modules/fundamental/src/sbom/model/mod.rs
i386x: After addressing your points about `delete_sboms`, this will no longer be needed.

modules/fundamental/src/sbom/endpoints/mod.rs
i386x: This also appears in the original code base but yes we can move it to the service.

modules/fundamental/src/sbom/endpoints/mod.rs
i386x: I am not sure either. This appears also in the original code base. Maybe it is useful for further analysis? Like an admin can perform additional manual checks over the storage server based on the logged error messages?

docs/adrs/00012-sbom-pruning-api.md
i386x: Well, lets see what can happen on DB side when deleting bunch of items at one transaction:
- All requested items are deleted, no error occurs.
  - Return `HttpResponse::NoContent`.
- Some requested items are deleted, no error occurs.
  - For one item, the situation is easy: item is either deleted or not, so for the latter case we can just return `HttpResponse::NotFound`.
  - For more items, though, the situation is a bit more complicated: we can either return `NoContent` and a list of ids that were successfully deleted or we can return `NotFound` and a list of skipped ids (which will probably need to do the difference of two sets of ids). We can discuss which one to prefer.
- Some requested items are deleted, error occurs.
  - I hope that at this point the DB back-end performs a rollback to recover from the failed transaction and at the end of the day nothing will be deleted (otherwise, the DB back-end is probably broken and we should report a bug for this).
  - The error is propagated and handled by the Actix Web framework.
- No one from the requested items is deleted, error occurs.
  - The error is propagated and handled by the Actix Web framework.

modules/fundamental/src/sbom/service/sbom.rs
i386x: To match the type in the response model (`DeleteContext`), but this can be changed to `usize`. Probably this line can go away.

modules/fundamental/src/sbom/service/sbom.rs
i386x: Why not.

modules/fundamental/src/sbom/service/sbom.rs
i386x: As mentioned in the answer to the similar question, this probably will not be needed or can be changed to `usize`.

cli/src/common/mod.rs
i386x: Original code base has `u32`, `Vector::len()` has `usize`. Probably we can also use `usize`.

cli/src/common/mod.rs
i386x: I think we can. However this part will be removed/reworked as we are going away from `DeleteContext`.

cli/Cargo.toml
i386x: I agree.

docs/adrs/00012-sbom-pruning-api.md
i386x: I agree.

docs/adrs/00012-sbom-pruning-api.md
i386x: I agree.

docs/adrs/00012-sbom-pruning-api.md
i386x: Good point, I will simplify the responses. Since we agreed on using `fetch_sboms`, the whole `GET` endpoint lost the purpose.

modules/fundamental/src/sbom/service/sbom.rs
i386x: Why not.

modules/fundamental/src/sbom/service/sbom.rs
i386x: I agree.

i386x:
> Why there are [`Database` not found in this scope](https://github.com/guacsec/trustify/actions/runs/26821525026/job/79077135135?pr=2364#step:8:1448) errors when `Database` is clearly imported and used in previous endpoints?
> 
> Also it compiles successfully on my machine.

~~Oh, I see why. GitHub is trying to merge against the top of the `main` branch, no matter whether the PR branch is rebased on top or not (read: new commits arrive when rebasing on my machine and force pushing, need to do another local rebase).~~

Answered by Jens, did not see his answer before.

i386x:
> I do understand there is the transaction gap between the S3 storage and the DB. However, from a user perspective, the DB either deletes all documents, or none. Any mishap with the S3 storage should be logged and flagged for an administrator. Not for the user.

So we also need to log errors from `storage.delete()` more properly, not just to terminal. Do we have/use some advanced logging facility other than just the one from the Rust's standard library?

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -412,22 +412,65 @@ pub async fn delete(
...
-                    delete_doc(&SourceDocument::from_entity(&source_document), i.storage()).await
-                {
-                    log::warn!("Ignoring {e}");
-                }
-                Ok(HttpResponse::NoContent().finish())
-            }
-        },
+        Some((v, _, _)) => Ok(
+            match service
+                .delete_sboms(vec![v.sbom_id], tx, i.storage())
```
ctron: I don't see the transaction being committed.

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -412,22 +412,65 @@ pub async fn delete(
...
         None => Ok(HttpResponse::NotFound().finish()),
     }
 }
 
+/// Delete multiple SBOMs
+#[utoipa::path(
+    tag = "sbom",
+    operation_id = "deleteSboms",
+    request_body(
+        content = Vec<String>,
```
ctron, line 442: I think this should be `Vec<Id>` for the spec.

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -412,22 +412,65 @@ pub async fn delete(
...
+    i: web::Data<IngestorService>,
+    service: web::Data<SbomService>,
+    db: web::Data<db::ReadWrite>,
+    web::Json(body): web::Json<Vec<String>>,
+    _: Require<DeleteSbom>,
+) -> actix_web::Result<impl Responder, Error> {
+    let tx = db.begin().await?;
+    let ids = body
+        .into_iter()
+        .map(|x| {
```
ctron: If we only accept the ID, then why do we force the parsing and conversion through `Id`? Why not just use `Uuid`. And threat an un-parsable UUIID as "not found", because that's what it is.

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -412,22 +412,65 @@ pub async fn delete(
...
-                    log::warn!("Ignoring {e}");
-                }
-                Ok(HttpResponse::NoContent().finish())
-            }
-        },
+        Some((v, _, _)) => Ok(
+            match service
+                .delete_sboms(vec![v.sbom_id], tx, i.storage())
+                .await?
+                .len()
```
ctron: I guess `is_empty` is what would be sufficient here.

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -412,22 +412,65 @@ pub async fn delete(
...
+            Id::parse_uuid(&x).map_err(Error::IdKey).and_then(|x| {
+                x.try_as_uid().ok_or(Error::Internal(
+                    "Id::parse_uuid does not returned Id::Uuid variant".to_string(),
+                ))
+            })
+        })
+        .collect::<Result<_, _>>()?;
+
+    let rids = service.delete_sboms(ids, tx, i.storage()).await?;
+    Ok(match rids.len() {
```
ctron: I guess `is_empty` is what would be sufficient here.

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -412,22 +412,65 @@ pub async fn delete(
...
+        .map(|x| {
+            Id::parse_uuid(&x).map_err(Error::IdKey).and_then(|x| {
+                x.try_as_uid().ok_or(Error::Internal(
+                    "Id::parse_uuid does not returned Id::Uuid variant".to_string(),
+                ))
+            })
+        })
+        .collect::<Result<_, _>>()?;
+
+    let rids = service.delete_sboms(ids, tx, i.storage()).await?;
```
ctron: I don't see the transaction being committed.

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -412,22 +412,65 @@ pub async fn delete(
...
+    Ok(match rids.len() {
+        0 => HttpResponse::NoContent().finish(),
+        _ => HttpResponse::NotFound().json(
+            rids.into_iter()
+                .map(|x| x.urn().encode_lower(&mut Uuid::encode_buffer()).to_string())
```
ctron: Why can't we use `to_string`? Also, if we'd use plain IDs, that might just go away.

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -114,80 +122,115 @@ impl SbomService {
...
-    pub async fn delete_sbom<C: ConnectionTrait>(
+    /// delete multiple sboms
+    #[instrument(skip(self, tx, storage), err(level=tracing::Level::INFO))]
+    pub async fn delete_sboms<S: StorageBackend>(
         &self,
-        id: Uuid,
-        connection: &C,
-    ) -> Result<bool, Error> {
+        ids: Vec<Uuid>,
+        tx: DatabaseTransaction,
```
ctron: This should be a trait, e.g. `impl ConnectionTrait`.

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -114,80 +122,115 @@ impl SbomService {
...
+        let mut source_document_ids = vec![];
+
         for row in &result {
-            let source_document = row.try_get_by_index::<Option<Uuid>>(0)?;
-            if let Some(doc) = source_document {
-                source_document::Entity::delete_by_id(doc)
-                    .exec(connection)
-                    .await?;
-            }
+            deleted_ids.insert(row.try_get_by_index::<Uuid>(0)?);
```
ctron: Why don't we use a tuple?  -> https://www.sea-ql.org/SeaORM/docs/1.1.x/advanced-query/custom-select/#unstructured-tuple

ctron: I don't think that can be `null`.

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -114,80 +122,115 @@ impl SbomService {
...
-            let array_param = sea_query::Value::Array(
-                sea_query::ArrayType::Uuid,
-                Some(Box::new(
-                    qualified_purl_ids
-                        .iter()
-                        .map(|&id| sea_query::Value::Uuid(Some(Box::new(id))))
-                        .collect(),
-                )),
-            );
+        let digests = match source_document_ids.len() {
```
ctron: `is_empty`

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -114,80 +122,115 @@ impl SbomService {
...
+            0 => vec![],
+            _ => source_document::Entity::delete_many()
+                .filter(source_document::Column::Id.is_in(source_document_ids))
+                .exec_with_returning(&tx)
+                .await?
+                .iter()
+                .map(|x| format!("sha256:{}", x.sha256))
```
ctron: Maybe add a comment here "why" we're picking the sha256.

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -114,80 +122,115 @@ impl SbomService {
...
-                connection.get_database_backend(),
+                tx.get_database_backend(),
                 // it looks much more readable in an SQL file
                 include_str!("gc_purls_after_sbom_deletion.sql"),
-                vec![array_param],
+                vec![qualified_purl_ids.into()],
             );
 
-            let gc_result = connection.execute(gc_stmt).await?;
+            let gc_result = tx.execute(gc_stmt).await?;
```
ctron: As this might take some time, add some instrumentation: https://github.com/guacsec/trustify/blob/main/docs/design/log_tracing.md#instrumentation

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -114,80 +122,115 @@ impl SbomService {
...
-                "Cleaned up {} orphaned purl records after SBOM {} deletion",
+                "Cleaned up {} orphaned purl records after SBOMs {:?} deletion",
                 gc_result.rows_affected(),
-                id
+                ids,
             );
         }
 
-        Ok(result.len() == 1)
+        tx.commit().await?;
```
ctron: This should be handled outside the service method. Creation as well is not part of the `SbomService`. The service also doesn't hold an instance of the storage. This is mixing up things. I'm not saying the current model is ideal. But creating a new mix doesn't make it easier.

My suggestion would be:

* Just return here if what was deleted, and what was missing. Any error should propagate. But an unparsable ID should not fail (but be not found).
* Create a "delete many" function on the `Storage` trait. That may actually allow the trait implementation to perform "multi delete" if the backend would support that. Not sure S3 does (spoiler alert, is does: https://docs.aws.amazon.com/AmazonS3/latest/API/s3_example_s3_DeleteObjects_section.html).

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -114,80 +122,115 @@ impl SbomService {
...
+
+        // Remove blobs from the storage. Only do this when everything that was
+        // requested to be deleted have been actually deleted
+        if nres == nids && nres == digests.len() {
+            for key in &digests {
+                if let Err(e) = match Id::from_str(key)
+                    .map_err(Error::IdKey)
+                    .and_then(|id| id.try_into().map_err(Error::StorageKey))
```
ctron: I think, instead of that back and forth between `Id` and `StorageKey` we should add a dedicated function which makes it clear what's happening. Suggestion: add `StorageKey::from_sha256(String)`. Or create a function which allows getting a `StorageKey` from a `&SourceDocument`.

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -114,80 +122,115 @@ impl SbomService {
...
+                {
+                    Ok(k) => storage
+                        .delete(k)
+                        .await
+                        .map_err(|x| Error::Storage(anyhow!("{x}"))),
```
ctron: Not sure why we need to involve `anyhow` here. I think this could be logged right away.

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -114,80 +122,115 @@ impl SbomService {
...
+                ids,
             );
         }
 
-        Ok(result.len() == 1)
+        tx.commit().await?;
+
+        // Remove blobs from the storage. Only do this when everything that was
+        // requested to be deleted have been actually deleted
+        if nres == nids && nres == digests.len() {
```
ctron: This seems to lead to cases where documents are not properly being deleted. Assuming the user requests to delete A, B, C. But only A, C where in the database. Then this would have deleted, and committed A, C. But none of them would be deleted from the storage?

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -114,80 +122,115 @@ impl SbomService {
...
+            for key in &digests {
+                if let Err(e) = match Id::from_str(key)
```
ctron: I'm not sure we need this other if-let. Can't we just use `inspect_err` on the call? Assuming we can drop the `Id::from_str`.

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -114,80 +122,115 @@ impl SbomService {
...
+                    .map_err(Error::IdKey)
+                    .and_then(|id| id.try_into().map_err(Error::StorageKey))
+                {
+                    Ok(k) => storage
+                        .delete(k)
+                        .await
+                        .map_err(|x| Error::Storage(anyhow!("{x}"))),
+                    Err(e) => Err(e),
+                } {
+                    log::warn!(
```
ctron: I think this may even be `error`.

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -114,80 +122,115 @@ impl SbomService {
...
+                        "Failed to delete the source document from the storage: {}",
```
ctron: I guess you can use `{e}` here too

openapi.yaml
```diff
@@ -2761,6 +2761,31 @@ paths:
...
+            schema:
+              type: array
+              items:
+                type: string
+        required: true
+      responses:
+        '204':
+          description: All requested SBOMs have been deleted
+        '404':
+          description: Some requested SBOMs have not been deleted
```
ctron: IIRC that misaligns with existing DELETE endpoints. The idea is to have the delete endpoint being idempotent. So deleting something that doesn't exists results in 2xx as well. Unless you delete with an `IfMatch` header, which we don't for SBOMs.

Also, the description seems vague: not delete why? Because they were not found. There's also a 500 error in case of DB errors etc.

ctron (changes requested):
Huge improvement. Some nits and issues need to be ironed out.

Most importantly, I don't see endpoint tests. Which would be required to test the whole functionality. Service level tests are fine, but here we need more for complete coverage.

modules/fundamental/src/sbom/endpoints/mod.rs
i386x: It's done inside `delete_sboms`, but after addressing your other suggestions I will move it back outside.

modules/fundamental/src/sbom/endpoints/mod.rs
i386x: It's done inside `delete_sboms`, but after addressing your other suggestions I will move it back outside.

modules/fundamental/src/sbom/service/sbom.rs
i386x: Will be reverted back if a transaction goes outside.

modules/fundamental/src/sbom/service/sbom.rs
i386x: Nice catch.

i386x: You think `None`? I am not using `Option<...>` which will automatically falls back to the blank implementation of `TryGetable` trait that will silently ignores `null` rows, turning them into `None`. Without `Option<...>`, this will issue an database `Error` when UUID is null, signaling that our DB is probably corrupted and immediate action should be then taken.

modules/fundamental/src/sbom/service/sbom.rs
i386x: Because other formats fails the conversion to `StorageKey`. I will add a note, even this is clear from the storage backend infra source code.

modules/fundamental/src/sbom/service/sbom.rs
i386x: Yes, I can. I don't know what conventions are preferred in Rust community, but in other languages communities (Python, C/C++), the preferred way is not to include variables to format strings in logging functions and instead pass them as arguments for the reason that such a strings can be more easily gathered using tools like `gettext`.

ctron: I don't think that's relevant here.

bxf12315 (commented):
This endpoint can also be called by multiple threads, so this piece of code may have thread safety issues. I previously described this in that [issue](https://github.com/guacsec/trustify/issues/2191). If possible, you could consider making these null qualified_purls a separate endpoint. This is just an initial idea and may not necessarily help you. I’m glad to still be able to provide some input.

openapi.yaml
i386x: > IIRC that misaligns with existing DELETE endpoints. The idea is to have the delete endpoint being idempotent. So deleting something that doesn't exists results in 2xx as well.

Well in that case our API is not consistent:
- `DELETE /api/v3/sbom/{id}` returns 404 when not found
- the same for advisories
- `PUT /api/v3/importer/{name}` returns 409 when not found
- `DELETE /api/v3/product/{id}` returns 404 when not found
- etc.

Any thoughts? From 2xx status codes only 200 OK or 204 NO_CONTENT fits the situation, but almost everything else in our API uses 404 NOT_FOUND.

ctron: > Well in that case our API is not consistent:

I agree. And we should fix this.

`PUT` is a different use case. The idea of DELETE being idempotent would mean that you can call delete multiple times, and if the resource is gone afterwards, return "something OK-ish".

Here's the analysis of claude:

```
  ┌─────────────────────────────────┬─────────┬────────────────────────────┬─────────────┐
  │            Endpoint             │ Success │    Non-Existent Object     │ Idempotent? │
  ├─────────────────────────────────┼─────────┼────────────────────────────┼─────────────┤
  │ DELETE /v3/importer/{name}      │ 204     │ 204                        │ Yes         │
  ├─────────────────────────────────┼─────────┼────────────────────────────┼─────────────┤
  │ DELETE /v3/userPreference/{key} │ 204     │ 204 (or 412 with if-match) │ Conditional │
  ├─────────────────────────────────┼─────────┼────────────────────────────┼─────────────┤
  │ DELETE /v3/sbom/{id}            │ 204     │ 404                        │ No          │
  ├─────────────────────────────────┼─────────┼────────────────────────────┼─────────────┤
  │ DELETE /v3/product/{id}         │ 200     │ 404                        │ No          │
  ├─────────────────────────────────┼─────────┼────────────────────────────┼─────────────┤
  │ DELETE /v3/advisory/{key}       │ 200     │ 404                        │ No          │
  ├─────────────────────────────────┼─────────┼────────────────────────────┼─────────────┤
  │ DELETE /v3/group/sbom/{id}      │ 204     │ 204                        │ Yes         │
  └─────────────────────────────────┴─────────┴────────────────────────────┴─────────────┘


```

I'd prefer fixing this behavior, rather than continuing with it. The two endpoints SBOM and advisory are the oldest ones. The product isn't really being used.

ctron: I'm preparing a PR to align this.

i386x:
@ctron There are still some tests missing, but I have a plan to do them in the follow up PR, if you don't mind.

modules/storage/src/service/mod.rs
```diff
@@ -30,6 +30,12 @@ pub enum StoreError<B: Debug> {
 #[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct StorageKey(String);
 
+impl StorageKey {
+    pub fn from_sha256(digest: &str) -> Self {
```
ctron, line 49: Please document.

modules/storage/src/service/mod.rs
```diff
@@ -112,4 +118,30 @@ pub trait StorageBackend {
     /// (i.e., return `Ok(())`) and not result in an error. This ensures consistent
     /// behavior across all backends.
     fn delete(&self, key: StorageKey) -> impl Future<Output = Result<(), Self::Error>>;
+
+    /// Batch variant of `delete`, with the same requirements.
+    fn delete_many(&self, keys: &[StorageKey]) -> impl Future<Output = Result<(), String>> {
```
ctron: `String` is not a proper error type.

I think having a default implementation is good. The question is indeed how to handle "multi errors"? It may be easier having `delete` being a default re-using `delete_many`.

That way, we could introduce common error types for this (just outlining, bad naming):

```rust
enum DeleteManyError<E> {
  /// The whole operation failed
  Generic(E),
  /// Individual delete errors
  Individual(HashMap<StorageKey, E>)
}

trait Storage {
  type Error;

  fn delete () -> Result<(), Self::Error>;
  fn delete_many() -> Result<(), DeleteManyError<Self::Error>>;
}
```

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -411,21 +424,60 @@ pub async fn delete(
...
+
+    let (digests, not_found) = service.delete_sboms(ids, &tx).await?;
+    missing.extend(not_found.iter().map(|v| Id::Uuid(*v)));
+
+    if !digests.is_empty() {
+        tx.commit().await?;
+        delete_blobs(&digests, i.storage()).await;
     }
+
+    Ok(HttpResponse::NoContent().json(missing))
```
ctron: We need to find a way to only return those missing if someone is interested it that information. So we need a flag. Otherwise we are repeating the mistake if returning potentially a lot of information no one is interested about.

modules/fundamental/src/sbom/service/sbom.rs
```diff
@@ -114,80 +119,101 @@ impl SbomService {
...
+                .filter(source_document::Column::Id.is_in(source_document_ids))
+                .exec_with_returning(connection)
+                .await?
+                .iter()
+                .map(|x| x.sha256.clone())
+                .collect(),
+        };
+
+        // Cleanup orphaned PURLs if deletion succeeded and we had PURLs to check
+        if !qualified_purl_ids.is_empty() && nres == nids {
```
ctron: Doesn't that mean: if we have less documents deleted than we expected (e.g. because one is already missing) the cleanup would not be performed?

ctron (changes requested):
This looks way better now.

I still think we must have some tests on the endpoint level. Things like deleting a set of documents, where one exists and one doesn't. Checking the outcome, if PURLs are gone, if documents (in storage) are gone, or still there if the deletion could not be performed.

I also think that @bxf12315's comment is valid, we need some concurrency test.

modules/fundamental/src/sbom/service/sbom.rs
i386x: Good catch.

modules/fundamental/src/sbom/endpoints/mod.rs
i386x: @ctron What about not return anything? If something is missing than the missing/ill-formed ids are already part of the request so user have that information already.

ctron: I'm fine with not returning anything.

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -411,21 +424,55 @@ pub async fn delete(
...
+    }
+    Ok(HttpResponse::NoContent().finish())
+}
+
+/// Delete multiple SBOMs
+#[utoipa::path(
+    tag = "sbom",
+    operation_id = "deleteSboms",
+    request_body(
+        content = Vec<Id>,
```
ctron: Sorry for noticing that late. We accept ID, but then actually don't support any of the other forms (aside from Uuid). That's confusing. Especially as we document we do.

I think just accepting `Vec<String>` is fine. And then turning that into `Uuid::from_str`, treating unparsable as "not found".

ctron (changes requested):
We are really on the right track! Much improved.

We have an endpoint test now. Covering the default path plus some extras. I think we should consider some corner cases as well. Like deleting an empty array. Deleting an ID which is not a UUID. the not-so-happy-path.

I also missed a case where `Id` is being used, but it doesn't seem to be the right. Sorry for that.

Also, it looks like we are still missing tests for the `delete_many` function added to the storage. Same here: happy path, but then, corner cases. What if a document gets delete twice. A document that is not there is being deleted, …

What is also still seems missing is a test "in parallel" as @bxf12315 suggested. Maybe similar to the existing test, ingest a bunch of documents, issue parallel deletes, examine results.

modules/fundamental/src/sbom/endpoints/test.rs
```diff
@@ -1232,7 +1237,115 @@ async fn delete_sbom(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
...
+    Ok(())
+}
+
+/// Test deleting sboms in a batch
+#[test_context(TrustifyContext)]
+#[test(actix_web::test)]
+async fn delete_sboms(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
+    let app = caller(ctx).await?;
+    let storage = &ctx.storage;
+    let results = ctx
```
ctron: Suggestion: Instead of taking the vec, and then picking random indexes later, you should be able to leverage a trait we have (`IngestionResult`):

```suggestion
let [a, b, c, d] = ctx.ingest_documents([]).await?.into_uuid();
```

The number of entries of both arrays (input/output) must be the same otherwise this will panic. Which is ok in tests. You might want to choose better names than `a`, `b`, …

ctron: e.g. here: https://github.com/guacsec/trustify/blob/1845ce87ceb8a318ffd01f177cd334def48c258c/modules/analysis/src/service/load/rank.rs#L510

i386x:
@ctron I added a simple concurrency test, lets see how it will run in the GitHub CI.

modules/fundamental/src/sbom/endpoints/test.rs
```diff
@@ -1241,6 +1245,133 @@ async fn delete_sbom(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
...
+
+        app.call_and_read_body_json::<PaginatedResults<PurlSummary>>(req)
+            .await
+            .total
+    }
+
+    async fn delete<const N: usize>(app: &impl CallService, ids: [&Uuid; N]) -> ServiceResponse {
+        let req = TestRequest::delete()
+            .uri("/api/v3/sbom")
+            .set_json(ids.into_iter().map(|x| Id::Uuid(*x)).collect::<Vec<_>>())
```
ctron, line 1298: Is this still correct? Shouldn't this just be the UUID?

ctron:
I think we are close. But we also lack test coverage. We only have a single test. And that tests only fractions. Like we seem to test multi-concurrent delete, but don't expect it to delete anything. As it's already deleted. Which makes it hard to see if the concurrent actually works (leaving not garbage behind).

modules/fundamental/src/sbom/endpoints/test.rs
i386x: Probably `x.to_string()` should suffice here, or just `.set_json(ids)`, however all cases should produce identical JSON payload (`Id::Uuid(*x).to_string()` should be identical to `x.to_string()` when serialized).

ctron (approved):
I think we need some more test coverage for the endpoints and the S3/Storage trait changes.

However, I also think it's ok take the risk and move on. Increasing test coverage in a new PR.

@i386x Thanks for all the effort!

modules/fundamental/src/sbom/endpoints/test.rs
ctron: I guess `x.to_string()` is sufficient

PR #2365, opened 2026-05-19
test: add a test to ensure loop references don't cause a problem

ctron:
## Summary by Sourcery

Tests:
- Add an async test that ingests cyclically referencing SPDX SBOMs and verifies `find_external_refs` completes within a bounded time using a timeout guard.

### DIFF
--- a/modules/analysis/src/service/load/rank.rs
+++ b/modules/analysis/src/service/load/rank.rs
@@ -472,7 +472,9 @@ mod test {
     use chrono::{TimeZone, Utc};
     use futures::{StreamExt, TryStreamExt, stream};
     use rstest::rstest;
+    use std::time::Duration;
     use test_context::test_context;
+    use tokio::time::timeout;
     use trustify_entity::cpe;
     use trustify_test_context::{IngestionResult, TrustifyContext};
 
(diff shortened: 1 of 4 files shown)


modules/analysis/src/service/load/rank.rs
```diff
@@ -615,4 +617,29 @@ mod test {
...
+
+        let mut visited = HashSet::new();
+
+        // no native test timeout in Rust, using tokio's timeout as a deadlock guard
+        let result = timeout(
+            Duration::from_secs(10),
+            find_external_refs(id_a, "SPDXRef-Root-A".into(), &ctx.db, &mut visited),
+        )
+        .await
+        .expect("find_external_refs should not loop infinitely")?;
```
helio-frota, line 640: Do you think that makes sense to add some basic assertions like this?

```rust
assert!(!result.is_empty());
assert_eq!(result.len(), 2);
assert!(result.contains(&ResolvedSbom {
    sbom_id: _id_b,
    node_id: "SPDXRef-Leaf-B".into()
}));
assert!(result.contains(&ResolvedSbom {
    sbom_id: id_a,
    node_id: "SPDXRef-Leaf-A".into()
}));
```

ctron: It's not so much about the content, but about the fact that it doesn't stall. Sure, I can add it.

ctron: I did a full compare to see the full diff (in case of error). But that should work too, right?

modules/analysis/src/service/load/rank.rs
```diff
@@ -615,4 +617,48 @@ mod test {
...
+
+        log::info!("resolved external SBOMs: {result:?}");
+
+        // assert
+
+        result.sort();
+
+        assert_eq!(
```
helio-frota, line 648: thanks I understood your point, but at least for me adding this assert helps as some sort of documentation lets say...

modules/analysis/src/service/load/rank.rs
helio-frota: you mean the diff between the 2 json files?

helio-frota (approved):

PR #2366, opened 2026-05-19
docs: Update CONVENTIONS.md file

mrrajan:
- Expands CONVENTIONS.md with comprehensive anti-pattern conventions identified in ADR-00018, covering: N+1 queries, unbounded queries, in-memory filtering, application-side counting, missing batch operations, recursive traversal, missing indexes, swallowed errors, stringly-typed APIs, code duplication, oversized functions, public API documentation, magic numbers, and raw SQL parameterization
  - Adds database resource conventions (table/column naming, column types, enums, indexes, foreign keys, migrations)
  - Refines existing tracing/instrumentation guidance with clearer scoping rules and logging framework migration policy (log:: → tracing::)

Reference PR - https://github.com/guacsec/trustify/pull/2358

JIRA - https://redhat.atlassian.net/browse/TC-4289

Preview - https://github.com/mrrajan/trustify/blob/TC-4289-convention/CONVENTIONS.md

## Summary by Sourcery

Expand and refine engineering conventions around observability and database usage, incorporating anti-pattern guidance and schema best practices into CONVENTIONS.md.

Documentation:
- Clarify tracing span usage, instrumentation scope, and logging framework migration from log:: to tracing:: for new and existing code.
- Document additional engineering conventions covering N+1 and unbounded queries, batching, counting, error handling, enums, magic numbers, and raw SQL safety requirements.
- Add database schema conventions for table/column naming, types, enums, indexes, foreign keys, and migrations, complementing existing migration and entity model patterns.

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -417,6 +417,67 @@ tracing::debug!("Processing vulnerability {id}");
 log::debug!("Processing vulnerability {id}");
 ```
 
+## Data Ingestion: Creator Pattern
+
+**Convention**: Use the `*Creator` batch pattern for data ingestion; do not use the
+deprecated `*Context` per-entity pattern.
+
+### Creator pattern (required for new code)
+
+A `*Creator` struct accumulates entries in memory (via `.add()`), then writes them all in
+a single `.create()` call using `INSERT ... ON CONFLICT DO NOTHING` with batch chunking.
+This approach is concurrency-safe, avoids N+1 queries, and uses consistent lock ordering
+to prevent deadlocks.
+
+Key characteristics:
+- **Batch-oriented** — collects entries in a deduplicating collection, then inserts in bulk
+- **Consuming** — `.create(self, ...)` takes ownership; the creator is single-use
+- **Deduplication** — entries are keyed by deterministic UUID or natural key before insert
+- **Atomic** — uses `ON CONFLICT DO NOTHING` so concurrent transactions cannot conflict
+- **Lock-safe** — sorted/chunked inserts where lock ordering matters (e.g.,
+  `OrganizationCreator` sorts by name; others rely on `BTreeMap` key order or the batch
+  helper's own ordering)
+
+```rust
+// Approved: Creator pattern — batch accumulation then bulk insert
+let mut purl_creator = PurlCreator::new();
+for purl in &purls {
+    purl_creator.add(purl.clone());
+}
+purl_creator.create(&connection).await?;
+```
+
+Examples include `PurlCreator`, `PurlStatusCreator`, `OrganizationCreator`,
+`VulnerabilityCreator`, `PackageCreator`, `RelationshipCreator`, `LicenseCreator`,
+`CpeCreator`, and `ScoreCreator` — see `modules/ingestor/src/graph/**/creator.rs` for the
+full set. When adding ingestion for a new entity type, follow this pattern.
+
+Use creators for batch ingest; use the
+[Shared Table Insert Pattern](#shared-table-insert-pattern-duplicate-key-handling) when you
+need the existing row ID back on conflict (e.g., `create_doc`).
+
+### Context pattern (deprecated)
+
+The `*Context` structs (e.g., `ProductContext`, `OrganizationContext`) wrap a single loaded
+entity model with a reference to the parent `Graph` and perform immediate per-entity
+database operations via methods like `ingest_*()`. This design causes:
+
+- **N+1 queries** — each method call triggers individual DB round-trips inside loops
+- **Race conditions** — SELECT-then-INSERT is not atomic; concurrent ingestion of the same
+  entity causes unique-constraint violations
+- **Mixed responsibility** — combines entity representation with ingestion logic
+
+Remaining `*Context` usage is tech debt. Do not extend or add new `*Context` structs. Not
+every entity has a `*Creator` yet, so when modifying code that uses a `*Context`, prefer
+migrating it to the corresponding `*Creator` in the same change when practical — creating
+a new `*Creator` if needed.
+
+```rust
+// Avoid: Context pattern — per-entity immediate insert
+let org_ctx = graph.ingest_organization(name, info, &connection).await?;
+let product_ctx = org_ctx.ingest_product(name, &connection).await?;
+```
+
 ## Shared Table Insert Pattern (Duplicate Key Handling)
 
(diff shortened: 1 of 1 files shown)


CONVENTIONS.md
```diff
@@ -436,3 +473,400 @@ including **ingestor**, **advisory**, **sbom**, and **risk_assessment**. All cod
 that insert into `source_document` must use the nested-transaction duplicate-handling
 pattern described above. Failing to do so will cause unhandled constraint violations
 under concurrent ingestion.
+
+## Additional Conventions
+
+The following anti-patterns were identified in
+[ADR-00018](docs/adrs/00018-conventions-file.md); each entry states the agreed convention.
```
ctron: I'd not expect an additional reference from this file to an ADR. I'd expect the content to be present and evolved in *this* file, according to the definition from the ADRs. The most recent active one.

CONVENTIONS.md
```diff
@@ -436,3 +473,400 @@ including **ingestor**, **advisory**, **sbom**, and **risk_assessment**. All cod
...
+All public API list endpoints MUST use `PaginatedResults` (see
+[Endpoint Patterns](#endpoint-patterns)) and enforce a maximum pagination limit
+([ADR-00017](docs/adrs/00017-efficient-pagination.md)). Internal service queries may be
+unbounded if the caller controls scope and the table is known to be bounded in practice
+(e.g., admin-configured importers).
+
+### In-Memory Filtering Instead of SQL WHERE
+
+**Convention (provisional)**: Push filters to SQL when the unfiltered dataset can exceed
+~100 rows; small known-bounded collections may filter in Rust for readability.
```
ctron: I don't think readability is an issue here. Small data sets can also be fast in PSQL, as it can cache too, and have the execution planner take into consideration the amount of data and the amount of memory available.

CONVENTIONS.md
```diff
@@ -436,3 +473,400 @@ including **ingestor**, **advisory**, **sbom**, and **risk_assessment**. All cod
...
+
+```rust
+// Preferred: filter in SQL
+advisory_vulnerability_score::Entity::find()
+    .filter(advisory_vulnerability_score::Column::AdvisoryId.eq(advisory_id))
+    .all(tx)
+    .await?;
+
+// Avoid: fetch-then-filter
+let all_scores = advisory_vulnerability_score::Entity::find().all(tx).await?;
```
ctron, line 599: The reason for this was that by doing an in-memory search, a duplicate query was eliminated.

CONVENTIONS.md
```diff
@@ -436,3 +473,400 @@ including **ingestor**, **advisory**, **sbom**, and **risk_assessment**. All cod
...
+**Convention**: Evaluate per-column with a pros/cons analysis.
+
+No single rule for every column. Before adding (or skipping) an index, weigh these factors:
+
+| For adding an index | Against adding an index |
+|---------------------|-------------------------|
+| Column is filtered or sorted often in production | Table is small or query volume is low |
+| Slow queries or full scans show up in logs | Writes are heavy; extra indexes slow inserts/updates |
+| Public API / user-facing latency matters | Existing indexes already cover the access pattern |
+| Table is large and still growing | Index would duplicate a UNIQUE constraint |
```
ctron, line 684: I think an additional factor is the storage size of indexes.

CONVENTIONS.md
```diff
@@ -436,3 +473,400 @@ including **ingestor**, **advisory**, **sbom**, and **risk_assessment**. All cod
...
+   `return Err(...)`) — do not swallow it.
+4. **Expected failure** — If failure is normal for this path (e.g., optimistic lock miss,
+   optional field missing), ignoring it is fine; no log required.
+5. **Log before dropping** — If you must drop the error and cannot report it upstream, log
+   it compactly (e.g., `.inspect_err(|e| tracing::debug!("…: {e}"))`) rather than a verbose
+   `match`.
+
+```rust
+// Approved: log the error before discarding
+match serde_json::from_value::<Report>(report) {
```
ctron: I'd say this is exact opposite of "compact" as described above.

mrizzi (commented):
I can see nothing about the `*Creator` pattern for data management as the one to be used and implemented.
For example the [`PurlStatusCreator`](https://github.com/guacsec/trustify/blob/480d2bd8224ba04a9567f2055c7d45fcdfdefb03/modules/ingestor/src/graph/purl/status_creator.rs) I had to introduce in https://github.com/guacsec/trustify/pull/2080.

The "forbidden" pattern is the `*Context` one that, if and where still used, should be, at this point, replaced by the `*Creator` pattern to fix the tech debt.
For example [`ProductContext`](https://github.com/guacsec/trustify/blob/480d2bd8224ba04a9567f2055c7d45fcdfdefb03/modules/ingestor/src/graph/product/mod.rs#L27).

@mrrajan I think it could be useful if you could run an agent against code and discussion in https://github.com/guacsec/trustify/pull/2080 itself to derive the appropriate code conventions.

CONVENTIONS.md
```diff
@@ -436,3 +473,400 @@ including **ingestor**, **advisory**, **sbom**, and **risk_assessment**. All cod
...
+Avoid CamelCase, pluralized table names, or abbreviated column names (e.g., `vulnId`,
+`Advisories`).
+
+#### Column Types
+
+Use the narrowest correct PostgreSQL type:
+
+| Data | PostgreSQL Type | Rust / SeaORM |
+|------|----------------|---------------|
+| Primary keys (default) | `UUID` | `Uuid` with `#[sea_orm(primary_key)]` |
```
ctron: Use v7 for random IDs. Use v5 if an ID can be derived from the data of a object. This allows pre-generating the ID, and implement an "upsert" pattern vs a "select then insert" approach. We currently use this in various locations already.

v7 IDs are better for PSQL compared to pure random v4: https://dev.to/umangsinha12/postgresql-uuid-performance-benchmarking-random-v4-and-time-based-v7-uuids-n9b

ctron (commented):
A few minor things. In general it's ok to merge. But I'll leave that button for someone else to press.

mrrajan:
@ctron @mrizzi Updated the PR with the changes requested.

mrizzi (approved):
@mrrajan thanks for the changes, they reflect my expectations, especially in the race conditions generated by the `SELECT-then-INSERT` not atomic workflow implemented in the `*Context` pattern.

PR #2385, opened 2026-06-09
feat(tls): add configurable TLS security profile

jcrossley3:
Replace the hardcoded mozilla_modern_v5 (TLS 1.3 only) SSL acceptor with a configurable TLS security profile aligned with OpenShift TLS Security Profiles. The new --http-server-tls-security-profile flag supports four profiles: old (TLS 1.0+), intermediate (TLS 1.2+, new default), modern (TLS 1.3 only), and custom (user-defined min version and cipher suites).

This changes the default from Modern to Intermediate to match the OpenShift default and broaden client compatibility.

Implements TC-3426

Assisted-by: Claude Code

## Summary by Sourcery

Introduce configurable TLS security profiles for the HTTP server and align defaults with OpenShift’s intermediate profile.

New Features:
- Add a configurable TLS security profile option (old, intermediate, modern, custom) for the HTTP server TLS configuration.
- Allow custom TLS configuration via minimum TLS version and configurable TLS 1.2/1.3 cipher lists.

Enhancements:
- Change the default TLS configuration from a hardcoded modern profile to the intermediate profile to broaden client compatibility.
- Refactor TLS acceptor construction into a profile-aware builder that logs the active TLS profile when enabling TLS.

Documentation:
- Document new HTTP server TLS environment variables for security profile, minimum TLS version, and custom cipher configuration.

Tests:
- Add tests to verify the default TLS profile, validation of custom profile requirements, and acceptance of all supported TLS profiles.

### DIFF
--- a/common/infrastructure/src/app/http.rs
+++ b/common/infrastructure/src/app/http.rs
@@ -13,10 +13,10 @@ use actix_web::{
 use anyhow::{Context, anyhow};
 use bytesize::ByteSize;
 use clap::{Arg, ArgMatches, Args, Command, Error, FromArgMatches, value_parser};
-use openssl::ssl::SslFiletype;
+use openssl::ssl::{SslFiletype, SslVersion};
 use opentelemetry_instrumentation_actix_web::{RequestMetrics, RequestTracing, RouteFormatter};
 use std::{
-    fmt::Debug,
+    fmt::{self, Debug},
     marker::PhantomData,
     net::{IpAddr, Ipv6Addr, SocketAddr, TcpListener},
     ops::Deref,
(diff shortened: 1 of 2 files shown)


ctron:
Why would we default to a less secure profile?

common/infrastructure/src/app/http.rs
```diff
@@ -264,13 +361,24 @@ where
             .json_limit(value.json_limit.0.0 as _);
 
         if value.tls_enabled {
+            if value.tls_security_profile == TlsSecurityProfile::Custom
```
ctron, line 364: Why not use clap's support for this?

jcrossley3:
> Why would we default to a less secure profile?

Claude recommended it because it's OpenShift's default and it broadens our compatibility with TLS 1.2 clients.

ctron:
> > Why would we default to a less secure profile?
> 
> Claude recommended it because it's OpenShift's default and it broadens our compatibility with TLS 1.2 clients.

I am sure compatibility with TLS 1.0 would be even higher. But that's the wrong direction. If the user wants a less secure profile, they can choose that. But I'd not default to it.

Did anyone ever ask us to do this?

jcrossley3:
@ctron I have changed the default per your request. The ultimate ask in TC-2437 is that TPA fetches and uses the policy configured in OpenShift, but I don't think it's appropriate to add any k8s libs to TPA. Instead, I think our operator should fetch the policy and then set the corresponding environment variables that this PR introduces. If you agree/approve, I think we can go ahead and merge this.

ctron (approved):
Yes, I agree that the backend should only provide options. The operator should translate and set those options.

jcrossley3:
For completeness, the operator side: https://github.com/trustification/trusted-profile-analyzer-operator/pull/850

PR #2387, opened 2026-06-09
fix!: make all DELETE endpoints idempotent

ctron:
Align DELETE behavior across product, sbom, and advisory endpoints to always return 204 No Content, making them idempotent. Previously some returned 404 on missing resources and 200 with a JSON body on success.

Document the DELETE idempotency convention in CONVENTIONS.md, including the If-Match revision check exception.

## Summary by Sourcery

Align DELETE behavior across advisory, product, and SBOM endpoints to be idempotent and consistently return 204 No Content without a response body, while updating documentation and tests accordingly.

Bug Fixes:
- Ensure deleting non-existent advisory, product, and SBOM resources returns 204 No Content instead of 404 or 200 with a body, making DELETE operations idempotent.

Enhancements:
- Simplify delete handlers for advisory, product, and SBOM endpoints to always respond with 204 No Content regardless of prior existence.

Documentation:
- Document the DELETE idempotency convention and its If-Match revision check exception in CONVENTIONS.md.

Tests:
- Update and extend advisory, product, and SBOM deletion tests to assert 204 No Content responses, empty bodies, and idempotent behavior on repeated or non-existent deletes.

### DIFF
--- a/modules/fundamental/src/advisory/endpoints/mod.rs
+++ b/modules/fundamental/src/advisory/endpoints/mod.rs
@@ -141,8 +141,7 @@ pub async fn get(
         ("key" = Id, Path),
     ),
     responses(
-        (status = 200, description = "Matching advisory", body = AdvisoryDetails),
-        (status = 404, description = "The advisory could not be found"),
+        (status = 204, description = "The advisory was deleted or did not exist"),
     ),
 )]
 #[delete("/v3/advisory/{key}")]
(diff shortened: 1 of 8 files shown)


ctron:
@sourcery-ai review

ctron:
@sourcery-ai review

i386x:
I added some of your changes to my PR (#2364).

ruromero (commented):
I agree with the changes. My only comment is that as this is a breaking change, the conventional commit `!` and become  `fix!: make all DELETE...`

ctron:
> I agree with the changes. My only comment is that as this is a breaking change, the conventional commit `!` and become `fix!: make all DELETE...`

You're right! Changed it. And made it a bit shorter.

modules/fundamental/src/advisory/endpoints/mod.rs
```diff
@@ -157,19 +156,15 @@ pub async fn delete(
...
-                Ok(HttpResponse::Ok().json(v))
-            }
-        },
-        None => Ok(HttpResponse::NotFound().finish()),
+    if let Some(v) = service.fetch_advisory(id, &tx).await?
+        && service.delete_advisory(v.head.uuid, &tx).await?
+    {
+        tx.commit().await?;
+        if let Err(e) = delete_doc(&v.source_document, i.storage()).await {
+            log::warn!("Ignoring {e}");
```
i386x: Shoud we use: `log::error!("Removing advisory from the storage failed: {e}");`?

modules/fundamental/src/product/endpoints/mod.rs
```diff
@@ -102,19 +101,7 @@ pub async fn delete(
...
-    match state.fetch_product(*id, &tx).await? {
-        Some(v) => {
-            let rows_affected = state.delete_product(v.head.id, &tx).await?;
-            match rows_affected {
-                0 => Ok(HttpResponse::NotFound().finish()),
-                1 => {
-                    tx.commit().await?;
-                    Ok(HttpResponse::Ok().json(v))
-                }
-                _ => Err(Error::Internal("Unexpected number of rows affected".into())),
```
i386x, line 115: Can this ever happen?

i386x (approved):
LGTM. There are some work that need to be done on the `trustify-cli` (like adjusting `DeleteResult` and `delete_list`) side, but this can be done in the follow up PRs to not further prolong 3.0 cut off.

modules/fundamental/src/advisory/endpoints/mod.rs
ctron: Yup, good point.

modules/fundamental/src/product/endpoints/mod.rs
ctron: Nope, as it has a primary unique key.

PR #2394, opened 2026-06-11
Fix recursion problem with rust 1.96.0

desmax74:
## Summary by Sourcery

Update the Rust toolchain version and adjust the ingestor crate recursion limit to remain compatible with the new compiler.

Enhancements:
- Set a crate-level recursion limit for the ingestor module to prevent compiler recursion issues with Rust 1.96.0.

Build:
- Bump the Rust toolchain channel from 1.91.1 to 1.96.0.

### DIFF
--- a/modules/analysis/src/lib.rs
+++ b/modules/analysis/src/lib.rs
@@ -1,3 +1,4 @@
+#![recursion_limit = "256"]
 pub mod config;
 pub mod endpoints;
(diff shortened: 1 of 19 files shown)


ctron (changes requested):
Instead of ignoring clippy warnings, please fix them. In most cases it's just applying the suggestion.

PR #2395, opened 2026-06-12
Configure a specific rust version 

desmax74:
## Summary by Sourcery

Pin the Rust toolchain to a specific stable version and configure a consistent formatting style for the project.

Enhancements:
- Add rustfmt configuration to standardize formatting using the 2024 style edition.

Build:
- Add rust-toolchain configuration to use Rust 1.96.0 with rustfmt and clippy components.

### DIFF
--- a/.github/actions/setup-rust/action.yaml
+++ b/.github/actions/setup-rust/action.yaml
@@ -1,0 +1,18 @@
+name: Setup Rust
+description: Setup Rust toolchain with the project's pinned version
+
+inputs:
+  components:
+    description: Comma-separated list of additional components to install (e.g. clippy, rustfmt)
+    required: false
+    default: ""
+
+runs:
+  using: "composite"
+
+  steps:
+    - uses: actions-rust-lang/setup-rust-toolchain@v1
+      with:
+        toolchain: 1.96.0
+        components: ${{ inputs.components }}
+        cache: false
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -46,6 +46,10 @@ jobs:
           df -h
 
       - uses: actions/checkout@v5
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: 1.96.0
+          cache: false
       - uses: Swatinem/rust-cache@v2
 
       - name: Setup cargo-binstall
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -46,10 +46,7 @@ jobs:
           df -h
 
       - uses: actions/checkout@v5
-      - uses: actions-rust-lang/setup-rust-toolchain@v1
-        with:
-          toolchain: 1.96.0
-          cache: false
+      - uses: ./.github/actions/setup-rust
       - uses: Swatinem/rust-cache@v2
 
       - name: Setup cargo-binstall
--- a/.github/workflows/build-binary.yaml
+++ b/.github/workflows/build-binary.yaml
@@ -15,10 +15,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v5
 
-      - uses: actions-rust-lang/setup-rust-toolchain@v1
-        with:
-          toolchain: 1.96.0
-          cache: false
+      - uses: ./.github/actions/setup-rust
 
       - name: Create vendor archive
         run: |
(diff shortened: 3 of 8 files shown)


ctron (changes requested):
Please don't do this. We removed this explicitly in a previous PR. None of these changes are necessary.

desmax74:
@ctron how we can use a specific rust version if not in this way ?

mrizzi:
> Please don't do this. We removed this explicitly in a previous PR. None of these changes are necessary.

@ctron we're trying to gain again control over the Rust version used to build the project. 
My understanding is that the `main` branch is built with whatever the GitHub hosted runner provides while we should be in control of the version we want the project to be built, updating it through a managed process that allows us to test if everything is fine as expected when updating the version.
In this context, the changes in this PR make sense and they're consistent with how this is achieved in other release branches.

ctron:
Control comes from the Rust version you execute with. The toolchain file was used in the past to ensure the "minimum supported rust version" (MSRV). This works differently now. We also don't require/enforce the installation of any nightly or set of components.

Having a toolchain file will trigger automatic downloads (which fails in network restricted environments) and it also not honored by the new resolver. It's a thing from the past.

We removed this on main to be able to control this actually. The fact that it's still on other branches means we didn't backport it.

ctron:
> Hey - I've left some high level feedback:
> 
>     * Consider aligning `style_edition = "2024"` with the actual crate edition (or current stable support) to avoid rustfmt incompatibilities on toolchains that don’t yet support the 2024 style.
> 
>     * Using a fixed `channel = "1.96.0"` may increase maintenance overhead; consider whether pinning via `rust-version` in Cargo.toml and using `channel = "stable"` would better balance reproducibility with ease of upgrades.
> 
> 
> Prompt for AI Agents
> 
> Sourcery is free for open source - if you like our reviews please consider sharing them ✨
> 
> 
> Help me be more useful! Please click 👍 or 👎 on each comment and I'll use the feedback to improve your reviews.

And I think the AI agrees with that.

mrizzi:
> Control comes from the Rust version you execute with. The toolchain file was used in the past to ensure the "minimum supported rust version" (MSRV). This works differently now. We also don't require/enforce the installation of any nightly or set of components.
> 
> Having a toolchain file will trigger automatic downloads (which fails in network restricted environments) and it also not honored by the new resolver. It's a thing from the past.
> 
> We removed this on main to be able to control this actually. The fact that it's still on other branches means we didn't backport it.

@ctron thanks and I take your point about rust-toolchain.toml approach being the old way to handle this.

That said, the underlying concern remains: right now `main` has no Rust version pinning at all as far as I can see.
CI builds with whatever Rust version ships on the GitHub runner image, which changes when GitHub updates the image, outside our control.
The `rust-version` field in `Cargo.toml` only sets a floor and it won't prevent a runner upgrade from silently changing the compiler version we ship with.

A middle ground: pin the version in CI using [`actions-rust-lang/setup-rust-toolchain`](https://github.com/actions-rust-lang/setup-rust-toolchain) in the workflow files. This:

  - Controls the exact build version where you said it should be controlled ("Control comes from the Rust version you execute with")
  - Does not affect network-restricted builds
  - Replaces the existing `Swatinem/rust-cache` step (it bundles it)
  - Keeps rust-toolchain.toml off main

 Would that approach work for you?

ctron:
> That said, the underlying concern remains: right now main has no Rust version pinning at all as far as I can see.

And that's correct. It has a minimum version: https://github.com/guacsec/trustify/blob/36d0b5d019832414537f817f2c59274c462e054e/Cargo.toml#L31 and future Rust versions are backwards compatible. That's a guarantee of Rust which they so far kept. Forcing a specific Rust version on to someone is an anti-pattern.

> CI builds with whatever Rust version ships on the GitHub runner image, which changes when GitHub updates the image, outside our control.

As I said, the CI is in control. We control the CI. So it is in our control. Either through `rustup` some action. Like the one you linked.

desmax74:
@ctron are you ok with these latest changes ?

ctron:
I still think that pinning anything is a mistake. As that lead to the problem in the first place.

But yes, that's the way it pinning it.

ctron (approved):

desmax74:
@ctron we can switch easily in a single point now without the toolchain file https://github.com/actions-rust-lang/setup-rust-toolchain#inputs

PR #2396, opened 2026-06-12
Align trustify-ui dependency

desmax74:
Align Trustify-ui branch release/0.5.z

## Summary by Sourcery

Build:
- Raise minimum supported Rust version to 1.94.0.


ctron (changes requested):
Please remove the toolchain and rustfmt file. We removed those for a good reason in the past.

desmax74:
@ctron without the toolchain isn't used the rust 1.96.0 (supposed to be the latest)
https://github.com/guacsec/trustify/actions/runs/27412029286/job/81015385984#step:6:781 how we can declare the specific rust version we want to use ?

PR #2401, opened 2026-06-16
perf(analysis): batch CPE resolution pipeline to fix latest/component

mrrajan:
Replace sequential N+1 query patterns in the CPE resolution pipeline with batched operations:

- Replace iterative find_node_ancestors loop with a single recursive CTE (N queries per node → 1 query)
- Convert find_external_refs from recursive DFS to iterative BFS with batch queries per level (eliminates exponential query growth)
- Add batch versions: find_node_ancestors_batch, describing_cpes_batch, resolve_rh_external_sbom_ancestors_batch
- Update enrich_external_sboms to use batch CPE lookup (N queries → 1)
- Add composite index on sbom_node_checksum(node_id, sbom_id) to eliminate bitmap-AND on separate single-column indexes
- Fix missing tokio feature on async-compression in common crate

Implements TC-4364

Assisted-by: Claude Code

## Summary by Sourcery

Batch CPE resolution and ancestor traversal in the analysis pipeline to eliminate per-node N+1 queries and improve performance.

New Features:
- Introduce batched variants for node ancestor lookup, CPE description lookup, and external SBOM ancestor resolution, allowing multiple SBOM nodes to be processed in a single query pass.

Enhancements:
- Rewrite node ancestor traversal to use recursive SQL CTEs instead of iterative per-level queries, reducing query count and adding in-database cycle protection.
- Refactor external SBOM reference resolution from recursive DFS to iterative BFS with level-wise batch queries.
- Update external SBOM enrichment to use batched CPE lookups for all matched SBOMs in one step.
- Add regression tests to ensure batch APIs return the same results as the existing per-node implementations.

Build:
- Enable the "tokio" feature for the async-compression dependency in the common crate.

Tests:
- Add tests verifying equivalence between batched and individual ancestor resolution and external SBOM ancestor resolution APIs.

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -9,7 +9,7 @@ rust-version.workspace = true
 [dependencies]
 actix-web = { workspace = true }
 anyhow = { workspace = true }
-async-compression = { workspace = true, features = ["gzip", "lzma"] }
+async-compression = { workspace = true, features = ["gzip", "lzma", "tokio"] }
 bytes = { workspace = true }
 bytesize = { workspace = true, features = ["serde"] }
 chrono = { workspace = true }
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -58,6 +58,7 @@ mod m0002160_fix_ref_fk;
 mod m0002170_drop_cvss_tables;
 mod m0002180_advisory_fk_indexes;
 mod m0002190_vulnerability_base_score_advisory;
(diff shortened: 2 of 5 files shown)


rh-jfuller:
@mrrajan considering current and ongoing work on 0.4.z (which will push to main) I think this PR can be closed ... do you agree ?

mrrajan:
Agree @rh-jfuller

PR #2402, opened 2026-06-16
perf(analysis): paginate before hydration to fix component search timeout

mrrajan:
Split graph node processing into three phases:
1. Count matching nodes across cached graphs (cheap in-memory filter)
2. Apply offset/limit to the globally-flattened node index iterator
3. Run expensive Collector hydration only for the paged subset

This reduces per-request work from ~39,280 Collector runs to ~25 (the default page limit), a ~1,500x improvement.

Also:
- Add deterministic graph ordering via ORDER BY sbom_id in load_graphs_subquery and sorted/deduped Vec in load_latest_graphs_query
- Migrate log::debug to tracing::debug in retrieve_latest
- Fix missing tokio feature on async-compression in common crate
- Add 4 pagination tests: limits, offset, cross-graph, endpoint

Implements TC-4365

Assisted-by: Claude Code

## Summary by Sourcery

Apply pagination to graph node queries before expensive hydration to dramatically reduce per-request analysis workload and prevent timeouts.

New Features:
- Add paginated graph query execution that flattens nodes across graphs, applies global offset/limit, and hydrates only the selected subset.

Bug Fixes:
- Ensure async-compression is correctly configured with the tokio feature in the common crate to fix runtime issues.

Enhancements:
- Introduce deterministic graph and node ordering via sorted SBOM IDs and (sbom_id, node_id) sorting to make pagination stable across requests.
- Refactor analysis retrieval paths to reuse a shared paginated graph query helper instead of paginating in-memory results.
- Switch debug logging in graph loading to use tracing for consistency with the rest of the codebase.

Tests:
- Add service-level tests to verify pagination limits hydration, honors offsets, and spans results across multiple SBOM graphs without duplication.
- Add endpoint-level test to validate paginated HTTP responses return correct totals and distinct items across pages.

### DIFF
--- a/modules/analysis/src/service/load/mod.rs
+++ b/modules/analysis/src/service/load/mod.rs
@@ -22,7 +22,7 @@ use sea_orm::{
 use sea_query::{JoinType, SelectStatement};
 use serde_json::Value;
 use std::{
-    collections::{HashMap, HashSet, hash_map::Entry},
+    collections::{BTreeSet, HashMap, HashSet, hash_map::Entry},
     fmt::Debug,
     str::FromStr,
     sync::Arc,
@@ -403,13 +403,11 @@ impl InnerService {
         log::debug!("ranked sboms: {:?}", ranked_sboms);
 
         // retrieve only ranked_sboms with rank = 1, sorted for deterministic pagination
-        let mut latest_ids: Vec<_> = ranked_sboms
+        let latest_ids: BTreeSet<_> = ranked_sboms
             .into_iter()
             .filter(|item| item.rank == Some(1))
             .map(|item| item.matched_sbom_id)
             .collect();
-        latest_ids.sort();
-        latest_ids.dedup();
 
         tracing::debug!("latest sboms: {:?}", latest_ids);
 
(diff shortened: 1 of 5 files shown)


modules/analysis/src/service/load/mod.rs
```diff
@@ -402,14 +402,16 @@ impl InnerService {
         apply_rank(&mut ranked_sboms);
         log::debug!("ranked sboms: {:?}", ranked_sboms);
 
-        // retrieve only ranked_sboms with rank = 1
-        let latest_ids: HashSet<_> = ranked_sboms
+        // retrieve only ranked_sboms with rank = 1, sorted for deterministic pagination
+        let mut latest_ids: Vec<_> = ranked_sboms
```
ctron: Why not use a `BTreeSet`?

mrrajan: [sdlc-workflow/verify-pr] Classified as **question** — this asks for clarification about the data structure choice (Vec+sort+dedup vs BTreeSet); no code change requested. No sub-task created.

mrrajan: Agree on this - updated Agentic workflow to prefer idiomatic types from the programming language

mrrajan:
@rh-jfuller This PR is based on main branch, but the changes are still considerable - I will raise a new PR based on `release/0.4.z`

PR #2403, opened 2026-06-16
perf(analysis): batch external SBOM resolution in Collector

mrrajan:
Add two-pass pre-fetch for external resolution data before collection:
- Pass 1: fetch checksums for initially loaded SBOMs
- Derive: cross-SBOM reverse lookup by checksum value (discovers ancestor SBOMs not in the initial graph set)
- Pass 2: fetch sbom_external_node entries for all discovered SBOMs

Store results in ExternalResolutionCache (shared via Arc across Collectors). collect_package uses in-memory HashMap lookups instead of per-node DB queries, with fallback to DB for transitively-loaded external graphs not in the pre-fetch set.

Reduces cache-bypass queries from ~250 per request to ~8.

Implements TC-4367

Assisted-by: Claude Code

## Summary by Sourcery

Batch and cache external SBOM resolution data to reduce per-node database lookups during analysis collection.

New Features:
- Introduce an ExternalResolutionCache shared across collectors to serve checksum-based cross-SBOM resolution from memory.
- Add prefetching of external resolution data for all relevant SBOMs before graph collection.

Enhancements:
- Update Collector construction and cloning to propagate the shared external resolution cache.
- Switch logging in analysis retrieval to use structured tracing debug logs instead of plain debug statements.

Build:
- Enable the tokio feature on the async-compression dependency to support the updated runtime environment.

Tests:
- Add an integration test to verify that the external resolution cache correctly performs cross-SBOM ancestor resolution based on checksums.

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -9,7 +9,7 @@ rust-version.workspace = true
 [dependencies]
 actix-web = { workspace = true }
 anyhow = { workspace = true }
-async-compression = { workspace = true, features = ["gzip", "lzma"] }
+async-compression = { workspace = true, features = ["gzip", "lzma", "tokio"] }
 bytes = { workspace = true }
 bytesize = { workspace = true, features = ["serde"] }
 chrono = { workspace = true }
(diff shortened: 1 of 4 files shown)


rh-jfuller:
@mrrajan considering current and ongoing work on 0.4.z (which will push to main) I think this PR can be closed ... do you agree ?

mrrajan:
Agree @rh-jfuller

PR #2405, opened 2026-06-16
[0.5.z] Configure a specific rust version

desmax74:
Backport of https://github.com/guacsec/trustify/pull/2395/

## Summary by Sourcery

Pin the Rust toolchain version in CI and automation to ensure consistent builds and checks.

Enhancements:
- Declare the workspace Rust version in Cargo.toml to document the required compiler version.

Build:
- Use the shared Rust setup action in binary build workflows to align build tooling with the pinned Rust version.

CI:
- Introduce a reusable composite action to set up the Rust toolchain at version 1.96.0 with optional components and use it across CI workflows for clippy, rustfmt, llvm-tools, and benchmarks.

### DIFF
--- a/.github/actions/setup-rust/action.yaml
+++ b/.github/actions/setup-rust/action.yaml
@@ -1,0 +1,18 @@
+name: Setup Rust
+description: Setup Rust toolchain with the project's pinned version
+
+inputs:
+  components:
+    description: Comma-separated list of additional components to install (e.g. clippy, rustfmt)
+    required: false
+    default: ""
+
+runs:
+  using: "composite"
+
+  steps:
+    - uses: actions-rust-lang/setup-rust-toolchain@v1
+      with:
+        toolchain: 1.96.0
+        components: ${{ inputs.components }}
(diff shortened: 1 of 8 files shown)


ctron (approved):
Again, in the future, this might just lead to the same case where we are not aware of things that break when newer Rust versions are being released.

But if we want pinning, then this works.

desmax74:
@ctron now we can choose to put _stable_ here https://github.com/guacsec/trustify/pull/2405/changes#diff-11f0ae113d663d6f9c68799547596c0c2eb2bd0437b5b45c357bfe3043e4d963R16 on main branch and a number on release/0.y.z branches, or _stable,1.96_ on main, is allowed to use multiple values

ctron:
Right we can use `stable`, but that's what you get from GH anyway.

desmax74:
Not exactly, now we can ask something that will be available on GH but we can ask before will be available in the runners, and in a controlled way we can choose, before we was just consumer without option to choose.

PR #2410, opened 2026-06-23
build: don't exclude download, it's required

ctron:
This only changes the release process part, not actual code.

Claude added an exclude for `/download`, I missed that, and we need it. That's why the last parts of the release build fail.

## Summary by Sourcery

Update project metadata and container ignore configuration for the new release.

Build:
- Adjust container ignore configuration to ensure required download assets are included in release builds.

Chores:
- Bump workspace and API specification versions from 0.4.14 to 0.4.15.

### DIFF
--- a/.containerignore
+++ b/.containerignore
@@ -8,7 +8,6 @@
 /.idea
 /.claude
 /docs/book
-/download
 /vendor
 **/node_modules
 **/.terraform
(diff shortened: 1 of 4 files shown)


rh-jfuller (approved):
LGTM

helio-frota (approved):

PR #2408, opened 2026-06-19
Materialize ancestor information

ctron:
~~**Note:** this PR builds on top of https://github.com/guacsec/trustify/pull/2407, only the last few commits are relevant.~~

It creates a table materializing the ancestor information of the RH model. Thus saving us from re-exploring that over and over again.

This brings down the "pypi request" query from days to ~13 mins:

```
➜  trustify git:(feature/atlas_perf_1_mat) time xh GET localhost:8080/api/v2/analysis/latest/component?q=purl=pkg%3Apypi%2Frequests (oidc token -Hf trusty)
HTTP/1.1 200 OK
Access-Control-Allow-Credentials: true
Access-Control-Expose-Headers: content-type
Content-Encoding: br
Content-Type: application/json
Date: Fri, 19 Jun 2026 06:49:53 GMT
Transfer-Encoding: chunked
Vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
Vary: accept-encoding

{
    "items": [],
    "total": 0
}



________________________________________________________
Executed in  747.02 secs      fish           external
   usr time   24.16 millis    0.00 micros   24.16 millis
   sys time   20.16 millis  929.00 micros   19.23 millis

```

However, this is a draft and it may break stuff. So let's be optimistic, but cautious.

## Summary by Sourcery

Materialize SBOM ancestor and describing-CPE relationships to avoid expensive runtime graph walks and improve analysis query performance.

New Features:
- Introduce materialized sbom_describing_cpe and sbom_ancestor tables plus corresponding ORM entities to store SBOM describing CPEs and cross-SBOM ancestor links.
- Populate describing CPEs and ancestor links during SBOM ingestion for CycloneDX and SPDX documents.
- Add a Containerfile and .containerignore for building and packaging the trustd binary into a container image.

Enhancements:
- Refactor SBOM CPE and ancestor resolution to use batched queries and the new materialized tables instead of recursive in-memory graph traversal.
- Optimize product advisory and context CPE queries to read from sbom_describing_cpe, simplifying SQL and reducing join cost.
- Improve logging with truncated iterable formatting and richer instrumentation for graph loading and ranking.
- Add a helper to migrate the database up to a specific migration and extend the trustd db CLI to support targeted migrations.
- Update several dependencies (moka, petgraph, sea-orm, tokio, uuid) and align uuid usage in the storage module.
- Adjust tests and entities to reflect the new ancestor and describing-CPE materialization behavior.

Tests:
[...]

PR #2407, opened 2026-06-19
Improve performance of latest endpoint in large data deployments

ctron:
## Summary by Sourcery

Optimize latest SBOM endpoint and related analysis by batching and caching graph queries, materializing describing CPE mappings, and tightening database indexing and migrations to scale better on large datasets.

New Features:
- Add materialized sbom_describing_cpe entity and population logic to speed up CPE lookups for describing components.
- Introduce configurable published-date cutoff for latest SBOM selection via environment variable.
- Provide CLI support for migrating the database up to a specific migration by name.
- Add container build artifacts (Containerfile and .containerignore) for building and running the trustd service.

Bug Fixes:
- Ensure ancestor walking and external SBOM resolution avoid cycles and handle multiple parents correctly while maintaining ranking semantics.
- Fix advisory sorting in tests by simplifying sort key usage to avoid inconsistent ordering.

Enhancements:
- Refactor latest-graph loading and CPE resolution to use batched queries, iterative DFS with caching, and reduced N+1 database calls.
- Add parallel graph loading with bounded concurrency and improved logging using truncated iter formatting.
- Replace inline describing CPE queries with reads from the new materialized table and adjust advisory product queries accordingly.
- Improve tracing and debug output for latest SBOM selection, cache hit/miss statistics, and ancestor graph traversal.
- Update time handling from chrono to time::OffsetDateTime in ranking logic and tests.

Build:
- Bump moka, petgraph, sea-orm, tokio, and uuid versions to newer releases.
- Add async-compression dependency for common utilities and unify uuid versions across crates.

Deployment:
- Introduce a multi-stage UBI-based container build for the trustd binary.

Documentation:
- Document the new TRUSTIFY_LATEST_SBOM_PUBLISHED_CUTOFF_DAYS environment variable in the env-vars reference.

Tests:
- Extend and adapt analysis, ingestor, and advisory tests to cover new ancestor traversal, CPE materialization, and external SBOM resolution behaviors.

Chores:
- Remove obsolete rust-toolchain and rustfmt configuration files and adjust recursion limits in several crates to accommodate deeper traversals.

PR #2412, opened 2026-06-24
Materialize ancestor information

ctron:
This is a forward-ported version of PR #2407 and #2408 

It creates a table materializing the ancestor information of the RH model. Thus saving us from re-exploring that over and over again.

This brings down the "pypi request" query from days to ~13 mins:

```
➜  trustify git:(feature/atlas_perf_1_mat) time xh GET localhost:8080/api/v2/analysis/latest/component?q=purl=pkg%3Apypi%2Frequests (oidc token -Hf trusty)
HTTP/1.1 200 OK
Access-Control-Allow-Credentials: true
Access-Control-Expose-Headers: content-type
Content-Encoding: br
Content-Type: application/json
Date: Fri, 19 Jun 2026 06:49:53 GMT
Transfer-Encoding: chunked
Vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
Vary: accept-encoding

{
    "items": [],
    "total": 0
}



Executed in  747.02 secs      fish           external
   usr time   24.16 millis    0.00 micros   24.16 millis
   sys time   20.16 millis  929.00 micros   19.23 millis

```


rh-jfuller:
so ... running scale tests we encounter the old issue with those license queries eg. I see many of the same following license queries stacking up:

```
WITH "licensing_infos_mappings" AS (SELECT ARRAY_AGG(CAST((ROW("license_id", "name")) AS license_mapping)) AS "license_mapping", "licensing_infos"."sbom_id" FROM "licensing_infos" GROUP BY "licensing_infos"."sbom_id") , "unique_license_sbom" AS (SELECT DISTINCT "license"."text", "sbom_package_license"."sbom_id", "license"."id" AS "license_id" FROM "sbom_package_license" JOIN "license" ON "sbom_package_license"."license_id" = "license"."id") , "expanded" AS (SELECT "unique_license_sbom"."sbom_... |
```

which tie up and rob resources for any other queries being performed -  making the entire set of rest api endpoints slow down.

migration/src/m0002130_sbom_ancestor.rs
```diff
@@ -0,0 +1,124 @@
...
+        manager
+            .get_connection()
+            .execute_unprepared(
+                r#"
+                CREATE INDEX IF NOT EXISTS sbom_node_checksum_value_sbom_id_idx
+                ON sbom_node_checksum (value, sbom_id)
+                "#,
+            )
+            .await
+            .map(|_| ())?;
```
rh-jfuller: we might consider a composite index as well

```
CREATE INDEX sbom_ancestor_sbom_id_ancestor_idx
  ON sbom_ancestor (sbom_id, ancestor_sbom_id);
```

modules/analysis/src/service/load/rank.rs
```diff
@@ -110,26 +110,43 @@ async fn batch_resolve_direct_cpe_matches(
...
-                    .eq(PgFunc::any(all_node_ids)),
-            )
-            .all(connection)
-            .instrument(tracing::info_span!("batch lookup nodes").or_current())
-            .await
-            .map_err(Error::from)?;
+        let nodes = NodeWithAncestor::find_by_statement(Statement::from_sql_and_values(
+            DatabaseBackend::Postgres,
+            r#"
+            SELECT sn.sbom_id, sn.node_id, sn.name,
```
rh-jfuller, line 136: this is fine ... though this area of the codebase is likely to migrate to the db ... there is a way to rewrite this query to be ensure stable query plan (ex. using unnest on supplied arrays) but not for consideration now.

migration/src/m0002130_sbom_ancestor.rs
ctron: Claude says:

>   The PK on sbom_ancestor is (sbom_id, ancestor_sbom_id) (line 22-25), which in PostgreSQL automatically creates a B-tree index on those columns in that order.
  The suggested index (sbom_id, ancestor_sbom_id) is identical to the PK index — it would be redundant.

rh-jfuller (approved):
LGTM

migration/src/m0002130_sbom_ancestor.rs
rh-jfuller: hehe, claude needs to learn more - its ok to leave it for now ...

PR #2424, opened 2026-06-29
fix: optimize expanded_license migration backfill

ctron:
## Summary

* Split migration backfill queries into two passes (LicenseRef- vs non-LicenseRef-), avoiding the expensive LEFT JOIN with `licensing_infos` and PL/pgSQL function calls for the majority of rows
* Add `ANALYZE` before backfill and use `MATERIALIZED` CTE to fix planner misestimates and prevent double function evaluation
* Replace `regexp_replace()` with boundary-aware `replace()` calls in `expand_license_expression_with_mappings()` for a ~2.16x speedup

### Query plan improvement

The original single-query backfill had a planner cost of ~1.8B due to a 170x row overestimate (7B vs ~40M) on the Hash Left Join. After splitting and adding ANALYZE, the LicenseRef- pass drops to ~15M cost (~120x reduction).

### Function benchmark (17 expression patterns x 1000 iterations, 8 mappings)

```
regex   : run1=226.3 ms  run2=226.8 ms  run3=226.4 ms  avg=226.5 ms
replace : run1=105.3 ms  run2=104.8 ms  run3=104.6 ms  avg=104.9 ms
speedup : 2.16x
```

## Test plan

- [ ] Run `cargo build -p trustify-migration` (passes)
- [ ] Run the benchmark script against a test database: `psql -f migration/src/m0002120_normalize_expanded_license/bench_expand.sql`
- [ ] Verify migration applies cleanly on a database with existing SBOM data
- [ ] Verify rollback (`down.sql`) restores the regex-based function

## Summary by Sourcery

Optimize the expanded license migration backfill for performance and planner stability.

New Features:
- Add a standalone SQL benchmark script to compare regex-based vs replace()-based license expression expansion.

Enhancements:
- Replace the regex-based expand_license_expression_with_mappings() implementation with a boundary-aware replace()-based version for faster expansion.
- Split the migration backfill into separate passes for LicenseRef and non-LicenseRef licenses to avoid unnecessary joins and function calls.
- Add ANALYZE statements and use a MATERIALIZED CTE in the backfill to improve query planner estimates and avoid duplicate function evaluations.

### DIFF
--- a/migration/src/m0002120_normalize_expanded_license/bench_expand.sql
+++ b/migration/src/m0002120_normalize_expanded_license/bench_expand.sql
@@ -1,0 +1,253 @@
+-- Benchmark: regexp_replace vs replace() for license expression expansion
+-- Run with: psql -f bench_expand.sql <database>
+--
+-- Creates temporary functions and test data, runs both implementations,
+-- verifies they produce identical results, and reports timings.
+
+-- Ensure the license_mapping type exists
+DO $$
+BEGIN
+    CREATE TYPE license_mapping AS (license_id TEXT, name TEXT);
+EXCEPTION WHEN duplicate_object THEN NULL;
+END $$;
+
+----------------------------------------------------------------------
+-- 1. Define both function variants under benchmark-specific names
+----------------------------------------------------------------------
+
+-- Old version: regex-based (dynamically compiled pattern per mapping)
+CREATE OR REPLACE FUNCTION bench_expand_regex(
+    license_text TEXT,
+    license_mappings license_mapping[]
+) RETURNS TEXT AS $$
+DECLARE
+    result_text TEXT := license_text;
+    mapping license_mapping;
+BEGIN
+    IF license_text IS NULL
+       OR POSITION('LicenseRef-' IN license_text) = 0
+       OR license_mappings IS NULL
+       OR array_length(license_mappings, 1) IS NULL THEN
+        RETURN license_text;
+    END IF;
+
+    FOREACH mapping IN ARRAY license_mappings
+    LOOP
+        IF POSITION('LicenseRef-' IN result_text) = 0 THEN
+            EXIT;
+        END IF;
+        result_text := regexp_replace(
+            result_text,
+            '\m' || mapping.license_id || '(?![a-zA-Z0-9.\-])',
+            mapping.name, 'g'
+        );
+    END LOOP;
+
+    RETURN result_text;
+END;
+$$ LANGUAGE plpgsql IMMUTABLE PARALLEL SAFE;
+
+-- New version: replace()-based (no regex, boundary-aware via SPDX delimiters)
+CREATE OR REPLACE FUNCTION bench_expand_replace(
+    license_text TEXT,
+    license_mappings license_mapping[]
+) RETURNS TEXT AS $$
+DECLARE
+    result_text TEXT;
+    mapping license_mapping;
+BEGIN
+    IF license_text IS NULL
+       OR POSITION('LicenseRef-' IN license_text) = 0
+       OR license_mappings IS NULL
+       OR array_length(license_mappings, 1) IS NULL THEN
+        RETURN license_text;
+    END IF;
+
+    result_text := ' ' || license_text || ' ';
+
+    FOREACH mapping IN ARRAY license_mappings
+    LOOP
+        IF POSITION('LicenseRef-' IN result_text) = 0 THEN
+            EXIT;
+        END IF;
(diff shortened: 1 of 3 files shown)


migration/src/m0002120_normalize_expanded_license/up.sql
```diff
@@ -25,36 +25,114 @@ CREATE TABLE IF NOT EXISTS sbom_license_expanded (
...
+    END LOOP;
+
+    RETURN substring(result_text FROM 2 FOR length(result_text) - 2);
+END;
+$$ LANGUAGE plpgsql IMMUTABLE PARALLEL SAFE;
+
+-- Update planner statistics before backfill for accurate cost estimates
+ANALYZE sbom_package_license;
+ANALYZE licensing_infos;
+ANALYZE sbom_license_expanded;
```
rh-jfuller, line 73: isnt this an empty table ? maybe we run this after things are populated

migration/src/m0002120_normalize_expanded_license/bench_expand.sql
```diff
@@ -0,0 +1,253 @@
+-- Benchmark: regexp_replace vs replace() for license expression expansion
+-- Run with: psql -f bench_expand.sql <database>
```
rh-jfuller, line 2: this is a manual check ... which probably does not need a separate test/CI in case of future changes but maybe we add some comment in migration.rs ... also if funcs are just for migration do they need to continue to live after migration ?

rh-jfuller (approved):
there maybe some minor diffs with old and new approach ex.  tab / new linedelimited expr (ex. LicenseRef-x\tAND ...) expands under old regex  but is unexpanded by new code ... is the new code matching _literal ASCII space only_ intentional ?

otherwise LGTM !

ctron:
> there maybe some minor diffs with old and new approach ex. tab / new linedelimited expr (ex. LicenseRef-x\tAND ...) expands under old regex but is unexpanded by new code ... is the new code matching _literal ASCII space only_ intentional ?
> 
> otherwise LGTM !

Claude said: it works.

● > Good question — let me walk through it precisely.
  >
  > **The regex pattern:** `\m` + `license_id` + `(?![a-zA-Z0-9.\-])`
  >
  > - `\m` — PostgreSQL word-start anchor: matches where the previous character is NOT in `[a-zA-Z0-9_]` (or at start of string)
  > - `(?![a-zA-Z0-9.\-])` — negative lookahead: the next character must NOT be alphanumeric, dot, or hyphen
  >
  > This means the regex matches `license_id` when it's bounded by "not a word/identifier character" on both sides.
  >
  > **The replace() version** uses sentinel spaces (to handle start/end of string) and checks 6 explicit boundary combinations:
  >
  > | Before | After | Example |
  > |--------|-------|---------|
  > | space | space | `MIT AND LicenseRef-foo AND BSD` |
  > | space | `)` | `LicenseRef-foo)` |
  > | space | `+` | `LicenseRef-foo+` |
  > | `(` | space | `(LicenseRef-foo AND MIT)` |
  > | `(` | `)` | `(LicenseRef-foo)` |
  > | `(` | `+` | `(LicenseRef-foo+)` |
  >
  > **Why this covers all valid SPDX expressions:** The SPDX expression grammar only allows license identifiers to be surrounded by:
  > - Before: space, `(`, or start of string
  > - After: space, `)`, `+`, or end of string
  >
  > Operators (`AND`, `OR`, `WITH`) are always space-separated. There are no other valid delimiters in the grammar. So the 6 combinations + sentinels are exhaustive for well-formed expressions.
  >
  > **Where they genuinely differ:** The regex `\m` is broader — it treats *any* non-word character as a valid boundary (tab, comma, colon, semicolon, etc.). Two cases where they'd diverge:
  >
  > 1. **`DocumentRef-xxx:LicenseRef-yyy`** — if a mapping targets just `LicenseRef-yyy`, the regex would match after `:` (non-word char), but replace() would not. However, replacing inside a `DocumentRef:LicenseRef` pair would likely
  produce broken output, so not matching is arguably more correct here.
  >
  > 2. **Malformed expressions** with unusual delimiters (tabs, commas, semicolons) — the regex would match, replace() would not. These aren't valid SPDX syntax.
  >
  > So the replace() version is equivalent for valid SPDX input and more conservative for malformed input. The benchmark's correctness check passed on all 17 test patterns, but those are all well-formed. If there's concern about malformed
  data in the database, it's worth checking.

ctron:
And yes, we do prevent illegal SPDX license expressions from being ingested

PR #2429, opened 2026-06-30
perf: focus on main entities used by ux.

rh-jfuller:
## Summary by Sourcery

Optimize SBOM summary construction and license search queries to reduce per-entity database calls and improve performance.

New Features:
- Add batch queries to compute package counts and describes-packages for multiple SBOMs at once.

Enhancements:
- Refactor SBOM summary construction to use batched queries instead of per-entity lookups.
- Introduce an SbomHead::from_parts constructor for building summaries from pre-fetched data.
- Adjust license search union query to avoid DISTINCT, use UNION ALL, and conditionally apply ordering while ensuring correct count queries.


rh-jfuller:
will be superseded by #2430

PR #2446, opened 2026-07-03
perf(analysis): paginate before hydration to fix component search timeout

mrrajan:
## Summary
   - **Paginate before hydration**: moves pagination from post-hydration array slicing to pre-hydration index-level skip/take via `collect_graph_paged`, eliminating the timeout caused by hydrating thousands of nodes only to discard most of them.
   - **Deterministic ordering**: uses `BTreeSet` for ranked SBOM IDs and adds `order_by_asc(sbom_id)` to subquery-based loading, ensuring stable pagination across requests.
   - **Adds `count_matching_nodes`** for efficient total counting without hydrating any nodes.

   ## Test plan
   - [x] Unit test `test_pagination_limits_hydration` — verifies only the requested number of nodes are hydrated while `total` reflects all matches
   - [x] Unit test `test_pagination_offset` — verifies different offsets produce different results and union matches combined query
   - [x] Unit test `test_pagination_cross_graph` — verifies pagination spans multiple SBOM graphs without duplicates or gaps
   - [x] Endpoint test `test_pagination_through_endpoint` — verifies pagination via HTTP returns correct totals and distinct pages

## Summary by Sourcery

Apply pagination at the graph index level before node hydration to improve performance and provide deterministic, cross-graph component retrieval with accurate totals.

New Features:
- Add paginated graph query execution that counts matching nodes without hydrating them and applies global offset/limit across graphs.

Enhancements:
- Ensure deterministic pagination by sorting ranked SBOM IDs and graph nodes by SBOM and node identifiers.
- Wire all component retrieval paths to the new paginated graph query flow instead of array-based pagination.

Tests:
- Add service-level tests to verify hydration is limited by pagination, offsets yield distinct pages, and pagination spans multiple SBOM graphs without gaps or duplicates.
- Add endpoint-level test to ensure HTTP pagination returns correct totals and non-overlapping pages.

### DIFF
--- a/modules/analysis/src/endpoints/tests/mod.rs
+++ b/modules/analysis/src/endpoints/tests/mod.rs
@@ -172,9 +172,7 @@ async fn test_quarkus_retrieve_analysis_endpoint(
 /// Verify pagination through the HTTP endpoint returns correct total and limited items.
 #[test_context(TrustifyContext)]
 #[test(actix_web::test)]
-async fn test_pagination_through_endpoint(
-    ctx: &TrustifyContext,
-) -> Result<(), anyhow::Error> {
+async fn test_pagination_through_endpoint(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
     let app = caller(ctx).await?;
     ctx.ingest_documents([
         "spdx/quarkus-bom-3.2.11.Final-redhat-00001.json",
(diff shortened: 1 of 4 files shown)


rh-jfuller:
The scale test reports improvements for rest api endpoints but results are mixed eg. max time (95 percentile) increases but we also see a higher RPS which is positive - however the slow rest api performance is negatively impacted. 

At face value the impact is >25% improvement (that is prob biased by the perf tests, improvement might be better then that).

I think this PR is definitely an improvement we want, though needs more analysis (I will run newer perf tests and look more at the code changes) ... lets pick it up next week - I do not think its a candidate yet for **0.5.z**.

PR #2454, opened 2026-07-07
fix: deduplicate vulnerabilities across advisories in batch severity counts

ctron:
## Summary

- The batch `?advisories=true` query on `GET /v3/sbom` double-counted vulnerabilities that appeared in multiple advisories. The `scored` CTE used `DISTINCT ON (sbom_id, advisory_id, vulnerability_id)`, so the same CVE reported by two advisories produced two rows and inflated severity counts.
- Changed to `DISTINCT ON (sbom_id, vulnerability_id)` to count each vulnerability once, matching the per-SBOM endpoint behavior.
- Derived `Ord` on `Severity` and `AffectedSeverity` to enable `.max()` instead of manual rank functions.

## Test plan

- [x] Added `batch_vs_per_sbom_advisory_severity_counts` test that ingests two advisories for the same CVE and asserts the batch count is `{Medium: 1}` (not `{Medium: 2}`)
- [x] Verified no performance regression via `EXPLAIN ANALYZE` on a 263k-SBOM database — same plan, same execution time
- [ ] Run full test suite (`cargo xtask precommit`)

## Summary by Sourcery

Deduplicate vulnerability severity counts in the batch SBOM advisories endpoint to align with per-SBOM behavior.

Bug Fixes:
- Prevent double-counting vulnerabilities in batch advisory severity aggregation when the same CVE appears in multiple advisories.

Enhancements:
- Derive ordering on Severity and AffectedSeverity to support consistent severity comparison and aggregation.

Tests:
- Add an integration test and CSAF fixture to verify batch advisory severity counts match per-SBOM results when multiple advisories reference the same CVE.

### DIFF
--- a/modules/fundamental/src/sbom/model/raw_sql.rs
+++ b/modules/fundamental/src/sbom/model/raw_sql.rs
@@ -170,7 +170,8 @@ pub fn batch_severity_counts_sql() -> &'static str {
 
     -- Pick the highest severity per (sbom, vulnerability), collapsing
     -- across advisories and CVSS versions into one row per unique vuln.
-    -- Unknown (no CVSS score) is treated as the highest severity.
+    -- Unknown (no CVSS score) is treated as the lowest severity so that
+    -- a real score from any advisory always wins.
     scored AS (
         SELECT DISTINCT ON (a.sbom_id, a.vulnerability_id)
             a.sbom_id,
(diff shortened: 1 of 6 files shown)


ctron:
@sourcery-ai review

modules/fundamental/src/sbom/model/raw_sql.rs
rh-jfuller: I wonder if something is 'unknown' could mask things .... unknown only wins when no advisory has a score for that vuln maybe this should be ELSE 0 ?

rh-jfuller (approved):
apart from existing comments, LGTM

PR #2463, opened 2026-07-08
perf: resolve rh external SBOM ancestors perf

rh-jfuller:
When performing latest queries with ancestor resolution there were a few issues causing less then optimal performance.

> http://localhost:8080/api/v3/analysis/latest/component/clevis?total=false&ancestors=5

**resolve_rh_external_sbom_ancestors** (modules/analysis/src/service/mod.rs:270) is performing 2 seq DB queries per PackageNode visited during ancestor traversal (and in other places) - the fix was to ensure we further use materialised **sbom_ancestor** table.

Needed to also fix Ancestor cache which was not working as efficiently due to race condition by using a coalescing pattern so concurrent requests, for the same key, share a single 'in flight' query instead of each hitting the DB independently (causing resource contention).

Lastly we needed to ensure that the AncestorCache is created only once for a request, in **run_graph_query**, and then shared across all Collector instances for the entire request:
- All nodes in the result set share one cache
- Both ancestor and descendant collectors for the same node share the cache

## Summary by Sourcery

Optimize resolution of Red Hat external SBOM ancestors and share ancestor resolution results across collectors via a per-request cache.

Enhancements:
- Introduce an AncestorCache layer using OnceCell to coalesce concurrent ancestor-resolution queries and reuse results across graph traversal.
- Refactor resolve_rh_external_sbom_ancestors to query the materialized sbom_ancestor table and return graph node identifiers, reducing per-node database roundtrips.
- Extend ResolvedSbom to carry optional graph node IDs and adjust collectors to avoid redundant sbom_external_node lookups when this data is available.

Tests:
- Update existing external SBOM resolution tests to align with the extended ResolvedSbom structure.

**Before**
image

**After**
image

your eye should look at the times for **resolve_rh_external_sbom_ancestors** where we should now have much less invokes (and no competing equiv invokes) and better run times. 

Roughly a 75% reduction in aggregated time.

### DIFF
--- a/modules/analysis/src/service/collector.rs
+++ b/modules/analysis/src/service/collector.rs
@@ -54,17 +54,11 @@ impl AncestorCache {
     ) -> Result<AncestorResult, Error> {
         let cell = {
             let mut map = self.cache.lock();
-            map.entry((sbom_id, node_id.clone()))
-                .or_default()
-                .clone()
+            map.entry((sbom_id, node_id.clone())).or_default().clone()
         };
 
         cell.get_or_try_init(|| async {
-            let result =
-                resolve_rh_external_sbom_ancestors(
-                    sbom_id, node_id, connection,
-                )
-                .await?;
+            let result = resolve_rh_external_sbom_ancestors(sbom_id, node_id, connection).await?;
             Ok(Arc::new(result))
         })
         .await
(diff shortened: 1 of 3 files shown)


modules/analysis/src/service/collector.rs
```diff
@@ -23,6 +23,55 @@ impl DiscoveredTracker {
...
+type AncestorResult = Arc<Vec<ResolvedSbom>>;
+type AncestorCell = Arc<tokio::sync::OnceCell<AncestorResult>>;
+type AncestorMap = HashMap<(Uuid, String), AncestorCell>;
+
+/// Cache for `resolve_rh_external_sbom_ancestors` results.
+///
+/// Keyed by `(sbom_id, node_id)` so that repeated lookups for the
+/// same node during ancestor traversal are served from memory.
+///
+/// Uses `tokio::sync::OnceCell` per key so that concurrent callers
```
ctron: Maybe that's off. But I don't see a `OnceCell`, but a `Mutex`.

ctron: Now I got it.

ctron (approved):
LGTM. I'd appreciate some "step-by-step" in-function comments, explaining the `OnceCell` part.

modules/analysis/src/service/mod.rs
```diff
@@ -260,53 +268,68 @@ struct ChecksumWithCpes {
...
+          ON sen.sbom_id = sa.ancestor_sbom_id
+        JOIN sbom_node_checksum snc_ref
+          ON snc_ref.sbom_id = sen.sbom_id
+         AND snc_ref.node_id = sen.external_node_ref
+        JOIN sbom_node_checksum snc_self
+          ON snc_self.sbom_id = $1
+         AND snc_self.node_id = $2
+         AND snc_self.value = snc_ref.value
+        LEFT JOIN sbom_describing_cpe sdc
+          ON sdc.sbom_id = sen.sbom_id
```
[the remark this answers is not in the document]
rh-jfuller: sbom_describing_cpe is intentionally SBOM-level, not node-level ... I think we are fine.

PR #2435, opened 2026-07-01
feat(entity): add exploit intelligence job tracking entities and migration

Strum355:
## Summary

Add database entities and migrations for tracking Exploit Intelligence analysis jobs in trustify. This is the data-layer foundation for integrating the Exploit Intelligence (EI) service, which performs automated vulnerability reachability analysis and produces VEX advisories.

## Jira

Implements [TC-4675](https://redhat.atlassian.net/browse/TC-4675)

### What changed

**New entities:**
- `exploit_intelligence_job` — tracks the lifecycle of each EI analysis request (submission → completion), including scan ID, status, associated SBOM/vulnerability, and the result (VEX advisory link or error). Supports both single-component (CycloneDX) and multi-component (SPDX product) flows via `product_id`/`total_components`.
- `exploit_intelligence_job_component` — tracks per-component analysis results within a multi-component SPDX product job. Each row represents one container image component independently analysed by the EI service.

**New enums:**
- `ExploitIntelligenceJobStatus` — `Pending`, `Running`, `Completed`, `Failed`
- `Finding` — `Vulnerable`, `NotVulnerable`, `Uncertain`

**Migration (`m0002200`):**
- Creates both tables in a single migration
- FK constraints to `sbom` and `advisory` tables (both nullable)
- CASCADE delete from job → components
- UNIQUE constraint on `scan_id` (both tables)
- FK indexes on all foreign key columns

[TC-4675]: https://redhat.atlassian.net/browse/TC-4675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Introduce database support for tracking Exploit Intelligence analysis jobs and their per-component results.

New Features:
- Add exploit_intelligence_job entity to represent Exploit Intelligence analysis jobs with lifecycle status, findings, and links to SBOMs and advisories.
- Add exploit_intelligence_job_component entity to capture per-component analysis details for multi-component Exploit Intelligence product jobs.

Enhancements:
- Register new entities in the entity module and add a migration to create the exploit_intelligence_job and exploit_intelligence_job_component tables with appropriate indexes and foreign keys.

PR #2464, opened 2026-07-08
feat: add exploit-intelligence job polling, configuration & API endpoints

Strum355:
## Summary

Add the Exploit Intelligence orchestration module to trustify. This module provides API endpoints for the UI to trigger automated vulnerability reachability analysis via the EI client service (agent-morpheus-client), and a background worker that submits SBOMs, polls for results, and ingests returned CSAF VEX documents into trustify's advisory system.

## Jira

Implements [TC-4677](https://redhat.atlassian.net/browse/TC-4677)

### What changed

**New module: `exploit_intelligence`** (`modules/fundamental/src/exploit_intelligence/`)

- `endpoints/` — HTTP handlers for triggering analysis and querying job status
- `runner/` — background processing: SBOM upload with retry, polling for results, orphaned job recovery, heartbeat coordination
- `service/` — CRUD operations on job and component entities, recovery queries
- `model/` — request/response DTOs and API models
- `auth.rs` — OIDC client credentials and static token authentication for the EI service
- `ei_client.rs` — response type definitions for the EI client API

**API endpoints:**

- `POST /v3/exploit-intelligence/analyze` — trigger analysis for a CVE against an SBOM. Returns existing active job if one already exists (deduplication).
- `GET /v3/exploit-intelligence/jobs` — paginated job list, filterable by `sbom_id` and `status`
- `GET /v3/exploit-intelligence/jobs/{id}` — job details including per-component findings, advisory links, and report URLs

**Analysis flows:**

- **CycloneDX** (single-component): upload to EI → poll `/api/v1/reports/by-scan-id/{scanId}` → extract finding → ingest VEX if available
- **SPDX** (multi-component): upload to EI → poll `/api/v1/products/{productId}` → fetch per-component reports → extract findings and ingest VEX documents individually

**Retry and polling:**

- Upload retries use the [`backon`](https://crates.io/crates/backon) crate (already a transitive dependency via `walker-common`) with configurable exponential backoff
- Polling uses `backon::ConstantBuilder` with a `Cell`-based consecutive failure counter for distinguishing transient errors from permanent failures
- All retry/poll parameters are configurable via environment variables using humantime format

**Multi-instance safety:**

- Heartbeat-based optimistic locking prevents multiple trustify instances from polling the same job. A `heartbeat` column (nanosecond timestamp) on the job table serves as a CAS lock — `UPDATE ... WHERE heartbeat = <expected>` ensures mutual exclusion.
- `SELECT ... FOR UPDATE SKIP LOCKED` during startup recovery atomically claims orphaned jobs
- `CancellationToken` for cooperative work cancellation when the heartbeat is lost
- All terminal status updates (`update_job_completed`, `update_job_failed`) use CAS against the heartbeat value — only the owning instance can transition job state
- Partial unique index `(sbom_id, vulnerability_id) WHERE status IN ('pending', 'running')` prevents duplicate active jobs at the database level

**EI-generated advisory labeling:**

- Ingested VEX documents are labeled `source: exploit-intelligence` to distinguish AI-generated findings from vendor-published advisories

**Configuration** (environment variables):

| Variable | Description | Default |
|---|---|---|
| `EXPLOIT_INTELLIGENCE_URL` | Base URL of the EI client service | — |
| `EXPLOIT_INTELLIGENCE_UI_URL` | EI web UI base URL for report deep-links | — |
| `EXPLOIT_INTELLIGENCE_POLL_INTERVAL` | Polling interval (humantime) | `30s` |
| `EXPLOIT_INTELLIGENCE_MAX_POLL_DURATION` | Max polling duration (humantime) | `30m` |
| `EXPLOIT_INTELLIGENCE_UPLOAD_MAX_RETRIES` | Upload retry attempts | `3` |
| `EXPLOIT_INTELLIGENCE_UPLOAD_RETRY_DELAY` | Initial retry delay (humantime, exponential) | `1s` |
| `EXPLOIT_INTELLIGENCE_MAX_CONSECUTIVE_POLL_FAILURES` | Consecutive poll failures before abort | `5` |
| `EXPLOIT_INTELLIGENCE_AUTH_TOKEN` | Static auth token (fallback) | — |
| `EXPLOIT_INTELLIGENCE_OIDC_TOKEN_URL` | OIDC token endpoint | — |
| `EXPLOIT_INTELLIGENCE_OIDC_CLIENT_ID` | OIDC client ID | — |
| `EXPLOIT_INTELLIGENCE_OIDC_CLIENT_SECRET` | OIDC client secret | — |

**Migration changes** (`m0002230`):

- Adds `heartbeat DECIMAL NULL` column to `exploit_intelligence_job` table
- Adds partial unique index `idx_ei_job_active_dedup` on `(sbom_id, vulnerability_id)` for active jobs

### Dependencies

- Depends on [TC-4675 / PR #2435](https://github.com/guacsec/trustify/pull/2435) (entity and migration for job tracking)
- Adds `backon` as a direct workspace dependency (already transitive via `walker-common`)
- Adds `humantime` to the server crate for duration configuration
- Adds `tokio-util` to fundamental for `CancellationToken`

[TC-4677]: https://redhat.atlassian.net/browse/TC-4677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
```diff
@@ -0,0 +1,1244 @@
...
+use trustify_module_storage::service::StorageBackend;
+use uuid::Uuid;
+
+// ---------------------------------------------------------------------------
+// Retry helpers for HTTP requests to the EI service
+// ---------------------------------------------------------------------------
+
+/// Retry categories for HTTP responses.
+#[allow(dead_code)]
+enum RetryDecision {
```
ruromero: The retry and polling patterns are implemented locally and I'm sure other parts of the code could benefit from it. Why not create a `common::retry` module or something like this? You can define a `RetryPolicy` with the `max_attempts`, `max_consecutive_failures` and `backoff_strategy` (fixed duration or exponential).
Then you can use a `retry_with_policy` that has the `RetryPolicy` and the `operation`.

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
```diff
@@ -0,0 +1,1244 @@
...
+}
+
+/// Returns `true` when a `reqwest::Error` looks transient (connection,
+/// timeout, or request-building problem on the network layer).
+fn is_transient_network_error(err: &reqwest::Error) -> bool {
+    err.is_connect() || err.is_timeout() || err.is_request()
+}
+
+/// Maximum number of consecutive transient poll failures before we give up.
+const MAX_CONSECUTIVE_TRANSIENT_POLL_FAILURES: u32 = 5;
```
ruromero: In my experience is better to have configurable values for values like this.

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
```diff
@@ -0,0 +1,1244 @@
...
+        }
+    };
+
+    let is_spdx = sbom_format == "spdx";
+
+    if is_spdx {
+        // SPDX multi-component flow: upload to /api/v1/products/upload-spdx
+        let upload_url = format!("{}/api/v1/products/upload-spdx", config.url);
+
+        const UPLOAD_MAX_ATTEMPTS: u32 = 3;
```
ruromero: I think this should be configurable as well

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
```diff
@@ -0,0 +1,1244 @@
...
+/// Poll the EI service for an in-flight single-component analysis and ingest the result.
+///
+/// This is the second half of the analysis lifecycle: it assumes the SBOM has
+/// already been uploaded and a `scan_id` obtained.  It is called both from
+/// `run_analysis` (normal path) and from the startup recovery path for
+/// orphaned Running jobs.
+///
+/// Updates both the component record (with finding/advisory) and the parent
+/// job record (status only) upon completion.
+pub(crate) async fn poll_for_result(
```
ruromero: Adding a distinction between `RetryableFailure` vs `PermanentFailure` would help end users distinguish between something that can be reattempted or something that requires an action before retry.

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
```diff
@@ -0,0 +1,1244 @@
...
+        }
+    }
+}
+
+/// Poll the EI service for a multi-component product and process per-component results.
+///
+/// This is the SPDX counterpart of `poll_for_result`. It polls the product-level
+/// endpoint until all components are complete, then fetches each component's
+/// report individually.
+pub(crate) async fn poll_for_product_result(
```
ruromero: I see a lot of duplication between this method and `poll_for_result`. I sure this can be refactored

modules/fundamental/src/exploit_intelligence/auth.rs
```diff
@@ -0,0 +1,133 @@
...
+        Self::Static(token)
+    }
+
+    /// Create an OIDC client credentials provider.
+    ///
+    /// Builds an internal HTTP client with sensible timeouts for token
+    /// exchange requests.
+    pub fn oidc(token_url: String, client_id: String, client_secret: String) -> Self {
+        let client = reqwest::Client::builder()
+            .connect_timeout(Duration::from_secs(10))
```
ruromero: I'm in favour of configuring this kind of values with reasonable defaults

modules/fundamental/src/exploit_intelligence/auth.rs
```diff
@@ -0,0 +1,133 @@
...
+            .timeout(Duration::from_secs(30))
```
ruromero: Same for configurable values with default instead of magic numbers

docs/env-vars.md
```diff
@@ -62,6 +57,14 @@
 | `UI_ISSUER_URL`                          | Issuer URL used by the UI                                                           | `http://localhost:8090/realms/trustify` |
 | `UI_LOAD_USER`                           | Whether to load user info                                                           | `true`                                  |
 | `UI_SCOPE`                               | Scopes to request                                                                   | `openid`                                |
+| `EXPLOIT_INTELLIGENCE_URL`               | Base URL of the Exploit Intelligence client service                                 |                                         |
+| `EXPLOIT_INTELLIGENCE_UI_URL`            | Base URL of the EI web UI for deep-linking to reports                               |                                         |
+| `EXPLOIT_INTELLIGENCE_POLL_INTERVAL_SECS`| Polling interval in seconds for EI analysis completion                              | `30`                                    |
```
ctron: Please don't use seconds, but `humantime` format, like with the others.

docs/env-vars.md
```diff
@@ -62,6 +57,14 @@
...
+| `EXPLOIT_INTELLIGENCE_MAX_POLL_DURATION_SECS`| Maximum duration in seconds before EI polling is considered timed out           | `1800`                                  |
```
ctron: Same here.

ctron:
Please consider that this code can run multiple times on the same database, in multiple instances.

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
```diff
@@ -0,0 +1,1244 @@
...
+    let job = ei_service.fetch_job(*id, &tx).await?;
+
+    match job {
+        Some(details) => Ok(HttpResponse::Ok().json(details)),
+        None => Ok(HttpResponse::NotFound().finish()),
+    }
+}
+
+/// Background task: submit SBOM to EI client, poll for results, ingest VEX.
+async fn run_analysis(
```
ruromero: Business logic should not live in the endpoints layer.
The `endpoints` layer should only hold simple HTTP adapters. The CRUD operations should be in the `services` layer.
The background processing like `run_analysis`, `poll_for_result`, `poll_for_result_product` and `recover_and_resume` belongs to the `runner` module.

ruromero (changes requested):
The functionality is there and should work as expected but all the logic ended up in the endpoints module which is an anti-pattern. Look how other functionalities are implemented. I wonder why Claude didn't check for these patterns.

Another concerning point is the fact that there are many hardcoded values that could be configurable for operational reasons. I know there should be a balance between what should be configurable vs hardcoded. It's just something to take into account.

Pay attention at the long functions (+200 lines) doing multiple steps. This goes against the defined conventions.

Finally, the whole retry/polling logic can be moved to a common module so other parts of the code could benefit from it. I haven't checked if there are other candidates for future refactoring.

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
Strum355: Implemented here for your review :bow: https://github.com/guacsec/trustify/pull/2464/changes/3b028cb8d8520da0524d46518e665ef7847d0ad5

common/src/retry.rs
```diff
@@ -0,0 +1,241 @@
+use std::time::Duration;
+
+use anyhow::bail;
+use reqwest::StatusCode;
+use tokio::time::sleep;
+use tracing::{debug, warn};
+
+/// How to compute delay between retry attempts.
+#[derive(Clone, Debug)]
+pub enum BackoffStrategy {
```
jcrossley3: Are we reinventing the wheel here? Can something like https://crates.io/crates/tokio-retry or https://crates.io/crates/backon mitigate some of this complexity?

Strum355: I should definitely have taken a look at some crates first, good callout

helio-frota: @Strum355 we used backon for csaf-walker https://github.com/scm-rs/csaf-walker/pull/48/changes :+1: 

```
➜  trustify git:(main) cargo tree --target all -i backon --depth 1
backon v1.6.0
└── walker-common v0.17.0
```

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
ctron: Instead of implementing that ourselves, whey not re-use existing logic for that. IIRC there was the `backon` crate which we already use.

common/src/retry.rs
Strum355: @helio-frota perfect, thanks! I was just considering backon, so this seals it :ok_hand:

Strum355: Added in https://github.com/guacsec/trustify/pull/2464/changes/caabea116e4802db6b57cffb02fc884d3c830769

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
Strum355: Helio mentioned backon further down below as well, amended the commit to use it instead here: https://github.com/guacsec/trustify/pull/2464/changes/caabea116e4802db6b57cffb02fc884d3c830769

Strum355:
> Please consider that this code can run multiple times on the same database, in multiple instances.

Thanks for pointing this out @ctron. Commit https://github.com/guacsec/trustify/pull/2464/commits/a7f1426ad05a50c418eebfc75cd81c0f576a4125 adds a heartbeat ownership mechanism similar to what exists for the ingestor. Theres a couple differences of course as the processes arent the same (EI jobs are request-driven), feel free to take a look and let me know if you come across any glaring issues with the implementation, or if anything could use inline clarification :slightly_smiling_face:

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
Strum355: That would require a lot of redesigning across the stack including the UI. Ultimately, all failures are retryable to some degree, Im not sure how much value add there is to adding a typed distinction that the error message itself doesnt show

entity/src/exploit_intelligence_job.rs
```diff
@@ -72,6 +73,9 @@ pub struct Model {
     /// Expected number of components in the product.
     pub total_components: Option<i32>,
 
+    /// Unix time in nanos, for coordinating multi-instance job ownership.
```
ctron: What speaks against using a `Timestamp`?

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    operation_id = "analyzeExploitIntelligence",
+    request_body = AnalyzeRequest,
+    responses(
+        (status = 200, description = "Active analysis job already exists", body = AnalyzeResponse),
+        (status = 201, description = "Analysis job created", body = AnalyzeResponse),
+        (status = 404, description = "SBOM not found"),
+        (status = 503, description = "Exploit Intelligence service not configured"),
+    ),
+)]
+#[post("/v3/exploit-intelligence/analyze")]
```
ctron, line 63: Not sure we're using that pattern so far. Maybe check.

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+use trustify_auth::{CreateExploitIntelligence, ReadExploitIntelligence, authorizer::Require};
+use trustify_common::{
+    db,
+    model::{Paginated, PaginatedResults},
+};
+use trustify_module_ingestor::service::IngestorService;
+use uuid::Uuid;
+
+/// Register exploit intelligence endpoints and spawn recovery for orphaned jobs.
+pub fn configure(
```
ctron, line 24: If EI is not available, how do we ensure those endpoints are not present?

server/src/profile/api.rs
```diff
@@ -92,6 +92,70 @@ pub struct Run {
     )]
     pub scan_limit: BinaryByteSize,
 
+    /// Base URL of the Exploit Intelligence client service.
+    #[arg(long, env = "EXPLOIT_INTELLIGENCE_URL")]
+    pub exploit_intelligence_url: Option<String>,
```
ctron, line 147: This should go into a flattened struct.

server/src/profile/api.rs
```diff
@@ -194,6 +258,9 @@ struct InitData {
     config: ModuleConfig,
     analysis: AnalysisService,
     read_only: bool,
+    ei_config: Option<
+        trustify_module_fundamental::exploit_intelligence::service::ExploitIntelligenceConfig,
```
ctron: Import namespace

server/src/profile/api.rs
```diff
@@ -298,6 +365,51 @@ impl InitData {
...
+            let token_provider = match oidc_args {
+                (Some(token_url), Some(client_id), Some(client_secret)) => {
+                    Some(TokenProvider::oidc(token_url, client_id, client_secret)?)
+                }
+                (None, None, None) => run
+                    .exploit_intelligence_auth_token
+                    .map(TokenProvider::static_token),
+                _ => {
+                    log::warn!(
+                        "Incomplete EI OIDC configuration: all three of \
```
ctron: Should be handled by clap

server/src/profile/api.rs
```diff
@@ -298,6 +365,51 @@ impl InitData {
...
+                         EXPLOIT_INTELLIGENCE_OIDC_CLIENT_ID, and \
+                         EXPLOIT_INTELLIGENCE_OIDC_CLIENT_SECRET must be set together. \
+                         Falling back to static token or no auth."
+                    );
+                    run.exploit_intelligence_auth_token
+                        .map(TokenProvider::static_token)
+                }
+            };
+
+            Some(trustify_module_fundamental::exploit_intelligence::service::ExploitIntelligenceConfig {
```
ctron: Same for others: import namespace

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
ctron: Or, what's our strategy here?

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    // The job was created with an initial heartbeat — use it for the heartbeat loop.
+    let initial_heartbeat = job.heartbeat;
+
+    // Spawn background task to submit and poll, wrapped with heartbeat.
+    let ei_svc = ei_service.as_ref().clone();
+    let ingestor_svc = ingestor.as_ref().clone();
+    let db_rw_clone = db_rw.as_ref().clone();
+    let db_hb = db_rw.as_ref().clone();
+
+    actix_web::rt::spawn(async move {
```
ctron: We shouldn't spawn tasks in handler methods. That might overload the instance.

modules/fundamental/src/exploit_intelligence/runner/analysis.rs
```diff
@@ -0,0 +1,257 @@
+use crate::exploit_intelligence::{
+    runner::heartbeat::SharedHeartbeat,
+    runner::{EiRequestError, classify_response_error},
```
ctron, line 3: please nest imports

modules/fundamental/src/exploit_intelligence/runner/analysis.rs
```diff
@@ -0,0 +1,257 @@
...
+///
+/// The `heartbeat` parameter is a shared state updated by the heartbeat loop.
+/// It is passed through to the polling functions for CAS-protected terminal
+/// status updates.
+#[allow(clippy::too_many_arguments)]
+pub(crate) async fn run_analysis(
+    ei_service: &ExploitIntelligenceService,
+    ingestor: &IngestorService,
+    db_rw: &db::ReadWrite,
+    config: &crate::exploit_intelligence::service::ExploitIntelligenceConfig,
```
ctron: please import

modules/fundamental/src/exploit_intelligence/runner/analysis.rs
```diff
@@ -0,0 +1,257 @@
...
+    client: &reqwest::Client,
+    job_id: Uuid,
+    sbom_id: Uuid,
+    vulnerability_id: &str,
+    heartbeat: SharedHeartbeat,
+    token: CancellationToken,
+) -> Result<(), anyhow::Error> {
+    // Retrieve the raw SBOM document from storage (read-only – no mutations needed).
+    let tx = db_rw
+        .begin_with_config(None, Some(AccessMode::ReadOnly))
```
ctron: We are we using the R/W instance for a R/O transaction?

modules/fundamental/src/exploit_intelligence/runner/analysis.rs
```diff
@@ -0,0 +1,257 @@
...
+            .ok_or_else(|| anyhow::anyhow!("source document not found for SBOM: {sbom_id}"))?;
+
+    // Build storage key from source document SHA256.
+    // The DB stores the raw hex digest without algorithm prefix, but Id::from_str expects "sha256:…".
+    let sha256_with_prefix = format!("sha256:{}", source_doc.sha256);
+    let storage_key = trustify_common::id::Id::from_str(&sha256_with_prefix)
+        .map_err(|e| anyhow::anyhow!("invalid sha256: {e}"))?;
+    let storage_key: trustify_module_storage::service::StorageKey = storage_key
+        .try_into()
+        .map_err(|e: trustify_module_storage::service::StorageKeyError| {
```
ctron: Let the compiler evaluate types

modules/fundamental/src/exploit_intelligence/runner/analysis.rs
```diff
@@ -0,0 +1,257 @@
...
+            anyhow::anyhow!("storage key error: {e}")
+        })?;
+
+    let stream = ingestor
+        .storage()
+        .retrieve(storage_key)
+        .await
+        .map_err(|e| anyhow::anyhow!("storage retrieve error: {e}"))?
```
ctron, line 61: Import

modules/fundamental/src/exploit_intelligence/runner/analysis.rs
```diff
@@ -0,0 +1,257 @@
...
+        .ok_or_else(|| anyhow::anyhow!("SBOM document not found in storage"))?;
+
+    // Collect the stream into bytes.
+    let sbom_bytes: Vec<u8> = stream
+        .try_fold(Vec::new(), |mut acc, chunk| async move {
```
ctron: I'm sure there is a function that we can re-use

modules/fundamental/src/exploit_intelligence/runner/analysis.rs
```diff
@@ -0,0 +1,257 @@
...
+            acc.extend_from_slice(&chunk);
+            Ok(acc)
+        })
+        .await
+        .map_err(|e| anyhow::anyhow!("failed to read SBOM stream: {e}"))?;
+
+    drop(tx);
+
+    // Detect SBOM format from labels.
+    let sbom_format = match sbom_model.labels.0.get("type") {
```
ctron: We can't trusty this, users can modify this.

modules/fundamental/src/exploit_intelligence/runner/analysis.rs
```diff
@@ -0,0 +1,257 @@
...
+                sbom_id = %sbom_id,
+                "SBOM has no 'type' label, defaulting to CycloneDX format"
+            );
+            "cyclonedx"
+        }
+    };
+
+    let is_spdx = sbom_format == "spdx";
+
+    if is_spdx {
```
ctron: Please use enums

modules/fundamental/src/exploit_intelligence/runner/analysis.rs
```diff
@@ -0,0 +1,257 @@
...
+        let upload_url = format!("{}/api/v1/products/upload-spdx", config.url);
+
+        let response = (|| {
+            let url = upload_url.clone();
+            let sbom = sbom_bytes.clone();
+            let vuln_id = vulnerability_id.to_string();
+            let cli = client.clone();
+            let tp = config.token_provider.clone();
+            async move {
+                let form = reqwest::multipart::Form::new().text("cveId", vuln_id).part(
```
ctron, line 96: Create a client and cache it.

modules/fundamental/src/exploit_intelligence/runner/heartbeat.rs
```diff
@@ -0,0 +1,190 @@
...
+pub enum HeartbeatError {
+    /// The heartbeat CAS failed — another instance owns this job.
+    /// The caller MUST NOT write to the job row after receiving this.
+    LockLost,
+    /// The work future returned an error. The work future is responsible
+    /// for having already called terminal status updates with CAS protection.
+    Work(anyhow::Error),
+}
+
+impl std::fmt::Display for HeartbeatError {
```
ctron: Please use `thiserror`

ctron:
Ok, I just peeked a bit into the implementation. I think this is a bit overcomplicated. Why not follow a pattern like this: https://aminediro.com/posts/pg_job_queue/

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
Strum355: Is not mounting endpoints the approach we want to go with? Afaik that would be a first for the codebase, but Im fine with that if its an approach you feel strongly about. 503 Unavailable is an existing pattern in a few places in trustify, but the conditions are probably distinct enough that we can deviate here if you wish

entity/src/exploit_intelligence_job.rs
Strum355: `Decimal` is the same type used by the ingestors heartbeat method, thats all. This point doesnt matter though if we change to the pg job queue approach you mentioned below (which I most likely will, it seems considerably cleaner than the hefty approach Claude devised here)

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
Strum355: Will be inherently handled by the pg job queue approach

modules/fundamental/src/exploit_intelligence/endpoints/mod.rs
Strum355: What pattern specifically is this referring to? The endpoint path?

modules/fundamental/src/exploit_intelligence/runner/analysis.rs
Strum355: Couldnt find any personally (nor could Claude), which surprises me. I did make an improvement here though to pre-allocate the vec based on `trustify_entity::source_document::Model` though

Strum355:
> Ok, I just peeked a bit into the implementation. I think this is a bit overcomplicated. Why not follow a pattern like this: https://aminediro.com/posts/pg_job_queue/

Thanks for this, https://github.com/guacsec/trustify/pull/2464/commits/f272d7dc9187dbdeffbf5996dae128b48a3e358b should address this, similar to the importer I believe (but with the visibility timeout approach mentioned in that article instead of the heartbeat approach that importer uses). Some of the style comments were address in the next commit, so you might see some things that should be addressed that have been fixed later :smile: the EI job polling worker runs in the API profile which I hope is okay, outside the initial loading of the SBOM (to check for SBOM type to route the doc to the correct EI endpoint) this process should be fairly lightweight (majority of the time its just polling EI and then ingesting a very small VEX once the work is completed).

common/auth/src/authenticator/default.rs
```diff
@@ -23,7 +23,9 @@ pub const DEFAULT_SCOPE_MAPPINGS: &[(&str, &[&str])] = &[
         "read:document",
         &[
             "ai",
+            "create.exploitIntelligence",
             "read.advisory",
+            "read.exploitIntelligence",
```
Strum355, line 28: I believe this is the default we want: if you can view the SBOM details page, you can view & request EI findings information in the vulns table

server/src/profile/api.rs
```diff
@@ -129,6 +138,166 @@ pub struct Run {
...
+    pub refresh_before: humantime::Duration,
+
+    #[arg(
+        id = "ei_oidc_tls_insecure",
+        long = "ei-oidc-tls-insecure",
+        env = "EXPLOIT_INTELLIGENCE_OIDC_TLS_INSECURE",
+        default_value = "false"
+    )]
+    pub tls_insecure: bool,
+}
```
Strum355, line 272: TBD whether we use separate env vars for EI or re-use `OIDC_PROVIDER_*` env vars (issuer URL etc will be the same afaik)

ctron: Yes, we should have separate env-vars for this. Not sure what else is using OIDC as a client right now, but if there would be something else, it wouldn't mean they would use the same OIDC provider as EI.

It should also be ensured (using clap), that if EI is configured, all values are (required).

Strum355: > Not sure what else is using OIDC as a client right now

Right now, apparently nothing. EI would be the only user.

> it wouldn't mean they would use the same OIDC provider as EI

Thats fair, although Im not sure why the existing OIDC env vars are so generically named then if they are specific to a certain use-case (that was a source of my confusion I think). I'll stick with separate env-vars anyway, thanks!

ruromero (approved):
LGTM

PR #2469, opened 2026-07-09
build: pin openssl/openssl-sys to avoid bindgen/clang issue

ctron:
## Summary
- Pin `openssl` to `=0.10.80` and `openssl-sys` to `=0.9.116` to work around a build issue with bindgen and clang in `openssl-sys` 0.9.117
- Add `openssl-sys` as a direct dependency in `common/infrastructure` to ensure the workspace pin applies to transitive resolution

## Test plan
- [x] `cargo update` no longer bumps `openssl-sys` past 0.9.116

## Summary by Sourcery

Pin OpenSSL crates in the workspace to avoid build issues with bindgen and clang.

Build:
- Pin `openssl` and `openssl-sys` versions in the root Cargo.toml to prevent incompatible updates.
- Add `openssl-sys` as a direct workspace dependency in `common/infrastructure` to ensure the pinned version is used transitively.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1431,9 +1431,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.12.0"
+version = "1.12.1"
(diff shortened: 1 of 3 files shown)


ctron:
We should not have this in `main`.

PR #2521, opened 2026-07-16
fix(purl): add VersionMatches filter to product status query

### LINKED ISSUE #2522: False-positive vulnerability matches due to missing VersionMatches filter in product status query
## Description

The `get_product_statuses_for_purl` function in `modules/fundamental/src/purl/model/details/purl.rs` is missing a `VersionMatches` filter on version ranges. This causes false-positive vulnerability matches when querying PURL details:

1. **RPM version ranges returned for non-RPM packages** — Version ranges with RPM-specific schemes (e.g., `vers:rpm/...`) are included in results for packages of other ecosystems (npm, PyPI, etc.), producing spurious vulnerability matches.
2. **Already-fixed CVEs reported as affected** — Without version filtering, advisory product statuses are returned regardless of whether the queried package version falls within the affected version range, so CVEs that have been fixed in the queried version are still reported.

## Root Cause

The sibling query (`purl_statuses`) in the same function already applies a `VersionMatches` filter, but `get_product_statuses_for_purl` does not. The fix is to add the same filter to the product status query.

## Fix

PR #2521

ruromero:
## Summary

- Adds the missing `VersionMatches` filter to the `get_product_statuses_for_purl` function, which was returning false-positive vulnerability matches
- Without this filter, RPM-specific version ranges were being returned for non-RPM packages, and already-fixed CVEs were reported as still affected
- The fix mirrors the same `VersionMatches` pattern already used by the `purl_statuses` query in the same file

Fixes #2522

## Jira

[TC-5170](https://issues.redhat.com/browse/TC-5170)

## Test plan

- [x] `cargo fmt` — clean
- [x] `cargo clippy -p trustify-module-fundamental` — zero warnings
- [x] All purl unit tests pass (individually; bulk run has pre-existing DB contention flakes unrelated to this change)
- [ ] Verify with a non-RPM SBOM that product statuses no longer include RPM-only version ranges
- [ ] Verify that fixed CVEs are no longer reported as affected

[TC-5170]: https://redhat.atlassian.net/browse/TC-5170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/fundamental/src/purl/model/details/purl.rs
+++ b/modules/fundamental/src/purl/model/details/purl.rs
@@ -106,6 +106,7 @@ impl PurlDetails {
             qualified_package.id,
             &package.name,
             package.namespace.as_deref(),
+            &package_version.version,
         )
         .await?;
 
@@ -157,6 +158,7 @@ async fn get_product_statuses_for_purl<C: ConnectionTrait>(
     qualified_package_id: Uuid,
     purl_name: &str,
     namespace_name: Option<&str>,
+    version: &str,
 ) -> Result<Vec<ProductStatusCatcher>, Error> {
     // Subquery to get all SBOM IDs for the given purl
     let sbom_ids_query = sbom::Entity::find()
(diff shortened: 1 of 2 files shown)


rh-jfuller (approved):
LGTM but can we consider adding a unit test that confirms the fix ?

ruromero:
@rh-jfuller test added, is it good to go?

PR #2523, opened 2026-07-16
fix(purl): filter purl_statuses by SBOM describing CPEs (TC-5171)

### LINKED ISSUE #2524: purl_statuses returns entries with RPM version ranges and RHEL CPE contexts for non-RPM packages
## Issue Description

When Trustify returns vulnerability data for a non-RPM package (e.g., a Maven or OCI PURL), some `purl_statuses` entries use an `rpm` version scheme with RHEL CPE contexts (e.g., `cpe:/o:redhat:enterprise_linux:6`) whose version ranges (e.g., `[6.0.0, 7.0.0)`) refer to RHEL product versions — not to the queried package version. These entries are meaningless when applied to non-RPM artifacts.

This stems from CSAF advisories that describe packages as shipped in specific RHEL products being matched against unrelated PURLs without checking whether the CPE context applies.

## Steps to Reproduce

1. Ingest CSAF advisories that cover multiple Red Hat products (e.g., advisories that bundle both Go and kernel updates for OpenShift)
2. Query Trustify for a package PURL that exists in a different product context
3. Inspect the `purl_statuses` — entries with unrelated CPE contexts appear

## Expected Result

`purl_statuses` entries whose `context_cpe_id` does not match the SBOM's describing CPEs should not be returned. Only statuses with:
- No CPE context (`context_cpe_id IS NULL`)
- A CPE context matching the SBOM's describing CPEs
- Or a fallback when the SBOM has no describing CPEs at all

## Actual Result

All matching `purl_status` rows are returned regardless of their `context_cpe_id`, causing cross-product false positives.

## Impact

In large SBOM analyses (e.g., full RHEL images with ~49,000 components), cross-product advisory matching contributes to thousands of false-positive vulnerability rows.

## Root Cause

The `PurlDetails::from_entity` query in `purl.rs` filters by `base_purl_id` and `version_matches()` but never validates `context_cpe_id` against the SBOM's describing CPEs. The raw SQL batch severity query (`raw_sql.rs`) already had this three-way filter, but the SeaORM query path did not.

Tracker: [TC-5171](https://issues.redhat.com/browse/TC-5171)

[TC-5171]: https://redhat.atlassian.net/browse/TC-5171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ruromero:
## Summary

Filter `purl_statuses` and `/vulnerability/analyze` results to exclude entries whose `context_cpe_id` does not match the SBOM's describing CPEs. Uses generalized CPE matching so that all CPEs sharing the same vendor, product, and major version are included in the allowed set — this correctly handles cases where the SBOM describes one edition (e.g. `appstream`) but advisories reference another edition of the same product (e.g. `baseos`).

## Problem

When querying vulnerability data for a package, `purl_statuses` and `/vulnerability/analyze` returned entries with CPE contexts from unrelated products (e.g., JBoss, RHSCL) because `context_cpe_id` was never validated against the SBOM's describing CPEs. This caused thousands of false-positive vulnerability rows in large SBOM analyses.

### 1. Three-way `context_cpe_id` filter

Added to all three query paths (SeaORM purl endpoint, raw SQL analyze endpoint, raw SQL SBOM batch counts):

1. **`context_cpe_id IS NULL`** — always allow statuses without a CPE context
2. **`context_cpe_id IN (allowed CPEs)`** — allow statuses whose CPE matches the SBOM's describing CPEs (with generalized matching)
3. **SBOM has no describing CPEs** — fallback: allow everything

### 2. Generalized CPE matching

The allowed CPE set is expanded beyond exact describing CPEs to include all CPEs sharing the same `(vendor, product, major_version)`. This handles the common case where an SBOM describes `enterprise_linux:8::appstream` but advisory entries reference `enterprise_linux:8::baseos` — both are the same product (RHEL 8) and should be included.

## Changes

- **`modules/fundamental/src/purl/model/details/purl.rs`**: Added three subqueries (`sbom_ids_for_purl`, `generalized_cpe_ids`, `sbom_has_cpes`) and a `Condition::any()` filter to the SeaORM `purl_statuses` query
- **`modules/fundamental/src/vulnerability/service/mod.rs`**: Added generalized CPE matching CTEs and three-way filter to both `build_query()` variants (by `base_purl.id` and by `{product_bp_condition}`)
- **`modules/fundamental/src/sbom/model/raw_sql.rs`**: Updated `CONTEXT_CPE_FILTER_SQL`, `batch_severity_counts_sql()`, and `product_advisory_info_sql()` to use generalized matching without edition constraints

## Testing

- Verified with `expat@2.2.5-11.el8` (22 version-matching purl_status entries across 7+ products): 2 baseos entries correctly pass, 20 cross-product entries correctly filtered
- `cargo clippy` and `cargo fmt` pass
- Existing test suite passes

Fixes #2524
Fixes: [TC-5171](https://issues.redhat.com/browse/TC-5171)

[TC-5171]: https://redhat.atlassian.net/browse/TC-5171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/fundamental/tests/vuln/mod.rs
+++ b/modules/fundamental/tests/vuln/mod.rs
@@ -61,7 +61,7 @@ async fn issue_1840(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
         .filter(|status| status.purl_status.status == "affected")
         .collect();
 
-    assert_eq!(status_entries.len(), 2);
+    assert_eq!(status_entries.len(), 4);
     let json = serde_json::to_value(status_entries).expect("must serialize");
     assert!(
         json.contains_subset(json!([{
@@ -80,7 +80,6 @@ async fn issue_1840(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
             },
             "scores": [{"type": "3.1", "value": 5.3, "severity": "medium", "vector": "CVSS:3.1/AV:N/AC:H/PR:L/UI:N/S:U/C:H/I:N/A:N"}],
             "status": "affected",
-            "context": null,
             "version_range": {
                 "version_scheme_id": "rpm",
                 "high_version": "3.7.6-23.el9_3.4",
@@ -102,7 +101,6 @@ async fn issue_1840(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
             },
             "scores": [{"type": "3.1", "value": 5.3, "severity": "medium", "vector": "CVSS:3.1/AV:N/AC:H/PR:L/UI:N/S:U/C:H/I:N/A:N"}],
             "status": "affected",
-            "context": null,
             "version_range": {
                 "version_scheme_id": "rpm",
                 "high_version": "3.8.3-4.el9_4",
--- a/modules/fundamental/tests/vuln/mod.rs
+++ b/modules/fundamental/tests/vuln/mod.rs
@@ -114,3 +114,63 @@ async fn issue_1840(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
 
     Ok(())
 }
+
+/// Proves that `version_matches` filtering works on the purl_status path:
+/// a version ABOVE the advisory's affected range must not be reported as affected.
+#[test_context(TrustifyContext)]
+#[test(tokio::test)]
+async fn version_filtering(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
+    ctx.ingest_dataset(Dataset::DS3).await?;
+
+    let service = VulnerabilityService::new(PaginationCache::for_test());
+
+    // Version BELOW the fix threshold — known affected by CVE-2024-28834
+    // (3.7.6-23.el9 < 3.7.6-23.el9_3.4, the fix version)
+    let affected_result = service
+        .analyze_purls_v3(["pkg:rpm/redhat/gnutls@3.7.6-23.el9?arch=aarch64"], &ctx.db)
+        .await?;
+
+    let affected_entry = &affected_result["pkg:rpm/redhat/gnutls@3.7.6-23.el9?arch=aarch64"];
+    let affected_count = affected_entry
+        .details
+        .iter()
+        .flat_map(|d| &d.purl_statuses)
+        .filter(|s| {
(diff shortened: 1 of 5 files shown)


modules/fundamental/tests/vuln/mod.rs
```diff
@@ -61,7 +61,7 @@ async fn issue_1840(ctx: &TrustifyContext) -> Result<(), anyhow::Error> {
         .filter(|status| status.purl_status.status == "affected")
         .collect();
 
-    assert_eq!(status_entries.len(), 2);
+    assert_eq!(status_entries.len(), 4);
```
rh-jfuller, line 64: not sure this change confirms the fix eg. increasing number of matched entries seems counter intuitive) ... perhaps we need to add another test ?

rh-jfuller:
can we force push to re-align commits on this PR ?

Also the generalized match is not quite as described eg. desc/comment say allowed CPEs are those sharing "the same vendor, product, and major version." 

What this code does "must exactly equal the major of a describing CPE". That is kinda asymmetric ...  a candidate enterprise_`linux:8.8::baseos` against a describing enterprise_`linux:8.6::appstream` shares major but has version "8.8" ne "8", so it is excluded, likewise describing "8" never admits candidate "8.8".

rh mainline advisory CPEs are **major only** which is why (I think) the expat verification passed but EUS/Z-stream advisories carry minor-versioned CPEs which I think become silent fp. I am guessing we do not have EUS/Z tests for this endpoint ?

I think the fix is maybe to compare split_part(...) on both sides ?

rh-jfuller:
with these changes we still need to observe at scale impact

rh-jfuller (approved):
I did a manual perf-test ... things look ok (as in no significant change from main branch). I think this whole area needs looking at (which https://github.com/guacsec/trustify/pull/2528 more comprehensively achieves) ... LGTM

PR #2525, opened 2026-07-20
fix(ingestor): compute CVSS scores from vector string instead of JSON fields

ctron:
## Summary

* Always parse CVSS from `vectorString` via `FromStr` + `calculated_base_score()`, removing the dependency on `baseScore`/`baseSeverity` JSON fields that many CVE producers omit (root cause of TC-5272)
* Consolidate the two divergent CVSS parsing paths (`extract_scores` and `extract_base_score`) into shared helpers in `mod.rs`, deleting the legacy `CvssMetric` struct, `get_score()`, and `score_from_metric()`
* Add schema migration `m0002250` to backfill `vulnerability.base_score/base_severity/base_type` from `advisory_vulnerability_score` for existing data
* Close testing gap: `CveBuilder` now builds metrics with only `vectorString`, plus explicit test cases for missing and invalid vector strings

### Data refresh on upgrade

The backfill migration (`m0002250`) runs automatically and updates the denormalized vulnerability columns from `advisory_vulnerability_score`.

If the `m0002010_add_advisory_scores` data migration was skipped or needs re-running with the fixed parser, run it first:

```sh
trustd db data m0002010_add_advisory_scores
```

Then re-run the backfill SQL manually if needed (idempotent):

```sh
psql -d <database> -f migration/src/m0002250_backfill_vulnerability_base_score/up.sql
```

Implements TC-5272

## Test plan

- [x] All existing CVE loader unit tests pass with vector-string-only test fixtures
- [x] New test: `missing_vector_string_yields_none` — metric with `baseScore` but no `vectorString` returns `None`
- [x] New test: `invalid_vector_yields_none` — garbage `vectorString` returns `None`
- [x] CSAF integration tests unaffected (separate code path)
- [x] `cargo xtask precommit` passes cleanly

## Summary by Sourcery

Parse CVSS scores from vector strings across the CVE ingestor and use them to compute and backfill vulnerability base scores.

Bug Fixes:
- Ensure CVSS base scores are derived from vectorString data instead of relying on often-missing baseScore/baseSeverity JSON fields, preventing missing scores for certain CVEs.

Enhancements:
- Unify CVSS parsing and base-score selection logic into shared helpers that work across CNA and ADP metrics and prefer higher CVSS versions.
- Relax base-score comparison in tests and switch CVE test fixtures to vector-string-only metrics to align with the new parsing approach.

Tests:
- Add unit and migration tests covering vector-string-only CVSS metrics, missing/invalid vector strings, and the base-score backfill behavior.

### DIFF
--- a/docs/design/cvss_scoring.md
+++ b/docs/design/cvss_scoring.md
@@ -1,0 +1,213 @@
+# CVSS Score Model
+
+This document describes how CVSS scores are stored, selected, and surfaced
+across the system. A single vulnerability can receive scores from multiple
+advisories (CVE, CSAF, OSV), each carrying one or more CVSS versions
+(v2.0, v3.0, v3.1, v4.0). The rules below govern which score appears in
+each context.
+
+## Storage
+
+Scores are stored at two levels:
+
+### Per-advisory scores (`advisory_vulnerability_score`)
+
+Every advisory that reports a vulnerability may contribute CVSS scores.
+These are stored in the `advisory_vulnerability_score` table, keyed by
+`(advisory_id, vulnerability_id)`:
+
+| Column           | Type       | Description                         |
+|------------------|------------|-------------------------------------|
+| `id`             | UUID       | Primary key (v7)                    |
+| `advisory_id`    | UUID       | FK to `advisory`                    |
+| `vulnerability_id` | String  | FK to `vulnerability`               |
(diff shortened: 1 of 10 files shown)


rh-jfuller (approved):
in context of just this PR **LGTM** ... I do have some broader concerns about computing CVSS but those can be followed up separately.

ctron:
Claude gone in circles and contradicted itself a few times now, trying to prove this PR works. By now I have zero confidence this is the right PR. I'll scrap it and start fresh.

PR #2526, opened 2026-07-20
fix: return HTTP 404 instead of 400 for non-existent SBOM IDs

rh-jfuller:
Return HTTP 404 for SBOM ID key lookup failures and align tests with the new behavior.

Bug Fixes:
- Change SBOM ID key errors to respond with HTTP 404 Not Found instead of HTTP 400 Bad Request.

Tests:
- Update SBOM endpoint tests to expect HTTP 404 for invalid or unsupported SBOM identifiers and non-existent SBOM IDs.

addresses TC-5278


jcrossley3 (approved):
Makes sense to me. Invalid or not, it's not there. 😄

PR #2534, opened 2026-07-22
feat: exploit intelligence integration

Strum355:
## Summary

Feature branch merge for Exploit Intelligence integration.

### Tasks implemented:
- [TC-4675](https://redhat.atlassian.net/browse/TC-4675): #2435
- [TC-4677](https://redhat.atlassian.net/browse/TC-4677): #2464

Implements [TC-4441](https://redhat.atlassian.net/browse/TC-4441)

Related trustify-ui PR: https://github.com/guacsec/trustify-ui/pull/1125

[TC-4675]: https://redhat.atlassian.net/browse/TC-4675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TC-4677]: https://redhat.atlassian.net/browse/TC-4677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TC-4441]: https://redhat.atlassian.net/browse/TC-4441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


migration/src/m0002290_create_exploit_intelligence_job.rs
```diff
@@ -0,0 +1,377 @@
...
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        create_enum_if_not_exists(
+            manager,
+            ExploitIntelligenceJobStatus::Table,
+            ExploitIntelligenceJobStatus::VARIANTS
+                .iter()
+                .skip(1)
```
ctron, line 19: Why are we skipping the first?

migration/src/m0002290_create_exploit_intelligence_job.rs
```diff
@@ -0,0 +1,377 @@
...
+                .copied(),
+        )
+        .await?;
+
+        create_enum_if_not_exists(
+            manager,
+            EiFinding::Table,
+            EiFinding::VARIANTS.iter().skip(1).copied(),
```
ctron, line 27: Why are we skipping the first?

migration/src/m0002290_create_exploit_intelligence_job.rs
ctron: I think I understand now. It's actually a local type, which we need to keep it stable.

migration/src/m0002290_create_exploit_intelligence_job.rs
Strum355: Because the first variant in the enums is the `Table` marker. You can see another example in the advisory scores migration of yours here: https://github.com/guacsec/trustify/blob/main/migration/src/m0002010_add_advisory_scores.rs#L30

ctron: See above.

modules/fundamental/src/lib.rs
```diff
@@ -4,6 +4,7 @@ pub mod advisory;
 pub mod common;
 pub mod endpoints;
 pub mod error;
+pub mod exploit_intelligence;
```
ctron: I wonder if this is the right location. I think this should go into it's own module.

server/src/profile/api.rs
```diff
@@ -129,6 +138,166 @@ pub struct Run {
...
+#[derive(clap::Args, Debug)]
+#[command(next_help_heading = "Exploit Intelligence")]
+pub struct ExploitIntelligenceArgs {
+    /// Base URL of the Exploit Intelligence client service.
+    #[arg(long, env = "EXPLOIT_INTELLIGENCE_URL")]
+    pub exploit_intelligence_url: Option<String>,
+
+    /// Base URL of the Exploit Intelligence web UI for deep-linking to reports.
+    #[arg(long, env = "EXPLOIT_INTELLIGENCE_UI_URL")]
+    pub exploit_intelligence_ui_url: Option<String>,
```
ctron, line 157: I assume if `exploit_intelligence_url` is set, so must be this one (and others). We need to make sure this information is available. Clap supports this.

server/src/profile/api.rs
```diff
@@ -129,6 +138,166 @@ pub struct Run {
...
+        env = "EXPLOIT_INTELLIGENCE_OIDC_CLIENT_SECRET"
+    )]
+    pub client_secret: Option<String>,
+
+    #[arg(
+        id = "ei_oidc_issuer_url",
+        long = "ei-oidc-issuer-url",
+        env = "EXPLOIT_INTELLIGENCE_OIDC_ISSUER_URL"
+    )]
+    pub issuer_url: Option<String>,
```
ctron, line 246: Same here, if client_id is present, so must be issuer_url

server/src/profile/api.rs
```diff
@@ -315,13 +487,48 @@ impl InitData {
...
             read_only: run.read_only,
+            ei_config,
         })
     }
 
     #[allow(unused_mut)]
     async fn run(mut self) -> anyhow::Result<()> {
         let ui = Arc::new(UiResources::new(&self.ui)?);
 
+        // Build the EI worker task before the HTTP server captures `self`.
```
ctron: This should at least be pulled out into it's own function.

server/src/profile/api.rs
```diff
@@ -351,6 +559,10 @@ impl InitData {
         #[allow(unused_mut)]
         let mut tasks = vec![http];
 
+        if let Some(worker) = ei_worker_task {
+            tasks.push(worker);
```
ctron: maybe use .extend()

ctron (changes requested):
I think that, unless impossible, this should go into it's own module. It doesn't seem to be "fundemental".

Also, how can the UI evaluate if this feature is configured or not?

server/src/profile/api.rs
Strum355: > I assume if exploit_intelligence_url is set, so must be this one (and others).

Not necessarily true, although it may be the most common case in practice. The EI client both serves the UI and provides the API (or an API gateway, whichever) so its entirely possible that the same URL is used.

modules/fundamental/Cargo.toml
```diff
@@ -21,6 +21,8 @@ trustify-query-derive = { workspace = true }
 
 actix-http = { workspace = true }
 actix-web = { workspace = true }
+backon = { workspace = true }
```
ruromero: **Convention: dependency ordering** — `backon` and `bytes` are inserted before `anyhow`, breaking the alphabetical ordering convention used throughout the workspace. They should be placed after `async-trait` and `bytesize` respectively.

modules/fundamental/src/endpoints.rs
```diff
@@ -6,13 +6,16 @@ use trustify_module_ingestor::service::IngestorService;
...
+use crate::exploit_intelligence::service::{ExploitIntelligenceConfig, ExploitIntelligenceService};
+
 #[derive(Clone, Debug, Eq, PartialEq, Default)]
 pub struct Config {
     pub sbom_upload_limit: usize,
     pub advisory_upload_limit: usize,
     pub max_group_name_length: usize,
 }
 
+#[allow(clippy::too_many_arguments)]
```
ruromero, line 16: **Convention: `#[allow(clippy::too_many_arguments)]`** — This lint is also suppressed on `poll_for_result` and `poll_for_product_result` in `polling.rs`. The codebase pattern is to introduce a context/config struct instead of suppressing the lint. Consider a `ConfigureContext` here and a `PollContext` for the polling functions.

modules/fundamental/src/endpoints.rs
```diff
@@ -45,7 +49,17 @@ pub fn configure(
...
-    crate::sbom_group::endpoints::configure(svc, db_rw, db_ro, config.max_group_name_length, cache);
+    crate::sbom_group::endpoints::configure(
+        svc,
+        db_rw.clone(),
+        db_ro.clone(),
+        config.max_group_name_length,
+        cache,
+    );
+
+    let ei_service = ExploitIntelligenceService::new(ei_config)
```
ruromero: **Convention: panic in server init path** — Other modules in Trustify propagate errors with `?` through `configure()` or use fallible initialization outside the HTTP configure closure. `unwrap_or_else(|e| panic!(...))` diverges from that pattern and makes this failure mode harder to handle gracefully.

modules/exploit-intelligence/src/runner/polling.rs
```diff
@@ -0,0 +1,630 @@
...
+) -> Result<(), anyhow::Error> {
+    let poll_url = format!(
+        "{}/api/v1/reports/by-scan-id/{}",
+        config.url,
+        urlencoding::encode(scan_id)
+    );
+    let poll_interval = config.poll_interval.max(Duration::from_secs(5));
+    let max_consecutive = config.max_consecutive_poll_failures;
+    let consecutive_failures = Cell::new(0u32);
+
```
ruromero: **Code quality: `Cell<u32>` for failure tracking** — Using `Cell` to share mutable state with the retry closure works in single-threaded tokio but is fragile and non-obvious. A straightforward `loop` with a `&mut u32` counter would be simpler and clearer. Same pattern appears at line ~260 for `poll_for_product_result`.

modules/exploit-intelligence/src/runner/polling.rs
```diff
@@ -0,0 +1,630 @@
...
+                .report
+                .as_ref()
+                .and_then(|r| r.output.as_ref())
+                .and_then(|o| o.analysis.as_ref())
+                .and_then(|a| a.first())
+                .and_then(|e| e.justification.as_ref())
+                .and_then(|j| j.status.as_deref())
+                .map(|s| match s.to_ascii_lowercase().as_str() {
+                    "false" => Finding::NotVulnerable,
+                    "true" => Finding::Vulnerable,
```
ruromero: **Code quality: magic strings for finding classification** — `"false"` → NotVulnerable, `"true"` → Vulnerable, `_` → Uncertain. If the EI service changes these strings, the code silently degrades to `Uncertain` with no warning. Consider defining constants or an explicit mapping with a `tracing::warn!` on unexpected values. Same pattern at line ~488.

modules/fundamental/src/exploit_intelligence/runner/polling.rs
```diff
@@ -0,0 +1,630 @@
...
+                .report
+                .as_ref()
+                .and_then(|r| r.output.as_ref())
+                .and_then(|o| o.vex.as_ref());
+
+            let advisory_id = if let Some(vex) = vex_json {
+                if !vex.is_null() {
+                    // Ingest the VEX document as an advisory with source label.
+                    let vex_bytes = serde_json::to_vec(vex)?;
+                    let mut labels = Labels::default();
```
ruromero: **Code quality: duplicated VEX ingestion** — The VEX → advisory ingestion block (fetch VEX JSON, serialize, create labels, begin transaction, call `ingestor.ingest()`, commit) is duplicated nearly verbatim between `poll_for_result` (~here) and `poll_for_product_result` (~line 520). Extract a shared `ingest_vex_advisory()` helper to reduce duplication and ensure consistent behavior.

modules/exploit-intelligence/src/service/mod.rs
```diff
@@ -0,0 +1,946 @@
...
+                        s.total_components = Some(c.total);
+                    }
+                    s
+                } else {
+                    ExploitIntelligenceJobSummary::from_entity(model, self.ui_url())
+                };
+
+                // Set report_url from component scan_id for single-component jobs.
+                if model.product_id.is_none()
+                    && let Some(scan_id) = scan_id_map.get(&model.id)
```
ruromero, line 410: **Convention: `let` chains (nightly feature)** — `if model.product_id.is_none() && let Some(scan_id) = ... && let Some(base) = ...` requires `#![feature(let_chains)]` (stabilized only in Rust 1.87.0, June 2025). If the project's MSRV is below 1.87, this won't compile on stable. Can be rewritten as nested `if let` blocks. Verify the project's MSRV/toolchain policy.

server/src/profile/api.rs
```diff
@@ -315,13 +487,48 @@ impl InitData {
...
+                let db_ro = self.db_ro.clone();
+
+                Some(
+                    async move {
+                        run_worker(
+                            ei_service,
+                            ingestor_for_worker,
+                            db_rw,
+                            db_ro,
+                            Duration::from_secs(5), // worker poll interval
```
ruromero: **Convention: hardcoded configuration** — `Duration::from_secs(5)` for the worker poll interval is hardcoded, unlike every other EI timing parameter which is configurable via env vars in `ExploitIntelligenceArgs`. Add a `EXPLOIT_INTELLIGENCE_WORKER_POLL_INTERVAL` env var for consistency.

server/src/profile/api.rs
```diff
@@ -315,13 +487,48 @@ impl InitData {
...
+        // Build the EI worker task before the HTTP server captures `self`.
+        // The worker needs an IngestorService, which is also created inside
+        // the module configuration — build one here and clone it for the worker.
+        let ei_worker_task = if let Some(ref ei_config) = self.ei_config {
+            if !self.read_only {
+                let ei_service = ExploitIntelligenceService::new(Some(ei_config.clone()))?;
+                let ingestor_for_worker = trustify_module_ingestor::service::IngestorService::new(
+                    Graph::new(),
+                    self.storage.clone(),
+                    Some(self.analysis.clone()),
```
ruromero: **Code quality: duplicate `Graph::new()`** — A new `Graph::new()` is created here for the worker's `IngestorService`, while another one is created inside `configure()` → `endpoints.rs`. If `Graph` holds any shared state (caches, counters), these two instances will diverge. Consider sharing a single `Graph` instance, or document that `Graph::new()` is intentionally stateless.

modules/exploit-intelligence/src/model/mod.rs
```diff
@@ -0,0 +1,435 @@
...
+    ) -> Self {
+        let mut summary = Self::from_entity(model, ui_url);
+        summary.completed_components = Some(counts.completed);
+        summary.failed_components = Some(counts.failed);
+        summary.excluded_components = Some(counts.excluded);
+        summary.vulnerable_components = Some(counts.vulnerable);
+        summary.not_vulnerable_components = Some(counts.not_vulnerable);
+        summary.uncertain_components = Some(counts.uncertain);
+
+        // Compute aggregate finding from component counts.
```
ruromero, line 288: **Code quality: fragile aggregate finding** — The aggregate finding logic here uses explicit if/else priority (Vulnerable > Failed > Uncertain > NotVulnerable). In `ExploitIntelligenceJobDetails::from_components()` (~line 380), a similar computation uses `.max()` on the enum ordinal. The enum variant ordering in `entity/exploit_intelligence_job.rs` implicitly defines this priority — if someone reorders the variants, the `.max()` path silently changes the business logic while this explicit path stays correct. Consider adding a `#[repr(u8)]` with explicit discriminants, or always using the explicit if/else approach.

ruromero (changes requested):
## Review: Exploit Intelligence Integration

Solid feature with well-designed job queue, comprehensive tests, and good resilience patterns. The convention and code quality items below should be addressed before merge.

See inline comments for details.

Strum355:
> Also, how can the UI evaluate if this feature is configured or not?

Thanks for pointing this out, seems my changes disappeared into a git stash. Im taking the approach of adding a flag to `/.well-known/trustify`, any concerns with that? Seems reasonably straightforward and clean
```json
{"version":"0.5.0-rc.1","readOnly":false,"exploitIntelligence":false,"build":{...}}
```

> I think that, unless impossible, this should go into it's own module. It doesn't seem to be "fundemental".

Ill have a look into this :+1:

modules/fundamental/src/endpoints.rs
Strum355: `#[allow(clippy::too_many_arguments)]` is used in a number of other places in the codebase already, so this isnt necessarily a deviation from existing conventions in the codebase.

modules/exploit-intelligence/src/service/mod.rs
Strum355: Our rust version is 1.92.0, this is a non-issue

modules/exploit-intelligence/src/runner/polling.rs
Strum355: Afaik `Cell<u32>` is needed because `backon`s `.when()` takes an `Fn` not an `FnMut`, so `Cell` is the intended way to mutate state inside it

modules/exploit-intelligence/src/runner/worker.rs
```diff
@@ -0,0 +1,114 @@
+//! Background worker loop for processing Exploit Intelligence jobs.
```
rh-jfuller: consider how this loop drives the whole thing ... maybe consider adding some tests ?

ctron:
The command line arguments of EI mix with others and are inconsistently named:

```
| Exploit Intelligence:
|       --exploit-intelligence-url <EXPLOIT_INTELLIGENCE_URL>
|           Base URL of the Exploit Intelligence client service
|           
|           [env: EXPLOIT_INTELLIGENCE_URL=]
| 
|       --exploit-intelligence-ui-url <EXPLOIT_INTELLIGENCE_UI_URL>
|           Base URL of the Exploit Intelligence web UI for deep-linking to reports
|           
|           [env: EXPLOIT_INTELLIGENCE_UI_URL=]
| 
|       --exploit-intelligence-poll-interval <EXPLOIT_INTELLIGENCE_POLL_INTERVAL>
|           Polling interval for EI analysis completion (humantime, e.g. "30s")
|           
|           [env: EXPLOIT_INTELLIGENCE_POLL_INTERVAL=]
|           [default: 30s]
| 
|       --exploit-intelligence-max-poll-duration <EXPLOIT_INTELLIGENCE_MAX_POLL_DURATION>
|           Maximum duration before EI polling is considered timed out (humantime, e.g. "30m")
|           
|           [env: EXPLOIT_INTELLIGENCE_MAX_POLL_DURATION=]
|           [default: 30m]
| 
|       --exploit-intelligence-upload-max-retries <EXPLOIT_INTELLIGENCE_UPLOAD_MAX_RETRIES>
|           Maximum number of upload retry attempts for EI analysis
|           
|           [env: EXPLOIT_INTELLIGENCE_UPLOAD_MAX_RETRIES=]
|           [default: 3]
| 
|       --exploit-intelligence-upload-retry-delay <EXPLOIT_INTELLIGENCE_UPLOAD_RETRY_DELAY>
|           Initial delay between upload retries (humantime, exponential backoff)
|           
|           [env: EXPLOIT_INTELLIGENCE_UPLOAD_RETRY_DELAY=]
|           [default: 1s]
| 
|       --exploit-intelligence-max-consecutive-poll-failures <EXPLOIT_INTELLIGENCE_MAX_CONSECUTIVE_POLL_FAILURES>
|           Maximum consecutive poll failures before giving up
|           
|           [env: EXPLOIT_INTELLIGENCE_MAX_CONSECUTIVE_POLL_FAILURES=]
|           [default: 5]
| 
|       --worker-poll-interval <WORKER_POLL_INTERVAL>
|           How often the background worker checks for new EI jobs
|           
|           [env: EXPLOIT_INTELLIGENCE_WORKER_POLL_INTERVAL=]
|           [default: 5s]
| 
|       --exploit-intelligence-auth-token <EXPLOIT_INTELLIGENCE_AUTH_TOKEN>
|           Authentication token for the Exploit Intelligence service (static token, for backward compatibility)
|           
|           [env: EXPLOIT_INTELLIGENCE_AUTH_TOKEN=]
| 
| Exploit Intelligence OIDC:
|       --ei-oidc-client-id <ei_oidc_client_id>
|           [env: EXPLOIT_INTELLIGENCE_OIDC_CLIENT_ID=]
| 
|       --ei-oidc-client-secret <ei_oidc_client_secret>
|           [env: EXPLOIT_INTELLIGENCE_OIDC_CLIENT_SECRET=]
| 
|       --ei-oidc-issuer-url <ei_oidc_issuer_url>
|           [env: EXPLOIT_INTELLIGENCE_OIDC_ISSUER_URL=]
| 
|       --ei-oidc-refresh-before <ei_oidc_refresh_before>
|           [env: EXPLOIT_INTELLIGENCE_OIDC_REFRESH_BEFORE=]
|           [default: 30s]
| 
|       --ei-oidc-tls-insecure
|           [env: EXPLOIT_INTELLIGENCE_OIDC_TLS_INSECURE=]
| 
|       --max-cache-size <max-cache-size>
|           Maximum size of the graph cache.
|           
|           [env: TRUSTD_MAX_CACHE_SIZE=8GiB]
|           [default: "200 MiB"]
| 
|       --concurrency <CONCURRENCY>
|           The number of concurrent tasks for analysis (must be > 0).
|           
|           [env: TRUSTIFY_ANALYSIS_CONCURRENCY=]
|           [default: 10]
| 
|       --sbom-published-cutoff-days <SBOM_PUBLISHED_CUTOFF_DAYS>
|           Only consider SBOMs published within this many days. Disabled by default.
|           
|           [env: TRUSTIFY_LATEST_SBOM_PUBLISHED_CUTOFF_DAYS=]
```

ctron:
When setting the EI URL, but not the EI UI URL, what's expected to happen?

ctron:
Are `--exploit-intelligence-auth-token <EXPLOIT_INTELLIGENCE_AUTH_TOKEN>` and the OIDC configuration mutually exclusive? Then we need to enforce this using clap.

modules/exploit-intelligence/src/runner/analysis.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    // Retrieve the raw SBOM document from storage (read-only -- no mutations needed).
+    let tx = db_ro.begin().await?;
+    let sbom_model = trustify_entity::sbom::Entity::find()
+        .filter(trustify_entity::sbom::Column::SbomId.eq(sbom_id))
+        .one(&tx)
+        .await?
+        .ok_or_else(|| anyhow::anyhow!("SBOM not found: {sbom_id}"))?;
+
+    let source_doc =
```
ctron: Why don't we use a join here?

modules/exploit-intelligence/src/runner/analysis.rs
```diff
@@ -0,0 +1,227 @@
...
+    let storage_key = StorageKey::from_sha256(&source_doc.sha256);
+
+    let stream = ingestor
+        .storage()
+        .retrieve(storage_key)
+        .await
+        .map_err(|e| anyhow::anyhow!("storage retrieve error: {e}"))?
+        .ok_or_else(|| anyhow::anyhow!("SBOM document not found in storage"))?;
+
+    let sbom_bytes: Vec<u8> = stream
```
ctron: IIRC, there was a function doing this already, instead of re-implementing this.

modules/exploit-intelligence/src/runner/worker.rs
```diff
@@ -0,0 +1,367 @@
...
+        };
+
+        if let Err(e) = processor.sweep_exhausted().await {
+            tracing::error!(error = %e, "failed to mark exhausted EI jobs as failed");
+        }
+
+        for job in jobs {
+            let p = processor.clone();
+            let job_id = job.id();
+            actix_web::rt::spawn(async move {
```
rh-jfuller: I may have missed something but this looks open ended to me eg. under load we might accumulate unbounded concurrent HTTP polling loops + DB connections ... consider constraining ?

modules/exploit-intelligence/src/model/mod.rs
```diff
@@ -0,0 +1,450 @@
...
+use utoipa::ToSchema;
+use uuid::Uuid;
+
+/// Request body for triggering an Exploit Intelligence analysis.
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct AnalyzeRequest {
+    /// The SBOM identifier to analyze.
+    pub sbom_id: Uuid,
+    /// The vulnerability identifier (e.g., "CVE-2024-9680").
+    pub vulnerability_id: String,
```
rh-jfuller, line 16: can we not give this a type - constrain it ?

modules/exploit-intelligence/src/runner/analysis.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    // Detect SBOM format from content rather than labels (which users can modify).
+    let format = DocumentDetector::detect(&sbom_bytes)
+        .map_err(|e| anyhow::anyhow!("failed to detect SBOM format: {e}"))?
+        .format();
+
+    // Wrap in Bytes for O(1) cloning inside the retry closures.
+    let sbom_data = bytes::Bytes::from(sbom_bytes);
+
+    if format == Format::SPDX {
```
ctron: this is too big and should be split.

modules/exploit-intelligence/src/runner/analysis.rs
```diff
@@ -0,0 +1,227 @@
...
+            ingestor,
+            db_rw,
+            config,
+            client,
+            job_id,
+            &product_id,
+        )
+        .await
+    } else {
+        // CycloneDX single-component flow.
```
ctron: This else doesn't mean it's a CDX file. It means it is not a SPDX file. This should be a match with an `other` catch.

modules/exploit-intelligence/src/runner/analysis.rs
```diff
@@ -0,0 +1,227 @@
...
+        // Create the component record before uploading so it is tracked.
+        let component_ref = sbom_model
```
ctron: Why are we cloning here if we then pass a reference?

modules/exploit-intelligence/src/runner/analysis.rs
```diff
@@ -0,0 +1,227 @@
...
+        let upload_url = format!("{}/api/v1/products/upload-cyclonedx", config.url);
+
+        let response = (|| {
+            let url = upload_url.clone();
+            let sbom = sbom_data.clone();
+            let vuln_id = vulnerability_id.to_string();
+            let cli = client.clone();
+            let tp = config.token_provider.clone();
+            async move {
+                let form = reqwest::multipart::Form::new().text("cveId", vuln_id).part(
```
ctron: Please import types

modules/exploit-intelligence/src/runner/worker.rs
```diff
@@ -0,0 +1,367 @@
...
+                if let Err(e) = p.process(job).await {
+                    tracing::error!(job_id = %job_id, error = %e, "EI analysis failed");
+                    let _ = p.mark_failed(job_id, &e.to_string()).await;
```
rh-jfuller: maybe this needs logging ? If DB is down then doesnt this mean jobs stay in "running" state forever ?

modules/exploit-intelligence/src/endpoints/mod.rs
rh-jfuller, line 1: any new endpoints should consider reusing generic query/filtering approach (eg. standard sorting, no FTS) ..

modules/exploit-intelligence/src/runner/polling.rs
```diff
@@ -0,0 +1,610 @@
...
+                }
+            }
+        })
+        .retry(
+            ConstantBuilder::default()
+                .with_delay(poll_interval)
+                .without_max_times(),
+        )
+        .when(|e| match e {
+            EiRequestError::Transient(reason) if reason == "pending" => {
```
rh-jfuller: why not define an enum instead of baking in raw string "pending" ?

modules/exploit-intelligence/src/error.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+impl ResponseError for Error {
+    fn error_response(&self) -> HttpResponse<BoxBody> {
+        match self {
+            Self::Unavailable => {
+                HttpResponse::ServiceUnavailable().json(ErrorInformation::new("Unavailable", self))
+            }
+            err => {
+                log::warn!("{err}");
```
rh-jfuller: I think we are using tracing::warn! everywhere now adays ...

modules/exploit-intelligence/src/endpoints/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+        .await
+        .map_err(Error::from)?;
+
+    if sbom.is_none() {
+        return Ok(HttpResponse::NotFound().finish());
+    }
+
+    // Check for an existing active job (deduplication).
+    if let Some(active_job) = ei_service
+        .find_active_job(body.sbom_id, &body.vulnerability_id, &tx)
```
rh-jfuller: I wonder if there is a race possible under READ COMMITTED here eg. `find_active_job check` and `create_job insert`  run inside a single transaction, but find_active_job is a plain SELECT without FOR UPDATE. 

I tihnk under pg READ COMMITTED two concurrent analyze calls for the same (sbom_id, vulnerability_id) can both see no active job ?

ctron:
The read only flag is not honored, I guess that might be a UI bug actually: 

image

modules/exploit-intelligence/src/runner/polling.rs
```diff
@@ -0,0 +1,610 @@
...
+        .map_err(|e| anyhow::anyhow!("VEX ingestion failed: {e}"))?;
+    tx.commit().await?;
+
+    tracing::info!(
+        job_id = %job_id,
+        advisory_id = %result.id,
+        "ingested VEX advisory from exploit intelligence"
+    );
+
+    Ok(Uuid::from_str(result.id.strip_prefix("urn:uuid:").unwrap_or(&result.id)).ok())
```
rh-jfuller: `strip_prefix` fix for bare UUIDs is good but I think .ok() silently converts any parse failure into `None` ... maybe we should comment, log or do something else eg. if ingestor ever returns something that isn't a valid UUID ?

modules/exploit-intelligence/src/runner/polling.rs
```diff
@@ -0,0 +1,610 @@
...
+                match poll.status.as_str() {
+                    "completed" | "failed" | "expired" => Ok(poll),
+                    _ => {
+                        tracing::debug!(
+                            status = poll.status,
+                            "exploit intelligence analysis still in progress"
+                        );
+                        // Extend visibility so other instances know we're alive.
+                        if let Err(e) = ei_svc
+                            .extend_visibility(job_id, std::time::Duration::from_secs(5 * 60), &*db)
```
rh-jfuller: If spawned analysis outlives visibility window because `extend_visibility` keeps failing or I guess a poll blocks long enough for the window to expire - would not another instance reclaim the job via `dequeue_jobs` and start a second run_analysis while the first is still running ? 

Not entirely sure about this but thought it worth raising a comment on.

modules/exploit-intelligence/src/runner/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+            let status = resp.status();
+            if status.is_success() {
+                Ok(resp)
+            } else {
+                let body = resp.text().await.unwrap_or_default();
+                let body = body.trim();
+                let detail = if body.is_empty() || body.starts_with('<') {
+                    format!("{operation}: HTTP {status}")
+                } else if body.len() <= 512 {
+                    format!("{operation}: HTTP {status}: {body}")
```
rh-jfuller, line 103: do we really want to expose exploit IQ details to trustify API users ? I would prefer some generic applicable error, appropriate logging and not expose to end users some unrelated system.

server/src/profile/api.rs
rh-jfuller, line 1: consider more tests  ... ex. `well_known_trustify` with EI enabled which asserts resp["exploitIntelligence"] == true as this is flag UI uses to show/hide the EI feature ... another ex. verify POST /v3/exploit-intelligence/analyze returns 503 in read-only mode ... there might be more

ctron:
Clicking on the "request" button, nothing happens at first, I can click multiple times, then I get a visual feedback after a while. In the meantime I see the following in the logs:

```
| 2026-07-27T09:28:22.562210Z INFO  create_job: trustify_module_exploit_intelligence::service: error=database error: Query Error: error returned from database: duplicate key value violates unique constraint "idx_ei_job_active_dedup"
| 2026-07-27T09:28:22.562288Z WARN  trustify_module_exploit_intelligence::error: database error: Query Error: error returned from database: duplicate key value violates unique constraint "idx_ei_job_active_dedup"
```

Which gets reported as a `500 Internal` error.

I think:

1. once the user clicks in the UI, the button should be disabled
2. in the case of a duplicate issue, this should be reported as some form of 4xx, which a proper message

modules/exploit-intelligence/src/runner/polling.rs
ctron: In a case like that you can consider using an `Arc<AtomicUsize>`. Or of you need locking `Mutex<usize>`.

ctron:
General observation, import types, rather than fully prefixing them.

modules/exploit-intelligence/src/client.rs
```diff
@@ -0,0 +1,119 @@
+use serde::Deserialize;
```
ctron, line 1: Please rename to just `client`. Also consider splitting it up.

ctron: Ok, splitting it up might not make sense, I was expecting the actual "client" to be here. This seems just to be the client model. `auth` (which is a sibling) contains the auth part of the client. I think this might need some structure.

modules/exploit-intelligence/src/service/mod.rs
```diff
@@ -0,0 +1,949 @@
...
+        ComponentCounts, ExploitIntelligenceJobDetails, ExploitIntelligenceJobSummary, JobStatus,
+    },
+};
+use sea_orm::sea_query::SimpleExpr;
+use sea_orm::{
+    ActiveModelTrait, ColumnTrait, Condition, ConnectionTrait, DbErr, EntityTrait, PaginatorTrait,
+    QueryFilter, QueryOrder, QuerySelect, Set, TransactionTrait,
+    sea_query::{Expr, LockBehavior, LockType},
+};
+use std::collections::HashMap;
```
ctron: Please nest all imports

modules/exploit-intelligence/src/service/mod.rs
```diff
@@ -0,0 +1,949 @@
...
+use trustify_entity::exploit_intelligence_job_component;
+use uuid::Uuid;
+
+/// Cast a job status string to the PostgreSQL enum type.
+/// `update_many` with `Expr::value()` sends a plain text value, which
+/// PostgreSQL rejects because the column is a custom enum. This helper
+/// wraps the value in an explicit CAST.
+fn status_expr(status: ExploitIntelligenceJobStatus) -> SimpleExpr {
+    let s = match status {
+        ExploitIntelligenceJobStatus::Pending => "pending",
```
ctron: Unless there's a reason, this should be done by e.g. `strum`

modules/exploit-intelligence/src/service/mod.rs
```diff
@@ -0,0 +1,949 @@
...
+
+    /// Atomically dequeue eligible jobs using `FOR UPDATE SKIP LOCKED`.
+    ///
+    /// At most `limit` jobs are claimed per call to avoid one instance
+    /// starving others. Each claimed job gets a short visibility window;
+    /// the worker must periodically call `extend_visibility` to prove it
+    /// is still alive. If it crashes, the window expires and another
+    /// instance reclaims the job on its next tick.
+    #[instrument(skip_all, err(level = tracing::Level::INFO))]
+    pub async fn dequeue_jobs(
```
ctron: This feels a bit like overkill. And I'm not sure the "lock for update, skip locked" is used in the way it makes sense. It seems like there are two locking strategies mixed here.

If we do "lock for update, skip locked", and would only fetch one (which we want to process), then we would hold that lock until we processed. That would be fine. No other mechanism required.

If we do this time locking thing (which I'm not a fan of) then we don't need the "update lock, skip locked"?

Also, I think it would be fine just picking one item after the other. Especially if we can run multiple workers in parallel. But that would reduce code complexity.

modules/exploit-intelligence/src/runner/analysis.rs
ctron:
```rust
stream.try_collect::<BytesMut>().await.unwrap()
```

modules/exploit-intelligence/src/service/mod.rs
Strum355: > If we do "lock for update, skip locked", and would only fetch one (which we want to process), then we would hold that lock until we processed. That would be fine. No other mechanism required.

This is not an approach I would ever take, given that EI analysis can take anywhere from minutes to potentially hours. We've had a lot of issues with long-lived `FOR UPDATE SKIP LOCKED` job queues (and just long-lived open transactions in general such as with advisory locks or whatever else) in the past at my previous company, so I am strongly against this. Its an operational nightmare on multiple fronts.

> If we do this time locking thing (which I'm not a fan of) then we don't need the "update lock, skip locked"?

The time-locking thing is an approach that you linked to me in the previous PR here https://github.com/guacsec/trustify/pull/2464#issuecomment-4977805271. If we don't have `FOR UPDATE SKIP LOCKED` then it may be possible for multiple trustify instances to be able to claim a job and both process them, which we obviously wouldn't want.

Strum355:
> The read only flag is not honored, I guess that might be a UI bug actually:
> image

Thanks for calling this out. The analyze endpoint should be gated behind having `create:document` permissions by default (or more granularly there is `create.exploitIntelligence`), so Im presuming this should be blocked by trustify on the backend, unless read-only side-steps the permissions model? @ctron The UI should definitely be updated to account for this though

modules/exploit-intelligence/src/service/mod.rs
ctron: But the lock would only be held for requesting the update from the remote side, right? Fetching the update would be quick.

ctron:
> > The read only flag is not honored, I guess that might be a UI bug actually:
> > image
> 
> Thanks for calling this out. The analyze endpoint should be gated behind having `create:document` permissions by default (or more granularly there is `create.exploitIntelligence`), so Im presuming this should be blocked by trustify on the backend, unless read-only side-steps the permissions model? @ctron The UI should definitely be updated to account for this though

No, this is the "read only" system state. I't independent of the permission model.

modules/exploit-intelligence/src/service/mod.rs
Strum355: I think I see what you mean, I read it as holding the transaction for the duration of the EI-side analysis. My main concern then is how fast can we expect VEX ingestion to be once analysis has completed and EI has returned a VEX? Is it possible for the ingestor to get backed-up, or whats the queueing situation there? Taking into account that in the future we will be looking at aggregating the individual VEXs returned by EI for a multi-component SPDX file (if that has any impact on the ingestion)

Strum355:
> When setting the EI URL, but not the EI UI URL, what's expected to happen?

EI UI URL should fallback to EI URL, but now that you mention it, it seems like theres been a regression on this. Will be fixed, thanks! 

> Once the user clicks in the UI, the button should be disabled

Correct, I had this planned todo but forgot to get around to it. 

> Are `--exploit-intelligence-auth-token <EXPLOIT_INTELLIGENCE_AUTH_TOKEN>` and the OIDC configuration mutually exclusive? Then we need to enforce this using clap.

They should be yes, TIL clap can enforce this! Neat

modules/exploit-intelligence/src/endpoints/mod.rs
Strum355: There is a partial unique index on `(sbom_id, vuln_id)` that should catch this, although I need to double check whether we're handling that error cleanly when bubbling up the result to the API

modules/exploit-intelligence/src/runner/polling.rs
Strum355: This should be resolvable with just `AtomicU32`, no need for Arc or Mutex. Theres no shared ownership here, The local context is single-threaded (so `!Send + !Sync` works fine here)

modules/exploit-intelligence/src/runner/worker.rs
Strum355: It is open-ended yes. Actual load on our end is relatively small, none of our operations are heavy outside of initial loading & uploading of the SBOM to EI. We don't hold any long-held DB connections (unless something slipped through, which I will double check for). If you think its worth constraining the SBOM upload stage, I can definitely do that (could be a reasonable change given the size of some SBOMs). 

I wouldn't want to constrain any more than that though. EI has its own queueing mechanism, and analysis can take anywhere from minutes to hours, so I dont want jobs queued up on our end waiting for slow analysis of other SBOMs when EI has its own queuing system in place. Thoughts?

modules/exploit-intelligence/src/runner/analysis.rs
Strum355: You mentioned this on the previous PR but I couldn't find an example of what you were referencing. Thanks for digging it up! Much cleaner

modules/exploit-intelligence/src/model/mod.rs
Strum355: I dont think we have an existing type for this in the codebase, or at least, every other field named `vulnerability_id` in the codebase also uses `String`

modules/exploit-intelligence/src/service/mod.rs
ctron: Maybe @rh-jfuller can help out with some experience here. But I'd say, that holding the lock (of the job entry) throughout the ingestion shouldn't be a problem at all. As the row would be skipped by others.

On the other side, if you do this "extends" patttern, while performing the ingestion, how would to extend the time?

modules/exploit-intelligence/src/model/mod.rs
```diff
@@ -0,0 +1,450 @@
...
+    fn from(finding: Finding) -> Self {
+        match finding {
+            Finding::Vulnerable => Self::Vulnerable,
+            Finding::NotVulnerable => Self::NotVulnerable,
+            Finding::Uncertain => Self::Uncertain,
+        }
+    }
+}
+
+impl From<JobStatus> for ExploitIntelligenceJobStatus {
```
ctron: Why do we need to duplicate this?

modules/exploit-intelligence/src/endpoints/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+            status: active_job.status,
+        }));
+    }
+
+    // Create the job record within the same transaction.
+    let job = ei_service
+        .create_job(body.sbom_id, &body.vulnerability_id, &tx)
+        .await?;
+
+    tx.commit().await.map_err(Error::from)?;
```
ctron: Why not just use `tx.commit().await?`?

modules/exploit-intelligence/src/endpoints/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+)]
+#[post("/v3/exploit-intelligence/analyze")]
+pub async fn analyze(
+    ei_service: web::Data<ExploitIntelligenceService>,
+    db_rw: web::Data<db::ReadWrite>,
+    body: web::Json<AnalyzeRequest>,
+    _: Require<CreateExploitIntelligence>,
+) -> actix_web::Result<impl Responder> {
+    // Verify EI is configured before doing any work.
+    let _ = ei_service.runtime().ok_or(Error::Unavailable)?;
```
ctron: nit: if runtime() would return `Result<Self, Error>`, this could be:

```rust
let _ = ei_service.runtime()?;
```

modules/exploit-intelligence/src/endpoints/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    // Use a single read-write transaction for the existence check, active-job
+    // deduplication, and job creation to prevent TOCTOU race conditions where
+    // two concurrent requests could both pass the check and create duplicate jobs.
+    let tx = db_rw.begin().await.map_err(Error::from)?;
```
ctron: The `map`s are not required if the handler returns `Result<_, Error>`.

modules/exploit-intelligence/src/endpoints/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+    responses(
+        (status = 200, description = "Job details", body = ExploitIntelligenceJobDetails),
+        (status = 404, description = "Job not found"),
+    ),
+)]
+#[get("/v3/exploit-intelligence/jobs/{id}")]
+pub async fn get_job(
+    ei_service: web::Data<ExploitIntelligenceService>,
+    db: web::Data<db::ReadOnly>,
+    id: web::Path<Uuid>,
```
ctron: IDs should be strings. So if the user provides something that is not parsable as `Uuid`, it should return 404.

modules/exploit-intelligence/src/endpoints/mod.rs
ctron: I don't think we are, I think I saw this resulting in a 500 (Internal error).

Why not just use create, catch the error using `.is_duplicate()` and return the proper code then?

ctron: `trustify_common::db::DatabaseErrors::is_duplicate`

modules/exploit-intelligence/src/ei_client.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub name: Option<String>,
+    pub version: Option<String>,
+    pub image: Option<String>,
+    pub error: Option<String>,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ProductSummary {
+    pub product_state: String,
```
ctron: This should be an enum

modules/exploit-intelligence/src/runner/mod.rs
ctron: I'd say we shouldn't. But log it instead.

modules/exploit-intelligence/src/runner/polling.rs
```diff
@@ -0,0 +1,610 @@
...
+///
+/// Each successful poll tick extends the job's visibility window so that
+/// other instances know this job is still being actively processed. EI
+/// analysis can take hours; without extension the short initial window
+/// would expire and another instance would re-claim the job.
+///
+/// Updates both the component record (with finding/advisory) and the parent
+/// job record (status only) upon completion.
+#[allow(clippy::too_many_arguments)]
+pub(crate) async fn poll_for_result(
```
ctron, line 102: I'm wondering if we really not all this complexity. I think, this would be a very simple loop of:

```
fn do_work() {
  ei_start().await?;

  let result = while (!timeout) {
    break check_result().await;
    sleep().await?;
  };

  match result {
    Ok(result) => ingest(result).await;
   Err(err) => report_failure(err).await;
  }
}

fn run_jobs() {

   loop {
     let tx = begin();     
     let next = find_next_job();
     do_work().await;
     tx.commit();
  }

}

fn find_next_job(tx: C) -> Result<Option<Job>, DbErr>{
  jobs::Enityt::find()
   .next_filter()
   .lock(ForUpdate, SkipNext)
   .limit(1)
   .one()
   .await;
}
```

modules/exploit-intelligence/src/endpoints/mod.rs
Strum355: > Why not just use create, catch the error using .is_duplicate() and return the proper code then?

Absolutely, thats the plan

modules/exploit-intelligence/src/runner/polling.rs
Strum355: Is this suggesting processing analysis requests one-by-one sequentially, or is some concurrency inherently expected here? This also still has the issue that I mentioned in the other discussion thread that it holds a FOR UPDATE lock for the duration of EI analysis, which can take anywhere from minutes to hours. I want to caution against that kind of pattern

modules/exploit-intelligence/src/service/mod.rs
```diff
@@ -0,0 +1,949 @@
...
+    ) -> Result<Vec<exploit_intelligence_job::Model>, Error> {
+        let runtime = self.runtime.as_ref().ok_or(Error::Unavailable)?;
+        // Short initial window — the worker extends it while processing.
+        let visibility_timeout = Duration::from_secs(5 * 60);
+        let max_retries = runtime.config.upload_max_retries as i32;
+
+        let tx = db.begin().await?;
+        let now = OffsetDateTime::now_utc();
+
+        let jobs = exploit_intelligence_job::Entity::find()
```
rh-jfuller: not sure if this is covered by an index eg. postgres must seqscan to check lock state ... at first this might not be bad but over time when exploit_intelligence_job grows bigger might turn into a problem ... note this kind of thing might also impact `fail_exhausted_jobs`

modules/exploit-intelligence/src/service/mod.rs
```diff
@@ -0,0 +1,949 @@
...
+        Ok(())
+    }
+
+    /// Mark jobs that have exhausted their retry budget as Failed.
+    ///
+    /// Without this, retry-exhausted jobs would remain in Running/Pending
+    /// status forever — excluded from dequeue (retry_count >= max) but
+    /// never transitioned to a terminal state.
+    #[instrument(skip_all, err(level = tracing::Level::INFO))]
+    pub async fn fail_exhausted_jobs(&self, db: &db::ReadWrite) -> Result<(), Error> {
```
rh-jfuller: I think this transitions jobs to **Failed** via a bare `update_many` (/w no TX  and no component cascade). `update_job_failed()` wraps both job | component updates in a single TX.

modules/exploit-intelligence/src/service/mod.rs
rh-jfuller: +1 to what @ctron said ... I think holding lock throughout ingestion should be carefree and fine.

modules/exploit-intelligence/src/service/mod.rs
Strum355: This (the `dequeue_jobs` scenario at least) _should_ be somewhat covered by [the index here](https://github.com/guacsec/trustify/pull/2534/changes#diff-4799bc61e175f469cc75235d416e8d16a291c9b8174e459be0b820fa2641962bR122-R124), so it should in theory only perform a scan of the index (see bottom for an example I had claude whip up)

Index:
```sql
CREATE UNIQUE INDEX IF NOT EXISTS idx_ei_job_active_dedup
ON exploit_intelligence_job (sbom_id, vulnerability_id)
WHERE status IN ('pending', 'running')
```

Example plan:
```
      Bitmap Heap Scan on exploit_intelligence_job  (cost=21.85..750.86 rows=64 width=21)
        Recheck Cond: (status = ANY ('{pending,running}'::text[]))
        Filter: (priority = 3)
        ->  Bitmap Index Scan on idx_ei_job_active_dedup  (cost=0.00..21.84 rows=313 width=0)
```

modules/exploit-intelligence/src/runner/polling.rs
Strum355: This is expected, the job queue follows at-least-once semantics here. If `extend_visibility` keeps failing (DB temporarily unreachable), the window expires and another instance reclaims the job via `dequeue_jobs`. Both might run the analysis concurrently, but terminal updates use a `WHERE status = Running` guard so only the first to complete actually transitions the job, the other is a no-op. Does this seem reasonable or did you spot anything more specific that this answer doesnt cover?

modules/exploit-intelligence/src/endpoints/mod.rs
Strum355: I kept `runtime()` returning Option since it's also called in `build_ei_worker_task` where `?` on `Option` gives the early-return-None behavior directly. Changing to `Result` would benefit the endpoint call site but make that one worse.

modules/exploit-intelligence/src/runner/worker.rs
Strum355: > If DB is down then doesnt this mean jobs stay in "running" state forever ?

It shouldn't, the visibility window will mean a worker will pick it up again after the window has expired

modules/exploit-intelligence/src/runner/mod.rs
Strum355: Is there some degree of EI details that we do want to expose e.g. EI rejected the SBOM because it didnt meet a set of requirements (certain fields were missing for example). I decided against doing validation on our side because of the risk for drift, so EI would return us an error indicating the SBOM won't work for them, and then I surface that to the user.
For more internal ones (like DB error/EI connectivity errors) we can definitely have something more generic. 

> and not expose to end users some unrelated system.

I think this is antithetical to the goal of the integration to consider the systems completely unrelated. We construct a URL to the EI UI in the API that specifically gets displayed in the TPA UI, we aren't a blackbox/white-label front for EI

rh-jfuller: I disagree ... this is not a product decision this is an arch decision - separation of concerns, etc ...as long as EI is separate we should abstract away internal operation details. Those who maintain and run instance of trustify will get information from logging.

modules/exploit-intelligence/src/client.rs
Strum355: Ive renamed this file for the time being, but Ill take another look at the structure tomorrow

Strum355:
@rh-jfuller @ctron most of the comments have addressed, what remains are:
- [unbound analysis request concurrency](https://github.com/guacsec/trustify/pull/2534#discussion_r3655500595)
- [`fail_exhausted_jobs` doesnt cascade to components](https://github.com/guacsec/trustify/pull/2534#discussion_r3658260954)
- two separate discussions around the locking strategy [here](https://github.com/guacsec/trustify/pull/2534#discussion_r3656137090) and [here]9https://github.com/guacsec/trustify/pull/2534#discussion_r3657083980)
- [structure of the "client" file](https://github.com/guacsec/trustify/pull/2534#discussion_r3656047389)

Some UI changes have been pushed to trustify-ui as well in the meantime

modules/exploit-intelligence/src/runner/worker.rs
rh-jfuller: This is not a question of distributed arch mechanics eg. just want a 'cap' on how many spawned jobs run at once without any worker-pool machinery - defining some sort of max concurrency with tokio::sync::Semaphore .. this does not even need to be exposed as an ENV var.

server/src/profile/api.rs
ctron: So if the UI URL is absent, then the normal URL is used? If that's the case, then this should be documented here.

modules/exploit-intelligence/src/runner/polling.rs
ctron: With that, you could just spawn multiple loops (of `run_jobs`). But that would be bounded then to the number of concurrent jobs to run.

But to my understanding, the request is not waiting for completion of the analysis. But for the request checking for the completion, right? That should be below one minute, and so it's fine. The lock would be held for those seconds. You can still have a "last_checked" timestamp, which postpones checking "too early". But this is about delaying then, not locking.

Strum355: Ah, I see what youre recommending yes. That seems like an approach that could work, I'll take a deeper look and see how it turns out.

Strum355: @ctron @rh-jfuller ive reworked the loop in https://github.com/guacsec/trustify/pull/2534/changes/3d6d0cd2942b3244d70fd3ec812144b61b264b88

TLDR: The worker now holds a database row lock for the duration of each processing step, replacing the old visibility-timeout approach. Jobs progress through ticks: one tick uploads the SBOM (transitions state from `pending` to `running`), subsequent ticks poll EI until completion (transitions from `running` to `completed`). If anything fails, the transaction rolls back cleanly and the job is retried on the next tick (up to a configurable limit). Permanent errors like 400s fail immediately without wasting retries. Multiple worker loops can be run concurrently capped by a config setting.

modules/exploit-intelligence/src/runner/mod.rs
Strum355: image

Error message no longer exposed in API, UI displays a (timed) banner as shown and with the "Retry analysis" button from before (slightly reworded), with the backend changes in https://github.com/guacsec/trustify/pull/2534/commits/245257f5ce0dd0f9111e91cdb5a835b98f9f2351 and the frontend changes in https://github.com/guacsec/trustify-ui/pull/1125/commits/1aad5641c91580b55d8fde0f9580b3c38bcb1056

rh-jfuller (approved):
thx for all the changes (nice work!), **LGTM** though would still like @ctron blessing as well  (and rebasing and CI job passing).

modules/exploit-intelligence/src/runner/mod.rs
ctron: (not a blocker) I'm not a fan of timed error messages. You click, it takes a bit, you blink, … "nothing" happened. Because you missed that notification.

If it's ok to miss a notification (like "everything fine"), then timed is ok. If something failed and needs attention, it should not go away automatically.

Strum355: Not my preference either, but this was what I was given from design team

modules/exploit-intelligence/src/endpoints/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+        .service(analyze)
+        .service(list_jobs)
+        .service(get_job);
+}
+
+/// Filterable/sortable fields for the job list endpoint.
+#[derive(trustify_query_derive::Query)]
+#[allow(dead_code)]
+struct ExploitIntelligenceJobQuery {
+    id: uuid::Uuid,
```
ctron: Import the type.

modules/exploit-intelligence/src/endpoints/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+    sbom_id: uuid::Uuid,
+    vulnerability_id: String,
+    status: String,
+    created: time::OffsetDateTime,
```
ctron: Import the type

modules/exploit-intelligence/src/endpoints/mod.rs
ctron: If we have this check, we do we still need this check?

modules/exploit-intelligence/src/endpoints/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+            tx.commit().await?;
+            Ok(HttpResponse::Created().json(AnalyzeResponse {
+                job_id: job.id,
+                status: JobStatus::Pending,
+            }))
+        }
+        Err(e) if e.is_duplicate() => {
+            drop(tx);
+            let active = ei_service
+                .find_active_job(body.sbom_id, &body.vulnerability_id, &**db_rw)
```
ctron: Why does the R/O operation require the R/W connection?

modules/exploit-intelligence/src/endpoints/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+)]
+#[get("/v3/exploit-intelligence/jobs/{id}")]
+pub async fn get_job(
+    ei_service: web::Data<ExploitIntelligenceService>,
+    db: web::Data<db::ReadOnly>,
+    id: web::Path<String>,
+    _: Require<ReadExploitIntelligence>,
+) -> Result<impl Responder, Error> {
+    let id =
+        uuid::Uuid::parse_str(&id).map_err(|_| Error::NotFound(format!("invalid job id: {id}")))?;
```
ctron: Import the type

modules/exploit-intelligence/src/model/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+    pub excluded: i32,
+    pub total: i32,
+    pub vulnerable: i32,
+    pub not_vulnerable: i32,
+    pub uncertain: i32,
+}
+
+impl ComponentCounts {
+    /// Compute component counts from a slice of component entity models.
+    pub fn from_component_models(models: &[exploit_intelligence_job_component::Model]) -> Self {
```
ctron, line 123: Instead of iterating through it multiple times:

```rust
let mut result = models.iter().fold(
    Self {
        completed: 0,
        failed: 0,
        excluded: 0,
        total: models.len() as i32,
        vulnerable: 0,
        not_vulnerable: 0,
        uncertain: 0,
    },
    |mut counts, model| {
        if model.excluded {
            counts.excluded += 1;
        } else {
            match model.status {
                ExploitIntelligenceJobStatus::Completed => counts.completed += 1,
                ExploitIntelligenceJobStatus::Failed => counts.failed += 1,
                _ => {}
            }
        }

        match model.finding.as_ref() {
            Some(Finding::Vulnerable) => counts.vulnerable += 1,
            Some(Finding::NotVulnerable) => counts.not_vulnerable += 1,
            Some(Finding::Uncertain) => counts.uncertain += 1,
            None => {}
        }

        counts
    },
);
```

modules/exploit-intelligence/src/model/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+    Pending,
+    /// Analysis is in progress.
+    Running,
+    /// Analysis completed successfully.
+    Completed,
+    /// Analysis failed with an error.
+    Failed,
+}
+
+impl From<ExploitIntelligenceJobStatus> for JobStatus {
```
ctron: I think I asked that before, why do we need to copy this enum?

modules/exploit-intelligence/src/model/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+pub enum JobFinding {
+    /// The analysed component is vulnerable to the CVE.
+    Vulnerable,
+    /// The analysed component is not vulnerable to the CVE.
+    NotVulnerable,
+    /// The analysis was inconclusive.
+    Uncertain,
+}
+
+impl From<Finding> for JobFinding {
```
ctron: Same here

modules/exploit-intelligence/src/model/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+    }
+}
+
+/// Compute the EI report URL from a job's identifiers and the UI base URL.
+///
+/// For multi-component (SPDX) jobs with a `product_id`, builds a product URL.
+/// For single-component (CycloneDX) jobs, uses the component's `scan_id`
+/// passed via `first_scan_id` to build a component URL.
+fn compute_job_report_url(
+    model: &trustify_entity::exploit_intelligence_job::Model,
```
ctron:
```suggestion
    model: &exploit_intelligence_job::Model,
```

modules/exploit-intelligence/src/endpoints/mod.rs
Strum355: It should not, think this a hold-over from a previous change that I forgot to change

modules/exploit-intelligence/src/runner/analysis.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    let storage_key = StorageKey::from_sha256(&source_doc.sha256);
+
+    let stream = ingestor
+        .storage()
+        .retrieve(storage_key)
+        .await
+        .map_err(|e| anyhow::anyhow!("storage retrieve error: {e}"))?
+        .ok_or_else(|| {
+            AnalysisError::Permanent(anyhow::anyhow!("SBOM document not found in storage"))
```
ctron, line 115: Please use proper errors. Using anyhow is ok for outer layers (e.g. CLI)

modules/exploit-intelligence/src/runner/analysis.rs
```diff
@@ -0,0 +1,264 @@
...
+        })?;
+
+    let sbom_data: bytes::Bytes = stream
```
ctron: Let the compiler figure it out.

modules/exploit-intelligence/src/runner/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+pub(crate) enum AnalysisError {
+    /// Retryable failure — worth claiming the job again on the next tick.
+    #[error("{0}")]
+    Retryable(anyhow::Error),
+    /// Permanent failure — the job should be marked failed immediately.
+    #[error("{0}")]
+    Permanent(anyhow::Error),
+}
+
+impl From<anyhow::Error> for AnalysisError {
```
ctron: Use `thiserror` and `from` attribute

modules/exploit-intelligence/src/runner/polling.rs
```diff
@@ -0,0 +1,412 @@
+use crate::{
+    client::{PollResponse, ProductReportEntry, ProductState, ProductStatusResponse},
+    runner::{AnalysisError, EiRequestError, classify_response_error},
+    service::{ExploitIntelligenceConfig, ExploitIntelligenceService},
+};
+use sea_orm::{ConnectionTrait, TransactionTrait};
+use std::str::FromStr;
+use trustify_auth::client::TokenInjector;
+use trustify_entity::exploit_intelligence_job::{ExploitIntelligenceJobStatus, Finding};
```
ctron: Nest imports

modules/exploit-intelligence/src/client.rs
```diff
@@ -0,0 +1,130 @@
...
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ReportRequestId {
+    pub report_id: String,
+}
+
+/// Response from GET /api/v1/reports/by-scan-id/{scan_id}
+#[derive(Debug, Deserialize)]
+pub struct PollResponse {
+    pub status: String,
```
ctron: Make this an enum.

modules/exploit-intelligence/src/runner/polling.rs
```diff
@@ -0,0 +1,412 @@
...
+                .and_then(|r| r.output.as_ref())
+                .and_then(|o| o.vex.as_ref());
+
+            let advisory_id = match vex_json {
+                Some(vex) => ingest_vex_advisory(vex, ingestor, db, job_id).await?,
+                None => None,
+            };
+
+            ei_service
+                .update_component_completed(component_id, finding, advisory_id, db)
```
ctron, line 131: If that fails, what's our recovery procedure?

modules/exploit-intelligence/src/runner/polling.rs
```diff
@@ -0,0 +1,412 @@
...
+        }
+        _ => Ok(()),
+    }
+}
+
+/// Single-poll an SPDX product status. Returns `Ok(())` on all outcomes:
+/// - Terminal (completed/failed): processes components + updates job
+/// - Still in progress: returns Ok without changes (re-polled next tick)
+/// - Error: returns Err (caller increments retry count)
+pub(crate) async fn poll_for_product_result(
```
ctron, line 170: This looks like a lot of duplicate code, can we re-use that?

modules/exploit-intelligence/src/runner/polling.rs
```diff
@@ -0,0 +1,412 @@
...
+        ProductState::Failed => {
+            let msg = format!("EI product analysis failed for product {product_id}");
+            ei_service.update_job_failed(job_id, &msg, db).await?;
+            Ok(())
+        }
+        _ => Ok(()),
+    }
+}
+
+fn extract_finding(poll: &PollResponse) -> Finding {
```
ctron: Please comment

modules/exploit-intelligence/src/runner/worker.rs
```diff
@@ -0,0 +1,749 @@
+use crate::{
+    runner::{AnalysisError, analysis::run_analysis},
+    service::ExploitIntelligenceService,
+};
+use std::time::Duration;
+use tokio::task::LocalSet;
+use tokio::time::MissedTickBehavior;
```
ctron: Nest imports

modules/exploit-intelligence/src/runner/worker.rs
```diff
@@ -0,0 +1,749 @@
...
+
+struct EiWorker {
+    ei_service: ExploitIntelligenceService,
+    ingestor: IngestorService,
+    db_rw: db::ReadWrite,
+}
+
+impl EiWorker {
+    async fn process_one_tick(&self) -> anyhow::Result<()> {
+        let runtime = self
```
ctron: Do we need to run it if it is not configured?

modules/exploit-intelligence/src/runner/worker.rs
```diff
@@ -0,0 +1,749 @@
...
+            )
+            .await?;
+
+        let Some(job) = job else {
+            tx.commit().await?;
+            return Ok(());
+        };
+
+        let job_id = job.id;
+        let max_retries = runtime.config.max_retries as i32;
```
ctron: I'd do it the other way round. A user could configure somethign that doesn't fit into `i32`, but the DB would not return something negative.

modules/exploit-intelligence/src/runner/worker.rs
```diff
@@ -0,0 +1,749 @@
...
+            &tx,
+            &runtime.config,
+            &runtime.client,
+            &job,
+        )
+        .await
+        {
+            tx.rollback().await?;
+
+            let fail_immediately = matches!(e, AnalysisError::Permanent(_));
```
ctron: have this in the if

modules/exploit-intelligence/src/runner/worker.rs
```diff
@@ -0,0 +1,749 @@
...
+
+            if fail_immediately {
+                tracing::error!(job_id = %job_id, error = %e, "permanent failure, failing job");
+                let msg = e.to_string();
```
ctron: This is allocating a string just to take the reference. Can't you just pass `&msg`?

modules/exploit-intelligence/src/runner/worker.rs
```diff
@@ -0,0 +1,749 @@
...
+                if let Err(db_err) = self
+                    .ei_service
+                    .increment_retry_count(job_id, &*self.db_rw)
+                    .await
+                {
+                    tracing::warn!(job_id = %job_id, error = %db_err, "failed to increment retry count");
+                }
+                if new_count >= max_retries {
+                    tracing::error!(job_id = %job_id, error = %e, attempt = new_count, max_retries, "failing job: retries exhausted");
+                    let msg = format!("max retries exhausted: {e}");
```
ctron: inline this

modules/exploit-intelligence/src/runner/worker.rs
```diff
@@ -0,0 +1,749 @@
...
+            }
+            return Ok(());
+        }
+
+        tx.commit().await?;
+        Ok(())
+    }
+}
+
+pub async fn run_worker(
```
ctron: We need a OneShot shutdown signal here

Cargo.toml
```diff
@@ -48,6 +49,7 @@ aws-config = { version = "1.8.14", features = ["behavior-version-latest"] }
 aws-sdk-s3 = { version = "1.124.0", default-features = false, features = ["behavior-version-latest", "default-https-client", "http-1x", "rt-tokio", "sigv4a"] }
 aws-smithy-http-client = { version = "1.2.0", features = ["rustls-aws-lc"] }
 aws-smithy-types = { version = "1.4.5" }
+backon = "1.6"
```
ctron: Do we still use `backon`?

ctron:
Looks like a great improvement. Less complex. Lot of minor things to clean up. Error handling, types, imports and common stuff.

Also I noticed that the arguments are still mixed:

```
Exploit Intelligence OIDC:
      --ei-oidc-client-id <ei_oidc_client_id>
          [env: EXPLOIT_INTELLIGENCE_OIDC_CLIENT_ID=]

      --ei-oidc-client-secret <ei_oidc_client_secret>
          [env: EXPLOIT_INTELLIGENCE_OIDC_CLIENT_SECRET=]

      --ei-oidc-issuer-url <ei_oidc_issuer_url>
          [env: EXPLOIT_INTELLIGENCE_OIDC_ISSUER_URL=]

      --ei-oidc-refresh-before <ei_oidc_refresh_before>
          [env: EXPLOIT_INTELLIGENCE_OIDC_REFRESH_BEFORE=]
          [default: 30s]

      --ei-oidc-tls-insecure
          [env: EXPLOIT_INTELLIGENCE_OIDC_TLS_INSECURE=]

      --max-cache-size <max-cache-size>
          Maximum size of the graph cache.
          
          [env: TRUSTD_MAX_CACHE_SIZE=]
          [default: "200 MiB"]

      --concurrency <CONCURRENCY>
          The number of concurrent tasks for analysis (must be > 0).
          
          [env: TRUSTIFY_ANALYSIS_CONCURRENCY=]
          [default: 10]

      --sbom-published-cutoff-days <SBOM_PUBLISHED_CUTOFF_DAYS>
          Only consider SBOMs published within this many days. Disabled by default.
          
          [env: TRUSTIFY_LATEST_SBOM_PUBLISHED_CUTOFF_DAYS=]

```

EI is claiming other settings.

modules/exploit-intelligence/src/runner/polling.rs
Strum355: Transaction gets rolled back, retry count is incremented, if retry budget isnt exceeded then the job is picked up again later. EI result is always available for re-download at a later point

modules/exploit-intelligence/src/runner/worker.rs
Strum355: Would probably be good, but I dont see where we have any existing infrastructure in place to actually make use of it

modules/exploit-intelligence/src/runner/worker.rs
Strum355: `e` here is `AnalysisError`, so an allocation to make it a string is going to happen either way

modules/exploit-intelligence/src/runner/worker.rs
ctron: https://github.com/guacsec/trustify/blob/93e2f8e695d038940c34071c26295710d1609f18/common/infrastructure/src/health/checks/local.rs#L20-L28

ctron (approved):

ruromero (approved):

PR #2536, opened 2026-07-23
fix(migration): re-derive base_score from advisory_vulnerability_score

ctron:
## Summary

- Fix migration `m0002270` that re-derives `vulnerability.base_score`, `base_severity`, and `base_type` from `advisory_vulnerability_score`
- Fixes two issues discovered while investigating TC-5272:
  1. **NULL base_score ("Unknown" in UI)**: CVEs ingested on 0.2.x before `base_score` columns existed were never backfilled after upgrading
  2. **Wrong base_type**: Migration `m0002170` hardcoded `base_type = '3.0'` for all existing scores, even v4.0 and v2.0 ones
- Refactors CVE score extraction to use a single code path (removes duplicate `extract_base_score`/`get_score` logic)
- Fixes score computation when CVE JSON only contains `vectorString` without individual metric fields

## Workaround

To fix existing data without deploying, run this SQL directly on the database:

```sql
UPDATE vulnerability
SET base_score = best.score,
    base_severity = best.severity,
    base_type = best.type
FROM (
    SELECT DISTINCT ON (avs.vulnerability_id)
           avs.vulnerability_id,
           round(avs.score::numeric, 1) AS score,
           avs.severity,
           avs.type
    FROM advisory_vulnerability_score avs
    JOIN vulnerability v ON v.id = avs.vulnerability_id
    WHERE v.authoritative_advisory_id IS NOT NULL
      AND avs.advisory_id = v.authoritative_advisory_id
    ORDER BY avs.vulnerability_id, avs.type DESC, avs.score DESC
) best
WHERE best.vulnerability_id = vulnerability.id;
```

**Note:** This only works if the data migration `m0002010_add_advisory_scores` has been run and `advisory_vulnerability_score` is populated. If that table is empty, the data migration must be run first (`trustd db data m0002010_add_advisory_scores`).

## Test plan

- [x] `cvss_v4_score_visibility` — verifies all endpoints return correct scores after fresh ingestion
- [x] `cvss_v4_score_after_cve_history` — verifies re-ingestion of all 6 historical revisions preserves correct scores
- [x] `cvss_v4_score_after_migration::from_02z` — reproduces TC-5272: CVE ingested on 0.2.x → NULL base_score after migration → fix migration backfills it
- [x] `cvss_v4_score_after_migration::from_03z` — CVE ingested on 0.3.z → wrong base_type after migration → fix migration corrects it
- [x] `cvss_v4_score_after_migration::from_04z` — CVE ingested on 0.4.z → wrong base_type after migration → fix migration corrects it
- [x] `base_score_after_ingestion` — existing parameterized test with new CVSS v4.0 case, all pass
- [x] Full workspace test suite passes

Implements TC-5272

### DIFF
--- a/etc/test-data/cve/CVE-2025-6965-02z-dump.sql
+++ b/etc/test-data/cve/CVE-2025-6965-02z-dump.sql
@@ -1,0 +1,4228 @@
+--
+-- PostgreSQL database dump
+--
+
+\restrict A80f3n21FoIGiiKypGABItbD36memNlX9riFEgpgzTlJ3rjroiBNCIZjOM5CmjN
+
+-- Dumped from database version 17.2
+-- Dumped by pg_dump version 18.4
+
+SET statement_timeout = 0;
+SET lock_timeout = 0;
+SET idle_in_transaction_session_timeout = 0;
+SET transaction_timeout = 0;
+SET client_encoding = 'UTF8';
+SET standard_conforming_strings = on;
+SELECT pg_catalog.set_config('search_path', '', false);
+SET check_function_bodies = false;
+SET xmloption = content;
+SET client_min_messages = warning;
+SET row_security = off;
+
+--
+-- Name: pg_stat_statements; Type: EXTENSION; Schema: -; Owner: -
+--
+
+CREATE EXTENSION IF NOT EXISTS pg_stat_statements WITH SCHEMA public;
+
+
+--
+-- Name: EXTENSION pg_stat_statements; Type: COMMENT; Schema: -; Owner: -
+--
+
+COMMENT ON EXTENSION pg_stat_statements IS 'track planning and execution statistics of all SQL statements executed';
+
+
+--
+-- Name: pg_trgm; Type: EXTENSION; Schema: -; Owner: -
+--
+
+CREATE EXTENSION IF NOT EXISTS pg_trgm WITH SCHEMA public;
+
+
+--
+-- Name: EXTENSION pg_trgm; Type: COMMENT; Schema: -; Owner: -
+--
+
+COMMENT ON EXTENSION pg_trgm IS 'text similarity measurement and index searching based on trigrams';
+
+
+--
+-- Name: cvss3_a; Type: TYPE; Schema: public; Owner: -
+--
+
+CREATE TYPE public.cvss3_a AS ENUM (
+    'n',
+    'l',
+    'h'
+);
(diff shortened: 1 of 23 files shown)


migration/src/m0002270_fix_vulnerability_base_score_type.rs
```diff
@@ -0,0 +1,21 @@
+use sea_orm_migration::prelude::*;
```
rh-jfuller, line 1: FWIW other PR #2534 using migration m0002270

migration/src/m0002270_fix_vulnerability_base_score_type.rs
```diff
@@ -0,0 +1,21 @@
...
+            .get_connection()
+            .execute_unprepared(include_str!(
+                "m0002270_fix_vulnerability_base_score_type/up.sql"
+            ))
+            .await
+            .map(|_| ())
+    }
+
+    async fn down(&self, _manager: &SchemaManager) -> Result<(), DbErr> {
+        Ok(())
```
rh-jfuller, line 19: maybe consider leaving some comments about why this is a no op

modules/ingestor/src/graph/cvss.rs
```diff
@@ -43,8 +44,14 @@ impl From<ScoreInformation> for advisory_vulnerability_score::ActiveModel {
 
 impl From<(String, v2_0::CvssV2)> for ScoreInformation {
     fn from((vulnerability_id, cvss): (String, v2_0::CvssV2)) -> Self {
-        // Use calculated_base_score() to compute the actual score from metrics
-        let base_score = cvss.calculated_base_score().unwrap_or(0.0);
+        let base_score = cvss
```
rh-jfuller, line 47: I think as we reparse vector str and recompute score ...  if cvss-rs crate's computation differs slightly from the CNA's baseScore maybe we silently overwrite CNA-provided scores ? 

maybe a comment to explain this possibility or hint to check cvss-rs crate ?

etc/test-data/cve/CVE-2025-6965-02z-dump.sql
rh-jfuller, line 1: with these large sql dumps do we document how they were generated anywhere ?

rh-jfuller (approved):
great test cases, added a few comments generally LGTM (apart from failing CI which obvs needs addressing)

PR #2543, opened 2026-07-24
feat(notification): add change_log infrastructure and WebSocket endpoint

ctron:
## Summary

- Add `change_log` table with PostgreSQL `LISTEN/NOTIFY` trigger for real-time change event streaming
- Introduce `ChangeListener` (single-consumer, poll+notify) and `ChangeBroadcaster` (fan-out via `tokio::sync::broadcast`) in `common/src/db/change.rs`
- Add WebSocket endpoint at `GET /api/v3/notifications` with:
  - Cursor-based gap-free reconnection (`?after=<cursor>`)
  - `?token=` query parameter auth for browser clients
  - Per-entity permission filtering (`read.sbom` / `read.advisory`)
  - Connection message with current cursor on first connect
  - 30s heartbeat ping
- Record change events in the ingestor on advisory/SBOM ingestion
- Configurable retention via `--change-log-retention` / `TRUSTD_CHANGE_LOG_RETENTION` (humantime, default `1d`)

## Test plan

- [ ] `cargo check` — workspace compiles
- [ ] `cargo clippy --workspace` — no warnings
- [ ] `cargo test -p trustify-module-notification` — 17 tests (12 unit, 5 integration)
- [ ] Manual: connect via `websocat`, ingest a document, verify event arrives
- [ ] Manual: reconnect with `?after=<cursor>`, verify backfill

## Summary by Sourcery

Enable gap-free real-time advisory and SBOM change notifications over WebSockets.

New Features:
- Add a WebSocket notifications endpoint that streams advisory and SBOM change events with authentication, permission filtering, cursor-based replay, and heartbeat support.

Enhancements:
- Introduce shared change event listening and broadcasting infrastructure with PostgreSQL notification and polling support.
- Record advisory and SBOM ingestion events for real-time notification delivery.
- Support configurable change-log retention and listener timing through command-line and environment settings.

Build:
- Add the notification module and required WebSocket/PostgreSQL dependencies to the workspace.

Documentation:
- Document the notification WebSocket protocol, authentication options, reconnection behavior, and client examples.

Tests:
- Add unit and integration coverage for token handling, permission filtering, change-log operations, event serialization, ingestion events, and retention behavior.

Chores:
- Add the change-log database migration, trigger, and cleanup support.

### DIFF
--- a/common/src/db/change.rs
+++ b/common/src/db/change.rs
@@ -93,6 +93,49 @@ pub async fn record_change(
     Ok(())
 }
 
+async fn fetch_latest_id(pool: &sqlx::PgPool) -> Uuid {
+    let result: Result<Option<(Uuid,)>, _> =
+        sqlx::query_as("SELECT id FROM change_log ORDER BY id DESC LIMIT 1")
+            .fetch_optional(pool)
+            .await;
+    match result {
+        Ok(Some((id,))) => id,
+        Ok(None) => Uuid::nil(),
+        Err(err) => {
+            tracing::warn!(%err, "failed to fetch latest change_log id");
+            Uuid::nil()
+        }
+    }
+}
+
+async fn fetch_entries_after(
+    pool: &sqlx::PgPool,
+    cursor: &Uuid,
+) -> Result<Vec<ChangeEntry>, anyhow::Error> {
+    let rows: Vec<(Uuid, String, Option<Uuid>, String)> = sqlx::query_as(
+        "SELECT id, entity_type, entity_id, operation FROM change_log WHERE id > $1 ORDER BY id",
+    )
+    .bind(cursor)
+    .fetch_all(pool)
+    .await?;
+
+    let entries = rows
+        .into_iter()
+        .filter_map(|(cursor, r#type, id, operation)| {
+            let r#type = ChangeEntity::from_str(&r#type)?;
+            let operation = ChangeOperation::from_str(&operation)?;
+            Some(ChangeEntry {
+                cursor,
+                r#type,
+                id,
+                operation,
+            })
+        })
+        .collect();
+
+    Ok(entries)
+}
+
 /// Watches the change_log table via PostgreSQL LISTEN/NOTIFY with
 /// a periodic polling fallback. All sqlx types are encapsulated —
 /// callers only interact through the public API.
@@ -217,47 +260,12 @@ impl ChangeListener {
         }
     }
 
-    /// Fetches the latest change_log ID for cursor initialization.
     async fn fetch_max_id(&self) -> Uuid {
-        let result: Result<Option<(Uuid,)>, _> =
-            sqlx::query_as("SELECT id FROM change_log ORDER BY id DESC LIMIT 1")
-                .fetch_optional(&self.pool)
-                .await;
-
-        match result {
-            Ok(Some((id,))) => id,
-            Ok(None) => Uuid::nil(),
-            Err(err) => {
-                tracing::warn!(%err, "failed to fetch latest change_log id, starting from zero");
-                Uuid::nil()
-            }
-        }
+        fetch_latest_id(&self.pool).await
     }
 
-    /// Fetches all change_log entries with id > cursor.
     async fn fetch_after(&self, cursor: &Uuid) -> Result<Vec<ChangeEntry>, anyhow::Error> {
-        let rows: Vec<(Uuid, String, Option<Uuid>, String)> = sqlx::query_as(
-            "SELECT id, entity_type, entity_id, operation FROM change_log WHERE id > $1 ORDER BY id",
-        )
-        .bind(cursor)
-        .fetch_all(&self.pool)
-        .await?;
-
-        let entries = rows
-            .into_iter()
-            .filter_map(|(cursor, r#type, id, operation)| {
-                let r#type = ChangeEntity::from_str(&r#type)?;
-                let operation = ChangeOperation::from_str(&operation)?;
-                Some(ChangeEntry {
-                    cursor,
-                    r#type,
-                    id,
-                    operation,
-                })
-            })
-            .collect();
-
-        Ok(entries)
+        fetch_entries_after(&self.pool, cursor).await
     }
 
     /// Deletes change_log entries older than the retention period.
@@ -324,44 +332,11 @@ impl ChangeBroadcaster {
         self.tx.subscribe()
     }
 
-    /// Returns the latest event cursor, or `Uuid::nil()` if the change_log is empty.
     pub async fn fetch_latest_cursor(&self) -> Uuid {
-        let result: Result<Option<(Uuid,)>, _> =
-            sqlx::query_as("SELECT id FROM change_log ORDER BY id DESC LIMIT 1")
-                .fetch_optional(&self.pool)
-                .await;
-        match result {
-            Ok(Some((id,))) => id,
-            Ok(None) => Uuid::nil(),
-            Err(err) => {
-                tracing::warn!(%err, "failed to fetch latest change_log cursor");
-                Uuid::nil()
-            }
-        }
+        fetch_latest_id(&self.pool).await
     }
 
     pub async fn fetch_after(&self, cursor: &Uuid) -> Result<Vec<ChangeEntry>, anyhow::Error> {
-        let rows: Vec<(Uuid, String, Option<Uuid>, String)> = sqlx::query_as(
-            "SELECT id, entity_type, entity_id, operation FROM change_log WHERE id > $1 ORDER BY id",
-        )
-        .bind(cursor)
-        .fetch_all(&self.pool)
-        .await?;
-
-        let entries = rows
-            .into_iter()
-            .filter_map(|(cursor, r#type, id, operation)| {
-                let r#type = ChangeEntity::from_str(&r#type)?;
-                let operation = ChangeOperation::from_str(&operation)?;
-                Some(ChangeEntry {
-                    cursor,
-                    r#type,
-                    id,
-                    operation,
-                })
-            })
-            .collect();
-
-        Ok(entries)
+        fetch_entries_after(&self.pool, cursor).await
     }
 }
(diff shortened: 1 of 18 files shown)


ctron:
@sourcery-ai review

modules/notification/src/inject_token.rs
```diff
@@ -0,0 +1,57 @@
...
+
+        let fut = self.service.call(req);
+        Box::pin(fut)
+    }
+}
+
+pub(crate) fn extract_token(query: &str) -> Option<&str> {
+    query.split('&').find_map(|pair| {
+        let (key, value) = pair.split_once('=')?;
+        (key == "token").then_some(value)
```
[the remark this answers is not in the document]
ctron: Done — `extract_token` now uses `url::form_urlencoded::parse` and returns an owned `String`, so percent-encoded tokens are decoded correctly. Also added tests for percent-encoded values and empty/missing token values.

modules/notification/src/test.rs
```diff
@@ -0,0 +1,336 @@
...
+// -- Group C: QueryTokenInjector middleware ----------------------------------
+
+async fn echo_auth(req: HttpRequest) -> HttpResponse {
+    match req.headers().get("Authorization") {
+        Some(val) => HttpResponse::Ok().body(val.to_str().unwrap_or("bad").to_string()),
+        None => HttpResponse::NoContent().finish(),
+    }
+}
+
+#[test(actix_web::test)]
```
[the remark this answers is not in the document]
ctron: Done — added tests for `/test?token`, `/test?token=&foo=bar`, and `/test?token=` to cover malformed/partial query strings. The middleware correctly ignores all of these without panicking.

modules/notification/src/endpoints.rs
```diff
@@ -0,0 +1,190 @@
...
+                        if !is_allowed(&entry, can_read_sbom, can_read_advisory) {
+                            continue;
+                        }
+                        let json = serde_json::to_string(&entry)?;
+                        if session.text(json).await.is_err() {
+                            break;
+                        }
+                    }
+                    Err(broadcast::error::RecvError::Lagged(n)) => {
+                        tracing::warn!(n, "WebSocket notification client lagged");
```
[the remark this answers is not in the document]
ctron: Good catch — fixed to `tracing::warn!(lagged = n, ...)` so the actual count is recorded.

common/src/db/change.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    /// Deletes change_log entries older than the retention period.
+    async fn cleanup(&self) {
+        let retention_secs = self.retention.as_secs() as i64;
+        let result = sqlx::query(&format!(
+            "DELETE FROM change_log WHERE created_at < NOW() - INTERVAL '{retention_secs} seconds'"
+        ))
+        .execute(&self.pool)
+        .await;
```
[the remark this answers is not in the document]
ctron: Done — the cleanup query now uses a static SQL string with `$1` bind parameter instead of `format!`.

common/src/db/change.rs
```diff
@@ -0,0 +1,342 @@
...
+pub async fn record_change(
+    conn: &impl ConnectionTrait,
+    entity_type: ChangeEntity,
+    entity_id: Option<Uuid>,
+    operation: ChangeOperation,
+) -> Result<(), DbErr> {
+    let id = Uuid::now_v7();
+    conn.execute(Statement::from_sql_and_values(
+        DbBackend::Postgres,
+        "INSERT INTO change_log (id, entity_type, entity_id, operation) VALUES ($1, $2, $3, $4)",
```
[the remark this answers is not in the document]
ctron: Done — `record_change` now uses `conn.get_database_backend()` instead of hard-coding `DbBackend::Postgres`.

rh-jfuller (commented):
needs to be rebased

server/src/profile/api.rs
```diff
@@ -419,6 +432,7 @@ pub(crate) fn configure(svc: &mut utoipa_actix_web::service_config::ServiceConfi
 
     svc.configure(|svc| {
         endpoints::configure(svc, auth.clone(), read_only);
+        trustify_module_notification::endpoints::configure(svc, broadcaster, auth.clone());
```
rh-jfuller, line 656: I can see why having this separate from API is intentional - maybe explain rationale why it is

modules/notification/src/test.rs
```diff
@@ -0,0 +1,825 @@
...
+    assert_eq!(
+        crate::inject_token::extract_token("token=abc123").as_deref(),
+        Some("abc123")
+    );
+}
+
+#[test]
+fn extract_token_with_other_params() {
+    assert_eq!(
+        crate::inject_token::extract_token("after=xxx&token=jwt.val&foo=bar").as_deref(),
```
rh-jfuller, line 34: should we test with smuggling token in **Sec-WebSocket-Protocol** header ?

rh-jfuller (approved):
I see this area building up incrementally so no need to expand scope to be 100% comprehensive just yet - added a few comments otherwise LGTM. CI tests need to pass ;)

common/src/db/change.rs
```diff
@@ -0,0 +1,342 @@
+use sea_orm::{ConnectionTrait, DbBackend, DbErr, Statement};
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::sync::broadcast;
+use uuid::Uuid;
+
+const CHANNEL: &str = "trustify_changes";
+const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(30);
```
rh-jfuller, line 8: some point these will prob turn into env vars ;)

common/src/db/change.rs
```diff
@@ -0,0 +1,342 @@
...
+        self
+    }
+
+    /// Sets the retention period for cleaning old entries.
+    pub fn with_retention(mut self, retention: Duration) -> Self {
+        self.retention = retention;
+        self
+    }
+
+    /// Runs forever, calling `on_change` with batches of new entries.
```
rh-jfuller, line 204: do we ever need to provide a clean break command ?

common/src/db/change.rs
```diff
@@ -0,0 +1,342 @@
...
+
+/// The kind of entity that changed.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ChangeEntity {
+    Advisory,
+    Sbom,
+}
+
+impl ChangeEntity {
```
rh-jfuller, line 19: strum ?

rh-jfuller (commented):
a few pg related (mostly operational) comments:

*  `PgListener` holds session then issues LISTEN trustify_changes which might not work with things like pgbouncer ?
* we are 'paying the cost' for NOTIFY payload and throwing it away? maybe a future thing ?
* pg delivers async notifications in commit order which is ordering the id-cursor sweep lacks (v7 ids are assigned at insert time). If we consumee payload id and delivered that specific row, you'd inherit commit-ordering for free ... I think ?
* I think on bulk ingest ... we might get quite a 'storm' ... we should test when bulk ingesting stuff

modules/notification/src/test.rs
ctron: Good point — `Sec-WebSocket-Protocol` is a known vector for passing auth in browser WS connections. The current `QueryTokenInjector` only handles query string tokens. Supporting `Sec-WebSocket-Protocol` as a token transport would need new middleware — tracked as a follow-up.

common/src/db/change.rs
ctron: Agreed — the poll interval and cleanup interval could become configurable. `change_log_retention` is already a CLI/env arg; these two are good candidates for a follow-up.

common/src/db/change.rs
ctron: Not right now — the listener reconnects automatically and `cleanup()` handles retention. A graceful shutdown hook or admin endpoint to flush/reset the listener could be useful operationally; worth considering as the feature matures.

common/src/db/change.rs
ctron: Considered it — the manual `from_str` returns `Option` (to silently skip unknown values in `filter_map` during DB reads), which doesn't match strum's `FromStr` trait (`Result`). The manual impl is small (6 lines per enum) and keeps the silent-skip semantics explicit. Could revisit if more variants are added.

ctron:
Addressed the NOTIFY payload point — the trigger now sends the full row as JSON:

```sql
PERFORM pg_notify('trustify_changes', json_build_object(
    'id', NEW.id,
    'entity_type', NEW.entity_type,
    'entity_id', NEW.entity_id,
    'operation', NEW.operation
)::text);
```

The listener parses the payload directly into a `ChangeEntry`, skipping the follow-up `SELECT` query. Falls back to a sweep on parse failure (e.g., unknown entity type) or on the poll-interval path.

Also addressed in this push:
- `Sec-WebSocket-Protocol: access_token, <jwt>` token extraction (as a fallback to `?token=`)
- Poll interval and cleanup interval are now configurable via `TRUSTD_CHANGE_LOG_POLL_INTERVAL` and `TRUSTD_CHANGE_LOG_CLEANUP_INTERVAL`
- Review fixes from sourcery-ai (parameterized cleanup query, `conn.get_database_backend()`, correct `tracing::warn!` lag field)
- Comment explaining why the notification endpoint lives outside `/api` scope
- Redundant borrow clippy fixes

Remaining points for follow-up:
- **PgListener + pgbouncer**: worth testing — LISTEN requires a real PG session, not transaction-pooled connections
- **Bulk ingest storm**: should test with large batch imports to see if the broadcast channel (1024 capacity) handles the volume

ctron:
Re: commit-ordering — good observation. With the NOTIFY payload change (`f70b98a`), the happy path now delivers entries in PostgreSQL's commit order, which is the correct causal order. UUIDv7 IDs are assigned at insert time (inside the transaction), but `pg_notify` fires after commit, so two concurrent transactions that insert in one order but commit in another are delivered in commit order.

The sweep fallback (poll interval, reconnection backfill) still uses `ORDER BY id` (insert-time order), but that path is for catch-up completeness — it just needs to return everything after the cursor, not maintain strict causal ordering.

PR #2544, opened 2026-07-28
chore: address PR review comments from #2536

ctron:
## Summary
- Renumber migration m0002270 → m0002280 to avoid collision with PR #2534
- Add comment explaining no-op `down()` in data-fix migration
- Document CVSS score preference order in `cvss.rs`
- Add generation provenance to SQL test dump files

Follow-up to #2536 — these review-response changes were pushed after the PR was merged.

## Test plan
- [ ] CI passes (no logic changes, only renaming + comments)
- [ ] Migration numbering does not collide with existing migrations

## Summary by Sourcery

Renumber a data-fix migration to avoid a numbering collision and clarify related behavior and test fixtures.

Enhancements:
- Renumber the vulnerability base score data-fix migration and wire it into the migrator under the new identifier.
- Clarify that the migration’s down() operation is a no-op because prior incorrect data cannot be reconstructed.
- Document the preferred CVSS score source when both a computed vector score and CNA-provided base_score are available.
- Annotate CVE SQL dump fixtures with their generation process and intended use as known-state migration test data.


rh-jfuller (approved):
LGTM

Strum355:
heads up, I changed the migration number in #2534 to m0002280 yesterday after seeing #2536, seeing as your PR landed on main first :slightly_smiling_face:

ctron:
Ok, reverted the rename and pushed the change. This should also fix the CI.

Issue #2545, opened 2026-07-28
Support known-exploited-vulnerability catalogs (CISA KEV) as an enrichment source


waldemar-kindler:
# Support known-exploited-vulnerability catalogs (CISA KEV) as an enrichment source

## Summary

Trustify currently answers "which of my SBOM components are affected by which vulnerabilities", but offers no signal for the natural follow-up: "which of these are actually being exploited in the wild?" The [CISA Known Exploited Vulnerabilities (KEV) catalog](https://www.cisa.gov/known-exploited-vulnerabilities-catalog) is the authoritative, curated source for that signal (~1,300 CVEs, each with confirmed active exploitation). Ingesting it would let users prioritize the handful of findings that matter most, and supports concrete compliance drivers:

- **BOD 22-01** (US federal): mandated remediation deadlines for KEV-listed CVEs (`dueDate` is part of the feed).
- **FDA premarket cybersecurity guidance** (medical devices): KEV-listed vulnerabilities "should be designed out of the device".
- **EU CRA** reporting on actively exploited vulnerabilities.

## Why KEV is not redundant with CVSS/EPSS

Empirical analysis in DependencyTrack/dependency-track#2267 showed EPSS scores frequently *decrease* over time (e.g. CVE-2020-13158 dropped 0.97 → 0.018 in two days), so probabilistic scores cannot substitute for the durable "was exploited in the wild" fact. CVSS measures theoretical severity, not exploitation. KEV is a trailing indicator that complements both.

## Design considerations (informed by prior art)

Lessons from [DependencyTrack #2267](https://github.com/DependencyTrack/dependency-track/issues/2267) (open since 2022) and [google/osv-scalibr #2194](https://github.com/google/osv-scalibr/issues/2194):

1. **Don't be CISA-specific.** Model this as a generic "known exploited" enrichment with KEV as the first source. Realistic additional sources: [VulnCheck KEV](https://vulncheck.com/kev) (free community feed, superset of CISA KEV, adds ransomware-campaign usage and exploit references; requires prominent attribution), a future ENISA/EU vulnerability database under CRA, [inthewild.io](https://inthewild.io/feed).
[...]

PR #2553, opened 2026-07-30
feat: support CISA KEV catalog ingestion and expose known-exploited data

waldemar-kindler:
**Motivation:**
Trustify ingests applicability data (VEX/CSAF product status: affected / not affected / fixed) and severity (CVSS), but carries no curated signal for whether a vulnerability is being **actively exploited in the wild**. The CISA KEV catalog is the authoritative source for that signal and complements the existing data: applicability narrows findings to what affects you; exploitation status ranks them by real-world urgency (and carries BOD 22-01 remediation deadlines). See #2545 for the full design discussion.

Relation to Exploit Intelligence (#2534): EI answers "is this CVE exploitable *in this specific product*?" via an external analysis service, per SBOM+CVE job. KEV answers "is this CVE exploited in the wild, globally?" from a public catalog — no external service, no product context, no auth. The signals are complementary.

**Changes**

Ingestor
- New `known_exploited_vulnerability` table (`m0002300`), keyed by `(source, cve_id)` so additional catalog providers (e.g. VulnCheck KEV) can be supported later. Intentionally no FK to `vulnerability`: catalog entries and CVE documents can be ingested in any order; consumers join on `vulnerability.id`.
- `Format::CisaKev` with JSON content detection (`catalogVersion` + `vulnerabilities`), wired through the `DocumentDetector`.
- `KevLoader` with full-sync semantics: each load atomically replaces the source's entry set, so entries removed from KEV (this has happened) are removed from the database. Unparsable dates are dropped with a warning instead of failing the catalog.

Importer
- New `kev` importer kind fetching the KEV JSON feed, skipping unchanged content via the `Last-Modified` header (same pattern as the CWE catalog importer), with a 64 MiB download cap.
- Disabled-by-default sample importer, daily period.

API
- Vulnerability details now include `known_exploited` entries (source, dateAdded, dueDate, requiredAction, knownRansomwareCampaignUse).

**Validation**: exercised end-to-end against the live CISA feed (1,656 entries): import, `Last-Modified` skip on re-run, and the full-sync semantics confirmed against a real-world removal (CISA has since dropped CVE-2025-6965 from the catalog, and the entry disappears on reload as intended).

**Open questions for reviewers:**
1. Schema: `known_exploited_vulnerability` keyed by `(source, cve_id)` — source-qualified for future providers. Reasonable, or would you prefer a CISA-only shape for now?
2. No FK to `vulnerability` (ingestion-order independence, same as `vulnerability.cwes` handling) — acceptable?
3. Full-sync (delete + reinsert per source, in one transaction) instead of upsert-with-tombstones — acceptable for a ~1,300-row catalog?
4. Naming: `known_exploited` on the API vs. the new Exploit Intelligence module — happy to rename if this is too confusable.

**Non-goals (per #2545)**
- Risk-score adjustments, additional providers, dedicated UI views, and list-endpoint filtering by KEV membership are follow-ups. Happy to add list filtering in this PR if preferred — I kept it out to keep the data model reviewable first.

## Summary by Sourcery

Add CISA KEV ingestion and expose source-qualified known-exploitation records across the API.

New Features:
- Add ingestion of CISA Known Exploited Vulnerabilities catalog data with source-qualified exploit records.
- Expose exploit records through vulnerability details and dedicated paginated list and retrieval APIs.
- Support configurable KEV catalog importers with daily scheduling and unchanged-feed detection.

Bug Fixes:
- Ensure catalog resynchronization removes withdrawn entries and updates revised entries while preserving existing data when an empty catalog is received.
- Parse date query values using date types for correct database filtering.

Enhancements:
- Represent provider-specific exploitation data in extensible metadata with stable source-qualified identifiers and searchable fields.

Documentation:
- Document configuring a CISA KEV importer and accessing exploit records through the API.

Tests:
- Add unit, integration, API, and end-to-end coverage for KEV detection, ingestion, synchronization, querying, pagination, and vulnerability exposure.

Chores:
- Add the exploit data model, database migration, and API schema definitions.


rh-jfuller:
before we start adding more entities to trustify (which btw enriching with CISA data is great !)  - we need to first define an overarching strategy to associating metadata generically in trustify ... otherwise we are going to experience 'sprawl' with concomitant maintenance ... I started scratching out a few thoughts here https://gist.github.com/rh-jfuller/845c1b17fefb79c9dc3dc3df194ee6d9 - I would close this PR and lets have a conversation. The idea is that we would provide some enrichment internals in trustify to be able to develop such things.

waldemar-kindler:
> before we start adding more entities to trustify (which btw enriching with CISA data is great !) - we need to first define an overarching strategy to associating metadata generically in trustify ... otherwise we are going to experience 'sprawl' with concomitant maintenance ... I started scratching out a few thoughts here https://gist.github.com/rh-jfuller/845c1b17fefb79c9dc3dc3df194ee6d9 - I would close this PR and lets have a conversation. The idea is that we would provide some enrichment internals in trustify to be able to develop such things.

Thanks for sharing your thoughts! Generally I follow your sentiments. Having an entity / data field sprawl was also one of my concerns. Just a little push back from my side. In my mental model KEV is closer to the NVD and CWE than to Exploit Intelligence, EPSS or Conforma.

The distinction I'd draw is feed vs. service call. EI, Conforma and (to a lesser degree) EPSS are request/response: they need context (which SBOM, which component), they need auth, they're per-job, and they benefit from exactly the shared plumbing you describe: Job queue, HTTP client with retry/backoff, error classification. KEV has none of that shape. It's a single public static JSON document with a stable schema . Pretty much like the CWE catalog importer.

On process: I'm glad to have the conversation, and I'll convert this to whatever shape you land on. One request: Could we keep it open as a draft rather than closed? It's working end-to-end so I think it's useful as a concrete reference for what the data looks like while we discuss. If you'd rather have it closed, that's fine too. I'll reopen once there's a direction.

rh-jfuller:
@waldemar-kindler no worries, lets keep this PR open ... and your comments are **spot on** (will incorporate to enrichment doc working on)  ... generally we want to identify primary entities and give them an importer/exporter treatment - everything else is enrichment. I can be convinced that your approach (eg. exploits as primary) works but want to have a few more conversations with others on the team to get alignment.

waldemar-kindler:
> @waldemar-kindler no worries, lets keep this PR open ... and your comments are **spot on** (will incorporate to enrichment doc working on) ... generally we want to identify primary entities and give them an importer/exporter treatment - everything else is enrichment. I can be convinced that your approach (eg. exploits as primary) works but want to have a few more conversations with others on the team to get alignment.

Awesome! Let me know if I can support somehow!

rh-jfuller:
ok lets move ahead with this PR ... will review in more detail once you are done

modules/fundamental/src/vulnerability/model/details/mod.rs
```diff
@@ -24,6 +28,10 @@ pub struct VulnerabilityDetails {
     /// Advisories addressing this vulnerability, if any.
     pub advisories: Vec<VulnerabilityAdvisorySummary>,
 
+    /// Known-exploited-vulnerability catalog entries (e.g. CISA KEV)
+    /// referring to this vulnerability, if any.
+    pub known_exploited: Vec<KnownExploitation>,
```
rh-jfuller: I have a pref for a more generic exploit entity - where we might have an attribute of known (with attribution like CISA) ... if we are to add a new top level entity we should be able to represent exploits from more then just CISA - the importer/upload can map to specifics.

rh-jfuller (commented):
loader, importer, format detection, and test infra are solid - most of that code carries over if we consider a more generic exploit entity. The change I would propose is to  
 
* flatten CISA specific columns into JSONB metadata (stops specific things like due_date, required_action, known_ransomware_campaign_use leaking into core data model)
* rename table/entity to exploit
* rename the API field from known_exploited to just exploits. 
 
The KevLoader still exists though it would just map out CISA fields into the JSONB instead of into dedicated columns.

modules/fundamental/src/vulnerability/model/mod.rs
```diff
@@ -83,6 +85,67 @@ impl BaseScore {
...
+
+    /// The remediation action required by the catalog, if any.
+    #[schema(required)]
+    pub required_action: Option<String>,
+
+    /// Whether the vulnerability is known to be used in ransomware campaigns.
+    /// Deliberately a free-form string, not an enum: the set of values
+    /// ("Known" / "Unknown" today) is owned by the upstream catalog.
+    #[schema(required)]
+    pub known_ransomware_campaign_use: Option<String>,
```
rh-jfuller: this is a good example of leaking highly specific CISA into core data model - sometihng we prob want to avoid

rh-jfuller: I took the liberty of thinking how we might have a more generic entity

```
CREATE TABLE exploit (
    id              UUID PRIMARY KEY DEFAULT gen_random_uuid(),
    -- Which catalog/source reported this
    source          TEXT NOT NULL,           -- "cisa-kev", "vulncheck", "epss", ...
    -- The CVE it refers to (no FK, same rationale as CWEs)
    cve_id          TEXT NOT NULL,
    -- When the source first reported exploitation
    date_reported   DATE,
    -- Provider-specific structured data lives here
    metadata        JSONB NOT NULL DEFAULT '{}',
    -- Dedup key
    UNIQUE (source, cve_id)
);

CREATE INDEX idx_exploit_cve_id ON exploit (cve_id);
CREATE INDEX idx_exploit_source ON exploit (source);
```

then what goes in metadata

For CISA KEV:
```
{
  "vendor_project": "Apache",
  "product": "Log4j2",
  "vulnerability_name": "Apache Log4j2 Remote Code Execution",
  "short_description": "...",
  "required_action": "Apply updates...",
  "due_date": "2021-12-24",
  "known_ransomware_campaign_use": "Known",
  "notes": "https://nvd.nist.gov/vuln/detail/CVE-2021-44228",
  "cwes": ["CWE-917", "CWE-20"]
}
```
For a hypothetical VulnCheck KEV:
```
{
  "exploit_maturity": "weaponized",
  "references": ["https://..."],
  "tags": ["ransomware", "apt"]
}
```

For EPSS :
```
{
  "score": 0.97,
  "percentile": 0.99
}
```

thoughts ?

rh-jfuller:
@waldemar-kindler have you considered adding GET /api/v3/exploit, GET /api/v3/exploit/{id} ?

entity/src/known_exploited_vulnerability.rs
```diff
@@ -0,0 +1,32 @@
...
+    pub date_added: Option<Date>,
+    pub due_date: Option<Date>,
+    /// Deliberately a free-form string, not an enum: the set of values
+    /// ("Known" / "Unknown" today) is owned by the upstream catalog, and new
+    /// values must not break ingestion.
+    pub known_ransomware_campaign_use: Option<String>,
+    pub notes: Option<String>,
+    pub cwes: Option<Vec<String>>,
+}
+
```
rh-jfuller: if we go with a more generic exploit structure then we may want to think a bit about cascade deletes (eg. if a vuln is deleted) and avoid orphaned exploit rows ...

modules/ingestor/src/service/kev/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+                vulnerability_name: Set(entry.vulnerability_name),
+                short_description: Set(entry.short_description),
+                required_action: Set(entry.required_action),
+                known_ransomware_campaign_use: Set(entry.known_ransomware_campaign_use),
+                notes: Set(entry.notes),
+                cwes: Set(normalize(entry.cwes)),
+            })
+            .collect::<Vec<_>>();
+
+        // The catalog is authoritative for its source: entries can be removed
```
rh-jfuller: check out CONVENTIONS.md - for hints on how batch add() + create() with ON CONFLICT DO NOTHING pattern for new code.

migration/src/m0002300_create_known_exploited_vulnerability.rs
```diff
@@ -0,0 +1,96 @@
...
+                    .col(KnownExploitedVulnerability::CveId)
+                    .to_owned(),
+            )
+            .await?;
+
+        Ok(())
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
```
rh-jfuller: got to also drop any indexes created here

entity/src/known_exploited_vulnerability.rs
```diff
@@ -0,0 +1,32 @@
...
+    pub short_description: Option<String>,
+    pub required_action: Option<String>,
+    pub date_added: Option<Date>,
+    pub due_date: Option<Date>,
+    /// Deliberately a free-form string, not an enum: the set of values
+    /// ("Known" / "Unknown" today) is owned by the upstream catalog, and new
+    /// values must not break ingestion.
+    pub known_ransomware_campaign_use: Option<String>,
+    pub notes: Option<String>,
+    pub cwes: Option<Vec<String>>,
```
rh-jfuller: uneasy about a decorative CWE eg. we might want to link back to existing CWE entity ... but that opens up a lot of scope ... maybe this (be it here or in metadata) is the way to go for now.

waldemar-kindler:
@rh-jfuller Implemented your proposal rather than just replying, so there's something concrete to argue with: generic `exploit` entity, CISA fields in `metadata` JSONB, `known_exploited` → `exploits`, plus the two endpoints. Importer, walker, detection and test infra carried over unchanged. Rebased on main.

Four deviations from your sketch, each with a test:

**Full sync, not `ON CONFLICT DO NOTHING`.** CISA both withdraws entries (it dropped CVE-2025-6965 while this PR was open) and revises `dueDate`/`requiredAction` on existing ones. `DO NOTHING` expresses neither — it keeps stale values indefinitely and never drops a withdrawn entry. `ExploitCreator` keeps the pattern's shape (`add()` + consuming `create()`, chunked inserts) but replaces the source's rows. It refuses an empty set so a truncated-but-parsable download can't wipe a source, and that guard sits in the creator rather than the loader so a second source inherits it. If you'd rather avoid rewriting a source each run: `DO UPDATE` + delete-not-in-set.

**UUIDv5 from `(source, cve_id)`** instead of `gen_random_uuid()` — CONVENTIONS.md lists it for content-derived keys, and it keeps `/exploit/{id}` stable across the daily resync.

**`remediation_due_date` stayed a column.** Everything else CISA-shaped went into `metadata`, which also dissolves the `cwes` question — decidable later, no migration. But BOD 22-01 deadlines are what people will filter and sort on, and JSONB dates are awkward for both.

**EPSS left out.** I've scoped `exploit` to "a source asserts this CVE is exploited, or that working exploit code exists". EPSS is a daily-rescored probability, closer to CVSS scoring than to a catalog entry. If the table takes scores *and* assertions, `metadata` loses any shared meaning — the sprawl you opened this thread with. Your call though; it's docs and intent, not schema.

**Orphaned rows:** no FK means no cascade to configure, and a row whose CVE isn't in the DB isn't an orphan — it's the catalog running ahead of advisory ingestion, and it resolves when the CVE lands. Rows for a source we've *stopped* importing are a real case, but that's importer-deletion. Which did you mean?

Two questions: the endpoints use `read.advisory` like the vulnerability ones — want a dedicated `read.exploit` instead? And `exploit` now sits beside `exploit_intelligence_job`; reads fine to me, but it was my open question 4 and the rename sharpens it.

Also fixed along the way, all tested: `catalogVersion` is now required, since detection keyed on it while the schema had it optional — a catalog could parse yet never be auto-detected. Dropped a redundant index on `source` (the `(source, cve_id)` composite serves it as a prefix, confirmed with EXPLAIN). Added connect/read timeouts to the walker: `reqwest::get` has none, and a blocked run keeps heartbeating, so the stale-job reaper treats it as alive and the importer never runs again. Capped the dropped-date warnings, which were one per bad date and so ~2600 near-identical strings for one fault. Plus a shared-layer bug: `date` columns were bound as text, so any comparison failed with `operator does not exist: date > text`.

The timeout and the continuation `unwrap_or_default()` apply to the other walkers too — I've left those alone rather than making KEV the odd one out, happy to file an issue. When a third catalog lands, I think what wants extracting is a *fetcher* (conditional GET, size cap, timeouts, continuation), not a walker.

Vulnerability-list filtering by exploit membership is still a follow-up, per the non-goals.

rh-jfuller:
@waldemar-kindler I think you judged this perfectly (really great work) ... would you consider changing this PR from WIP/draft and i will provide final review and see if we cannot land this week.

waldemar-kindler:
> @waldemar-kindler I think you judged this perfectly (really great work) ... would you consider changing this PR from WIP/draft and i will provide final review and see if we cannot land this week.

Done! Would love to land this soon. Thanks for your great support!

waldemar-kindler:
One question that came out of running this against the live feed, and it applies to the CWE catalog as much as to KEV.

`POST /api/v3/advisory` with a KEV catalog and no `?format=` returns:

```
400 {"error":"UnsupportedFormat","message":"... Unable to detect document format for hint advisory"}
```

The endpoint hints `Format::Advisory`, and `matches_hint` maps that to `CSAF | CVE | OSV`, so the `CisaKev` branch in `detect_format` is never reached. `?format=cisakev` works fine (201). `CweCatalog` has the same property, so both catalogs currently require an explicit format on manual upload — the importer is unaffected either way, since it passes a concrete format.

I tried adding `CisaKev` to the `Advisory` hint group to see what it costs. It works cleanly: identical `id` and `document_id` to the explicit-format upload, CSAF still detected as CSAF, 209/209 tests green. So it's mechanically a one-liner.

I've **not** included it, for two reasons:

1. It would make KEV the only auto-detecting catalog while CWE keeps the restriction — an asymmetry I'd rather not introduce unilaterally.
2. Neither catalog creates an advisory row. You already get `201` from `/api/v3/advisory` with nothing in the advisory list; widening the `Advisory` category to cover catalogs leans further into that.

So: should the advisory endpoint auto-detect catalog documents — both KEV and CWE — or neither? If neither, the thing worth fixing is the error, which doesn't hint that `?format=` is the answer.

Either way I've corrected the tests. `detect_cisa_kev` was using `Format::Unknown`, which no caller passes, so it was green while the real endpoint 400s — that false confidence is why I described this gap inaccurately earlier in the thread. It now detects through a concrete hint like the importer does, plus a test pinning the `Advisory` behaviour so changing it has to be deliberate.

common/src/db/query/columns.rs
```diff
@@ -307,7 +307,10 @@ fn parse(s: &str, ct: &ColumnType) -> Result<SimpleExpr, Error> {
                 s.to_owned(),
             ))))),
         ),
-        ColumnType::TimestampWithTimeZone => {
+        // `Date` shares this arm: it already tries a plain date before falling
```
rh-jfuller, line 310: thx for adding this code comment ... I do not think there are any impacts elsewhere

entity/src/exploit.rs
```diff
@@ -0,0 +1,34 @@
...
+use time::Date;
+
+/// An assertion by some source that a vulnerability is exploited in the wild,
+/// or that working exploit code for it exists.
+///
+/// The columns are the fields every source can be expected to carry; anything
+/// specific to one source lives in [`Model::metadata`].
+#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
+#[sea_orm(table_name = "exploit")]
+pub struct Model {
```
rh-jfuller, line 11: just a thought - if exploits are updated ... maybe we need to link to importer time (or include something like a 'last successfully synced' dt) ?

rh-jfuller (commented):
I left a few comments ... nothing blocking ... a PR with many commits we normally ask to squash them down, but I think its ok to leave asis (esp w/ some folks on holidays).  A big THANK YOU for the solid submission LGTM.

rh-jfuller (approved):
LGTM

rh-jfuller:
@waldemar-kindler ping me if you plan to make any more changes, otherwise I will merge today.

entity/src/exploit.rs
waldemar-kindler: Too be honest I don't think that its worth it. Information already exists in `/api/v3/importer/<name>`. It gives lastSuccess and lastRun, and the report gives startDate/endDate/numberOfItems per run

waldemar-kindler:
> @waldemar-kindler ping me if you plan to make any more changes, otherwise I will merge today.

No more changes from my side.

rh-jfuller:
@waldemar-kindler  need to fix a dep (unrelated) before this merge can go in first (https://github.com/guacsec/trustify/pull/2583)

PR #2556, opened 2026-07-31
feat: enable clippy::absolute_paths lint to enforce import style

mrizzi:
## Summary

Enable the `clippy::absolute_paths` restriction lint with `absolute-paths-max-segments = 3`
to flag fully-qualified paths with 4+ segments at build time. This enforces import style
deterministically — no reviewer vigilance needed for the common case.

### Why `max-segments = 3` (not 2)?

The lint does not distinguish internal from external crate paths. With `max-segments = 2`,
every 3-segment external crate path (e.g. `std::fmt::Result`, `tokio::sync::Mutex`,
`cpe::component::Component`) would be flagged, requiring a 30+ crate allowlist
(`absolute-paths-allowed-crates`) that must be maintained as dependencies evolve.

Three configurations were tested against the current codebase:

| Configuration | Violations | Files | Allowlist maintenance |
|---|---|---|---|
| `max-segments = 2`, no allowlist | 526 | 165 | N/A — too many false positives |
| `max-segments = 2`, 30-crate allowlist | 196 | 63 | Ongoing — every new dependency with 3-segment paths must be added |
| **`max-segments = 3`, no allowlist** | **70** | **35** | **None** |

`max-segments = 3` was chosen because it:
1. **Catches the original problem** — 4+ segment paths like `crate::graph::db_context::parse_status` are flagged at build time
2. **Requires zero allowlist maintenance** — 3-segment external paths pass automatically
3. **Preserves build safety** — all 70 existing violations are fixed in this PR

**Trade-off:** 3-segment paths (e.g. `std::fmt::Result`, `entity::advisory_vulnerability::Model`)
are not caught. These are covered by a documented reviewer convention (see CONVENTIONS.md changes).

The threshold can be tightened to 2 later if the team decides to take on the allowlist burden —
going from 3 → 2 is additive work, not a breaking change.

### Changes

- `.clippy.toml` — add `absolute-paths-max-segments = 3`
- `xtask/src/precommit.rs` — add `-W clippy::absolute_paths` to clippy invocation
- `CONVENTIONS.md` — document lint (goal 1) and qualified-names convention (goal 2)
- 35 source files — fix existing 4+ segment violations by replacing with `use` imports

Implements [TC-4513](https://redhat.atlassian.net/browse/TC-4513)

[TC-4513]: https://redhat.atlassian.net/browse/TC-4513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Enforce consistent Rust import conventions by linting and removing deeply qualified absolute paths throughout the codebase.

Enhancements:
- Enforce a consistent Rust import style by enabling Clippy's absolute-path lint for paths exceeding three segments.
- Replace existing deeply qualified paths across the codebase with imports while preserving qualified naming for common or ambiguous types through documented conventions.

Build:
- Configure the maximum allowed absolute path length in Clippy.

CI:
- Apply the absolute-path lint during the precommit Clippy checks.

Documentation:
- Document the absolute-path lint threshold and conventions for qualified names.

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,2 +1,3 @@
 allow-unwrap-in-tests = true
 allow-expect-in-tests = true
+absolute-paths-max-segments = 3
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -36,6 +36,44 @@ use sea_orm::{ConnectionTrait, TransactionTrait};
 This is a manual convention — `rustfmt`'s `imports_granularity = "Crate"` option is not
 available on the stable channel. Reviewers should flag un-nested imports during code review.
 
+### Import Style — Absolute Paths
+
+Fully-qualified paths with 4+ segments are enforced by the `clippy::absolute_paths` lint
+(configured in `.clippy.toml` with `absolute-paths-max-segments = 3`). Paths exceeding the
+threshold must be replaced with `use` imports:
+
+```rust
+// Good — imported
+use crate::graph::db_context::parse_status;
+parse_status(value)?;
+
+// Lint error — 4+ segment absolute path
+crate::graph::db_context::parse_status(value)?;
+```
+
+3-segment paths (e.g. `std::fmt::Result`, `entity::advisory_vulnerability::Model`) are not
+flagged by the lint and are acceptable.
+
+### Import Style — Qualified Names for Common Types
+
+Common or ambiguous names should stay qualified with a 2-segment path, even when there is
+no conflict in the current file. This keeps the code self-documenting and avoids confusion
+when reviewing:
+
+```rust
+// Good — qualified with module prefix
+let model: advisory_vulnerability::Model = ...;
+let result: std::fmt::Result = ...;
+let value = serde_json::from_value(...);
+
+// Avoid — bare name is ambiguous across modules
+let model: Model = ...;
+```
+
+Names that benefit from qualification include `Model`, `Entity`, `Column`, `ActiveModel`,
+`Relation`, `from_value`, `info`, and similar names that appear in many modules. This is a
+reviewer convention, not lint-enforced.
+
 ## Naming Conventions
 
 - Structs: PascalCase (`SbomService`, `AdvisoryService`, `SbomSummary`)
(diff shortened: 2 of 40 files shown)


rh-jfuller:
apologies by the time I had time to get to this there were conflicts

mrizzi:
> apologies by the time I had time to get to this there were conflicts

@rh-jfuller I've solved the conflicts.

rh-jfuller (approved):
LGTM

ctron:
I'm wondering what the list would be (having a threshold of 2). Maybe it's worth reducing that? I'd like to see `core::future::ready` to away. As well as `std::fmt::Result`. Unless of course if there's a name clash.

Otherwise it looks like a step in the right direction.

mrizzi:
> With `max-segments = 2`,
> every 3-segment external crate path (e.g. `std::fmt::Result`, `tokio::sync::Mutex`,
> `cpe::component::Component`) would be flagged, requiring a 30+ crate allowlist
> (`absolute-paths-allowed-crates`) that must be maintained as dependencies evolve.

I've tried the `max-segments = 2` approach and the result is reported in the PR's description above:

```
With max-segments = 2, every 3-segment external crate path (e.g. std::fmt::Result, tokio::sync::Mutex,
cpe::component::Component) would be flagged, requiring a 30+ crate allowlist (absolute-paths-allowed-crates) 
that must be maintained as dependencies evolve.
```

I'm going to merge this PR and we can tackle further discussions and followups with new tasks.

ctron:
> > With `max-segments = 2`,
> > every 3-segment external crate path (e.g. `std::fmt::Result`, `tokio::sync::Mutex`,
> > `cpe::component::Component`) would be flagged, requiring a 30+ crate allowlist
> > (`absolute-paths-allowed-crates`) that must be maintained as dependencies evolve.
> 
> I've tried the `max-segments = 2` approach and the result is reported in the PR's description above:
> 
> ```
> With max-segments = 2, every 3-segment external crate path (e.g. std::fmt::Result, tokio::sync::Mutex,
> cpe::component::Component) would be flagged, requiring a 30+ crate allowlist (absolute-paths-allowed-crates) 
> that must be maintained as dependencies evolve.
> ```
> 
> I'm going to merge this PR and we can tackle further discussions and followups with new tasks.

I'd like to see things like `tokio::sync::Mutex` to go away anyway. Same for `cpe::component::Component`.

PR #2557, opened 2026-07-31
chore(deps): update deps for 0.6.0-rc.1 (and main)

rh-jfuller:
Updated `packageurl` crate from 0.6 to 0.7. 

This is a far reaching change which corrects usage of pURL throughout trustify.

The largest impact was that many tests had to be corrected for percent encoding ... for example `/` should be percent encoded in `repository_url` qualifier **urlparam** value now should be `https:%2F%2Fmaven.repository.redhat.com%2Fga%2F` instead of `https://maven.repository.redhat.com/ga/`. The same idiom applies to other **urlparam** (like `download_url`).

Also in tests:
*  `:` should NOT be %3A encoded -- packageurl 0.7 only encodes / and , in qualifier values, not :. Fixed all %3A%2F%2F back to :%2F%2F.
* Split dual-use purl variables in tests -- Where the same purl string was used as both a query input AND output assertion (quarkus_component_by_purl, cdx_ancestor_of), split into _query (unencoded) and _expected (encoded) variables.

Lastly, had to fix `Purl::like()` (common/src/purl.rs:41-46) -- The in-memory filter was decoding the search pattern but comparing against self.to_string() which now (due to updated `packageurl`) produces %2F-encoded output, this was a latent minor bug that has been around for a long time. Fixed by decoding both sides before comparison.

## Summary by Sourcery

Build:
- Update the packageurl crate dependency from 0.6 to 0.7 in Cargo.toml and refresh Cargo.lock.

## Summary by Sourcery

Update package URL handling across the codebase to be compatible with packageurl 0.7, including fixing purl matching and adjusting tests for new encoding behavior.

Bug Fixes:
- Fix Purl::like() to compare decoded values consistently after packageurl encoding changes.

Enhancements:
- Align in-memory purl filtering and matching behavior with updated packageurl encoding semantics.

Build:
- Update packageurl crate dependency to 0.7 and refresh Cargo.lock.

Tests:
- Update analysis, fundamental, and ingestor tests to expect percent-encoded URL qualifiers and separate query vs expected purl values where needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -428,7 +428,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -439,7 +439,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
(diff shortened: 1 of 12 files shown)


mrizzi (approved):

rh-jfuller:
@sourceryai summary

PR #2565, opened 2026-08-10
fix: correctly serialize RFC3339 datetimes in EI API types

Strum355:
OpenAPI spec currently states that these date fields should be strings, but the implementation didnt match. Not currently used anywhere so the issue wasn't caught. 

Fixes [TC-5538](https://redhat.atlassian.net/browse/TC-5538)

[TC-5538]: https://redhat.atlassian.net/browse/TC-5538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Ensure exploit intelligence API timestamps are serialized as RFC3339 strings and update tests accordingly.

Bug Fixes:
- Correct RFC3339 serialization of created and updated timestamp fields in exploit intelligence job and component responses.

Tests:
- Adjust exploit intelligence job completion test to assert string-based RFC3339 timestamps and other serialized fields via generic JSON inspection.

### DIFF
--- a/modules/exploit-intelligence/src/endpoints/test.rs
+++ b/modules/exploit-intelligence/src/endpoints/test.rs
@@ -11,9 +11,10 @@ use actix_web::{
     web,
 };
 use sea_orm::{ActiveModelTrait, Set, TransactionTrait};
+use serde_json::Value;
 use test_context::test_context;
 use test_log::test;
-use time::OffsetDateTime;
+use time::{OffsetDateTime, format_description::well_known::Rfc3339};
 use trustify_common::{db, error::ErrorInformation, model::PaginatedResults};
 use trustify_entity::{
     exploit_intelligence_job::{self, ExploitIntelligenceFinding, ExploitIntelligenceJobStatus},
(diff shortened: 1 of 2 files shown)


modules/exploit-intelligence/src/endpoints/test.rs
```diff
@@ -147,13 +148,18 @@ async fn get_job_completed(ctx: &TrustifyContext) -> anyhow::Result<()> {
...
-    let body: ExploitIntelligenceJobDetails = test::read_body_json(resp).await;
-    assert_eq!(body.id, job_id);
-    assert_eq!(body.status, ExploitIntelligenceJobStatus::Completed);
-    assert_eq!(body.vulnerability_id, "CVE-2024-1234");
+    let body: Value = test::read_body_json(resp).await;
+    assert_eq!(body["id"], job_id.to_string());
+    assert_eq!(body["status"], "completed");
+    assert_eq!(body["vulnerability_id"], "CVE-2024-1234");
+    assert_eq!(body["finding"], "not_vulnerable");
+    assert_eq!(
```
rh-jfuller: using unwrap is fine in tests (as per CONVENTIONS.md) but .err() is a bit indirect ... maybe consider 

```
assert!(
    OffsetDateTime::parse(body["created"].as_str().unwrap(), &Rfc3339).is_ok(),
    "created should be valid RFC3339"
);
```

rh-jfuller (approved):
you might have to run  >cargo xtask precommit to refresh openapi defs ... otherwise LGTM

Strum355:
> you might have to run >cargo xtask precommit to refresh openapi defs ... otherwise LGTM

This was bringing the implementation in-line with what openapi was declaring. CI would catch precommit yielding a diff though, so we're all good there

PR #2566, opened 2026-08-11
fix: configurable oauth scope instead of hardcoded 'openid' scope

Strum355:
AWS Cognito rejects client_credential grant requests containing system-reserved scopes such as `openid`, therefore we can't force the scope in the general case. From previous testing, the `openid` scope was required for Keycloak auth with EI, hence the original introduction of the hard-coded `openid` scope. This PR introduces the option to set a scope if necessary, leaving it unset by default (some cognito & keycloak setups may require custom scopes to be requested that cannot be set as default scopes for whatever reasons).

From the [AWS docs for machine-to-machine authorization](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-define-resource-servers.html#cognito-user-pools-define-resource-servers-about-m2m) "You can authorize only custom scopes from resource servers in access tokens for client credentials grants."

Fixes [TC-5584](https://redhat.atlassian.net/browse/TC-5584)

## Summary by Sourcery

Make OAuth scopes configurable for OpenID client_credentials flows instead of hardcoding the `openid` scope, with sensible defaults and propagation through configuration and provider code.

New Features:
- Add CLI and environment configuration options to specify OAuth scope(s) for EI OIDC and generic OpenID token providers.
- Support multiple space-separated OAuth scopes for client_credentials token requests, while allowing no scope parameter to be sent by default.

Enhancements:
- Propagate the configured OAuth scope from argument structs into OpenID provider configuration and runtime, including token refresh and initial acquisition logic.

Tests:
- Extend OIDC argument and configuration tests to cover scope defaults and propagation into the resulting config.

[TC-5584]: https://redhat.atlassian.net/browse/TC-5584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/common/auth/src/client/provider/openid.rs
+++ b/common/auth/src/client/provider/openid.rs
@@ -50,6 +50,10 @@ pub struct OpenIdTokenProviderConfigArguments {
         default_value = "false"
     )]
     pub tls_insecure: bool,
+    /// OAuth scope(s) to request in the client_credentials token request.
+    /// Space-separated when multiple scopes are needed. If unset, no scope parameter is sent.
+    #[arg(id = "oidc_scope", long = "oidc-scope", env = "OIDC_PROVIDER_SCOPE")]
+    pub scope: Option<String>,
 }
 
 impl OpenIdTokenProviderConfigArguments {
@@ -60,6 +64,7 @@ impl OpenIdTokenProviderConfigArguments {
             client_secret: Some(devmode::SSO_CLIENT_SECRET.to_string()),
             refresh_before: Duration::from_secs(30).into(),
             tls_insecure: false,
+            scope: None,
         }
     }
 }
@@ -89,6 +94,7 @@ pub struct OpenIdTokenProviderConfig {
     pub issuer_url: String,
     pub refresh_before: humantime::Duration,
     pub tls_insecure: bool,
+    pub scope: Option<String>,
 }
 
 impl OpenIdTokenProviderConfig {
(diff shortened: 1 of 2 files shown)


common/auth/src/client/provider/openid.rs
```diff
@@ -50,6 +50,10 @@ pub struct OpenIdTokenProviderConfigArguments {
         default_value = "false"
     )]
     pub tls_insecure: bool,
+    /// OAuth scope(s) to request in the client_credentials token request.
```
rh-jfuller, line 53: I wonder if this is a breaking change ? what happens with existing deployments (and have we tested )

common/auth/src/client/provider/openid.rs
```diff
@@ -243,7 +258,11 @@ impl OpenIdTokenProvider {
             Some(current_token) => {
                 log::debug!("Refreshing token ... ");
```
rh-jfuller, line 259: just noticed while in the area - I think this should be converted to tracing call (As per CONVENTIONS.md)

rh-jfuller (approved):
as long as this change has been tested with appropriate existing deployments LGTM

common/auth/src/client/provider/openid.rs
Strum355: This PR is essentially building upon https://github.com/guacsec/trustify/pull/2552, so there should be nothing breaking here. We're bringing back the default of no scope being explicitly requested, with an option to request scopes if necessary. What possible breaking change did you have in mind?

rh-jfuller: I guess my request is if we have tested with an existing env which has named scope and if there is any migration path concerns ... if you have none, then I have none ;)

PR #2568, opened 2026-08-11
chore: include underlying error when logging failed exploit intelligence job reason

Strum355:
Previously if an EI request job failed, we would log the following which doesnt give much information:
`failing job: retries exhausted job_id=019ff11e-3217-7473-938a-28743c159835 error=analysis service temporarily unavailable attempt=3 max_retries=3`

With this change, we will log something as such instead:
`failing job: retries exhausted job_id=019ff11d-c2c0-7151-82ff-9bb600403776 error=SPDX upload: HTTP 429 Too Many Requests: {"error": "mock upload failure (HTTP 429)"} attempt=3 max_retries=3`

Fixes [TC-5541](https://redhat.atlassian.net/browse/TC-5541)

[TC-5541]: https://redhat.atlassian.net/browse/TC-5541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Improve exploit intelligence job failure handling to surface detailed underlying errors in logs and error classification.

Enhancements:
- Preserve and propagate detailed EiRequestError information into AnalysisError instead of generic user-facing messages.
- Simplify EI upload error mapping by directly converting EiRequestError into AnalysisError without additional logging.

### DIFF
--- a/modules/exploit-intelligence/src/runner/mod.rs
+++ b/modules/exploit-intelligence/src/runner/mod.rs
@@ -19,19 +19,6 @@ pub(crate) enum EiRequestError {
     Permanent(anyhow::Error),
 }
 
-impl EiRequestError {
-    /// User-facing error message suitable for storing in the database.
-    /// Transient errors get a generic message; rejected errors surface
-    /// the EI response so users can fix their input.
-    pub(crate) fn user_message(&self) -> String {
-        match self {
-            Self::Transient(_) => "analysis service temporarily unavailable".to_string(),
-            Self::Rejected(msg) => msg.clone(),
-            Self::Permanent(_) => "internal analysis error".to_string(),
-        }
-    }
-}
-
 /// Error from a single analysis tick, classified for retry decisions.
 #[derive(Debug, thiserror::Error)]
 pub(crate) enum AnalysisError {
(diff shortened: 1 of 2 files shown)


modules/exploit-intelligence/src/runner/mod.rs
```diff
@@ -74,12 +61,10 @@ impl From<crate::Error> for AnalysisError {
 
 impl From<EiRequestError> for AnalysisError {
     fn from(e: EiRequestError) -> Self {
-        let msg = e.user_message();
         match e {
-            EiRequestError::Transient(_) => Self::Retryable(anyhow::anyhow!("{msg}")),
-            EiRequestError::Rejected(_) | EiRequestError::Permanent(_) => {
-                Self::Permanent(anyhow::anyhow!("{msg}"))
-            }
+            EiRequestError::Transient(detail) => Self::Retryable(anyhow::anyhow!("{detail}")),
```
rh-jfuller, line 65: This change seems counter to previous request to mask internals of EI and pipe through to end users .... we have no idea what EI internals might expose on error. 

Why are we now backing out from this decision with a new PR ?

These details can be logged or traced but I think any distributed system should be defensive ... I would rather we reuse observability in place (logging/tracing/OTEL) then depend on end user messages for debugging.

rh-jfuller (approved):
LGTM

rh-jfuller (approved):
LGTM

PR #2569, opened 2026-08-11
fix(cvss): honor Exploit Maturity threat metric in CVSS v4.0 scoring

mrrajan:
## Summary

Use `calculated_full_score()` instead of `calculated_base_score()` for CVSS v4.0 so the Exploit Maturity (E) threat metric is included in scoring. Fixes inflated scores for CVEs with `E:P` or `E:U` (e.g., 10.0 instead of 9.3 for CVE-2026-18236).

- Only the v4.0 `From` impl is changed; v2/v3 are unaffected
- No-op for CVEs without an E metric
- Adds 3 reproducer tests (E:P → 9.3, E:U → 9.1, no-E → 10.0)

Implements [TC-5626](https://redhat.atlassian.net/browse/TC-5626)

[TC-5626]: https://redhat.atlassian.net/browse/TC-5626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Honor CVSS v4.0 Exploit Maturity in score calculation to avoid inflated scores and validate behavior with new tests.

Bug Fixes:
- Fix CVSS v4.0 scoring to include the Exploit Maturity (E) threat metric instead of using pure base score, preventing inflated scores for partially or unreported exploits.

Tests:
- Add CVSS v4.0 regression tests covering Exploit Maturity values E:P, E:U, and absence of E to ensure expected scores and severities.

### DIFF
--- a/modules/ingestor/src/graph/cvss.rs
+++ b/modules/ingestor/src/graph/cvss.rs
@@ -93,11 +93,11 @@ impl From<(String, v3::CvssV3)> for ScoreInformation {
 impl From<(String, v4_0::CvssV4)> for ScoreInformation {
     fn from((vulnerability_id, cvss): (String, v4_0::CvssV4)) -> Self {
         let base_score = cvss
-            .calculated_base_score()
+            .calculated_full_score()
             .or_else(|| {
                 v4_0::CvssV4::from_str(&cvss.vector_string)
                     .ok()
-                    .and_then(|p| p.calculated_base_score())
+                    .and_then(|p| p.calculated_full_score())
             })
             .unwrap_or(cvss.base_score);
         Self {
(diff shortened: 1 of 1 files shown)


PhilipCattanach:
Many thanks @mrrajan
It is something we need to backport into the 3.1after it is merged to main.

modules/ingestor/src/graph/cvss.rs
```diff
@@ -93,11 +93,11 @@ impl From<(String, v3::CvssV3)> for ScoreInformation {
 impl From<(String, v4_0::CvssV4)> for ScoreInformation {
     fn from((vulnerability_id, cvss): (String, v4_0::CvssV4)) -> Self {
         let base_score = cvss
```
rh-jfuller: consider renaming 'base_score' now that is 'full_score' .. .this change might have a larger blast radius (ok for a separate PR).

modules/ingestor/src/graph/cvss.rs
```diff
@@ -281,6 +281,56 @@ mod test {
         assert_eq!(info.severity, Severity::None);
     }
 
+    /// Verifies that CVSS v4.0 with E:P (ProofOfConcept) produces score 9.3, not 10.0.
```
rh-jfuller, line 284: seems to be missing an E:A test

rh-jfuller (approved):
LGTM

mrrajan:
Thanks for the review @rh-jfuller I just addressed the requested changes :)  Please let me know.

PR #2570, opened 2026-08-11
fix: persist total_components count for EI analysis when job terminates

Strum355:
`total_components` was always set to NULL at job creation time. No code ever wrote it back to the DB during the polling/completion flow. The list endpoint masked this with a fallback that computed the count from component rows, but the details endpoint, which didn't have this fallback, returned the raw DB value of null.

The fix: `update_job_completed` and `update_job_failed` now accept an optional `total_components` parameter and write it to the DB when provided. The polling code passes the known count (`Some(1)` for single-component, `Some(components.len())` for SPDX) at finalization time. The reason we do it when marking a job as failed/completed is that, to my understanding, EI doesn't give us a definitive list/number of components until EI itself marks a job as having terminated.

## Summary by Sourcery

Persist and surface accurate component counts for exploit intelligence analysis jobs when they complete or fail.

Bug Fixes:
- Ensure total_components is written to the database when EI jobs transition to completed or failed, so details responses no longer return null for terminated jobs.

Enhancements:
- Pass known component totals from polling logic into job completion/failure updates for both single-component and multi-component (SPDX) flows.
- Tighten exploit intelligence job details tests to validate product metadata, report URLs, component counters, and component statuses across success and failure scenarios.
- Simplify summary construction by relying on entity-with-counts rather than recomputing total_components in the list endpoint.
- Remove unused helper for creating jobs with preset product_id and total_components in the service layer.

Tests:
- Expand EI runner, service, and endpoint tests to cover total_components persistence and full job/component counter invariants across various analysis outcomes.

### DIFF
--- a/modules/exploit-intelligence/src/runner/polling.rs
+++ b/modules/exploit-intelligence/src/runner/polling.rs
@@ -131,7 +131,7 @@ pub(crate) async fn poll_for_result(
                 .update_component_completed(component_id, finding, advisory_id, db)
                 .await?;
 
-            ei_service.update_job_completed(job_id, db).await?;
+            ei_service.update_job_completed(job_id, Some(1), db).await?;
 
             tracing::info!(
                 job_id = %job_id,
@@ -154,7 +154,7 @@ pub(crate) async fn poll_for_result(
                 .update_component_failed(component_id, &error_msg, db)
                 .await?;
 
-            ei_service.update_job_completed(job_id, db).await?;
+            ei_service.update_job_completed(job_id, Some(1), db).await?;
 
             Ok(())
         }
@@ -288,8 +288,8 @@ pub(crate) async fn poll_for_product_result(
                     .await?;
             }
 
-            // Check whether any component actually has a finding.
             let components = ei_service.fetch_components(job_id, db).await?;
+            let total = Some(components.len() as i32);
 
             let has_finding = components.iter().any(|c| {
                 c.status == ExploitIntelligenceJobStatus::Completed && c.finding.is_some()
@@ -297,7 +297,7 @@ pub(crate) async fn poll_for_product_result(
             let all_excluded = !components.is_empty() && components.iter().all(|c| c.excluded);
 
             if has_finding || all_excluded {
-                ei_service.update_job_completed(job_id, db).await?;
+                ei_service.update_job_completed(job_id, total, db).await?;
 
                 tracing::info!(
                     job_id = %job_id,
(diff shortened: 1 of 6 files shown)


modules/exploit-intelligence/src/service/mod.rs
```diff
@@ -392,15 +394,7 @@ impl ExploitIntelligenceService {
             .iter()
             .map(|model| {
                 let mut summary = if let Some(c) = counts_map.get(&model.id) {
-                    let mut s = ExploitIntelligenceJobSummary::from_entity_with_counts(
-                        model,
-                        self.ui_url(),
-                        c,
-                    );
-                    if s.total_components.is_none() {
```
rh-jfuller, line 400: good to remove

modules/exploit-intelligence/src/runner/worker.rs
```diff
@@ -45,7 +45,7 @@ impl EiWorker {
         if job.retry_count as usize >= max_retries {
             tracing::warn!(job_id = %job_id, retry_count = job.retry_count, "failing job: max retries exhausted on claim");
             self.ei_service
-                .update_job_failed(job_id, "max retries exhausted", &tx)
+                .update_job_failed(job_id, "max retries exhausted", None, &tx)
```
rh-jfuller, line 48: maybe consider adding a code comment somewhere near here ...  

```
//  total_components is None here because failure occurs before
// (or independently of) component creation. If components already exist
// in DB (e.g. retry-exhausted during SPDX polling)  stored
// total_components will remain NULL. 
```

modules/exploit-intelligence/src/runner/polling.rs
```diff
@@ -288,16 +288,16 @@ pub(crate) async fn poll_for_product_result(
                     .await?;
             }
 
-            // Check whether any component actually has a finding.
             let components = ei_service.fetch_components(job_id, db).await?;
+            let total = Some(components.len() as i32);
```
rh-jfuller, line 292: its unlikely any EI job would have 2b+ components ... though maybe add a source comment to inform future human/agents we ok with this

rh-jfuller (approved):
LGTM

modules/exploit-intelligence/src/runner/worker.rs
Strum355: On Phils recommendation, I'll follow up on non-functional changes like this in a separate PR so we don't have to wait the multiple hours for CI to finish before we can get this merged

PR #2581, opened 2026-08-16
Analysis module performance: reduce allocs, CPU, and DB round-trips

rh-jfuller:
graph analysis module had several performance inefficiencies:

1. **Per-node JSON serialization in `filter()`** — every node scanned during a `GraphQuery::Query` built `serde_json::Value` maps for all purls and CPEs even if query only needed substring matching
2. **Deep cloning in `BaseSummary`** — every node in the result tree cloned all purls, CPEs, and string fields from the graph into owned types
3. **Double hash lookup in `Context::intern`** — `contains_key` + `get` instead of single `entry()` call during graph construction
4. **Per-node `HashSet<Relationship>` clone** — the relationship filter set was cloned for every matching node in the async closure
5. **Two seq DB queries in `resolve_rh_external_sbom_descendants`** — first fetched the checksum value, then searched for matches
6. **No caching for `resolve_external_sbom`** — every `ExternalNode` hit during descendant traversal triggered a fresh DB query, even for duplicate references

Addressing these positively impacts:
- **Memory**: eliminates tens of MB transient heap allocs per query on large graphs (JSON construction), plus ~500 bytes per result node (Arc sharing vs deep clone)
- **CPU**: common filter path avoids all JSON serialization
- **I/O**: saves a DB round-trip per RH external node; eliminates all duplicate external resolution queries per request

No API changes. No schema changes. No migration required.

## Summary by Sourcery

Improve performance of the analysis module by reducing per-node allocations and database round-trips, primarily through caching, shared ownership of summary data, and more efficient query and filtering paths.

New Features:
- Introduce a request-scoped cache for external SBOM resolution to deduplicate database lookups across descendant traversal.

Enhancements:
- Optimize RH external SBOM descendant resolution by replacing two sequential checksum queries with a single self-join query.
- Avoid unnecessary JSON construction in graph filtering by only materializing nested purl and CPE structures when the query string requires them.
- Reduce per-node memory usage in BaseSummary by sharing purl, CPE, and string fields via Arc instead of cloning.
- Improve string interning performance during graph loading by using a single hashmap entry operation instead of separate contains/get steps.
- Avoid cloning the relationship filter set for each node by sharing it across tasks via Arc.

Tests:
- Update analysis service tests to reflect shared purl and CPE storage using Arc-backed slices.

## Summary by Sourcery

Improve analysis performance by reducing transient allocations, sharing immutable data, and eliminating redundant database queries.

Enhancements:
- Reduce analysis query allocations and CPU usage by avoiding unnecessary nested purl and CPE materialization during filtering.
- Share graph-derived summary data through reference-counted ownership to reduce per-node memory usage.
- Improve graph loading and concurrent traversal efficiency through streamlined string interning and shared relationship filters.
- Deduplicate external SBOM lookups within each request using a concurrent resolution cache.
- Resolve RH external SBOM descendants with a single checksum-matching database query instead of sequential round-trips.

Tests:
- Update analysis service assertions for Arc-backed purl and CPE collections.

### DIFF
--- a/modules/analysis/src/model.rs
+++ b/modules/analysis/src/model.rs
@@ -79,14 +79,17 @@ pub struct CacheStatusEntry {
 pub struct BaseSummary {
     pub sbom_id: String,
     pub node_id: String,
-    pub purl: Vec<Purl>,
-    pub cpe: Vec<Cpe>,
+    pub purl: Arc<[Purl]>,
+    pub cpe: Arc<[Cpe]>,
     pub name: String,
     pub version: String,
     pub published: String,
-    pub document_id: String,
-    pub product_name: String,
-    pub product_version: String,
+    #[schema(value_type = String)]
+    pub document_id: Arc<String>,
+    #[schema(value_type = String)]
+    pub product_name: Arc<String>,
+    #[schema(value_type = String)]
+    pub product_version: Arc<String>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, ToSchema)]
--- a/modules/analysis/src/model.rs
+++ b/modules/analysis/src/model.rs
@@ -85,11 +85,11 @@ pub struct BaseSummary {
     pub version: String,
     pub published: String,
     #[schema(value_type = String)]
-    pub document_id: Arc<String>,
+    pub document_id: Arc<str>,
     #[schema(value_type = String)]
-    pub product_name: Arc<String>,
+    pub product_name: Arc<str>,
     #[schema(value_type = String)]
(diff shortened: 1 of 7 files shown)


ctron (approved):

ctron:
Regarding the `ExternalSbomCache` concern: the cache key is `node_id` alone because `resolve_external_sbom` only takes `node_id` and a DB connection as inputs — there is no SBOM ID, discriminator, or external type parameter that could vary the result. The cache key already reflects all inputs that affect resolution.

PR #2597, opened 2026-08-25
fix(purl): filter purl_statuses by SBOM describing CPEs

mrrajan:
Backport of f322d188 to `release/0.4.z`.

The `purl_status` query in `PurlDetails::from_entity` returned entries whose
`context_cpe_id` came from unrelated products (e.g. a hummingbird/RHEL advisory
context attaching to an el8 package by name), because the context CPE was never
validated against the SBOM's own describing CPEs.

This applies the same three-way CPE-context filter already used by
`get_product_statuses_for_purl`, reusing the existing `cpe_context_subqueries`
helper:

```
context_cpe_id IS NULL
OR context_cpe_id IN (SBOM's generalized describing CPEs)
OR the SBOM has no describing CPEs
```

Scope: `/purl` endpoint only (matches upstream `f322d188`). The generalized-CPE
and edition-constraint refinements (`c2fd9ffe`, `76b2732b`) are already present in
0.4.z's `cpe_context_subqueries` helper. TC-5630 CPE-only-node parity is a separate,
larger change (requires the `cpe_status` table).

Verified: `cargo check` clean; `advisory::csaf::reingest` (4/4) and `vuln` regression
tests pass.

## Summary by Sourcery

Restrict PURL advisory statuses to CPE contexts supported by the associated SBOMs.

Bug Fixes:
- Filter `/purl` advisory statuses so context CPEs are limited to CPEs describing the SBOM, preventing unrelated cross-product status matches.

Enhancements:
- Apply generalized CPE edition matching consistently when determining valid PURL status contexts.

Build:
- Add the rstest test dependency and supporting CPE edition test fixtures.

Tests:
- Add regression coverage for accepting null or wildcard editions while rejecting specific editions in PURL CPE-context matching.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9034,6 +9034,7 @@ dependencies = [
  "regex",
  "reqwest 0.12.28",
  "roxmltree",
+ "rstest",
  "sanitize-filename",
  "sea-orm",
  "sea-query",
--- a/etc/test-data/csaf/issues/cpe_edition/advisory-edition-el8.json
+++ b/etc/test-data/csaf/issues/cpe_edition/advisory-edition-el8.json
@@ -1,0 +1,92 @@
+{
+  "document": {
+    "category": "csaf_vex",
+    "csaf_version": "2.0",
+    "distribution": { "tlp": { "label": "WHITE" } },
+    "publisher": {
+      "category": "vendor",
+      "name": "Test Vendor",
+      "namespace": "https://test.example.com"
+    },
+    "title": "TEST-EDITION-EL8",
+    "tracking": {
+      "current_release_date": "2024-01-01T00:00:00Z",
+      "id": "TEST-EDITION-EL8",
+      "initial_release_date": "2024-01-01T00:00:00Z",
+      "revision_history": [
+        {
+          "date": "2024-01-01T00:00:00Z",
+          "number": "1",
+          "summary": "Initial"
+        }
+      ],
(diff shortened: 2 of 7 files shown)


ctron:
Looks ok to me. But I didn't check sourcery's comment. Which sounds reasonable.

ctron:
Confirmed — Sourcery's finding is valid.

The raw SQL in `CONTEXT_CPE_FILTER_SQL` (line 19) and `product_advisory_info_sql` (line 50) in `modules/fundamental/src/sbom/model/raw_sql.rs` both constrain generalized CPEs with:

```sql
WHERE (edition IS NULL OR edition = '*')
  AND (vendor, product, version) IN (...)
```

But the SeaORM equivalent in `cpe_context_subqueries` (lines 188–224 of `purl.rs`) omits the edition predicate on the candidate CPE alias `c`. This means edition-specific CPEs (e.g. `edition = "el8"`) can pass the generalized match when they shouldn't — exactly the cross-product false positive this PR is meant to prevent.

Fix: add the edition constraint to `generalized_cpe_ids`, after the `sbom_describing_cpe` join:

```rust
.and_where(
    Condition::any()
        .add(Expr::col((c.clone(), cpe::Column::Edition)).is_null())
        .add(Expr::col((c.clone(), cpe::Column::Edition)).eq("*")),
)
```

ctron (approved):

PR #2599, opened 2026-08-25
fix(ingestor): materialize OS-node CPEs into sbom_describing_cpe so the CPE-context filter engages for child-node product CPEs (TC-5749)

mrrajan:
## Summary

The vulnerability-correlation **CPE-context filter** suppresses wrong-product ([TC-5171](https://redhat.atlassian.net/browse/TC-5171)) and wrong-scheme ([TC-5170](https://redhat.atlassian.net/browse/TC-5170)) advisory matches by comparing an advisory's context CPE against the SBOM's *describing* CPEs (materialized in `sbom_describing_cpe`). It works today only when the product/OS CPE sits on the SBOM's `Describes` (root) node.

Real Red Hat RHEL image SBOMs (CycloneDX/SPDX) attach the product/OS CPE — e.g. `cpe:/o:redhat:enterprise_linux:8` — to a **child/OS component**, not the `Describes` node. `populate_describing_cpes` only captured CPEs on `Describes`-relationship (`relationship = 13`) nodes, so for those SBOMs `sbom_describing_cpe` was empty, the filter fell through its escape hatch (`OR sbom_id NOT IN sbom_has_cpes`), and cross-product / wrong-scheme matches leaked on `/sbom/{id}/advisory`, the SBOM list, and `/purl`. This is the **SBOM-scale root cause behind [TC-5170](https://redhat.atlassian.net/browse/TC-5170) / [TC-5171](https://redhat.atlassian.net/browse/TC-5171)**.

## Change

Broaden describing-CPE detection in **both** population sites so a product/OS CPE on a child node is materialized:

- **Ingest-time** `populate_describing_cpes` (`modules/ingestor/src/graph/sbom/mod.rs`): operating-system CPEs (`part = 'o'`) on **any** node now count as product/OS context, in addition to `Describes`-node CPEs of any part (a `LEFT JOIN` preserves the existing `relationship = 13` behaviour exactly).
- **Re-backfill migration** `m0002310_sbom_describing_cpe_child_nodes` for already-ingested SBOMs — additive, idempotent (`INSERT … ON CONFLICT DO NOTHING`), with a precise `down()` that removes only the OS-CPE rows not justified by a `Describes` relationship.

Component-identity CPEs (`part = 'a'`, used by `cpe_status` matching — [TC-5630](https://redhat.atlassian.net/browse/TC-5630) / S7 hummingbird) are **deliberately excluded**, so package identities are never mistaken for product context. The consumers (`CONTEXT_CPE_FILTER_SQL`, `product_advisory_info_sql`, `batch_severity_counts_sql`, `cpe_context_subqueries`) are unchanged — the fix is purely upstream population. The codebase already treats `part = 'a'` as the component-identity space (`cpe_advisory_info_sql` / `cpe_version_matches` gate on `part = 'a'`), so `part = 'o'` for describing context is consistent by construction.

## Tests

New integration test `sbom_details_describing_cpe_on_child_node` (`modules/fundamental/tests/sbom/details.rs`) with synthetic fixtures under `etc/test-data/cyclonedx/TC-5749/`:

- **child-cpe** SBOM (OS CPE on a child node) — the wrong-product advisory is now filtered, **identically to the root-cpe control**.
- **root-cpe** SBOM — parity control (already filtered).
- **no-cpe** SBOM — the escape hatch still applies (no product CPE ⇒ the match still surfaces), proving the fix is scoped and introduces no new false negatives for genuinely context-less SBOMs.

Migration `refresh()` (down+up) passes; the [TC-5630](https://redhat.atlassian.net/browse/TC-5630) purl-less regression control (`sbom_details_purlless_cpe_node_consistency`) still passes; `cargo xtask precommit` is clean (clippy `-D warnings`, fmt, check; no OpenAPI/schema drift).

## Scenario validation ([TC-5639](https://redhat.atlassian.net/browse/TC-5639) suite)

Against a build with this change, the child-node-CPE SBOMs now behave like their root-CPE counterparts on `/sbom/{id}/advisory` and `/purl`: **S10 child-cpe** flips to the golden `not_affected` for CVE-2024-6602, and the wrong-product (S3/S4) and cross-**major** (S1 el8→el9/el10) leaks are dropped.

## Known coupling — [TC-5732](https://redhat.atlassian.net/browse/TC-5732) (bare `known_affected`)

Engaging the CPE-context filter for these SBOMs surfaces a **documented, pre-existing** coupling with **[TC-5732](https://redhat.atlassian.net/browse/TC-5732)** (bare / version-less `known_affected` entries are unmatchable). Two scenario rows go false-negative because the genuine el8 match is a *bare* `known_affected` and was previously only "caught" by the very wrong-product/by-name leak this PR removes:

- **S3 / CVE-2025-13034** — el8 curl is genuinely affected but stated version-less; its only prior match was the spurious `hummingbird` row.
- **S11 / firefox `main_el8` / CVE-2023-6135** — el8 affected bare; only sub-stream `fixed` rows are versioned.

Per the scenario READMEs: *"the describing-CPE filter and the bare-entry bug are coupled — the same filter that rescues thunderbird (S10) breaks firefox — to get main-8 right you must also make bare `known_affected` entries matchable."* **[TC-5732](https://redhat.atlassian.net/browse/TC-5732) should land together with (or immediately after) this PR** to close that window.

## Follow-up (out of scope here)

The `/vulnerability/{cve}` backlink endpoint does not apply the CPE-context filter, so it over-reports relative to the now-correct `/sbom/advisory` and `/purl` paths — a separate 4th-endpoint-consistency follow-up.

Implements [TC-5749](https://redhat.atlassian.net/browse/TC-5749)
Root-cause bug: [TC-5750](https://redhat.atlassian.net/browse/TC-5750) · Symptoms: [TC-5170](https://redhat.atlassian.net/browse/TC-5170) / [TC-5171](https://redhat.atlassian.net/browse/TC-5171)

[TC-5749]: https://redhat.atlassian.net/browse/TC-5749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Ensure product-context filtering applies consistently when SBOM operating-system CPEs are attached to child nodes.

Bug Fixes:
- Materialize operating-system CPEs attached to child SBOM nodes so the CPE-context filter correctly suppresses wrong-product and wrong-scheme advisory matches.

Tests:
- Add integration coverage for child-node, root-node, and CPE-less SBOM scenarios to verify filtering behavior and preserve the context-less escape hatch.

Chores:
- Add an idempotent migration to backfill child-node OS CPE associations for existing SBOMs, with a targeted rollback.

[TC-5171]: https://redhat.atlassian.net/browse/TC-5171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TC-5170]: https://redhat.atlassian.net/browse/TC-5170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TC-5170]: https://redhat.atlassian.net/browse/TC-5170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/etc/test-data/cyclonedx/TC-5749/sbom_child_cpe.json
+++ b/etc/test-data/cyclonedx/TC-5749/sbom_child_cpe.json
@@ -1,0 +1,37 @@
+{
+  "bomFormat": "CycloneDX",
+  "specVersion": "1.5",
+  "serialNumber": "urn:uuid:tc5749-child-cpe-0000-000000000001",
+  "version": 1,
+  "metadata": {
+    "timestamp": "2025-01-01T00:00:00Z",
+    "component": {
+      "type": "application",
+      "name": "eus-target-app",
+      "bom-ref": "root-app",
+      "version": "1.0"
+    }
+  },
+  "components": [
+    {
+      "type": "operating-system",
+      "name": "rhel",
+      "bom-ref": "os-el8",
+      "version": "8",
+      "cpe": "cpe:2.3:o:redhat:enterprise_linux:8:*:*:*:*:*:*:*"
+    },
+    {
+      "type": "library",
+      "name": "mypkg",
+      "bom-ref": "pkg-mypkg",
+      "version": "1.0.0-1.el8",
+      "purl": "pkg:rpm/redhat/mypkg@1.0.0-1.el8?arch=x86_64"
+    }
+  ],
+  "dependencies": [
+    {
+      "ref": "root-app",
+      "dependsOn": ["os-el8", "pkg-mypkg"]
+    }
+  ]
+}
--- a/etc/test-data/cyclonedx/TC-5749/sbom_no_cpe.json
+++ b/etc/test-data/cyclonedx/TC-5749/sbom_no_cpe.json
@@ -1,0 +1,30 @@
+{
+  "bomFormat": "CycloneDX",
+  "specVersion": "1.5",
+  "serialNumber": "urn:uuid:tc5749-no-cpe-00000-000000000003",
+  "version": 1,
+  "metadata": {
+    "timestamp": "2025-01-01T00:00:00Z",
+    "component": {
+      "type": "application",
+      "name": "no-cpe-app",
+      "bom-ref": "root-nocpe",
+      "version": "1.0"
+    }
+  },
+  "components": [
+    {
+      "type": "library",
+      "name": "mypkg",
+      "bom-ref": "pkg-mypkg",
+      "version": "1.0.0-1.el8",
+      "purl": "pkg:rpm/redhat/mypkg@1.0.0-1.el8?arch=x86_64"
+    }
+  ],
+  "dependencies": [
+    {
+      "ref": "root-nocpe",
+      "dependsOn": ["pkg-mypkg"]
+    }
+  ]
+}
--- a/etc/test-data/cyclonedx/TC-5749/sbom_root_cpe.json
+++ b/etc/test-data/cyclonedx/TC-5749/sbom_root_cpe.json
@@ -1,0 +1,31 @@
+{
+  "bomFormat": "CycloneDX",
+  "specVersion": "1.5",
+  "serialNumber": "urn:uuid:tc5749-root-cpe-0000-000000000002",
+  "version": 1,
+  "metadata": {
+    "timestamp": "2025-01-01T00:00:00Z",
+    "component": {
+      "type": "operating-system",
+      "name": "rhel",
+      "bom-ref": "root-os",
+      "version": "8",
+      "cpe": "cpe:2.3:o:redhat:enterprise_linux:8:*:*:*:*:*:*:*"
+    }
+  },
+  "components": [
+    {
+      "type": "library",
+      "name": "mypkg",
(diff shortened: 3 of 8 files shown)


ctron (changes requested):
I am against a change like this: This mixes generic spec elements with very specific RH data.

`o` meaning "Operating System" as a trigger for special logic could be a problem for RH data alone. But other publishers out there might disagree with that special logic completely.

mrrajan:
We have to revisit this fix with proper vulnerability correlation ADR.

PR #2600, opened 2026-08-26
fix(sbom): version-filter product_status on the advisory and severity paths (TC-5641)

mrrajan:
## What this fixes

`GET /sbom/{id}/advisory` and the severity summary matched Red Hat `product_status` rows by package **name** only — reporting RPMs already **past the fix** as affected. `purl_status`/`cpe_status` and the `/purl` + `/analyze` endpoints already version-match; these two SBOM paths were missed.

## Change

In `modules/fundamental/src/sbom/model/raw_sql.rs`, add `version_matches` to the `product_status` CTEs in `product_advisory_info_sql` and `batch_severity_counts_sql` — joining `product_status → product_version_range → version_range`, mirroring the sibling `purl_version_matches` CTE (and adding `vp.version` to `sbom_purls`). `product_version_range_id` is NOT NULL and bare `known_affected` creates no `product_status` row, so the inner join drops nothing that would otherwise match.

## Tests

New `sbom_details_product_status_version_filtering` (`tests/sbom/details.rs`, fixtures in `etc/test-data/cyclonedx/TC-5751/`): a past-fix RPM is filtered, an in-range RPM is kept, and detail count == severity-summary count. `sbom::details` suite green; `cargo xtask precommit` clean.

Implements [TC-5641](https://redhat.atlassian.net/browse/TC-5641)

[TC-5641]: https://redhat.atlassian.net/browse/TC-5641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Apply version-range filtering to product-status matches in SBOM advisory details and severity summaries.

Bug Fixes:
- Filter Red Hat product-status advisory matches by the installed package version so past-fix packages are no longer reported as affected.
- Keep SBOM advisory details and severity summaries consistent when applying product-status version filtering.

Tests:
- Add coverage for past-fix and in-range product versions, including consistency checks between advisory details and severity summaries.

### DIFF
--- a/modules/fundamental/tests/sbom/details.rs
+++ b/modules/fundamental/tests/sbom/details.rs
@@ -835,6 +835,122 @@ async fn sbom_details_purlless_cpe_node_consistency(
     Ok(())
 }
 
+/// Reproducer for TC-5641: the SBOM `/advisory` detail and severity-summary
+/// paths must apply `version_matches` to Red Hat `product_status` matches, as
+/// the `/purl` and `/analyze` endpoints already do. A product whose installed
+/// version falls outside the advisory's version range must not be reported.
+///
+/// `CVE-2098-0001` marks package `widget` affected under `cpe:/a:example:product:8`
+/// (version range major 8 → `[8, 9)`). Two SBOMs, each with **no** describing CPE
+/// (so the CPE-context escape hatch is on and the installed version is the only
+/// discriminator):
+/// - `sbom_pastfix` — `widget 1.2.3-1.el8` (outside `[8, 9)`) → must NOT be
+///   reported (fails before the fix — no `version_matches` — passes after).
+/// - `sbom_inrange` — `widget 8.1.0-1.el8` (inside `[8, 9)`) → must still be
+///   reported (true positive preserved).
+///
+/// Also asserts detail count == severity-summary count for both SBOMs (the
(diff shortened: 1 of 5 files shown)


modules/fundamental/tests/sbom/details.rs
```diff
@@ -835,6 +835,122 @@ async fn sbom_details_purlless_cpe_node_consistency(
...
+///   reported (fails before the fix — no `version_matches` — passes after).
+/// - `sbom_inrange` — `widget 8.1.0-1.el8` (inside `[8, 9)`) → must still be
+///   reported (true positive preserved).
+///
+/// Also asserts detail count == severity-summary count for both SBOMs (the
+/// TC-5630 detail==list invariant).
+#[test_context(TrustifyContext)]
+#[test(tokio::test)]
+#[instrument]
+async fn sbom_details_product_status_version_filtering(
```
ctron, line 857: This looks like it should be a parametrized test.

mrrajan:
The CI failure reveals the real concern of this fix. TC-5751 producing false-negatives. The root cause is a semantic mismatch in the version filter I added.

PR #2602, opened 2026-08-26
fix(db): match a fully-unbounded version range against all versions (TC-5732)

mrrajan:
## What this fixes

A CSAF `known_affected` with **no version** (all versions affected) is stored as a fully-unbounded version range, but `version_matches` returned **false** for a range with neither a low nor a high bound — so the row was dropped everywhere and a genuinely-affected main-stream package went **false-negative** once the CPE-context filter engaged ([TC-5732](https://redhat.atlassian.net/browse/TC-5732) / scenario S11).

## Change

New migration `m0002310_version_matches_unbounded` adds a three-line guard at the top of the `version_matches` dispatch: a both-NULL range matches all versions, uniformly across schemes (dispatch `CASE` unchanged; `down()` restores it). Minimal and backport-friendly for release/0.4.z, 0.5.z, 0.6.z.

## Tests

- `tests/version/mod.rs` — both-NULL cases per scheme assert `true` (fail before, pass after).
- `sbom_details_versionless_known_affected` + `etc/test-data/cyclonedx/[TC-5752](https://redhat.atlassian.net/browse/TC-5752)/` — version-less `known_affected` under `enterprise_linux:8` is reported for a main-el8 SBOM; a patched-EUS control is not (stream-scoped, no over-report).

Migration refresh + correlation/CSAF/version suites green; `cargo xtask precommit` clean.

Implements [TC-5732](https://redhat.atlassian.net/browse/TC-5732)

## Summary by Sourcery

Fix version matching so version-less affected advisories correctly report impacted packages across supported version schemes.

Bug Fixes:
- Make fully unbounded version ranges match all candidate versions, preventing false negatives for version-less affected advisories.
- Ensure version-less known-affected advisories remain scoped to the correct product stream without over-reporting patched EUS packages.

Tests:
- Add cross-scheme coverage for fully unbounded version ranges.
- Add end-to-end SBOM coverage for version-less CSAF affected status and stream-specific filtering.

[TC-5732]: https://redhat.atlassian.net/browse/TC-5732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TC-5752]: https://redhat.atlassian.net/browse/TC-5752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TC-5732]: https://redhat.atlassian.net/browse/TC-5732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/etc/test-data/cyclonedx/TC-5752/advisory-firefox.json
+++ b/etc/test-data/cyclonedx/TC-5752/advisory-firefox.json
@@ -1,0 +1,106 @@
+{
+  "document": {
+    "aggregate_severity": { "text": "Important" },
+    "category": "csaf_vex",
+    "csaf_version": "2.0",
+    "distribution": { "tlp": { "label": "WHITE" } },
+    "lang": "en",
+    "publisher": {
+      "category": "vendor",
+      "name": "Red Hat",
+      "namespace": "https://www.redhat.com"
+    },
+    "title": "TC-5752 version-less known_affected reproducer (firefox)",
+    "tracking": {
+      "current_release_date": "2025-01-15T00:00:00Z",
+      "id": "TC-5752-FIREFOX",
+      "initial_release_date": "2025-01-15T00:00:00Z",
+      "revision_history": [
+        { "date": "2025-01-15T00:00:00Z", "number": "1", "summary": "Initial version" }
+      ],
+      "status": "final",
+      "version": "1"
+    }
+  },
+  "product_tree": {
+    "branches": [
+      {
+        "category": "vendor",
+        "name": "Red Hat",
+        "branches": [
+          {
+            "category": "product_name",
+            "name": "Red Hat Enterprise Linux 8",
+            "product": {
+              "name": "Red Hat Enterprise Linux 8",
+              "product_id": "rhel_8",
+              "product_identification_helper": { "cpe": "cpe:/o:redhat:enterprise_linux:8" }
+            }
+          },
+          {
+            "category": "product_name",
+            "name": "Red Hat Enterprise Linux 8.8 Extended Update Support",
+            "product": {
+              "name": "Red Hat Enterprise Linux 8.8 EUS",
+              "product_id": "rhel_eus_8.8",
+              "product_identification_helper": { "cpe": "cpe:/a:redhat:rhel_eus:8.8::appstream" }
+            }
+          },
+          {
+            "category": "product_version",
+            "name": "firefox",
+            "product": {
+              "name": "firefox",
+              "product_id": "firefox",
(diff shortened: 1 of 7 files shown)


ctron (commented):
## CSAF spec compliance concern

The `version_matches` SQL fix is mechanically correct — it makes the query layer consistent with what the ingestion layer stores. However, the underlying interpretation of "no version = all versions" is a vendor-specific convention, not what the CSAF 2.0 specification prescribes. Trustify should follow the spec.

### The spec requires explicit `vers:all/*` for "all versions"

Per CSAF 2.0 §3.1.2.3.2, the way to express "all versions affected" is a `product_version_range` branch with `vers:all/*`. There is no normative statement in the spec that omitting a version implies "all versions." A `product_version` branch (§3.1.2.1) represents exactly one version of a product.

### The ingestion code makes a vendor-specific assumption

Two ingestion paths create both-NULL (fully-unbounded) version ranges:

1. **PURL without a version** (`modules/ingestor/src/service/advisory/csaf/creator.rs:252–255`): a version-less PURL like `pkg:rpm/redhat/firefox` is interpreted as `Range(Unbounded, Unbounded)`, meaning "all versions." But per the PURL spec, a version-less PURL means "the package identity without pinning a version" — that is not the same as "all versions are affected."

2. **CPE with `version=*`** (`modules/ingestor/src/service/advisory/csaf/product_status.rs:88–94`): treated as unbounded. This is more defensible since CPE matching semantics define `*` as "any."

### The SQL fix masks the real issue

The `version_matches` guard enshrines a non-spec interpretation at the query layer. If the ingestion layer correctly followed the CSAF spec, both-NULL ranges for CSAF-sourced data should only exist for advisories that explicitly use `vers:all/*` (or the CPE `*` case).

### What spec compliance would look like

- Handle `product_version_range` branches with `vers:all/*` and create unbounded ranges for those — that is the spec-sanctioned path.
- A `product_version` with no version info should either be warned about or stored without a version range (meaning "version unspecified," not "all versions affected").
- The `version_matches` SQL fix may still be useful for CPE `version=*` and any future legitimate "all versions" scenarios, but it should not be motivated by a non-spec CSAF interpretation.

mrrajan:
Taken care by https://github.com/guacsec/trustify/pull/2643

PR #2604, opened 2026-08-26
docs(adr): add ADR-00020 unified importer credential resolution

ruromero:
## Summary

- Proposes a shared auth model (Basic, Bearer, API key) with a credential source abstraction (inline/env/file) for all HTTP-based importers
- Enables production deployments to reference secrets via env vars or mounted files instead of storing credentials as plaintext in the database
- Integrates with Kubernetes Secrets populated by Vault or External Secrets Operator

Refs [TC-5738](https://redhat.atlassian.net/browse/TC-5738)

[TC-5738]: https://redhat.atlassian.net/browse/TC-5738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Adopt a shared, reference-based credential resolution model for importer authentication without storing production secrets directly in the database.

Enhancements:
- Define a unified authentication and credential-source model for HTTP and Git-based importers, supporting inline, environment-variable, and file-backed secrets.
- Standardize authenticated client construction and provide a Kubernetes-compatible pattern for external secret injection and runtime credential rotation.
- Specify migration and backward-compatibility handling for existing Quay API tokens while preserving unauthenticated and legacy Git behavior.

Documentation:
- Add ADR-00020 documenting unified importer credential resolution, supported auth methods, deployment patterns, migration strategy, alternatives, and future scope.

### DIFF
--- a/docs/adrs/00020-importer-credential-resolution.md
+++ b/docs/adrs/00020-importer-credential-resolution.md
@@ -1,0 +1,283 @@
+# 00020. Unified importer credential resolution
+
+Date: 2026-08-24
+
+## Status
+
+APPROVED
+
+## Context
+
+Trustify importers fetch data from external sources that may require authentication. Today, each
+importer implements its own credential handling:
+
+* **Quay importer:** `api_token: Option<String>` — a bearer token stored as plaintext in the
+  importer's `configuration` JSONB column. Uses both an HTTP client (registry API) and an OCI
+  client (manifest/blob pulls), each with independent auth.
+* **HTTP-based importers** (NVD, CWE, ClearlyDefined, CSAF, SBOM): use `reqwest` or
+  walker-based HTTP clients with no shared auth model — each builds its own client.
+* **Git-based importers** (OSV, CVE, ClearlyDefined Curation): credential callback in the
+  git2 walker, no explicit auth model in the importer configuration.
+
+### Problems
+
+1. **Credentials stored in plaintext in the database.** The `importer.configuration` column is a
+   JSONB field that stores the full importer configuration, including secrets. Anyone with database
+   read access can extract credentials. This is acceptable for development but not for production.
+
+2. **No shared auth model.** Each importer defines its own auth types and client-building logic.
+   Adding a new HTTP-based importer means reimplementing auth from scratch.
+
+3. **No support for external secret stores.** Production deployments typically manage secrets via
+   Kubernetes Secrets (populated by Vault, External Secrets Operator, or similar). There is no way
+   to reference an external secret — the only option is to put the literal value in the API request
+   that creates the importer.
+
+## Decision
+
+### Unified importer auth model
+
+Introduce a shared authentication model for HTTP-based importers, with a parallel model for
+git-based importers (deferred — see below). Both share the same `CredentialSource` abstraction
+for secret resolution.
+
+The HTTP auth model supports three methods:
+
+* **Basic** — username and password
+* **Bearer** — token
+* **API key** — header name and value (e.g., `X-Api-Key`)
+
+Each importer exposes an optional `auth` field using this shared model. When absent, the importer
+accesses the source without authentication.
+
+### Credential source abstraction
+
+Instead of storing credential values directly, each credential is wrapped in a source descriptor
+that specifies *where* to obtain the actual secret at runtime. Three source types are supported:
+
+| Source | Description | Use case |
+|--------|-------------|----------|
+| **Inline** | Literal value in the configuration | Development and testing |
+| **Env** | Read from an environment variable | K8s `envFrom: secretRef` |
+| **File** | Read from a file path | K8s volume-mounted secrets |
+
+The file source reads a single value per file. This maps directly to Kubernetes Secret volume
+mounts, where each key in a Secret is projected as a separate file (e.g., `/run/secrets/username`
+and `/run/secrets/password`). No multi-key file parsing is needed because K8s handles the
+key-to-file mapping natively.
+
+**Security note:** The file source accepts a path from the importer configuration and reads it
+at runtime. Creating or updating importers requires authorization (OIDC + permissions), so only
+privileged users can configure file paths. In production Kubernetes deployments, the pod
+filesystem is constrained by the container runtime — only mounted volumes and the container
+image are accessible, limiting the scope of readable files.
+
+Credentials are resolved **at import time**, not at importer creation time. This means:
+
+* The database stores only references (env var name or file path), never the actual secret for
+  non-inline sources.
+* File-based credentials support rotation without pod restart — the importer reads the current
+  file contents on each run.
+* Environment variables work naturally with Kubernetes `envFrom: secretRef`.
+
+### Kubernetes deployment pattern
+
+The recommended production pattern uses Kubernetes Secrets as the bridge between a secret store
+(Vault, External Secrets Operator, cloud provider KMS) and Trustify:
+
+```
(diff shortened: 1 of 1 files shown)


docs/adrs/00020-importer-credential-resolution.md
```diff
@@ -0,0 +1,221 @@
...
+### Credential source abstraction
+
+Instead of storing credential values directly, each credential is wrapped in a source descriptor
+that specifies *where* to obtain the actual secret at runtime. Three source types are supported:
+
+| Source | Description | Use case |
+|--------|-------------|----------|
+| **Inline** | Literal value in the configuration | Development and testing |
+| **Env** | Read from an environment variable | K8s `envFrom: secretRef` |
+| **File** | Read from a file path | K8s volume-mounted secrets |
```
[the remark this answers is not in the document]
ruromero: Thanks for flagging the security concern. We considered a directory allowlist but decided against it for these reasons:

1. **Creating/updating importers already requires authorization** (OIDC + permissions) — only privileged users can configure file paths.
2. **In production K8s deployments, the pod filesystem is constrained by the container runtime** — only mounted volumes and the container image are accessible, which limits the scope of readable files. This is the standard security boundary for secret consumption in Kubernetes.
3. **No established precedent** — tools in this space (Vault Agent, External Secrets Operator, K8s CSI secret drivers) all mount secrets to arbitrary paths without an application-level allowlist. Adding one would complicate deployment without meaningful security gain in a containerized environment.

Updated the ADR with a security note documenting these mitigations.

docs/adrs/00020-importer-credential-resolution.md
```diff
@@ -0,0 +1,254 @@
...
```
ctron, line 62: That might be dangerous, unless we somehow limit file system access.

docs/adrs/00020-importer-credential-resolution.md
```diff
@@ -0,0 +1,254 @@
...
+
+The file source reads a single value per file. This maps directly to Kubernetes Secret volume
+mounts, where each key in a Secret is projected as a separate file (e.g., `/run/secrets/username`
+and `/run/secrets/password`). No multi-key file parsing is needed because K8s handles the
+key-to-file mapping natively.
+
+**Security note:** The file source accepts a path from the importer configuration and reads it
```
ctron, line 69: I'd say that's still too broad, and won't work for bare metal installations (ansible).

docs/adrs/00020-importer-credential-resolution.md
ruromero: This is a standard practice in k8s environments. Containers are expected to limit accessible paths via volume mounts, not application-level allowlists.

Adding an allowlist would add deployment complexity without meaningful security gain in containerized environments.

docs/adrs/00020-importer-credential-resolution.md
ruromero: As this is only related to non containerized environments. I'd just mention it in the documentation and recommend the environment variable approach instead.
Maybe the only thing I'd add to the ADR could be a base path restriction for bare metal environments. Would that be ok?

docs/adrs/00020-importer-credential-resolution.md
```diff
@@ -0,0 +1,254 @@
...
+`importer.configuration` JSONB column from the old `api_token` field to the new `auth`
+structure. The migration converts existing plaintext tokens to inline credential sources.
+Since migrations run on startup before the application serves requests, stored configurations
+are in the new format by the time the application starts.
+
+**API backward compatibility.** The database migration handles stored data, but external
+clients that create or update importers via the REST API would break if the old field is
+removed immediately. To avoid a breaking API change:
+
+1. The Quay importer continues to accept the legacy `api_token` field as a **deprecated**
```
ctron: I'm not sure we have to do this. Not sure if we ever agreed on this being stable. Not pushing back on this though.

docs/adrs/00020-importer-credential-resolution.md
```diff
@@ -0,0 +1,254 @@
...
+  }
+}
+```
+
+This can be defined in a separate piece of work. The `CredentialSource` abstraction introduced
+by this ADR is designed to support it without architectural changes.
+
+### Git-based importer credentials
+
+Git-based importers (OSV, CSAF, CVE, CWE, SBOM) use git2's credential callback, which operates
```
ctron: I think it would be worth exploring this, as we are designing something should unify things already.

ctron (commented):
This looks great! Some small comments. And I'd like to explore the git2 integration, I think that's something we might tackle at the same time. At least we should understand what that would mean.

docs/adrs/00020-importer-credential-resolution.md
ruromero: I wanted to keep backward compatibility. I prefer the reviewers to decide.

docs/adrs/00020-importer-credential-resolution.md
ruromero: Sounds good. I'll add a proposal for git as well

ruromero:
> This looks great! Some small comments. And I'd like to explore the git2 integration, I think that's something we might tackle at the same time. At least we should understand what that would mean.

Done @ctron , let me know what you think about the suggested approach.
Thanks for your comments and review!

docs/adrs/00020-importer-credential-resolution.md
rh-jfuller: FWIW ADR does not need to be comprehensive, just mark out broad decisions ... and defer to any future decisions as open questions (provided they are not blocking).

docs/adrs/00020-importer-credential-resolution.md
rh-jfuller: considering we have brittle binding to external services ... I think @ctron has a point, lets leave this out.

docs/adrs/00020-importer-credential-resolution.md
rh-jfuller: agreed, a git base class importer is something I was thinking about implementing.

docs/adrs/00020-importer-credential-resolution.md
```diff
@@ -0,0 +1,319 @@
...
+   Adding a new HTTP-based importer means reimplementing auth from scratch.
+
+3. **No support for external secret stores.** Production deployments typically manage secrets via
+   Kubernetes Secrets (populated by Vault, External Secrets Operator, or similar). There is no way
+   to reference an external secret — the only option is to put the literal value in the API request
+   that creates the importer.
+
+## Decision
+
+### Unified auth model
```
rh-jfuller: pedantry - _unified importer auth model_

rh-jfuller (approved):
LGTM - it seems best to stick to reference-based credential resolution for this ADR ... other related stuff like git transport model or OCI dual client stuff could be added in terms of addenda/additional information.

docs/adrs/00020-importer-credential-resolution.md
ruromero: done

ruromero:
> LGTM - it seems best to stick to reference-based credential resolution for this ADR ... other related stuff like git transport model or OCI dual client stuff could be added in terms of addenda/additional information.

Thanks @rh-jfuller I think I have addressed all the requested changes. Let me know if there's anything pending.

PR #2611, opened 2026-08-26
docs(adr): add ADR-00021 HTTP transport with pluggable file discovery

ruromero:
## Summary

- Proposes a generic HTTP transport with pluggable file discovery strategies
- Pulp Manifest is the first implementation — simple CSV format with SHA256 checksums, served by the open-source Pulp content distribution platform
- Transport is content-agnostic: format detection delegated to the ingestor
- Rejects HTML scraping (fragile) and WebDAV (requires server support) as primary strategies, leaving room for both as future extensions

## Dependencies

- Depends on [#2604](https://github.com/guacsec/trustify/pull/2604) (ADR-00020: unified credential resolution) for authentication

Refs [TC-5764](https://redhat.atlassian.net/browse/TC-5764)

[TC-5764]: https://redhat.atlassian.net/browse/TC-5764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Adopt a pluggable HTTP transport design for discovering and retrieving files from generic content repositories.

Enhancements:
- Document a proposed content-agnostic HTTP transport that separates shared file retrieval from pluggable file discovery strategies.
- Define Pulp Manifest discovery as the initial strategy, including integrity metadata, pattern filtering, and compatibility with shared authentication and ingestor format detection.
- Establish an extensible model for adding alternative discovery mechanisms without changing the transport core.

Documentation:
- Add ADR-00021 describing HTTP file discovery challenges, the manifest-based design, alternatives considered, and expected consequences.


docs/adrs/00021-http-transport-with-pluggable-discovery.md
```diff
@@ -0,0 +1,204 @@
...
+## Status
+
+PROPOSED
+
+## Context
+
+Trustify's existing importers rely on two transport mechanisms to fetch external data:
+
+* **Git** — clone a repository and walk the filesystem (used by OSV, CSAF, CVE, CWE, SBOM
+  importers).
```
jcrossley3: I believe it's only the `CVE`, `OSV`, and `ClearlyDefined` importers that use the "GitWalker". All others (except `Quay`) use HTTP, though it's abstracted by the [csaf-walker](https://github.com/scm-rs/csaf-walker) library.

docs/adrs/00021-http-transport-with-pluggable-discovery.md
```diff
@@ -0,0 +1,204 @@
...
+* **OCI registry** — pull artifacts from a container registry (used by the Quay importer).
+
+Neither supports importing data served over plain HTTP. This is a gap because advisory and
+vulnerability data providers increasingly publish content through HTTP-based content
+repositories rather than git repositories or container registries.
```
jcrossley3: Importing JSON docs over plain HTTP is exactly what `csaf-walker` does. Initially created specifically for CSAF docs, it was later extended to import SBOM's and, I believe, any JSON doc served by HTTP (@ctron can confirm).

docs/adrs/00021-http-transport-with-pluggable-discovery.md
```diff
@@ -0,0 +1,204 @@
...
+integrity data that the other approaches lack.
+
+### Pulp as a content distribution platform
+
+[Pulp](https://pulpproject.org/) is an open-source platform for managing and distributing
+content repositories. It is widely used across the Red Hat ecosystem and beyond for publishing
+software artifacts, advisories, and security data.
+
+Pulp file-content distributions configured to serve a publication expose a `PULP_MANIFEST` file
+at their root — a simple CSV format with three fields per line:
```
jcrossley3, line 50: I believe our CSAF source does something very similar. I expect the abstractions for discovering docs from a manifest file may already exist within csaf-walker.

jcrossley3 (commented):
I suspect that most if not all of this is a reimplementation of the behavior already provided by the csaf-walker library. I think I'd want its maintainer, @ctron, to inform the decision of whether reimplementing it is in the long-term best interest of the project.

ruromero:
> I suspect that most if not all of this is a reimplementation of the behavior already provided by the csaf-walker library. I think I'd want its maintainer, @ctron, to inform the decision of whether reimplementing it is in the long-term best interest of the project.

@jcrossley3 this is the architectural proposal, nothing about the implementation details. Indeed the `walker-common` can reuse the `Fetcher` (retry/rate-limit) and `RetrievedDigest` (hash verification). The authentication part is not implemented and will have to be contributed to the walker-common module or handled at the reqwest client level. The rest of the csaf-walker is specific to CSAF and for Pulp we need a different walker.

rh-jfuller (approved):
LGTM can you consider referencing what we already use/dep that could be considered for impl (agreed impl details are for PR implementing!)

jcrossley3:
> @jcrossley3 this is the architectural proposal, nothing about the implementation details. 

Ok, but the "Context" section is wrong. We fetch most docs via HTTP (using a manifest file) and only a few types with git/oci.

> The rest of the csaf-walker is specific to CSAF and for Pulp we need a different walker.

Not exactly. It fetches SBOM's, too. And it has extension points that could be used for Pulp. I'm not saying we should do that, but I think the existing context should be accurate for an ADR.

ruromero:
Thanks @jcrossley3 for your comments.

> > @jcrossley3 this is the architectural proposal, nothing about the implementation details.
> 
> Ok, but the "Context" section is wrong. We fetch most docs via HTTP (using a manifest file) and only a few types with git/oci.
> 
Updated.

> > The rest of the csaf-walker is specific to CSAF and for Pulp we need a different walker.
> 
> Not exactly. It fetches SBOM's, too. And it has extension points that could be used for Pulp. I'm not saying we should do that, but I think the existing context should be accurate for an ADR.

You're right. CSAF and SBOM already use HTTP via `sbom-walker`, and NVD, CWE, ClearlyDefined use plain reqwest. Fixed the context section.

PR #2615, opened 2026-09-01
fix(rpm): fix rpmver_cmp caret ordering and add epoch support

ctron:
## Summary

- **Fix caret (`^`) ordering bug** in `rpmver_cmp` — the PL/pgSQL function had `^` sorting *after* regular segments instead of *before*, contradicting upstream RPM's `rpmvercmp`. The bug was masked when `^` faced numeric segments (numeric check fires first) but produced wrong results against alpha segments.
- **Add epoch awareness** to `rpmver_cmp` — parses optional `N:version` prefix, compares epochs first (omitted = 0), falls through to version comparison when equal.
- **Add `Purl::effective_version()`** — folds the RPM epoch qualifier into the version string for UUID computation and storage without changing the `Purl` struct or its display/serialization. UI and exports continue to show `pkg:rpm/...@1.0?epoch=1`.
- **Use `effective_version()` in PurlCreator and CSAF creator** — ensures epoch consistency between SBOM and advisory version strings in `versioned_purl.version` and `version_range`.
- **Data migration** — backfills existing RPM `versioned_purl` entries with epoch > 0 by creating new rows with correct epoch-aware UUIDs and re-pointing `qualified_purl` FK references.
- **Port 47 test cases** from the upstream RPM test suite covering: basic numeric, alpha suffixes, mixed alpha-numeric, leading zeros, separator equivalence, tilde, caret, mixed tilde+caret, and epoch.

Implements TC-5733

## Test plan

- [x] `cargo xtask precommit` passes
- [x] All 47 `rpmver_cmp` test vectors pass (`cargo test -p trustify-module-ingestor --test ingestor rpmver`)
- [x] TC-5733 epoch mismatch test passes (`cargo test -p trustify-module-fundamental --lib s8_epoch -- --include-ignored`)
- [x] Existing purl tests unchanged (`cargo test -p trustify-common purl`)
- [x] All previously passing correlation tests still pass (S6, S7, S8, S11)

## Summary by Sourcery

Align RPM version comparison, PURL identity, and advisory matching with epoch- and caret-aware RPM semantics.

New Features:
- Add RPM epoch-aware version identity and comparison, including epoch-sensitive PURL storage and UUID generation.
- Support parsing `vers:` product version ranges in CSAF advisories.

Bug Fixes:
- Correct RPM caret ordering to match upstream RPM semantics.
- Fix unbounded version-range matching and enable previously ignored correlation scenarios.

Enhancements:
- Apply effective RPM versions consistently to SBOM and CSAF version records, including Red Hat fixed-version ranges.
- Backfill existing RPM versioned PURLs and update references for non-zero epochs.
- Validate unbounded version-range deserialization and expand RPM comparison coverage with upstream test vectors.

Build:
- Add the `vers-rs` dependency and required UUID system library installation in CI and benchmark workflows.

CI:
- Install the UUID system dependency in CI and benchmark jobs.

Tests:
- Add comprehensive RPM version comparison tests covering numeric, alpha-numeric, tilde, caret, separators, and epochs.
- Enable epoch mismatch and version-less known-affected correlation tests.

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -50,6 +50,9 @@ jobs:
       - uses: ./.github/actions/setup-rust
       - uses: Swatinem/rust-cache@v2
 
+      - name: Install system dependencies
+        run: sudo apt-get install -y libossp-uuid16
+
       - name: Setup cargo-binstall
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -50,9 +50,6 @@ jobs:
       - uses: ./.github/actions/setup-rust
       - uses: Swatinem/rust-cache@v2
 
-      - name: Install system dependencies
-        run: sudo apt-get install -y libossp-uuid16
-
       - name: Setup cargo-binstall
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/migration/src/m0002340_backfill_rpm_epoch.rs
+++ b/migration/src/m0002340_backfill_rpm_epoch.rs
@@ -1,34 +1,38 @@
+use sea_orm::{
+    ConnectionTrait, DatabaseBackend, EntityTrait, FromQueryResult, Set, Statement,
+    sea_query::OnConflict,
+};
 use sea_orm_migration::prelude::*;
+use std::collections::BTreeMap;
+use trustify_common::{db::chunk::EntityChunkedIter, purl::Purl};
+use trustify_entity::versioned_purl;
+use uuid::Uuid;
 
 #[derive(DeriveMigrationName)]
 pub struct Migration;
 
-const ENSURE_UUID_OSSP: &str = r#"CREATE EXTENSION IF NOT EXISTS "uuid-ossp""#;
+#[derive(Debug, FromQueryResult)]
+struct RpmEpochRow {
+    qp_id: Uuid,
+    epoch: String,
+    vp_version: String,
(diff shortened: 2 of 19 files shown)


.github/workflows/benchmark.yaml
```diff
@@ -50,6 +50,9 @@ jobs:
       - uses: ./.github/actions/setup-rust
       - uses: Swatinem/rust-cache@v2
 
+      - name: Install system dependencies
+        run: sudo apt-get install -y libossp-uuid16
```
rh-jfuller: this is a new system dep ... does this not need to flow into all envs (containers, helm, etc) ?

migration/src/m0002340_backfill_rpm_epoch.rs
```diff
@@ -0,0 +1,95 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+const ENSURE_UUID_OSSP: &str = r#"CREATE EXTENSION IF NOT EXISTS "uuid-ossp""#;
```
rh-jfuller: reaching out for uuid-ossp is usually an anti pattern ... I can see why we need to use it, but also would be great eventually to dep on default pg uuid funcs.

rh-jfuller (approved):
LGTM though ... though I wonder why the backfill is not being done in rust ... eg use `Purl::version_uuid()`  and drop new deps... for example do pg images have lib for uuid ossp installed ?

migration/src/m0002340_backfill_rpm_epoch.rs
ctron: Ugh, sorry. Didn't spot this.

PR #2616, opened 2026-09-01
Token auth AWS S3/RDS/IAM

desmax74:
Assisted-by: Claude: Opus 4.8

## Summary by Sourcery

Enable AWS token-based authentication for S3 and PostgreSQL, including OpenShift CCO Web Identity support and refreshed RDS IAM connection credentials.

New Features:
- Add AWS RDS/Aurora IAM database authentication using short-lived SigV4 tokens with automatic refresh for pooled connections.
- Enable S3 authentication through the AWS default credential provider chain, including OpenShift CCO Web Identity/STS credentials, when static keys are absent.

Bug Fixes:
- Avoid engaging AWS credential resolution when no AWS credential source is configured.
- Reject incompatible read-only database URL configuration when inherited RDS IAM authentication is enabled.

Enhancements:
- Enforce TLS for RDS IAM database connections and add shared AWS credential configuration detection.

Build:
- Add AWS credential and SigV4 dependencies required for token authentication.

Documentation:
- Document RDS/Aurora IAM and S3 CCO token authentication configuration and credential behavior.
- Document the new database IAM authentication environment variables.

Tests:
- Add coverage for AWS credential detection, RDS IAM token signing and encoding, TLS requirements, and IAM read-only database configuration.
- Add S3 construction coverage without static credentials.


common/src/db/mod.rs
```diff
@@ -138,6 +146,67 @@ impl Database {
...
+        .await
+        .context("failed to generate initial RDS IAM auth token")?;
+
+        let pool = PgPoolOptions::new()
+            .max_connections(database.max_conn)
+            .min_connections(database.min_conn)
+            .acquire_timeout(Duration::from_secs(database.acquire_timeout))
+            .max_lifetime(Duration::from_secs(database.max_lifetime))
+            .idle_timeout(Duration::from_secs(database.idle_timeout))
+            .connect_with(pg_connect_options(database, ssl_mode, &token))
```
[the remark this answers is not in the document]
desmax74: it's actually a false positive — applying connect_timeout to the IAM pool would have no effect, because there's no distinct connection-establishment timeout to configure at this layer.

common/src/config.rs
```diff
@@ -147,6 +161,11 @@ impl Database {
                     .map_err(|s| anyhow!("Failed to convert '{s}' to SslMode"))?,
                 _ => Default::default(),
             },
+            iam_auth: match env::var(ENV_DB_IAM_AUTH) {
+                Ok(s) => s.parse::<bool>().unwrap_or(false),
```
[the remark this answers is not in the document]
desmax74: Replaced with a mapping of the error

desmax74:
@sourcery-ai review

common/src/db/mod.rs
```diff
@@ -675,6 +752,87 @@ fn strip_password(url: String) -> String {
...
+    pool: PgPool,
+    database: crate::config::Database,
+    region: String,
+    ssl_mode: PgSslMode,
+) {
+    tokio::spawn(async move {
+        loop {
+            tokio::time::sleep(rds_iam::RDS_IAM_TOKEN_REFRESH).await;
+
+            if pool.is_closed() {
```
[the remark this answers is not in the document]
desmax74: Fixed

common/src/aws.rs
```diff
@@ -0,0 +1,82 @@
...
+const AWS_CREDENTIAL_ENV_VARS: &[&str] = &[
+    "AWS_ACCESS_KEY_ID",
+    "AWS_WEB_IDENTITY_TOKEN_FILE",
+    "AWS_ROLE_ARN",
+    "AWS_SHARED_CREDENTIALS_FILE",
+    "AWS_CONFIG_FILE",
+    "AWS_PROFILE",
+    "AWS_CONTAINER_CREDENTIALS_RELATIVE_URI",
+    "AWS_CONTAINER_CREDENTIALS_FULL_URI",
+];
```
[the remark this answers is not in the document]
desmax74: Fixed adding a new env vars to signal the ec2 instance with IMDS

common/src/db/mod.rs
```diff
@@ -108,6 +110,12 @@ pub struct Database {
 impl Database {
     #[instrument(skip(database), fields(database = ?crate::redact::HideString(database, &database.password.0)), err(level=tracing::Level::INFO))]
     pub async fn new(database: &crate::config::Database) -> Result<Self, anyhow::Error> {
+        // RDS/Aurora IAM authentication uses a short-lived, auto-refreshed token as the
+        // password and requires a dedicated, self-refreshing connection pool.
+        if database.iam_auth {
+            return Self::new_with_iam_auth(database).await;
+        }
+
         let url = database.to_url();
```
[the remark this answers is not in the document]
desmax74: Fixed

common/src/config.rs
```diff
@@ -234,6 +253,10 @@ impl DatabaseReadOnly {
             max_conn: self.max_conn.unwrap_or(fallback.max_conn),
             min_conn: self.min_conn.unwrap_or(fallback.min_conn),
             sslmode: self.sslmode.unwrap_or(fallback.sslmode),
+            // IAM authentication settings are inherited from the R/W config; a read-only
+            // replica reached with IAM auth uses the same mechanism and region.
+            iam_auth: fallback.iam_auth,
+            region: fallback.region.clone(),
```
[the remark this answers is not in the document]
desmax74: Fixed

common/src/config.rs
```diff
@@ -234,6 +257,10 @@ impl DatabaseReadOnly {
...
```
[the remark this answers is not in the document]
desmax74: Fixed

docs/design/rds-iam-cco-token-authentication.md
```diff
@@ -0,0 +1,156 @@
+# RDS/Aurora — CCO Token Authentication (IAM database authentication)
+
+## Purpose
+
+Trustify (RHTPA) stores its data in a PostgreSQL database. On AWS, that database can be an
```
ctron: this is upstream. please remove RHTPA.

common/src/config.rs
```diff
@@ -88,6 +90,18 @@ pub struct Database {
...
+    /// static password. When enabled, the password is generated (and periodically
+    /// refreshed) from AWS credentials and `region`, and TLS is forced on (RDS IAM
+    /// authentication mandates TLS). Set by the RHTPA operator when CCO-backed RDS IAM
+    /// authentication (`ccoRds`) is enabled.
+    #[arg(id="db-iam-auth", long, env = ENV_DB_IAM_AUTH, default_value_t = false, conflicts_with = "db-url")]
+    pub iam_auth: bool,
+
+    /// AWS region of the RDS/Aurora instance. Required when `iam_auth` is enabled.
+    #[arg(id="db-region", long, env = ENV_DB_REGION, conflicts_with = "db-url")]
+    pub region: Option<String>,
```
ctron, line 103: If that's required, then this should be enforced with clap.

docs/design/rds-iam-cco-token-authentication.md
```diff
@@ -0,0 +1,156 @@
...
+Both conflict with `--db-url` (`TRUSTD_DB_URL`): IAM authentication builds the connection
+from the individual host/port/user/name fields and injects the generated token as the
+password, so a pre-built DSN is not supported. The static `--db-password`
+(`TRUSTD_DB_PASSWORD`) is ignored when IAM auth is enabled.
+
+The read-only database config (`DatabaseReadOnly`) inherits `iam_auth` and `region` from
+the read-write config: a read replica reached with IAM auth uses the same mechanism and
+region.
+
+The RHTPA operator sets `TRUSTD_DB_IAM_AUTH=true` and `TRUSTD_DB_REGION` (and omits the DB
```
ctron: Please don't use RHTPA here, this is upstream. Some for all other locations.

docs/design/s3-cco-token-authentication.md
```diff
@@ -0,0 +1,131 @@
+# S3 Storage — CCO Token Authentication (OpenShift STS / Web Identity)
```
ctron, line 1: Reads like a lot of duplication. That should go into the same document.

docs/env-vars.md
```diff
@@ -73,6 +75,35 @@
 | `EXPLOIT_INTELLIGENCE_OIDC_REFRESH_BEFORE`| Duration an EI access token must still be valid before requesting a new one        | `30s`                                   |
 | `EXPLOIT_INTELLIGENCE_OIDC_TLS_INSECURE` | Allow insecure TLS connections with the EI OIDC issuer                              | `false`                                 |
 
+### S3 credentials
```
ctron: Feels like the wrong approach for this document,.

modules/storage/src/service/s3.rs
```diff
@@ -103,12 +104,53 @@ impl S3Backend {
             config.region(Region::new(region))
         };
 
-        // credentials
-
-        if let Some((key_id, access_key)) = access_key.zip(secret_key) {
-            let credentials = Credentials::new(key_id, access_key, None, None, "config");
-            config = config.credentials_provider(credentials);
-        }
+        // Credentials
```
ctron: This feels a bit verbose, can we extract that into a function?

modules/storage/src/service/s3.rs
```diff
@@ -103,12 +104,53 @@ impl S3Backend {
...
+            None if aws_credentials_configured() => {
+                let shared = aws_config::defaults(BehaviorVersion::latest()).load().await;
+                match shared.credentials_provider() {
+                    Some(provider) => config.credentials_provider(provider),
+                    None => config,
+                }
+            }
+            None => {
+                log::info!(
+                    "No S3 static credentials and no AWS credential environment detected; \
```
ctron: don't to do line breaks like this.

common/src/aws.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    // No credential env var is set; honour the explicit IMDS opt-in.
+    get(AWS_USE_IMDS_ENV_VAR)
+        .and_then(|value| value.into_string().ok())
+        .is_some_and(|value| is_truthy(&value))
+}
+
+/// Interpret an environment-variable value as a boolean opt-in flag.
+fn is_truthy(value: &str) -> bool {
```
ctron: We're never doing this. Don't start. Just `true` and `false`. Or just use the `SetTrue` action in clap.

common/src/aws.rs
```diff
@@ -0,0 +1,154 @@
...
+const AWS_USE_IMDS_ENV_VAR: &str = "TRUSTD_AWS_USE_IMDS";
+
+/// Returns `true` if the environment is configured for the AWS default credential provider
+/// chain: either one of [`AWS_CREDENTIAL_ENV_VARS`] is present, or the caller has opted into
+/// IMDS via [`AWS_USE_IMDS_ENV_VAR`].
+///
+/// Use this to decide whether to engage AWS at all: when it returns `false`, callers should
+/// skip loading the default chain and avoid making AWS-authenticated connections.
+pub fn aws_credentials_configured() -> bool {
+    credentials_configured_with(|name| std::env::var_os(name))
```
ctron: import

common/src/aws.rs
```diff
@@ -0,0 +1,154 @@
...
+/// the full chain, including IMDS.
+const AWS_USE_IMDS_ENV_VAR: &str = "TRUSTD_AWS_USE_IMDS";
+
+/// Returns `true` if the environment is configured for the AWS default credential provider
+/// chain: either one of [`AWS_CREDENTIAL_ENV_VARS`] is present, or the caller has opted into
+/// IMDS via [`AWS_USE_IMDS_ENV_VAR`].
+///
+/// Use this to decide whether to engage AWS at all: when it returns `false`, callers should
+/// skip loading the default chain and avoid making AWS-authenticated connections.
+pub fn aws_credentials_configured() -> bool {
```
ctron, line 58: Why do we need two functions for this?

common/src/db/rds_iam.rs
```diff
@@ -0,0 +1,211 @@
...
+    credentials: Credentials,
+    time: SystemTime,
+) -> anyhow::Result<String> {
+    let identity = credentials.into();
+
+    // Presign a request to the rds-db service, placing the signature in the query string
+    // and setting the token's validity window.
+    let mut settings = SigningSettings::default();
+    settings.signature_location = SignatureLocation::QueryParams;
+    settings.expires_in = Some(RDS_IAM_TOKEN_EXPIRY);
```
ctron, line 84: I guess this should be configurable.

common/src/db/rds_iam.rs
```diff
@@ -0,0 +1,211 @@
...
+};
+use reqwest::Url;
+use std::time::{Duration, SystemTime};
+
+/// Lifetime of a generated RDS IAM authentication token (AWS fixes this at 15 minutes).
+pub const RDS_IAM_TOKEN_EXPIRY: Duration = Duration::from_secs(15 * 60);
+
+/// How long before token expiry a pooled connection's token should be refreshed. Chosen
+/// comfortably below [`RDS_IAM_TOKEN_EXPIRY`] so new connections never see a stale token.
+pub const RDS_IAM_TOKEN_REFRESH: Duration = Duration::from_secs(10 * 60);
```
ctron, line 30: I guess this should be turned into a calculated value: refresh = expiry - grace

common/src/db/rds_iam.rs
```diff
@@ -0,0 +1,211 @@
...
+//! `manual` mode). See `docs/design/rds-iam-cco-token-authentication.md`.
+
+use anyhow::{Context, anyhow};
+use aws_config::BehaviorVersion;
+use aws_credential_types::{Credentials, provider::ProvideCredentials};
+use aws_sigv4::{
+    http_request::{SignableBody, SignableRequest, SignatureLocation, SigningSettings, sign},
+    sign::v4,
+};
+use reqwest::Url;
```
ctron: Do we need request? Or can we use `url::Url`?

common/src/db/mod.rs
```diff
@@ -675,6 +764,166 @@ fn strip_password(url: String) -> String {
...
+/// `is_closed()`, which would otherwise leak the task and its connections.
+fn spawn_iam_token_refresher(
+    pool: PgPool,
+    database: crate::config::Database,
+    region: String,
+    ssl_mode: PgSslMode,
+) -> tokio::task::JoinHandle<()> {
+    tokio::spawn(async move {
+        loop {
+            tokio::time::sleep(rds_iam::RDS_IAM_TOKEN_REFRESH).await;
```
ctron: Please import

common/src/db/mod.rs
```diff
@@ -675,6 +764,166 @@ fn strip_password(url: String) -> String {
...
+/// so the task is aborted once the last `Database` clone is dropped; this is required because
+/// the task owns a strong `PgPool` clone and a merely-dropped (never-closed) pool never reports
+/// `is_closed()`, which would otherwise leak the task and its connections.
+fn spawn_iam_token_refresher(
+    pool: PgPool,
+    database: crate::config::Database,
+    region: String,
+    ssl_mode: PgSslMode,
+) -> tokio::task::JoinHandle<()> {
+    tokio::spawn(async move {
```
ctron: Please import

common/src/db/mod.rs
```diff
@@ -675,6 +764,166 @@ fn strip_password(url: String) -> String {
...
+        .await
+        .context("failed to connect to database using RDS IAM authentication (admin)")?;
+
+    Ok(SqlxPostgresConnector::from_sqlx_postgres_pool(pool))
+}
+
+/// Build the postgres connect options for an IAM-authenticated connection, using the
+/// given short-lived IAM token as the password.
+fn pg_connect_options(
+    database: &crate::config::Database,
```
ctron: Please import. same for all others.

common/src/db/mod.rs
```diff
@@ -675,6 +764,166 @@ fn strip_password(url: String) -> String {
...
+        SslMode::Disable => PgSslMode::Disable,
+        SslMode::Allow => PgSslMode::Allow,
+        SslMode::Prefer => PgSslMode::Prefer,
+        SslMode::Require => PgSslMode::Require,
+        SslMode::VerifyCa => PgSslMode::VerifyCa,
+        SslMode::VerifyFull => PgSslMode::VerifyFull,
+    }
+}
+
+/// Ensure the SSL mode enforces encryption, as required by RDS IAM authentication.
```
ctron, line 771: I think it makes more sense failing, rather than just changing things.

common/src/db/mod.rs
```diff
@@ -675,6 +764,166 @@ fn strip_password(url: String) -> String {
     }
 }
 
+/// Map trustify's [`crate::config::SslMode`] to sqlx's [`PgSslMode`].
+fn to_pg_ssl_mode(mode: crate::config::SslMode) -> PgSslMode {
```
ctron: implement `From`

common/src/db/mod.rs
```diff
@@ -103,11 +106,37 @@ pub struct Database {
     db: DatabaseConnection,
     /// the database name
     name: String,
+    /// Keeps the RDS IAM token refresher task alive for exactly as long as any `Database`
+    /// clone lives. Non-IAM connections have no refresher, so this is `None` for them. When
+    /// the last clone is dropped, the [`AbortOnDrop`] aborts the task, which in turn drops
+    /// its `PgPool` clone and lets the pool (and its connections) be released.
+    _iam_refresher: Option<Arc<AbortOnDrop>>,
```
ctron: Instead of relying on drop support and "aborting", use a "one shot" channel to close it

common/src/db/mod.rs
```diff
@@ -675,6 +764,166 @@ fn strip_password(url: String) -> String {
...
+/// `database`.
+///
+/// Bootstrap/setup connect to the `postgres` maintenance database to create (or drop) the
+/// target database. When RDS IAM authentication is enabled the static password is absent or
+/// invalid, so this generates a one-off IAM token (with TLS forced on, as IAM mandates)
+/// rather than relying on [`crate::config::Database::to_url`]'s static password. Unlike
+/// [`Database::new`] no token refresher is spawned: the connection is used once, well within
+/// a single token's lifetime, and then closed.
+#[instrument(skip(database), fields(host = database.host, name = database.name), err(level=tracing::Level::INFO))]
+pub async fn connect_admin(
```
ctron, line 824: I'm not sure why we need another function to connect? Don't we already have one?

docs/env-vars.md
desmax74: moved is the specific docs

modules/storage/src/service/s3.rs
desmax74: It' the cargo fmt, if I put inlined the build fails

docs/design/rds-iam-cco-token-authentication.md
desmax74: cleaned

docs/design/rds-iam-cco-token-authentication.md
desmax74: cleaned

common/src/db/rds_iam.rs
desmax74: AWS fixes this at 15 minutes, it's an hardcoded value https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.Troubleshooting.html

modules/storage/src/service/s3.rs
desmax74: Done

common/src/aws.rs
desmax74: Removed

common/src/aws.rs
desmax74: Exists for testability, not because the logic needs two functions. 

  - aws_credentials_configured() is the real public API. It hardwires the variable lookup to std::env::var_os — i.e. it reads the actual process environment.
  - credentials_configured_with(get) holds all the actual logic, but takes the lookup strategy as a parameter (impl FnMut(&str) -> Option<OsString>).

  The whole point of that parameter is so the tests (lines 90–139) can inject a fake environment via set_vars(...) instead of touching the real one.

common/src/db/mod.rs
desmax74: Ok, refactored

common/src/db/mod.rs
desmax74: Done

common/src/db/mod.rs
desmax74: Agree, changed from `never drop below `Require`` to ` reject anything weaker than `Require``. Changed require_tls  to fail instead of silently promoting weaker SSL modes.

common/src/db/mod.rs
desmax74: Ok

common/src/db/mod.rs
desmax74: ok

common/src/db/mod.rs
desmax74: connect_admin is essentially "the IAM-aware, no-refresher, single-connection variant" needed for the DB-creation bootstrap step, which is a genuinely distinct scenario from opening the app's main pool.

common/src/db/rds_iam.rs
desmax74: Nope, we can just use `url::URL`. Refactored

common/src/aws.rs
desmax74: ok

common/src/db/rds_iam.rs
desmax74: Expiry is fixed to 15 minutes, so the duration is 5 minutes, do we want to let configure this duration of refresh ?

common/src/config.rs
desmax74: Yep, added

docs/design/s3-cco-token-authentication.md
desmax74: Moved all in this doc

common/src/aws.rs
ctron: Makes sense. Maybe a comment in the code helps understand.

common/src/aws.rs
```diff
@@ -0,0 +1,146 @@
...
+        .is_some_and(|value| value == "true")
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use std::ffi::OsString;
+
+    /// Build a lookup that reports the given variables as set to a non-empty value.
+    fn set_vars<'a>(vars: &'a [(&'a str, &'a str)]) -> impl FnMut(&str) -> Option<OsString> + 'a {
```
ctron: Not sure if the name is a good fit then, this doesn't "set" anything.

docs/design/s3-cco-token-authentication.md
ctron: I can still see the other doc.

docs/env-vars.md
```diff
@@ -43,9 +43,11 @@
...
-| `TRUSTD_DB_PASSWORD`                     | Database password                                                                   | `trustify`                              |
+| `TRUSTD_DB_PASSWORD`                     | Database password (ignored when IAM authentication is enabled)                      | `trustify`                              |
 | `TRUSTD_DB_PORT`                         | Database port                                                                       | `5432`                                  |
 | `TRUSTD_DB_USER`                         | Database username                                                                   | `postgres`                              |
+| `TRUSTD_DB_IAM_AUTH`                     | Authenticate with an AWS RDS/Aurora IAM token instead of a password                 | `false`                                 |
+| `TRUSTD_DB_REGION`                       | AWS region of the RDS/Aurora instance (required when `TRUSTD_DB_IAM_AUTH=true`)     |                                         |
```
ctron: If this is IAM specific, wouldn't it make sense to prefix this with `TRUSTD_DB_IAM_`?

common/src/db/rds_iam.rs
```diff
@@ -0,0 +1,211 @@
...
+use aws_credential_types::{Credentials, provider::ProvideCredentials};
+use aws_sigv4::{
+    http_request::{SignableBody, SignableRequest, SignatureLocation, SigningSettings, sign},
+    sign::v4,
+};
+use url::Url;
+use std::time::{Duration, SystemTime};
+
+/// Lifetime of a generated RDS IAM authentication token (AWS fixes this at 15 minutes).
+pub const RDS_IAM_TOKEN_EXPIRY: Duration = Duration::from_secs(15 * 60);
```
ctron: Please use: `https://doc.rust-lang.org/beta/std/time/struct.Duration.html#method.from_mins`

common/src/db/rds_iam.rs
ctron: It wouldn't be configurable. But it would explain what's going on. as everything is `const`, that would only be calculated during build time.

common/src/db/mod.rs
ctron: Ok, maybe it makes sense splitting this up into three functions:
* `connect_admin`: this one, with the if
* `connect_iam`: the IAM single connect version
* `connect`: the "normal" version

common/src/db/mod.rs
```diff
@@ -675,6 +753,182 @@ fn strip_password(url: String) -> String {
...
+    log::info!(
+        "connecting to {}:{} db '{}' as '{}' using RDS IAM authentication (admin, region: {region}, sslmode: {ssl_mode:?})",
+        database.host,
+        database.port,
+        database.name,
+        database.username,
+    );
+
+    let token =
+        rds_iam::generate_rds_iam_token(&database.host, database.port, &database.username, &region)
```
ctron: Please import `generate_rds_iam_token`

docs/design/s3-cco-token-authentication.md
desmax74: One is for S3 and one for rds which are two distinct components, going to collapse in a cco-token-aws

docs/env-vars.md
desmax74: Done

common/src/db/mod.rs
```diff
@@ -675,6 +753,178 @@ fn strip_password(url: String) -> String {
     }
 }
 
+/// Map trustify's [`crate::config::SslMode`] to sqlx's [`PgSslMode`].
+impl From<crate::config::SslMode> for PgSslMode {
```
ctron: Please import types.

common/src/db/mod.rs
```diff
@@ -675,6 +753,178 @@ fn strip_password(url: String) -> String {
...
+    let region = database.region.clone().ok_or_else(|| {
+        anyhow::anyhow!(
+            "'--db-region' (TRUSTD_DB_IAM_REGION) is required when IAM authentication is enabled"
+        )
+    })?;
+
+    // RDS IAM tokens are signed with AWS credentials from the default provider chain. If
+    // nothing in the environment points at a credential source, fail fast.
+    anyhow::ensure!(
+        crate::aws::aws_credentials_configured(),
```
ctron: Please import

common/src/db/mod.rs
```diff
@@ -675,6 +753,178 @@ fn strip_password(url: String) -> String {
...
+                    break;
+                }
+            }
+
+            if pool.is_closed() {
+                log::debug!("database pool closed; stopping RDS IAM token refresher");
+                break;
+            }
+
+            match rds_iam::generate_rds_iam_token(
```
ctron: Please import

common/src/db/mod.rs
```diff
@@ -675,6 +753,178 @@ fn strip_password(url: String) -> String {
...
+    pool: PgPool,
+    database: DatabaseConfig,
+    region: String,
+    ssl_mode: PgSslMode,
+) -> oneshot::Sender<()> {
+    let (shutdown_tx, mut shutdown_rx) = oneshot::channel::<()>();
+    spawn(async move {
+        loop {
+            tokio::select! {
+                _ = time::sleep(rds_iam::RDS_IAM_TOKEN_REFRESH) => {}
```
ctron: Please import

common/src/db/rds_iam.rs
```diff
@@ -0,0 +1,211 @@
...
+/// AWS SDKs' `generate_db_auth_token` helpers produce.
+pub async fn generate_rds_iam_token(
+    host: &str,
+    port: u16,
+    username: &str,
+    region: &str,
+) -> anyhow::Result<String> {
+    // Resolve AWS credentials from the default provider chain, pinned to the DB region.
+    let sdk_config = aws_config::defaults(BehaviorVersion::latest())
+        .region(aws_config::Region::new(region.to_string()))
```
ctron: Please import

common/src/db/rds_iam.rs
```diff
@@ -0,0 +1,211 @@
...
+/// for the `rds-db` service, with the scheme stripped — this is exactly the format the
+/// AWS SDKs' `generate_db_auth_token` helpers produce.
+pub async fn generate_rds_iam_token(
+    host: &str,
+    port: u16,
+    username: &str,
+    region: &str,
+) -> anyhow::Result<String> {
+    // Resolve AWS credentials from the default provider chain, pinned to the DB region.
+    let sdk_config = aws_config::defaults(BehaviorVersion::latest())
```
ctron: Please import

common/src/db/rds_iam.rs
```diff
@@ -0,0 +1,211 @@
...
+    let mut token_url =
+        Url::parse(&format!("https://{host}:{port}/")).context("failed to build RDS IAM url")?;
+    token_url
+        .query_pairs_mut()
+        .append_pair("Action", "connect")
+        .append_pair("DBUser", username);
+
+    let url = token_url.to_string();
+    let signable_request =
+        SignableRequest::new("GET", &url, std::iter::empty(), SignableBody::Bytes(&[]))
```
ctron: Please import

common/src/aws.rs
```diff
@@ -0,0 +1,146 @@
...
+///
+/// Use this to decide whether to engage AWS at all: when it returns `false`, callers should
+/// skip loading the default chain and avoid making AWS-authenticated connections.
+pub fn aws_credentials_configured() -> bool {
+    credentials_configured_with(|name| env::var_os(name))
+}
+
+/// Core of [`aws_credentials_configured`], parameterised over how a variable is looked up so it
+/// can be exercised without mutating process-wide environment state.
+fn credentials_configured_with(mut get: impl FnMut(&str) -> Option<std::ffi::OsString>) -> bool {
```
ctron: Please import

docs/env-vars.md
```diff
@@ -43,9 +43,11 @@
...
-| `TRUSTD_DB_PASSWORD`                     | Database password                                                                   | `trustify`                              |
+| `TRUSTD_DB_PASSWORD`                     | Database password (ignored when IAM authentication is enabled)                      | `trustify`                              |
 | `TRUSTD_DB_PORT`                         | Database port                                                                       | `5432`                                  |
 | `TRUSTD_DB_USER`                         | Database username                                                                   | `postgres`                              |
+| `TRUSTD_DB_IAM_AUTH`                     | Authenticate with an AWS RDS/Aurora IAM token instead of a password                 | `false`                                 |
+| `TRUSTD_DB_IAM_REGION`                       | AWS region of the RDS/Aurora instance (required when `TRUSTD_DB_IAM_AUTH=true`)     |                                         |
```
ctron: Pleas format the table

modules/storage/src/service/s3.rs
```diff
@@ -145,6 +141,59 @@ impl S3Backend {
...
+            config.credentials_provider(credentials)
+        }
+        // No static keys: only engage the AWS default credential provider chain when the
+        // environment actually points at an AWS credential source (e.g. CCO manual/STS
+        // mode sets `AWS_ROLE_ARN` + `AWS_WEB_IDENTITY_TOKEN_FILE`, or an EC2 deployment
+        // opts into IMDS with `TRUSTD_AWS_USE_IMDS=true`). Loading the chain otherwise just
+        // costs time and resolves to nothing, so skip AWS entirely — the client is left
+        // without a credentials provider, matching a non-AWS backend.
+        None if aws_credentials_configured() => {
+            let shared = aws_config::defaults(BehaviorVersion::latest()).load().await;
```
ctron: Please import

common/src/db/mod.rs
desmax74: Done

common/src/aws.rs
desmax74: `lookup_vars`

docs/env-vars.md
desmax74: Done

common/src/db/rds_iam.rs
desmax74: Done, both converted with `from_mins`

common/src/db/mod.rs
desmax74: Done

common/src/db/mod.rs
desmax74: Done

modules/storage/src/service/s3.rs
desmax74: Done

common/src/db/rds_iam.rs
```diff
@@ -0,0 +1,211 @@
...
+};
+use std::time::{Duration, SystemTime};
+use url::Url;
+
+/// Lifetime of a generated RDS IAM authentication token (AWS fixes this at 15 minutes).
+pub const RDS_IAM_TOKEN_EXPIRY: Duration = Duration::from_mins(15);
+
+/// How long before token expiry a pooled connection's token should be refreshed. Chosen
+/// comfortably below [`RDS_IAM_TOKEN_EXPIRY`] so new connections never see a stale token.
+pub const RDS_IAM_TOKEN_REFRESH: Duration = Duration::from_mins(10);
```
ctron: I still think this should be calculated.

common/src/aws.rs
desmax74: Done

common/src/db/rds_iam.rs
desmax74: Done

common/src/db/rds_iam.rs
desmax74: Done

common/src/db/mod.rs
desmax74: Done

common/src/db/rds_iam.rs
desmax74: Done

common/src/aws.rs
desmax74: Done

common/src/db/mod.rs
desmax74: Done

common/src/db/mod.rs
desmax74: Done

common/src/db/rds_iam.rs
desmax74: Why if is fixed and the only requisite is to be under 15 minutes ?

ctron: because it lays out the path how you got to 10 mins. And it allows tweaking it in the future without breaking it.

common/src/db/rds_iam.rs
```diff
@@ -0,0 +1,213 @@
...
+use anyhow::{Context, anyhow};
+use aws_config::{BehaviorVersion, Region, defaults};
+use aws_credential_types::{Credentials, provider::ProvideCredentials};
+use aws_sigv4::{
+    http_request::{SignableBody, SignableRequest, SignatureLocation, SigningSettings, sign},
+    sign::v4,
+};
+
+use std::iter;
+use std::time::{Duration, SystemTime};
```
ctron: Please nest imports.

common/src/aws.rs
```diff
@@ -0,0 +1,154 @@
...
+/// IMDS via [`AWS_USE_IMDS_ENV_VAR`].
+///
+/// Use this to decide whether to engage AWS at all: when it returns `false`, callers should
+/// skip loading the default chain and avoid making AWS-authenticated connections.
+/// NOTE Exists for testability, not because the logic needs two functions.
+///
+///  aws_credentials_configured() is the real public API. It hardwires the variable lookup to std::env::var_os — i.e. it reads the actual process environment.
+///  credentials_configured_with(get) holds all the actual logic, but takes the lookup strategy as a parameter (impl FnMut(&str) -> Option).
+///
+/// The whole point of that parameter is so the tests can inject a fake environment via set_vars(...) instead of touching the real one.
```
ctron: Guess it's `lookup_vars` now and could be linked.

common/src/aws.rs
```diff
@@ -0,0 +1,154 @@
...
+    fn imds_opt_in_is_configured() {
+        assert!(
+            credentials_configured_with(lookup_vars(&[(AWS_USE_IMDS_ENV_VAR, "true")])),
+            "expected \"true\" to opt into IMDS"
+        );
+    }
+
+    #[test]
+    fn imds_opt_in_falsey_is_not_configured() {
+        for value in ["false", "TRUE", "1", "yes", "on", " true ", ""] {
```
ctron: Looks like a candidate for `rstest`.

common/src/db/rds_iam.rs
```diff
@@ -0,0 +1,213 @@
...
+
+use anyhow::{Context, anyhow};
+use aws_config::{BehaviorVersion, Region, defaults};
+use aws_credential_types::{Credentials, provider::ProvideCredentials};
+use aws_sigv4::{
+    http_request::{SignableBody, SignableRequest, SignatureLocation, SigningSettings, sign},
+    sign::v4,
+};
+
+use std::iter;
```
ctron: Please nest imports.

common/src/db/rds_iam.rs
```diff
@@ -0,0 +1,213 @@
...
+use std::time::{Duration, SystemTime};
+use url::Url;
+
+/// Lifetime of a generated RDS IAM authentication token (AWS fixes this at 15 minutes).
+pub const RDS_IAM_TOKEN_EXPIRY: Duration = Duration::from_mins(15);
+
+/// How long before token expiry a pooled connection's token should be refreshed. Chosen
+/// comfortably below [`RDS_IAM_TOKEN_EXPIRY`] so new connections never see a stale token.
+pub const RDS_IAM_TOKEN_REFRESH: Duration = Duration::from_mins(10);
```
ctron: Please calculate: refresh = expiry - grace

common/src/db/rds_iam.rs
desmax74: Done

common/src/aws.rs
desmax74: Good spot, done.

common/src/db/rds_iam.rs
desmax74: Done

common/src/aws.rs
desmax74: Done

common/src/db/rds_iam.rs
desmax74: Done

common/src/db/rds_iam.rs
desmax74: Ok, Done

common/src/aws.rs
ctron: You still need to parametrize the test.

common/src/db/mod.rs
```diff
@@ -4,30 +4,43 @@ pub mod limiter;
...
 
 mod create;
 mod func;
 
 pub use create::*;
 pub use func::*;
 
+use crate::aws::aws_credentials_configured;
+use crate::config::Database as DatabaseConfig;
+use crate::config::SslMode;
```
ctron: Please nest imports.

common/src/db/mod.rs
```diff
@@ -675,6 +756,179 @@ fn strip_password(url: String) -> String {
...
+#[instrument(skip(database), fields(host = database.host, name = database.name), err(level=tracing::Level::INFO))]
+pub async fn connect_admin(database: &DatabaseConfig) -> Result<DatabaseConnection, anyhow::Error> {
+    if !database.iam_auth {
+        return connect_iam(database).await;
+    }
+
+    connect(database).await?
+}
+
+async fn connect(database: &DatabaseConfig) -> Result<Result<DatabaseConnection, Error>, Error> {
```
ctron: Why do we need a double wrapped result?

common/src/db/mod.rs
```diff
@@ -724,4 +978,27 @@ mod test {
...
+        for mode in [PgSslMode::Disable, PgSslMode::Allow, PgSslMode::Prefer] {
+            assert!(
+                require_tls(mode).is_err(),
+                "sslmode {mode:?} must be rejected under RDS IAM authentication"
+            );
+        }
+    }
+
+    #[test]
+    fn require_tls_accepts_require_and_stricter_modes() {
```
ctron: require_tls_rejects_modes_weaker_than_require and require_tls_accepts_require_and_stricter_modes can be combined in a single rstest.

common/src/db/mod.rs
desmax74: Rover automatic refactor..., now fixed.

common/src/db/mod.rs
desmax74: Done

common/src/db/mod.rs
desmax74: Done

common/src/db/mod.rs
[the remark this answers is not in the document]
ctron: I'm not sure that's correct. The `pg_connect_options` function never sets the connection timeout. Meaning, that there would be no timeout for the connect operation?

common/src/aws.rs
```diff
@@ -0,0 +1,155 @@
...
+    #[test]
+    fn imds_opt_in_is_configured() {
+        assert!(
+            credentials_configured_with(lookup_vars(&[(AWS_USE_IMDS_ENV_VAR, "true")])),
+            "expected \"true\" to opt into IMDS"
+        );
+    }
+
+    #[rstest]
+    fn imds_opt_in_falsey_is_not_configured() {
```
ctron: This needs to be parametrized

ctron:
@sourcery-ai review

common/src/db/mod.rs
```diff
@@ -4,30 +4,42 @@ pub mod limiter;
...
 use sea_orm::{
     AccessMode, ConnectOptions, ConnectionTrait, DatabaseConnection, DatabaseTransaction,
-    DbBackend, DbErr, ExecResult, IsolationLevel, QueryResult, RuntimeErr, Statement, StreamTrait,
-    TransactionError, TransactionTrait, prelude::async_trait,
+    DbBackend, DbErr, ExecResult, IsolationLevel, QueryResult, RuntimeErr, SqlxPostgresConnector,
+    Statement, StreamTrait, TransactionError, TransactionTrait, prelude::async_trait,
 };
+
+use rds_iam::generate_rds_iam_token;
+
```
ctron: remove those empty lines

common/src/db/rds_iam.rs
```diff
@@ -0,0 +1,218 @@
...
+//! `manual` mode). See `docs/design/rds-iam-cco-token-authentication.md`.
+
+use anyhow::{Context, anyhow};
+use aws_config::{BehaviorVersion, Region, defaults};
+use aws_credential_types::{Credentials, provider::ProvideCredentials};
+use aws_sigv4::{
+    http_request::{SignableBody, SignableRequest, SignatureLocation, SigningSettings, sign},
+    sign::v4,
+};
+
```
ctron: remove this empty line

desmax74: Done

common/src/db/mod.rs
desmax74: Done

common/src/db/rds_iam.rs
desmax74: Done

common/src/aws.rs
desmax74: Done

common/src/aws.rs
desmax74: Done

common/src/db/mod.rs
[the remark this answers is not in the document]
desmax74: Short explanation: `new_with_iam_auth` sets acquire_timeout, max_lifetime, idle_timeout but not connect_timeout — but it has no behavioral effect. `PgConnectOptions` has no connect-timeout knob at all, and in SeaORM `connect_timeout` is just an alias for sqlx's `acquire_timeout`.

Long explanation: ```
 The observation is literally true — new_with_iam_auth sets acquire_timeout, max_lifetime, idle_timeout but not connect_timeout — but it has no behavioral effect. PgConnectOptions has no connect-timeout knob at all, and in SeaORM connect_timeout is just an alias for sqlx's acquire_timeout that the next line (acquire_timeout) overwrites. So
  both paths end up with identical PoolOptions.

  Doc / source evidence

  1. SeaORM's connect_timeout is documented as an acquire timeout, not a TCP/connect timeout

  sea-orm-1.1.20/src/database/mod.rs:233:
  /// Set the timeout duration when acquiring a connection
  pub fn connect_timeout(&mut self, value: Duration) -> &mut Self {
  (docs.rs: https://docs.rs/sea-orm/1.1.20/sea_orm/struct.ConnectOptions.html#method.connect_timeout)

  2. SeaORM folds both fields into the same sqlx setting, with acquire_timeout applied last

  sea-orm-1.1.20/src/driver/sqlx_common.rs:51-59:
  if let Some(connect_timeout) = self.connect_timeout {
      opt = opt.acquire_timeout(connect_timeout);   // applied first
  }
  if let Some(idle_timeout) = self.idle_timeout { ... }
  if let Some(acquire_timeout) = self.acquire_timeout {
      opt = opt.acquire_timeout(acquire_timeout);   // overwrites the above
  }
  The order is fixed by this function, not by your call order in mod.rs:150-151. And since DatabaseConfig::acquire_timeout is a plain u64 (common/src/config.rs:107), it is always Some, so connect_timeout is always clobbered in the non-IAM path too.

  3. sqlx's PgConnectOptions has no connect timeout to apply

  There is no connect_timeout/timeout field anywhere in sqlx-postgres-0.8.6/src/options/ — the only timeout knob in sqlx is pool-level PoolOptions::acquire_timeout, whose docs explicitly say it caps the whole acquire path including opening a new connection (sqlx-core-0.8.6/src/pool/options.rs:248-260):

  ▎ Caps the total amount of time Pool::acquire() can spend waiting across multiple phases: … If a new connection needs to be opened, that will obviously require I/O, handshaking, and initialization commands.

  4. It also bounds the initial connect in the IAM path

  sqlx-core-0.8.6/src/pool/options.rs:541-542, inside connect_with() — which is exactly what common/src/db/mod.rs:201 calls:
  // Don't take longer than `acquire_timeout` starting from when this is called.
  let deadline = Instant::now() + self.acquire_timeout;

  Practical upshot

  Effective config is the same on both paths: acquire_timeout governs, connect_timeout is dead config. If you want it to be non-dead, the real fix isn't to add .connect_timeout() to the IAM path (no such method on PgPoolOptions) — it's either to drop --db-conn-timeout as redundant, or wrap the connect in a
  tokio::time::timeout(connect_timeout, ...) if you genuinely want a separate bound on the initial handshake. Worth noting mod.rs:135-139 also has a latent bug: the ENV_DB_CONNECT_TIMEOUT parse fallback uses DB_IDLE_TIMEOUT.
```

File: common/src/db/mod.rs [hunk 54]
```diff
@@ -4,30 +4,38 @@ pub mod limiter;
...
+pub mod rds_iam;
 
 mod create;
 mod func;
 
 pub use create::*;
 pub use func::*;
 
+use crate::aws::aws_credentials_configured;
+use crate::config::{Database as DatabaseConfig, SslMode};
```
ctron: Please nest imports

common/src/db/mod.rs
```diff
@@ -4,30 +4,38 @@ pub mod limiter;
...
 use actix_web::{HttpResponse, ResponseError};
-use anyhow::Context;
-use reqwest::Url;
+use anyhow::{Context, Error};
+use rds_iam::RDS_IAM_TOKEN_REFRESH;
```
ctron: Please nest imports

common/src/db/mod.rs
desmax74: Done

common/src/db/mod.rs
desmax74: Done

common/src/db/mod.rs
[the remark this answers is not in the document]
ctron: Sorry, but don't think this is true:

* https://docs.rs/sea-orm/latest/src/sea_orm/database/mod.rs.html#332-335
* https://docs.rs/sea-orm/latest/src/sea_orm/database/mod.rs.html#357-360

And I actually had the case with AWS where the connection took "too long" to establish. Leading to an irrecoverable acquire failure, because the connection was not established.

And the PgConnectOptions has `options`, which has https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-OPTIONS

I see the path of assigning both to the same field. Then again, there definitely is a difference between acquire and connect. And it makes sense that the acquire is part of the pool options, while connect is part of the connection options.

ctron (approved):
I still think there's something with the acquire vs connect timeout. But I leave that up to you.

PR #2617, opened 2026-09-02
chore: add security headers and ensure good default CORS

rh-jfuller:
addresses TC-5475

## Summary by Sourcery

Harden the HTTP server defaults with restricted CORS and configurable security response headers.

New Features:
- Add configurable CORS origins with a fail-closed same-origin default and an explicit permissive development mode.
- Add configurable defense-in-depth security response headers, including CSP and opt-in HSTS.

Enhancements:
- Apply security headers consistently to application responses, while allowing deployment-specific values and disabling them when necessary.
- Document the new CORS and security-header environment variables.

Documentation:
- Document configuration for allowed CORS origins, permissive CORS, CSP, HSTS, and security-header disabling.

Tests:
- Add coverage for fail-closed and opt-in CORS behavior, security-header defaults and overrides, TLS-only HSTS, and middleware response headers.


common/infrastructure/src/app/http.rs
```diff
@@ -282,13 +282,73 @@ where
...
+
+    /// The Strict-Transport-Security header value. An empty value omits the
+    /// header. Only honored by browsers over HTTPS.
+    #[arg(
+        id = "http-server-hsts",
+        long,
+        env = "HTTP_SERVER_HSTS",
+        default_value = default::HSTS
+    )]
+    pub hsts: String,
```
ctron, line 336: HSTS should only be enabled when TLS is offered. Ideally by the TLS proxy (if one is used). So that would make it part of the TLS configuration. If we *really* want that.

However, that should also not be the default. As first you should check your TLS setup, before enabling HSTS. Otherwise you can shoot yourself in the foot pretty quickly.

common/infrastructure/src/app/http.rs
```diff
@@ -282,13 +282,73 @@ where
...
+
+    /// Disable the defense-in-depth security response headers
+    /// (X-Content-Type-Options, X-Frame-Options, Referrer-Policy, CSP, HSTS).
+    #[arg(
+        id = "http-server-security-headers-disabled",
+        long,
+        env = "HTTP_SERVER_SECURITY_HEADERS_DISABLED",
+        default_value_t = false
+    )]
+    pub security_headers_disabled: bool,
```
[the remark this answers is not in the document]
ctron: Yup, we should definitely do that.

common/infrastructure/src/app/http.rs
```diff
@@ -282,13 +282,73 @@ where
...
+    /// Allow ANY origin (development only). This reflects the request origin and
+    /// permits any method/header. Do not use in production; prefer
+    /// --http-server-cors-allowed-origins.
+    #[arg(
+        id = "http-server-cors-permissive",
+        long,
+        env = "HTTP_SERVER_CORS_PERMISSIVE",
+        default_value_t = false
+    )]
+    pub cors_permissive: bool,
```
ctron, line 305: I think this also conflicts with the others.

ctron (changes requested):

rh-jfuller:
@ctron - HSTS is now empty by default:
* HSTS is only emitted when tls_enabled is true
* Explicit HSTS configuration is supported for TLS deployments
* five new HTTP environment variables documented in docs/env-vars.md.
* The “TLS proxy should ideally own HSTS” part was not implemented as proxy integration; app now safely omits HSTS unless its own TLS is enabled.

does that address your concerns ?

rh-jfuller:
@sourcery-ai summary

ctron (approved):
lgtm, thanks for paying attention to updating the env-var file!

Issue #2605, opened 2026-08-26
Add exploit format category and upload endpoint


ctron:
## Context

PR #2553 added CISA KEV catalog integration. The KEV format (`Format::CisaKev`) is not part of the `Advisory` or `SBOM` hint groups, so uploading a KEV catalog via `POST /api/v3/advisory` requires an explicit `?format=cisakev` parameter — without it, auto-detection is skipped because the `Advisory` hint only matches CSAF, CVE, and OSV.

The importer bypasses this entirely (it passes `Format::CisaKev` directly), but manual upload has no natural home.

## Proposal

1. Add a new `Format::Exploit` hint group (analogous to `Format::Advisory` and `Format::SBOM`) that matches `Format::CisaKev` (and future exploit catalog formats).
2. Add a corresponding `POST /api/v3/exploit` upload endpoint that passes `Format::Exploit` as the detection hint, so KEV catalogs are auto-detected on upload without requiring `?format=cisakev`.

## References

- Jira: [TC-5753](https://redhat.atlassian.net/browse/TC-5753)
- PR #2553 (KEV integration)
- #2596 (exploit permissions)
- Detection code: `modules/ingestor/src/service/format.rs` (`matches_hint`)
- Pinning test: `modules/ingestor/src/service/detect.rs` (`cisa_kev_needs_a_format_hint_from_the_advisory_endpoint`)

[TC-5753]: https://redhat.atlassian.net/browse/TC-5753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2629, opened 2026-09-03
feat(exploit): add dedicated exploit permissions

### LINKED ISSUE #2596: Create new permissions for Exploits
https://github.com/guacsec/trustify/blob/610b122b1a1049693eeb66ea3093d4c3cb132b09/modules/fundamental/src/exploit/endpoints/mod.rs#L44

This looks like a new entity type, so we should have dedicated permissions for this.

waldemar-kindler:
## Summary
- Add a dedicated `read.exploit` permission and require it on `GET /api/v3/exploit` and `GET /api/v3/exploit/{id}`, which previously reused `read.advisory`
- Map `read.exploit` into the default `read:document` scope so existing Keycloak setups keep working without realm changes
- Document the new permission in `docs/oidc.md`

Only a read permission is added: exploit entries are written by the KEV importer and dataset upload, and there are no create/update/delete endpoints yet. `create.exploit` should come with the upload endpoint proposed in #2605.

Closes #2596
Jira: [TC-5743](https://redhat.atlassian.net/browse/TC-5743)

## Test plan
- [x] serde round-trip for `read.exploit` in `trustify-auth`
- [x] Endpoint test with an active authorizer: no permissions → 403, `read.advisory` only → 403 on list and get, `read.exploit` → 200 on list and get
- [x] `cargo fmt --check`, `cargo clippy` with CI flags on the touched crates

[TC-5743]: https://redhat.atlassian.net/browse/TC-5743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Separate exploit endpoint authorization from advisory access with dedicated exploit permissions and default scope mappings.

New Features:
- Add dedicated exploit permissions and require `read.exploit` for exploit list and detail endpoints.

Enhancements:
- Include exploit permissions in the default document scope mappings so existing authorization configurations continue to grant access.

Documentation:
- Document exploit permissions in the OIDC scope mapping documentation.

Tests:
- Add authorization coverage confirming exploit endpoints reject missing or advisory-only permissions and accept `read.exploit`.

### DIFF
--- a/common/auth/src/authenticator/default.rs
+++ b/common/auth/src/authenticator/default.rs
@@ -11,6 +11,7 @@ pub const DEFAULT_SCOPE_MAPPINGS: &[(&str, &[&str])] = &[
         "create:document",
         &[
             "create.advisory",
+            "create.exploit",
             "create.exploitIntelligence",
             "create.importer",
             "create.metadata",
@@ -25,6 +26,7 @@ pub const DEFAULT_SCOPE_MAPPINGS: &[(&str, &[&str])] = &[
         &[
             "ai",
             "read.advisory",
+            "read.exploit",
             "read.exploitIntelligence",
             "read.importer",
             "read.metadata",
@@ -38,6 +40,7 @@ pub const DEFAULT_SCOPE_MAPPINGS: &[(&str, &[&str])] = &[
         "update:document",
         &[
             "update.advisory",
+            "update.exploit",
             "update.importer",
             "update.metadata",
             "update.sbom",
@@ -49,6 +52,7 @@ pub const DEFAULT_SCOPE_MAPPINGS: &[(&str, &[&str])] = &[
         "delete:document",
         &[
             "delete.advisory",
+            "delete.exploit",
             "delete.importer",
             "delete.metadata",
             "delete.sbom",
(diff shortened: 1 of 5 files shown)


ctron:
Thanks for the PR. I think we should create all CRUD permissions at the same time.

waldemar-kindler:
Done, thanks for the review. The commit now declares `create.exploit`, `read.exploit`, `update.exploit` and `delete.exploit`, maps all four into the default document scopes, and the docs list them. Only `read.exploit` is enforced so far; `create.exploit` is ready for the upload endpoint in #2605.

ctron (approved):

waldemar-kindler:
Just removed the serde_exploit test. It was little useless. 

Was thinking of adding this to _ common/auth/src/authenticator/default.rs_

`#[cfg(test)]
mod tests {
    use super::*;
    use crate::Permission;

    /// Every permission name in the default scope mappings must be a real permission.
    /// A typo here would silently grant nothing.
    #[test]
    fn default_scope_mappings_name_real_permissions() {
        for (scope, permissions) in DEFAULT_SCOPE_MAPPINGS {
            for name in *permissions {
                assert!(
                    name.parse::<Permission>().is_ok(),
                    "scope {scope} maps to unknown permission {name}"
                );
            }
        }
    }
}`

But it currently fails on delete.vulnerability which has no mapping in the Permission enum. What do you think. Worth adding?

PR #2540, opened 2026-07-24
docs(adr): describe how vuln correlation works today

rh-jfuller:
[view](https://github.com/rh-jfuller/trustify/blob/8690815c9c6635e2cf374130533ea22322ba0d02/docs/adrs/00020-vulnerability-correlation-engine.md)

## Summary by Sourcery

Documentation:
- Add ADR 00020 describing the status quo and future direction of the vulnerability correlation engine, including directional behaviors, ingestion constraints, and identified limitations.

## Summary by Sourcery

Add an architectural decision record documenting the current design and known limitations of the vulnerability correlation engine for SBOMs and advisories.

Documentation:
- Document the data model, query flows, and status semantics for SBOM-to-vulnerability and vulnerability-to-SBOM correlation directions.
- Capture ingestion constraints, version matching schemes, and ecosystem-specific behaviors impacting vulnerability correlation.
- Record a catalog of identified limitations and asymmetries between correlation directions as a basis for future refactoring.

PR #2528, opened 2026-07-21
feat: in-memory correlation engine and WebSocket notifications

ctron:
## Summary

- **Correlation engine** (`modules/correlation/`): new in-memory advisory-SBOM correlation service behind `/api/v4/` endpoints. Loads advisory and SBOM indexes into memory at startup, correlates PURLs and product names against them, and hydrates results via DB lookups. Replaces multi-join SQL queries with in-memory index scans for the hot path.
- **Dedup fix**: fixed a bug where different statuses for the same advisory+CVE pair were collapsed during `correlate_purls` (dedup key now includes `status_id`).
- **WebSocket notifications** (`modules/notification/`): new `GET /api/v3/notifications` WebSocket endpoint that streams `change_log` events (advisory/SBOM ingestion and deletion) in real time. Supports cursor-based reconnection (`?after=<uuid>`) and `?token=<jwt>` query parameter auth for browser WebSocket clients that can't set HTTP headers.

## Benchmarks (DS3 dataset, quarkus-bom)

Speedup of the in-memory correlation engine (v3) vs the existing SQL-based path (v3a):

| Endpoint | v3a (SQL) | v3 (in-memory) | Speedup |
|----------|-----------|----------------|---------|
| SBOM details (quarkus-bom, 22 advisories, 5488 matches) | 434ms | 120ms | **3.6x** |
| Vulnerability (CVE-2023-4853) | 89ms | 4ms | **23.9x** |
| PURL advisories | 89ms | 4ms | **23.9x** |
| SBOM details (ubi8, 1 advisory) | 57ms | 4ms | **13.3x** |
| Recommend | 25ms | 4ms | **6.7x** |

## Test plan

- [x] `cargo test -p trustify-module-correlation --test correctness` — 6 tests verify result parity with SQL
- [x] `cargo test -p trustify-module-correlation --test benchmark` — 6 benchmarks pass
- [x] `cargo test -p trustify-module-notification` — 17 tests (token extraction, permission filtering, middleware, endpoint auth gates, DB backfill)
- [x] `cargo xtask precommit` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2631, opened 2026-09-04
docs(adr): propose correlation engine redesign (WIP)

ctron:
## Summary

> **Work in progress** — this ADR is a starting point for discussion, not a final design.
> Feedback on the architectural direction, matching dimensions, and trade-offs is welcome.

- Add ADR 00021 proposing a centralized `modules/correlation/` module to replace the current five scattered query-time SQL correlation pathways
- Matches on component identity only using three spec-endorsed dimensions: PURL, CPE identity, and digest
- Drops `product_status` name-only matching (not CSAF-spec-valid for automated correlation)
- Drops `sbom_describing_cpe` scoping (doesn't work for most non-Red-Hat data)
- Persists materialized results with match evidence/provenance for transparency
- Introduces inbox queue pattern for tracking correlation progress
- Configurable policy knobs via DB table and CLI args

## Context

Based on analysis of:
- [Correlation test suite evaluation](https://gist.github.com/ctron/6bfcb575068f52a2fb90a5ab0c1176b9) — 18 scenarios, 59% pass rate
- 5 independent code paths with inconsistent matching logic
- 9 root-cause bugs across TC-5170, TC-5171, TC-5640, TC-5641, TC-5730–TC-5733, TC-5750
- CSAF v2.0 specification on `product_identification_helper`
- PR #2528 in-memory engine prototype
- PR #2540 current-state ADR

## Test plan

- [ ] Team review of architectural decisions
- [ ] Validate against all 18 test scenarios (S1–S18)

## Summary by Sourcery

Establish a centralized, spec-oriented correlation architecture to improve matching correctness, transparency, performance, and operational visibility.

Enhancements:
- Propose replacing fragmented query-time correlation logic with a centralized, CSAF-aligned engine that matches components by PURL, CPE identity, and digest.
- Define materialized correlation results with status resolution, match confidence, and inspectable evidence/provenance.
- Specify an entity-based correlation queue with background processing, atomic result updates, progress tracking, and configurable matching policies.
- Outline a new versioned correlation API and phased integration while preserving existing v3 endpoint behavior.

Documentation:
- Add ADR 00021 documenting the proposed correlation engine redesign, its rationale, schema, APIs, operational model, trade-offs, and delivery phases.

### DIFF
--- a/docs/adrs/00021-correlation-engine-redesign.md
+++ b/docs/adrs/00021-correlation-engine-redesign.md
@@ -64,7 +64,7 @@ causes:
 | RPM epoch ignored by `rpmver_cmp` PL/pgSQL function | TC-5733 | Latent |
 | Child-node CPEs not included in `sbom_describing_cpe` | TC-5750 | False positive |
 
-### Performance
+### Current Performance
 
 All correlation is computed at query time via complex multi-CTE SQL. Benchmarks from a
 prototype in-memory engine (PR #2528) show the SQL path takes 57–434ms per request
@@ -115,6 +115,14 @@ different stream via standard RPM comparison, that match stands. Stream scoping 
 advisory's responsibility (via precise version ranges or separate assertions per stream),
 not the engine's — consistent with the decision to match on component identity only.
 
+**Known gap — source-to-binary PURL mismatch:** An advisory referencing a source RPM
+(e.g., `pkg:rpm/redhat/kernel?arch=src`) will not match an SBOM containing the binary
+RPM (e.g., `pkg:rpm/redhat/kernel-core@...?arch=x86_64`) — these have different
+`base_purl_id`s, no shared digest, and CPE matching is too coarse. The join between
+source and binary only exists through SBOM relationships (SPDX `GENERATED_FROM`) or
+build-system metadata (srpm→binary mapping). A relationship-based matching dimension
+could address this but is out of scope for the initial engine.
+
 **CPE identity matching (confidence: 0.8)** — for CVE/NVD data and CSAF advisories
 that identify components by CPE. Matches by vendor+product (part='a') with version
 comparison using `COALESCE(cpe_version, package_version)` for the SBOM side.
@@ -122,6 +130,9 @@ comparison using `COALESCE(cpe_version, package_version)` for the SBOM side.
 **Digest matching (confidence: 1.0)** — the strongest possible match per spec.
 Exact hash-value lookup with no version comparison needed. Requires ingestor extension
 to extract `product_identification_helper.hashes` from CSAF documents (prerequisite work).
+Confidence may be adjusted in the future based on SBOM trust signals (e.g., whether the
+SBOM is signed) — the policy model should leave space for this without implementing it
+in the initial version.
 
 ### Status Resolution
 
@@ -135,8 +146,13 @@ same `(sbom_node, vulnerability)`:
 
 ### Inbox Queue Pattern
 
-On document ingestion, a change event (via the existing `ChangeBroadcaster`) enqueues
-an entry in `correlation_inbox` with status `pending`. A background worker:
+The correlation inbox is a general-purpose work queue: any component that produces
+documents can enqueue a `pending` entry. The initial implementation subscribes to the
+existing `ChangeBroadcaster`, but the inbox interface is not coupled to it — batch
+imports, database triggers, CLI commands, or external orchestrators can all enqueue
+work through the same `correlation_inbox` table.
+
+A background worker:
 
 1. Picks up `pending` entries, marks them `processing`
 2. Builds or updates the in-memory index for the entity
@@ -228,10 +244,15 @@ CREATE TYPE evidence_source AS ENUM ('purl_status', 'cpe_status');
 |--------|------|---------|---------|
 | id | UUID PK | | |
 | name | TEXT UNIQUE | 'default' | |
+| ecosystem | TEXT | null | nullable — when set (e.g., `rpm`, `maven`), this policy overrides the global default for that ecosystem |
 | purl_type_strict | BOOLEAN | true | require PURL type match |
 | not_affected_suppresses | BOOLEAN | true | not_affected beats affected |
 | fixed_suppresses | BOOLEAN | true | fixed beats affected |
 | digest_matching | BOOLEAN | true | enable hash-based matching |
+
+The engine resolves the most specific policy for a given PURL type: an ecosystem-scoped
+policy (where `ecosystem` matches the PURL type) takes precedence over the global
+default (where `ecosystem IS NULL`).
 
 ### In-Memory Index Architecture
 
@@ -280,6 +301,7 @@ New endpoints under `/api/v4/correlation/`:
 | GET | `/sbom/{id}/summary` | Severity counts from materialized data |
 | POST | `/analyze` | Analyze PURLs against materialized + real-time |
 | GET | `/vulnerability/{id}` | All SBOMs affected by a vulnerability |
+| GET | `/vulnerability/{id}/summary` | Affected-SBOM count and severity breakdown |
 | GET | `/match/{id}/evidence` | Evidence for a specific match |
 | GET | `/inbox` | Inbox queue status |
 | GET | `/status` | Engine status (last run, index sizes, backlog) |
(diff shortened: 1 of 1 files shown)


docs/adrs/00021-correlation-engine-redesign.md
```diff
@@ -0,0 +1,410 @@
...
+  manageable; for very large deployments the index may need partitioning or tiering.
+- **Migration effort**: existing v3 consumers continue working unchanged during the
+  transition period, but full cutover requires testing the v3 adapter.
+
+### What This Does NOT Change
+
+- **Ingestion pipeline**: SBOMs and advisories are still ingested by the existing ingestor
+  module. The `purl_status`, `product_status`, and `cpe_status` tables continue to be
+  populated at ingestion time — the correlation engine reads from them.
+- **Analysis module**: the graph-walking analysis service is orthogonal and unaffected.
```
xquery, line 495: I see this as complementary to whatever we eventually use for resolving transitive deps ... and its a good thing to keep out of scope for now.

docs/adrs/00021-correlation-engine-redesign.md
```diff
@@ -0,0 +1,410 @@
...
+### API Surface
+
+New endpoints under `/api/v4/correlation/`:
+
+| Method | Path | Description |
+|--------|------|-------------|
+| GET | `/sbom/{id}` | Materialized matches for an SBOM |
+| GET | `/sbom/{id}/summary` | Severity counts from materialized data |
+| POST | `/analyze` | Analyze PURLs against materialized + real-time |
+| GET | `/vulnerability/{id}` | All SBOMs affected by a vulnerability |
```
xquery, line 353: maybe this will need a /summary as well

docs/adrs/00021-correlation-engine-redesign.md
```diff
@@ -0,0 +1,410 @@
...
+| purl_type_strict | BOOLEAN | true | require PURL type match |
+| not_affected_suppresses | BOOLEAN | true | not_affected beats affected |
+| fixed_suppresses | BOOLEAN | true | fixed beats affected |
+| digest_matching | BOOLEAN | true | enable hash-based matching |
+
+### In-Memory Index Architecture
+
+Two index structures behind `ArcSwap` for lock-free concurrent reads:
+
+**`AdvisoryIndex`** — loaded at startup, updated incrementally on advisory ingest/delete:
```
xquery, line 311: what about an advisory that sez **pkg:rpm/redhat/kernel?arch=src** and the SBOM says **pkg:rpm/redhat/kernel-core@...?arch=x86_64**. Different base purls. No digest in common. Same coarse CPE, but that's far too broad an answer ...  the 'join' only exists as edge eg. **SPDX GENERATED_FROM** or via srpm→binary build relation? do we need a by_relationship as well ?

docs/adrs/00021-correlation-engine-redesign.md
```diff
@@ -0,0 +1,410 @@
...
+| status | correlation_status | raw status from the advisory |
+| description | TEXT | e.g., "PURL match: pkg:rpm/redhat/curl@1.2.3" |
+| sbom_identifier | TEXT | PURL, CPE, or hash from the SBOM side |
+| version_scheme | TEXT | nullable |
+| sbom_version | TEXT | nullable |
+| range_low | TEXT | nullable |
+| range_high | TEXT | nullable |
+| version_matched | BOOLEAN | nullable |
+
+**`correlation_policy`** — configurable matching knobs:
```
xquery, line 291: for the ADR I think this is great, in reality we may need different policies for different ecosystems.

docs/adrs/00021-correlation-engine-redesign.md
```diff
@@ -0,0 +1,410 @@
...
+On document ingestion, a change event (via the existing `ChangeBroadcaster`) enqueues
+an entry in `correlation_inbox` with status `pending`. A background worker:
+
+1. Picks up `pending` entries, marks them `processing`
+2. Builds or updates the in-memory index for the entity
+3. Runs the matching pipeline
+4. Persists results to `correlation_match` and `correlation_evidence`
+5. Marks the inbox entry `completed` (or `failed` with error)
+
+This decouples ingestion from correlation, provides progress visibility, enables retry
```
xquery: somewhat similarly - we may not want to couple correlation to change event ... eg. we ma y want to batch process or use database triggers, or some other mechanism. At this stage it just means ensuring interface is general.

docs/adrs/00021-correlation-engine-redesign.md
```diff
@@ -0,0 +1,410 @@
...
+### Matching Dimensions
+
+Three dimensions, all operating on component identity without product scoping:
+
+**PURL matching (confidence: 0.9)** — the primary path for CSAF+PURL and OSV data.
+Matches by `base_purl_id` with:
+- PURL type enforcement (an RPM version range cannot match a golang PURL)
+- Version comparison using the version scheme appropriate for the PURL type
+- Unbounded version ranges match all versions (fixes bare `known_affected`)
+
```
xquery, line 112: there are a few outliers to be considered in this area:

**Containers and layered products**: non-RPM content, where "component" means a Go module or Maven and the purl namespace conventions have to be agreed rather than derived.

**Static linking**:  one vendored module vulnerability fans out across every binary that embedded it, so the affect count explodes. The Golang stdlib case for example gets very complex.

**Name mismatch with upstream/NVD**:  one of the root cause of third-party scanner false positives, since name Red Hat uses isn't the one NVD publishes ... we might need some aliasing here.

docs/adrs/00021-correlation-engine-redesign.md
```diff
@@ -0,0 +1,410 @@
...
+No dist-tag stream isolation: if an advisory's version range catches a package from a
+different stream via standard RPM comparison, that match stands. Stream scoping is the
+advisory's responsibility (via precise version ranges or separate assertions per stream),
+not the engine's — consistent with the decision to match on component identity only.
+
+**CPE identity matching (confidence: 0.8)** — for CVE/NVD data and CSAF advisories
+that identify components by CPE. Matches by vendor+product (part='a') with version
+comparison using `COALESCE(cpe_version, package_version)` for the SBOM side.
+
+**Digest matching (confidence: 1.0)** — the strongest possible match per spec.
```
xquery, line 130: strongest would actually be this + signed sboms .. we need to leave a 'space' for considering (in policies) if sboms are signed or not.

docs/adrs/00021-correlation-engine-redesign.md
```diff
@@ -0,0 +1,410 @@
...
+| `product_status` matches without CPE product-scope check | TC-5171 | False positive |
+| RPM dist tags (.el8 vs .el9) ignored in version comparison | TC-5640 | False positive |
+| `product_status` does not apply `version_matches()` to the component version; the `product_version_range` stores the product version, not the component version, so the "naive fix" is semantically wrong | TC-5641 | False positive |
+| `known_not_affected` status is never honored — queries only read `affected` | TC-5730 | False positive |
+| OSV advisories without CVE alias are dropped — no advisory_vulnerability created | TC-5731 | False negative |
+| Bare `known_affected` (no version range) creates no matchable row | TC-5732 | False negative |
+| RPM epoch ignored by `rpmver_cmp` PL/pgSQL function | TC-5733 | Latent |
+| Child-node CPEs not included in `sbom_describing_cpe` | TC-5750 | False positive |
+
+### Performance
```
xquery: Current Performance

rh-jfuller:
FWIW I meant to comment as @rh-jfuller (not my other @xquery) !

rh-jfuller:
a few more observations

*  evidence_source enum can not represent digest matches ... maybe I am missing something
* I think it would be very useful to try and trace the scenarios in a matrix in this ADR ... WDYT ?
* also a first stab at a few charts .... not claiming they are correct (as an interpretation of current spec) or correct in desired end state ... though curious if we should include this kind of detail 

## End-to-end: event → state → worker → persist → API

```mermaid
flowchart TD
    subgraph INGEST[Ingestion transaction]
      A[SBOM ingested] -->|UPSERT pending| ST[(correlation_state<br/>one row per entity)]
      B[Advisory ingested / deprecated / deleted] -->|UPSERT pending| ST
      M[Manual trigger] -->|UPSERT pending| ST
    end

    ST -->|SELECT ... FOR UPDATE SKIP LOCKED| W{Worker claims row<br/>status→processing}

    W -->|entity_type = advisory| FO[Phase 1 · Advisory fan-out<br/>coarse PURL/CPE query,<br/>no version check]
    FO -->|UPSERT affected SBOMs → pending| ST
    FO --> AC[advisory row → completed]

    W -->|entity_type = sbom| RE[Phase 2 · SBOM re-evaluation<br/>delete-and-rebuild]
    RE --> IDX[Build SbomIndex • read AdvisoryIndex via ArcSwap<br/>excludes deprecated advisories]
    IDX --> MP[Matching pipeline<br/>3 dimensions]
    MP --> SR[Status resolution<br/>per sbom_node × vulnerability]
    SR --> SWAP[[Atomic swap tx<br/>DELETE matches for sbom_id →<br/>INSERT correlation_match + evidence →<br/>state=completed, correlated_as_of=now]]
    SWAP --> RDB[(correlation_match<br/>correlation_evidence)]
    RDB --> API[/v4 API reads materialized results/]
```
## Matching pipeline (component-identity only)
```mermaid
flowchart LR
    N[SBOM node] --> P{has PURL?}
    P -->|base_purl_id + type match<br/>+ version_matches| M1[PURL match<br/>conf 0.9]
    N --> C{has CPE part=a?}
    C -->|vendor+product<br/>+ version_matches| M2[CPE-identity match<br/>conf 0.8]
    N --> D{has digest?}
    D -->|algo+hash exact| M3[Digest match<br/>conf 1.0]
    M1 --> RES
    M2 --> RES
    M3 --> RES[Collect assertions for<br/> sbom_node, vulnerability ]
```
## Status resolution (precedence, all configurable)
```mermaid
flowchart TD
    IN[All assertions for a node×vuln] --> Q1{any not_affected?<br/>not_affected_suppresses}
    Q1 -->|yes| NA[effective = not_affected]
    Q1 -->|no| Q2{any fixed?<br/>fixed_suppresses}
    Q2 -->|yes| FX[effective = fixed]
    Q2 -->|no| Q3{any under_investigation?}
    Q3 -->|yes| UI[effective = under_investigation surfaced]
    Q3 -->|no| AF[effective = affected]
```
## correlation_state.status lifecycle
```mermaid
stateDiagram-v2
    [*] --> pending: ingest / fan-out / manual
    pending --> processing: worker claims (SKIP LOCKED)
    processing --> completed: swap tx commits
    processing --> failed: error (error_message set)
    processing --> pending: crash / timeout (tx rollback)
    completed --> pending: re-queue (new/changed advisory)
    failed --> pending: retry / re-queue
```

rh-jfuller (commented):
As mentioned ... I think we have 2 ADR in one ... 

* **Correlation semantics**: what a correct match is across ecosystems, and how it should degrade as identity data thins out.
* **Data-at-scale mechanics**: how you store/refresh/serve results (materialization, work queue, in-memory indexes, freshness, atomic swap).

I think we actually need to treat both of these concerns in a single ADR ... but if we can break it down now it would be useful ... WDYT ?

ctron:
Moving on.

PR #2637, opened 2026-09-05
#2636 Add arg to analyze and advisory endpoints to allow resolved statuses

### LINKED ISSUE #2636: Reconcile VEX not_affected status with SBOM vulnerability analysis
When a CSAF VEX advisory declares a package version as `not_affected` (via `product_status.known_not_affected` and flags with `vulnerable_code_not_present`), the SBOM vulnerability view should reflect that status — either by reducing the CVE count or clearly marking those CVEs as resolved. Currently, the VEX advisory is ingested and viewable on its own, but the SBOM analysis shows the same CVE matches regardless of VEX status. This forces users to manually cross-reference advisories against SBOMs to determine which vulnerabilities are actually relevant — exactly the workflow TPA should be automating.

Organizations consuming remediated open-source dependencies (e.g., vendor-patched builds with VEX attestations) have no way to see the effect of that remediation in TPA's SBOM view. A security team triaging vulnerabilities sees the same CVE list whether or not a VEX document has declared those CVEs resolved. At scale — hundreds of SBOMs, thousands of CVEs — manual reconciliation is not viable.

**Current Behavior**

1. Upload a CSAF VEX with product_status.known_not_affected and flags: `[{"label": "vulnerable_code_not_present"}]` for specific PURLs
2. Upload an SBOM containing those same PURLs
3. View the SBOM in the UI → vulnerabilities tab shows all CVE matches with no indication that VEX has resolved some of them

The advisory detail view displays _some_ VEX data: ID, Title, Discovery, Release, CWE all populate; but notably the individual fix versions are not browsable unless you download the JSON file and view it. Moreover, the SBOM-to-advisory reconciliation doesn't happen — the `/vulnerability/analyze` endpoint doesn't factor in VEX status when matching CVEs to SBOM contents.

**Expected Behavior**

When an SBOM contains a component whose PURL matches a VEX `not_affected` declaration, the SBOM vulnerability view should either:
- Exclude or visually distinguish those CVEs (e.g., strikethrough, "resolved by VEX" badge)
- Provide a filtered/unfiltered toggle so users can see the before/after difference
- At minimum, surface the VEX status alongside each CVE match

This is core to TPA's value as the central visibility layer for supply chain security. Without it, VEX data is stored but not actionable — users still have to build their own reconciliation tooling or rely on external pipeline steps to get an accurate picture of their vulnerability exposure.

andykrohg:
Resolves #2636

The SBOM advisory endpoint (`GET /v3/sbom/{id}/advisory`) and the analyze endpoint (`POST /v3/vulnerability/analyze`) currently hardcode filters that exclude `fixed` and `not_affected` statuses from results. This means consumers cannot see VEX resolutions — even though Trustify correctly ingests and stores them.

This PR adds a backwards-compatible `include_resolved` flag to both endpoints:

- Analyze endpoint: include_resolved field in the JSON request body (defaults to false)
- SBOM advisory endpoint: ?include_resolved=true query parameter (defaults to false)

When `include_resolved=true`, the `fixed` and `not_affected` statuses are included in the response alongside `affected`. The `recommended` status (a remediation hint, not a vulnerability status) remains excluded in both modes. When `include_resolved` is omitted or false, behavior is identical to before this change.

This enables VEX-aware consumers such as pipeline gates and UI components to see which vulnerabilities have been resolved. For example, a pipeline can call the analyze endpoint with `include_resolved: true` and gate on whether any entries have status: "affected", rather than just checking whether any entries exist at all.

## Summary by Sourcery

Allow consumers to opt in to resolved vulnerability statuses from advisory and analysis endpoints.

New Features:
- Add an optional `include_resolved` parameter to SBOM advisory and vulnerability analysis endpoints to expose `fixed` and `not_affected` statuses alongside `affected` results.

Enhancements:
- Preserve existing filtering behavior by default while continuing to exclude `recommended` statuses.

Documentation:
- Update the OpenAPI specification for the new advisory query parameter and analysis request field.

Tests:
- Add VEX reconciliation coverage for resolved statuses in SBOM advisory and vulnerability analysis responses.

### DIFF
--- a/modules/fundamental/src/sbom/endpoints/mod.rs
+++ b/modules/fundamental/src/sbom/endpoints/mod.rs
@@ -367,12 +367,21 @@ pub async fn get(
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Default, serde::Deserialize, IntoParams)]
+pub struct SbomAdvisoryParams {
+    /// When true, include resolved statuses (fixed, not_affected)
+    /// alongside affected.
+    #[serde(default)]
+    pub include_resolved: bool,
+}
+
 /// Get advisories for an SBOM
 #[utoipa::path(
     tag = "sbom",
     operation_id = "getSbomAdvisories",
     params(
         ("id" = Id, Path),
+        SbomAdvisoryParams,
     ),
     responses(
         (status = 200, description = "Matching SBOM", body = Vec<SbomAdvisory>),
@@ -384,12 +393,17 @@ pub async fn get_sbom_advisories(
     fetcher: web::Data<SbomService>,
     db: web::Data<db::ReadOnly>,
     id: web::Path<String>,
+    web::Query(SbomAdvisoryParams { include_resolved }): web::Query<SbomAdvisoryParams>,
     _: Require<GetSbomAdvisories>,
 ) -> Result<impl Responder, Error> {
     let id = Id::from_str(&id).map_err(Error::IdKey)?;
     let tx = db.begin().await?;
 
-    let statuses: Vec<String> = vec!["affected".to_string()];
+    let statuses: Vec<String> = if include_resolved {
+        vec![]
+    } else {
+        vec!["affected".to_string()]
+    };
     match fetcher.fetch_sbom_details(id, statuses, &tx).await? {
         Some(v) => Ok(HttpResponse::Ok().json(v.advisories)),
         None => Ok(HttpResponse::NotFound().finish()),
--- a/modules/fundamental/src/sbom/endpoints/mod.rs
+++ b/modules/fundamental/src/sbom/endpoints/mod.rs
@@ -400,7 +400,11 @@ pub async fn get_sbom_advisories(
     let tx = db.begin().await?;
 
     let statuses: Vec<String> = if include_resolved {
-        vec![]
+        vec![
+            "affected".to_string(),
+            "fixed".to_string(),
+            "not_affected".to_string(),
+        ]
     } else {
         vec!["affected".to_string()]
     };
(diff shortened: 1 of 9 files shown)


rh-jfuller:
running > cargo xtask precommit is always a good thing to do ;)

andykrohg:
Sorry about that! Pushed a new commit and should be addressed now

modules/fundamental/src/sbom/endpoints/mod.rs
```diff
@@ -384,12 +393,21 @@ pub async fn get_sbom_advisories(
...
     id: web::Path<String>,
+    web::Query(SbomAdvisoryParams { include_resolved }): web::Query<SbomAdvisoryParams>,
     _: Require<GetSbomAdvisories>,
 ) -> Result<impl Responder, Error> {
     let id = Id::from_str(&id).map_err(Error::IdKey)?;
     let tx = db.begin().await?;
 
-    let statuses: Vec<String> = vec!["affected".to_string()];
+    let statuses: Vec<String> = if include_resolved {
+        vec![
```
rh-jfuller, line 403: we define this kind of thing elsewhere (vulnerability/service/mod.rs:474)... maybe consider a single source of truth RESOLVED_STATUSES

rh-jfuller (approved):
nice work, LGTM

modules/fundamental/src/sbom/endpoints/mod.rs
andykrohg: Thanks @rh-jfuller - want me to make this change before merging or is this more of a note to come back to in the future?

PR #2639, opened 2026-09-07
fix(vuln corr): apply CPE-context filter to the backlink purl_status …

rh-jfuller:
The `GET /v**3/vulnerability/{id}**` backlink ("Related SBOMs") correlated advisory
statuses to SBOMs by base PURL name + version only, with no check that the
status' CPE context actually applied to the SBOM.

As a result, a `purl_status` scoped to one product stream (e.g. an el9 `curl`
advisory) was linked to an SBOM describing a different stream (e.g. el8),
producing wrong-product entries in the backlink and vulnerability report.

This applies the same three-way CPE-context filter already used by `/purl`
(`cpe_context_subqueries`) and `/sbom/{id}/advisory` (`CONTEXT_CPE_FILTER_SQL`)
to the backlink `sbom_status_query`.

Added two regression tests using the existing `cpe_edition` fixtures:

Refs: TC-5171

## Summary by Sourcery

Prevent vulnerability backlinks from associating advisory statuses with SBOMs whose CPE product context does not match.

Bug Fixes:
- Apply CPE-context filtering when correlating vulnerability backlink statuses to SBOMs, preventing advisories for unrelated product streams from appearing in Related SBOMs results.

Tests:
- Add regression coverage for matching and mismatching CPE edition contexts in vulnerability backlinks.

### DIFF
--- a/modules/fundamental/src/vulnerability/model/details/vulnerability_advisory.rs
+++ b/modules/fundamental/src/vulnerability/model/details/vulnerability_advisory.rs
@@ -33,6 +33,53 @@ use trustify_entity::{
 use utoipa::ToSchema;
 use uuid::Uuid;
 
+/// Correlated three-way CPE-context filter for the vulnerability backlink
+/// `purl_status` path.
+///
+/// Suppresses advisory statuses whose product CPE context does not apply to
+/// the SBOM being linked, so a status scoped to one product stream (e.g. an
+/// el9 advisory) is not correlated to an SBOM that describes a different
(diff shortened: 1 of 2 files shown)


ctron (approved):
With the whole complexity of this, I can't really tell if this is heading in the right direction. It looks good to me. And CI should decide.

Can we unlock more scenario tests with this?

rh-jfuller:
FWIW we do not have scenario tests in release/0.4.z branch and I am not advocating for (as part of this PR) any main branch fixes ... just yet which might enable some scenario tests to pass ... making a note so we do not miss in main.

PR #2643, opened 2026-09-07
fix(migration): make fully-unbounded version ranges match any version (TC-5732) on release/0.4.z

rh-jfuller:
A bare, version-less CSAF `known_affected` is ingested as an unbounded version
range (both bounds NULL). Every per-scheme comparator
(`rpmver/semver/maven/python_version_matches`) returns `false` when both bounds
are NULL, so the row is created but never matches — genuine vulnerabilities were
silently missed (false negative). Example: firefox main-el8 `known_affected` in
CVE-2023-6135.

Add an early guard to the top-level `version_matches` dispatcher: a range with
both bounds NULL means "all versions" → return `true`, before delegating to the
scheme-specific comparators.

Added both-unbounded regression tests for rpm, semver, maven, and python
(verified they fail with the guard removed, pass with it).

Refs: TC-5732

## Summary by Sourcery

Ensure version-less affected-product ranges are treated as matching all versions.

Bug Fixes:
- Make fully unbounded version ranges match every version across supported version schemes, preventing version-less CSAF affected entries from being silently missed.

Enhancements:
- Add an idempotent database migration that updates the version-matching dispatcher while preserving rollback support.

Build:
- Increase Rust recursion limits for migration crates and tests.

Tests:
- Add regression coverage for fully unbounded RPM, SemVer, Maven, and Python version ranges.

### DIFF
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -36,6 +36,7 @@ mod m0002100_analysis_perf_indexes;
 mod m0002110_sbom_describing_cpe;
 mod m0002120_ancestor_walk_index;
 mod m0002250_create_cpe_status;
+mod m0002260_version_matches_unbounded;
 
 pub struct Migrator;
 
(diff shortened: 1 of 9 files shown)


ctron (approved):

ctron:
We must ensure that the other branches have this migration too. even if it is a no-op. Otherwise, someone upgrading from this branch to any of the other will run into a failure.

rh-jfuller:
I will raise this on main and other release branches separately thx

PR #2645, opened 2026-09-07
feat(importer): unified auth model with shared HTTP client builder

ruromero:
## Summary

- Introduces `CredentialSource` (Inline/Env/File), `AuthMethod` (Basic/Bearer/ApiKey), and `AuthConfig` types in `modules/importer/src/model/auth.rs`
- Replaces `QuayImporter.api_token: Option<String>` with `auth: Option<AuthConfig>` (breaking API change)
- Adds shared `build_http_client(auth: Option<&AuthConfig>)` in `modules/importer/src/runner/common/http.rs`; updates Quay walker to use it
- Data migration `m0002350_importer_quay_auth` rewrites existing Quay JSONB `apiToken` → `auth` (Bearer+Inline); reversible `down()`
- Regenerates `modules/importer/schema/importer.json` and `openapi.yaml`

Env and File credential sources resolve at import time — the DB stores only references, enabling K8s Secret rotation without pod restarts.

## Test plan

- [x] `cargo xtask precommit` passes with no warnings
- [x] `cargo test -p trustify-module-importer --lib` — 35 tests pass including 7 new auth/http tests
- [ ] Integration test against a Quay instance with env-based credentials
- [ ] Verify data migration round-trip on a DB with existing Quay importers

Implements [TC-6090](https://redhat.atlassian.net/browse/TC-6090)

[TC-6090]: https://redhat.atlassian.net/browse/TC-6090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Unify importer authentication and migrate Quay credentials to support reusable, runtime-resolved authentication configuration.

New Features:
- Add a unified authentication model supporting Basic, Bearer, and API-key methods with inline, environment, or file-based credential sources.
- Provide shared authenticated HTTP client construction for importer runners, including import-time credential resolution.

Enhancements:
- Update Quay importer configuration to use the new authentication model instead of a dedicated API token field.
- Allow environment and file credential references to remain outside the database while supporting runtime secret rotation.

Documentation:
- Regenerate importer JSON schema and OpenAPI definitions for the new authentication configuration.

Tests:
- Add coverage for credential resolution, HTTP client authentication, and authenticated Quay walking.

Chores:
- Migrate existing Quay configurations between the legacy API-token format and the unified authentication format with reversible database migration.


migration/src/m0002350_importer_quay_auth.rs
```diff
@@ -0,0 +1,58 @@
...
+        manager
+            .get_connection()
+            .execute_unprepared(UP_SQL)
+            .await
+            .map(|_| ())
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .get_connection()
```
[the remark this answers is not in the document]
ruromero: [sdlc-workflow/verify-pr] Classified as **code change request** — the down migration SQL path `configuration->'auth'->'method'->>'token'` incorrectly used `->>` on an object, returning its JSON text representation instead of the scalar token value. Fixed in the current commit (b1bc6c5) to use `configuration->'auth'->'method'->'token'->>'value'`. No sub-task created — fix applied inline during conflict resolution.

ruromero:
[sdlc-workflow/verify-pr] Re: @sourcery-ai review — Classified as **code change request** — the down migration SQL incorrectly extracted the token as a JSON object text rather than the scalar value string. Fixed in commit b1bc6c5 by using `->'token'->>'value'` path and restricting the rollback WHERE clause to Bearer+Inline shape only. No sub-task created — fix applied inline.

ruromero:
## Verification Report for TC-6090 (commit b1bc6c5)

| Check | Result | Details |
|-------|--------|---------|
| Review Feedback | WARN | 1 code change request (down migration SQL bug) — fixed inline in b1bc6c5; no sub-task created |
| Root-Cause Investigation | DONE | TC-6097: Document PostgreSQL JSONB extraction patterns in CONVENTIONS.md |
| Scope Containment | WARN | All task files present; 4 justified companion files (Cargo.lock, migration/src/lib.rs, Cargo.toml, xtask/schema/generate-dump.json) |
| Diff Size | PASS | ~310 hand-written lines + ~790 auto-generated schema lines — proportionate |
| Commit Traceability | PASS | `Implements TC-6090` in commit body |
| Sensitive Patterns | PASS | No hardcoded credentials or secrets detected |
| CI Status | WARN | macos-15 ✓, ubuntu-24.04 ✓, supply-chain ✓; windows-2025 / ci / bench pending (force-push rebase) |
| Acceptance Criteria | WARN | 10/10 functional criteria met; 5/7 test requirements covered — 2 integration tests missing (QuayWalker auth path, migration round-trip) |
| Test Quality | PASS | All 7 new tests have `///` doc comments; no repetitive patterns; Eval Quality N/A |
| Test Change Classification | ADDITIVE | Both test files new; +7 tests, 0 removed |
| Verification Commands | PASS | `cargo xtask precommit` passed; ubuntu CI corroborates |

### Overall: WARN

**Issues requiring attention:**

1. **Missing integration tests** (Acceptance Criteria WARN): `QuayWalker::new()` authenticated path and migration up/down round-trip are not integration-tested. Existing walker tests use `auth: None`. Consider adding to `modules/importer/src/test.rs`.

2. **CI pending** (WARN, not FAIL): windows-2025, ci, and bench checks are pending after the force-push rebase. No failures at time of report.

3. **Scope WARN** (informational): 4 auto-generated/companion files outside the explicit task spec — all justified (`Cargo.lock` from dep add, `migration/src/lib.rs` for registration, `Cargo.toml` for `base64` dep, `xtask/schema/generate-dump.json` from `cargo xtask precommit`).

---
*This comment was AI-generated by [sdlc-workflow/verify-pr](https://github.com/RHEcosystemAppEng/sdlc-plugins) v0.13.9.*

migration/src/m0002350_importer_quay_auth.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// Rewrites Quay importer JSONB configurations to replace the legacy
+/// `apiToken` field with the new unified `auth` credential structure.
+///
+/// The `auth` field uses a Bearer + Inline credential source so that
+/// existing plaintext tokens continue to work unchanged after upgrade.
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+const UP_SQL: &str = r#"
```
ctron: We haven't done this so far. Why not just inline this?

modules/importer/src/model/auth.rs
```diff
@@ -0,0 +1,145 @@
...
+    PartialEq,
+    Eq,
+    serde::Serialize,
+    serde::Deserialize,
+    ToSchema,
+    schemars::JsonSchema,
+)]
+#[serde(tag = "type", content = "value", rename_all = "camelCase")]
+pub enum CredentialSource {
+    /// Literal value stored in the database (development use only).
```
ctron: Not sure I'd declare this as "development" use. But calling it out as a security thing would make more sense.

modules/importer/src/model/auth.rs
```diff
@@ -0,0 +1,145 @@
...
+impl CredentialSource {
+    /// Resolves the credential to its string value.
+    pub fn resolve(&self) -> anyhow::Result<String> {
+        match self {
+            Self::Inline(v) => Ok(v.clone()),
+            Self::Env(name) => {
+                std::env::var(name).map_err(|e| anyhow::anyhow!("env var {name}: {e}"))
+            }
+            Self::File(path) => {
+                std::fs::read_to_string(path).map_err(|e| anyhow::anyhow!("file {path}: {e}"))
```
ctron: import types, functions, ...

modules/importer/src/model/auth.rs
```diff
@@ -0,0 +1,145 @@
...
+    Inline(String),
+    /// Name of an environment variable read at import time.
+    Env(String),
+    /// Filesystem path to a file whose content is read at import time.
+    File(String),
+}
+
+impl CredentialSource {
+    /// Resolves the credential to its string value.
+    pub fn resolve(&self) -> anyhow::Result<String> {
```
ctron: As this is not directly reporting back to `main`, please use a real error (using `thiserror`).

modules/importer/src/runner/common/http.rs
```diff
@@ -0,0 +1,93 @@
...
+};
+use base64::{Engine, engine::general_purpose::STANDARD as BASE64_STANDARD};
+use reqwest::header;
+
+/// Builds an authenticated (or unauthenticated) `reqwest::Client`.
+///
+/// When `auth` is `None` a plain client with no default headers is returned.
+/// When `auth` is `Some`, credentials are resolved from their [`AuthConfig`]
+/// and set as default headers so every request the client makes carries them.
+pub fn build_http_client(auth: Option<&AuthConfig>) -> Result<reqwest::Client, Error> {
```
ctron: import

modules/importer/src/runner/common/http.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    let mut headers = header::HeaderMap::new();
+
+    match &auth.method {
+        AuthMethod::Bearer { token } => {
+            let resolved = token.resolve().map_err(|e| {
+                Error::Processing(anyhow::anyhow!("bearer token resolution failed: {e}"))
+            })?;
+            let value = format!("Bearer {resolved}");
+            let mut header_val = header::HeaderValue::from_str(&value)?;
```
ctron: import

modules/importer/src/runner/common/http.rs
```diff
@@ -0,0 +1,93 @@
...
+            header_val.set_sensitive(true);
+            headers.insert(header::AUTHORIZATION, header_val);
```
ctron, line 35: import, .. same for others

modules/importer/src/model/auth.rs
```diff
@@ -0,0 +1,145 @@
...
+        match self {
+            Self::Inline(v) => Ok(v.clone()),
+            Self::Env(name) => {
+                std::env::var(name).map_err(|e| anyhow::anyhow!("env var {name}: {e}"))
+            }
+            Self::File(path) => {
```
ctron, line 99: To my understanding, that would allow the user to define e.g `/var/run/secrets/kubernetes.io/serviceaccount/token` and extract the pods token with that.

I think in an ADR PR before, that was already raised as a concern. What we must do is to have a configuration parameter, which provides a set of base paths, which are eligible for such paths.

modules/importer/src/model/auth.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// Filesystem path to a file whose content is read at import time.
+    File(String),
+}
+
+impl CredentialSource {
+    /// Resolves the credential to its string value.
+    pub fn resolve(&self) -> anyhow::Result<String> {
+        match self {
+            Self::Inline(v) => Ok(v.clone()),
+            Self::Env(name) => {
```
ctron, line 80: Similar to file names, we should limit this. Otherwise one could e.g. use `TRUSTD_DB_PASSWORD`, and get the DB password, right?

ctron (changes requested):
Did a first pass. And I think we must fix some security concerns.

migration/src/m0002350_importer_quay_auth.rs
ruromero: I don't mind changing it to inline but I have seen it in other files like `m0002330_fix_rpmver_cmp.rs`

modules/importer/src/model/auth.rs
ruromero: 3 segments is allowed inline by `CONVENTIONS.md` should it be changed?

modules/importer/src/runner/common/http.rs
ruromero: the `CONVENTIONS.md` only requires 4 segments to use import. Should we change this in the conventions so that it doesn't happen again? [Ref](https://github.com/guacsec/trustify/blob/main/CONVENTIONS.md#import-style--absolute-paths)

modules/importer/src/runner/common/http.rs
ruromero: same, let's agree on this to prevent more issues with the imports.

modules/importer/src/model/auth.rs
ruromero: This change makes sense but is not covered by CONVENTIONS.md or Clippy. Is this something we want to enforce in the future? If so we can add something like this:

> internal library functions must use typed errors via thiserror, not anyhow.

modules/importer/src/model/auth.rs
ruromero: Good point. I will restrict only to `IMPORTER_AUTH_` prefixed env vars. Is that fine?
Example: `IMPORTER_AUTH_LW_USER` or `IMPORTER_AUTH_QUAY_TOKEN`

modules/importer/src/model/auth.rs
ruromero: Yes and that might be intended if the SA token is required for authentication. A configurable allow list will not help because the path is also configurable and a compromised privileged account will be able to update the allow list and the configuration parameter.

modules/importer/src/model/auth.rs
ruromero: Updated with more specific comment.

modules/importer/src/model/auth.rs
ctron: True, but that's what we do and should continue doing. If there's a way to enforce this, let's do it.

modules/importer/src/runner/common/http.rs
ctron: Import should always be used. Unless there's a clash, or possible clash.

modules/importer/src/model/auth.rs
ctron: That configuration should be on the application level. Not part of the importer configuration.

modules/importer/src/model/auth.rs
ctron: I think that should be configurable.

ruromero: I have to disagree, otherwise you allow configuring `TRUSTD_DB_` as prefix and end up having the same security problem we're trying to solve.

modules/importer/src/runner/common/http.rs
ruromero: I will create a follow up PR for the conventions file.

modules/importer/src/model/auth.rs
ruromero: I will add this as a follow up PR to the conventions file

modules/importer/src/model/auth.rs
ruromero: I see your point now. I was only focusing on any set of configuration. I will make the changes accordingly.

modules/importer/src/model/auth.rs
ruromero: Same principle as the accepted paths at application level. I will make the changes. Thanks for your comments

modules/importer/src/model/auth.rs
ruromero: Done

modules/importer/src/model/auth.rs
ruromero: Done, it is now configurable.

ruromero:
@ctron I have added the configurable parameters for the allowed paths and env var prefixes to the application (cli/env var) configuration.

modules/importer/src/model/auth.rs
```diff
@@ -0,0 +1,297 @@
+use std::{
+    env::{VarError, var},
+    fs::read_to_string,
+    io::Error,
+};
+
```
ctron: remove empty line

modules/importer/src/model/auth.rs
```diff
@@ -0,0 +1,297 @@
...
+/// Operator-level restrictions on credential resolution.
+///
+/// Set at application startup via CLI flags or environment variables.
+/// These constraints cannot be changed at runtime by users who have
+/// importer write permission, providing a security boundary above
+/// per-importer [`AuthConfig`].
+#[derive(Debug, Clone, Default)]
+pub struct CredentialConfig {
+    /// Required prefix for env var names. Empty = allow all.
+    pub allowed_prefix: String,
```
ctron: I think it makes sense to have a secure default. Alternatively, use `Option<String>` to indicate `None`, allowing none.

Same for the path. The default is "allow all", which opens up to the security issue mentioned, by default.

modules/importer/src/runner/quay/mod.rs
```diff
@@ -2,6 +2,7 @@ mod oci;
 mod walker;
 
 use crate::model::QuayImporter;
+use crate::model::auth::CredentialConfig;
```
ctron: Please nest imports

server/src/profile/importer.rs
```diff
@@ -32,6 +32,14 @@ pub struct Run {
     )]
     pub concurrency: usize,
 
+    /// Allowed base paths for file credential sources; files not under these directories are rejected (empty = allow all)
+    #[arg(long, env = "IMPORTER_CREDENTIAL_PATHS", value_delimiter = ',')]
+    pub allowed_credential_paths: Vec<String>,
+
+    /// The required prefix for environment variable credential sources (empty = allow all)
+    #[arg(long, env = "IMPORTER_ENV_PREFIX", default_value = "")]
+    pub env_prefix: String,
```
ctron: Nit: if we allow multiple paths, why don't we allow multiple prefixes?

ctron (changes requested):
Looks good. A few nits. And an insecure default that we should change.

modules/importer/src/model/auth.rs
ruromero: I prefer the `Option<String>` rather than any other default for this and for the paths. I'll make the changes

server/src/profile/importer.rs
ruromero: done

server/src/profile/importer.rs
```diff
@@ -32,6 +32,14 @@ pub struct Run {
     )]
     pub concurrency: usize,
 
+    /// Allowed base paths for file credential sources; files not under these directories are rejected. Default = empty list = allow none.
+    #[arg(long, env = "IMPORTER_CREDENTIAL_PATHS", value_delimiter = ',')]
```
ctron, line 36: Please add them to the env-var doc

server/src/profile/importer.rs
```diff
@@ -32,6 +32,14 @@ pub struct Run {
...
+    pub allowed_credential_paths: Option<Vec<String>>,
+
+    /// The required prefix for environment variable credential sources. Default = empty list = allow none.
+    #[arg(long, env = "IMPORTER_ENV_PREFIXES", value_delimiter = ',')]
+    pub env_prefixes: Option<Vec<String>>,
```
ctron: We can't we just use `Vec<String>`?

server/src/profile/importer.rs
```diff
@@ -32,6 +32,14 @@ pub struct Run {
     )]
     pub concurrency: usize,
 
+    /// Allowed base paths for file credential sources; files not under these directories are rejected. Default = empty list = allow none.
+    #[arg(long, env = "IMPORTER_CREDENTIAL_PATHS", value_delimiter = ',')]
+    pub allowed_credential_paths: Option<Vec<String>>,
```
ctron: Why can't we just use `Vec<String>`?

server/src/profile/importer.rs
```diff
@@ -95,6 +104,10 @@ impl InitData {
             working_dir: run.working_dir,
             concurrency: run.concurrency,
             read_only: run.read_only,
+            credential_config: CredentialConfig {
+                allowed_prefixes: run.env_prefixes.clone(),
```
ctron: Why do we need to clone here? And next line?

modules/importer/src/runner/common/http.rs
```diff
@@ -0,0 +1,141 @@
...
+        // Then an error is returned
+        assert!(result.is_err());
+    }
+
+    /// Verifies that `build_http_client` succeeds when the env var matches the required prefix.
+    #[test]
+    fn build_http_client_bearer_env_allowed_prefix_succeeds() {
+        // Given the env var is set with the correct prefix
+        // SAFETY: test-only mutation, single-threaded test binary
+        unsafe { std::env::set_var("IMPORTER_AUTH_HTTP_TEST_TOKEN", "test-value") };
```
ctron: This will not work. As tests run in parallel and must not modify the environment.

modules/importer/src/runner/common/http.rs
```diff
@@ -0,0 +1,141 @@
...
+        let credential_config = CredentialConfig {
+            allowed_prefixes: Some(vec!["IMPORTER_AUTH_".into()]),
+            allowed_paths: None,
+        };
+
+        // When building the client
+        let result = build_http_client(Some(&auth), &credential_config);
+
+        // SAFETY: test-only mutation, single-threaded test binary
+        unsafe { std::env::remove_var("IMPORTER_AUTH_HTTP_TEST_TOKEN") };
```
ctron: This will not work. As tests run in parallel and must not modify the environment.

ctron (changes requested):

modules/importer/src/model/auth.rs
```diff
@@ -0,0 +1,367 @@
...
+            "my-secret"
+        );
+    }
+
+    /// Verifies that an Env credential reads the named environment variable and there is no allowed prefix.
+    #[test]
+    fn credential_source_env_reads_variable() {
+        // Given an env var is set
+        // SAFETY: test-only mutation, single-threaded test binary
+        unsafe { set_var("IMPORTER_AUTH_TC6090_TEST_VAR", "env-value") };
```
ctron: We cannot use `set_var` in tests.

modules/importer/src/model/auth.rs
```diff
@@ -0,0 +1,367 @@
...
+            allowed_prefixes: vec!["IMPORTER_AUTH_".into()],
+            allowed_paths: vec![],
+        };
+        // When resolving
+        let src = CredentialSource::Env("IMPORTER_AUTH_TC6090_TEST_VAR".into());
+
+        // Then the value matches the env var
+        assert_eq!(src.resolve(&credential_config).unwrap(), "env-value");
+        // SAFETY: test-only mutation, single-threaded test binary
+        unsafe { remove_var("IMPORTER_AUTH_TC6090_TEST_VAR") };
```
ctron: We cannot use `remove_var` in tests.

modules/importer/src/model/auth.rs
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    /// Verifies that an Env credential errors when the variable is not set.
+    #[test]
+    fn credential_source_env_errors_when_not_set() {
+        // Given an env var is not set
+        // SAFETY: test-only mutation, single-threaded test binary
+        unsafe { remove_var("IMPORTER_AUTH_TC6090_MISSING_VAR") };
```
ctron: We cannot use `remove_var` in tests.

modules/importer/src/model/auth.rs
```diff
@@ -0,0 +1,367 @@
...
+        assert!(
+            matches!(src.resolve(&credential_config), Err(AuthError::EnvVarNotAllowed { name , allowed_prefixes }) if name == "NOT_ALLOWED_VAR" && allowed_prefixes == credential_config.allowed_prefixes)
+        );
+    }
+
+    /// Verifies that an Env credential is valid when the variable matches the allowed prefix.
+    #[test]
+    fn credential_source_env_is_valid_when_allowed() {
+        // SAFETY: test-only mutation, single-threaded test binary
+        unsafe { set_var("ALLOWED_VAR", "allowed-value") };
```
ctron: We cannot use `set_var` in tests.

modules/importer/src/model/auth.rs
```diff
@@ -0,0 +1,367 @@
...
+        let src = CredentialSource::Env("ALLOWED_VAR".into());
+        let credential_config = CredentialConfig {
+            allowed_prefixes: vec!["ALLOWED_".into()],
+            allowed_paths: vec![],
+        };
+        assert!(matches!(src.resolve(&credential_config), Ok(value) if value == "allowed-value"));
+        // SAFETY: test-only mutation, single-threaded test binary
+        unsafe { remove_var("ALLOWED_VAR") };
```
ctron: We cannot use `remove_var` in tests.

ctron (changes requested):
We can't use functions which manipulate the process for testing. The assumption in the comment (single thread) never holds up.

Alternatively, you can have a different source. One that you provide during the tests. I think the repo holds some examples already. But in a nutshell, you could add a field to the resolver which is a `impl Fn`, doing the job. Defaulting to `std::env:...`, but then override this with something like `HashMap::get`.

ruromero:
> We can't use functions which manipulate the process for testing. The assumption in the comment (single thread) never holds up.
> 
> Alternatively, you can have a different source. One that you provide during the tests. I think the repo holds some examples already. But in a nutshell, you could add a field to the resolver which is a `impl Fn`, doing the job. Defaulting to `std::env:...`, but then override this with something like `HashMap::get`.

Much better, thanks!

modules/importer/src/model/auth.rs
```diff
@@ -0,0 +1,386 @@
...
+        let dir = tempdir().unwrap();
+        let path = dir.path().join("secret.txt");
+        write(&path, "file-secret").unwrap();
+        let src = CredentialSource::File(path.to_string_lossy().into_owned());
+        let credential_config = CredentialConfig {
+            allowed_prefixes: vec![],
+            allowed_paths: vec![dir.path().to_string_lossy().into_owned()],
+        };
+        assert_eq!(
+            src.resolve(&credential_config, |name| std::env::var(name))
```
ctron:
```suggestion
            src.resolve(&credential_config, std::env::var)
```

modules/importer/src/model/auth.rs
```diff
@@ -0,0 +1,386 @@
...
+            "env-value".to_string(),
+        )]
+        .into();
+        let credential_config = CredentialConfig {
+            allowed_prefixes: vec!["IMPORTER_AUTH_".into()],
+            allowed_paths: vec![],
+        };
+        let src = CredentialSource::Env("IMPORTER_AUTH_TC6090_TEST_VAR".into());
+        assert_eq!(
+            src.resolve(&credential_config, |name| {
```
ctron: That feels like repetitive, verbose AI stuff.

Why not:

* create a trait `EnvSource` (unless we already have that)
* Give that one method: `fn resolve(&str) -> Result<String, VarError>`
* Implement that for `()` and for `HashMap<String,String>`, or better `&[(&'static str, &'static str)]`
* then just pass in a `()` when you want OS env-vars, and `[("foo", "bar")]` for tests.

add that to the common module to ensure we can re-use that in the future.

ctron (changes requested):
I think the approach is the right one, but instead of spreading the code in every function call, do it properly one.

rh-jfuller:
can this PR be revisited, updated or closed ?

ctron (approved):

PR #2649, opened 2026-09-09
build: Remove unused dependencies

rhredhat:
* Analysis done using `cargo-shear`.
* Verified using the `precommit` task using `cargo-xtask`.
* ~Removes the `vendored` feature flag from `trustd` which relied on unused dependencies. Note that they are used transitively, but the build issue mentioned in the removed comment did not occur on a local x86 linux machine.~

## Summary by Sourcery

Remove unused dependencies and the obsolete vendored build configuration.

Enhancements:
- Remove unused workspace, runtime, and development dependencies across the project to simplify dependency management.

Build:
- Update the lockfile and macOS binary builds to no longer use the removed vendored dependency configuration.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8532,7 +8532,6 @@ name = "trustify-db"
 version = "0.6.0-rc.1"
 dependencies = [
  "anyhow",
- "async-compression",
  "log",
  "postgresql_commands",
  "postgresql_embedded",
@@ -8548,7 +8547,6 @@ dependencies = [
 name = "trustify-entity"
 version = "0.6.0-rc.1"
 dependencies = [
- "anyhow",
  "cpe",
  "csaf",
  "cvss",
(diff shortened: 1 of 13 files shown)


trustd/Cargo.toml
```diff
@@ -36,13 +33,6 @@ default = ["pm"]
 bundled = ["postgresql_embedded/bundled"]
 garage-door = ["trustify-server/garage-door"]
 
-vendored = [
-    "openssl/vendored",
```
rh-jfuller, line 40: this vendored/openssl change prob needs a macOS build and its own discussion ... maybe consider leaving this change for a future PR ?

rh-jfuller (approved):
a good check to run always before commits is > **cargo xtask precommit** ... also as this PR is mainly about deps you might consider running >**cargo clean** first locally to check, otherwise LGTM

PR #2653, opened 2026-09-10
chore(migration): drop unused GiST index purl_status_vulnerability_id_gist

ruromero:
## Summary

- Adds migration `m0002360` that drops the `purl_status_vulnerability_id_gist` index
- The index uses `gist_trgm_ops` (LIKE/similarity matching) but every codebase access on `vulnerability_id` is an equality join served by existing B-tree indexes
- Confirmed 0 query planner scans and ~47 GB wasted disk in a production v2.2.6 instance
- `down()` re-creates the index for rollback safety

## Test plan

- [ ] Verify `purl_status_vulnerability_id_gist` absent from `pg_indexes` after migration runs
- [ ] Confirm B-tree (`purl_status_vuln_id_idx` / `purl_status_combo_idx`) still chosen by planner for equality joins on `vulnerability_id`
- [ ] Full integration test suite passes
- [ ] `down()` restores the index

Implements [TC-6169](https://redhat.atlassian.net/browse/TC-6169)

[TC-6169]: https://redhat.atlassian.net/browse/TC-6169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Remove the unused `purl_status.vulnerability_id` GiST index and support restoring it during rollback.

Enhancements:
- Remove the unused GiST trigram index on `purl_status.vulnerability_id` to reduce database storage overhead while retaining existing equality-query indexes.
- Provide rollback support that recreates the removed index.

Chores:
- Register the new database migration in the migration sequence.

### DIFF
--- a/migration/src/m0002350_drop_purl_status_gist_index.rs
+++ b/migration/src/m0002350_drop_purl_status_gist_index.rs
@@ -1,0 +1,43 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+#[allow(deprecated)]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        // Drop GiST trigram index on purl_status.vulnerability_id introduced
+        // in m0000050. The index uses gist_trgm_ops (substring/LIKE matching)
+        // but every access in the codebase uses equality (=), which is served
+        // by the existing B-tree indexes purl_status_vuln_id_idx and
+        // purl_status_combo_idx. The GiST index is never chosen by the
+        // query planner (0 scans in production) and wastes ~47 GB of disk.
+        manager
+            .drop_index(
+                Index::drop()
+                    .if_exists()
+                    .name("purl_status_vulnerability_id_gist")
+                    .table(PurlStatus::Table)
+                    .to_owned(),
+            )
+            .await
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .get_connection()
+            .execute_unprepared(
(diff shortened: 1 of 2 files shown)


rh-jfuller:
FWIW: these kind of proposed changes need to be manually tested with an instance of trustify, using EXPLAIN on specific queries and/or apply scale tests to 'stripe' load on instance and view if indexes are used ... we know we want to drop a few indexes ... this might be one of them - needs analysis to confirm.

migration/src/m0002350_drop_purl_status_gist_index.rs
```diff
@@ -0,0 +1,43 @@
...
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        // Drop GiST trigram index on purl_status.vulnerability_id introduced
+        // in m0000050. The index uses gist_trgm_ops (substring/LIKE matching)
+        // but every access in the codebase uses equality (=), which is served
+        // by the existing B-tree indexes purl_status_vuln_id_idx and
+        // purl_status_combo_idx. The GiST index is never chosen by the
+        // query planner (0 scans in production) and wastes ~47 GB of disk.
+        manager
+            .drop_index(
+                Index::drop()
```
rh-jfuller, line 18: FWIW this being a big index ... normally I would prefer using CONCURRENTLY but I think that interferes with migration (re TX).

rh-jfuller (approved):
when **m0001060_advisory_vulnerability_indexes** added a b-tree index this index is no longer used ... though this index was also doing _double-duty_ (used elsewhere) but subsequent changes to advisory/correlation means we no longer have any need for trigram operators ... this can be safely dropped and LGTM

rh-jfuller (approved):

PR #2658, opened 2026-09-15
[Backend] Return 503 FEATURE_UNCONFIGURED when /recommend called without TRUSTD_RECOMMEND_PATTERNS

ruromero:
## Summary

- Guards `/v2/purl/recommend` and `/v3/purl/recommend` with a 503 check when `TRUSTD_RECOMMEND_PATTERNS` is not configured on the server
- Returns structured `FEATURE_UNCONFIGURED` body so API consumers get a clear signal instead of empty results
- Makes `PurlService::recommend_patterns` `pub(crate)` for handler access
- Updates utoipa OpenAPI annotations to document the 503 response

## Test plan

- [ ] Call `/v2/purl/recommend` or `/v3/purl/recommend` without `TRUSTD_RECOMMEND_PATTERNS` set — expect HTTP 503 with `{"status":503,"code":"FEATURE_UNCONFIGURED","message":"..."}`
- [ ] Call the same endpoints with `TRUSTD_RECOMMEND_PATTERNS` configured — expect existing 200 behaviour unchanged
- [ ] `cargo check -p trustify-module-fundamental` passes

Implements [TC-6251](https://redhat.atlassian.net/browse/TC-6251)

[TC-6251]: https://redhat.atlassian.net/browse/TC-6251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Prevent unconfigured PURL recommendation endpoints from returning empty results by clearly reporting that the feature is unavailable.

Bug Fixes:
- Return a structured HTTP 503 FEATURE_UNCONFIGURED response from the v2 and v3 PURL recommendation endpoints when recommendation patterns are not configured.

Enhancements:
- Expose PURL service configuration through accessors and document the new unavailable response in the OpenAPI specifications.


modules/fundamental/src/purl/endpoints/mod.rs
```diff
@@ -108,7 +108,8 @@ mod v2 {
         tag = "purl",
         request_body = RecommendRequest,
         responses(
-            (status = 200, description = "Get recommendations and remediations for provided purls", body = RecommendResponse)
+            (status = 200, description = "Get recommendations and remediations for provided purls", body = RecommendResponse),
+            (status = 503, description = "Endpoint disabled — TRUSTD_RECOMMEND_PATTERNS not configured"),
```
rh-jfuller, line 116: ya cargo xtask precommit should be run

modules/fundamental/src/purl/service/mod.rs
```diff
@@ -122,7 +122,7 @@ impl InputPurl {
 
 pub struct PurlService {
     cache: PaginationCache,
-    recommend_patterns: Vec<Regex>,
+    pub(crate) recommend_patterns: Vec<Regex>,
```
rh-jfuller: I would prefer a func ...  instead of leaking all this to answer a boolean
impl PurlService {
    pub fn recommend_enabled(&self) -> bool {
        !self.recommend_patterns.is_empty()
    }
}

rh-jfuller (commented):
added some comments ... also I wonder if this is a change in behviour to v2 endpoints ... are we sure about this eg. that its all good ? Also might be good to add a test that exercises all this

modules/fundamental/src/purl/endpoints/mod.rs
ruromero: done

modules/fundamental/src/purl/service/mod.rs
ruromero: Change done.
Do you think this is something worth adding to `CONVENTIONS.md` ?

rh-jfuller: Not particularly - I know it is a grey area of what to put or what not to put in conventions ... I think initially people thought of using CONVENTIONS.md similar to deterministic linting ... I would rather use a deterministic linter with lots of rules like this then spend tokens on ... otherwise I see conventions.md as useful for persisting local domain (eg. this projects) idiomatic patterns modulo rust.

rh-jfuller (approved):
will need to be rebased after https://github.com/guacsec/trustify/pull/2656#pullrequestreview-5220453241 otherwise LGTM

PR #2660, opened 2026-09-16
fix(openapi): recommendReportRequest description

gildub:
The schema description documents the endpoint as POST /v3/recommend/report, but the actual path is /api/v3/recommend/report, so generated API documentation describes a non-existent URL.

Change the description to POST /api/v3/recommend/report.

## Summary by Sourcery

Align the recommend report endpoint implementation, tests, and OpenAPI documentation with its actual `/api/v3/purl/recommend/report` route.

Bug Fixes:
- Correct the recommend report endpoint to use the `/api/v3/purl/recommend/report` path consistently across the service and generated OpenAPI documentation.

Documentation:
- Update endpoint and request schema descriptions to reference the correct recommend report URL.

Tests:
- Update recommend report endpoint tests to use the corrected URL.


ruromero (changes requested):
Thanks for checking it.
I think you are missing `cargo xtask precommit`

ctron (changes requested):
And I think the AI is right. The path is wrong.

openapi.yaml
```diff
@@ -30,6 +30,27 @@ paths:
                     type: boolean
                   version:
                     type: string
+  /api/api/v3/recommend/report:
```
ruromero: This doesn't seem right

modules/fundamental/src/purl/endpoints/mod.rs
```diff
@@ -183,7 +183,7 @@ mod v3 {
             (status = 413, description = "Total package count across requested SBOMs exceeds the configured limit"),
         )
     )]
-    #[post("/v3/recommend/report")]
+    #[post("/api/v3/recommend/report")]
```
ruromero: This causes the double `/api` in the openapi.yaml file
```suggestion
    #[post("/v3/recommend/report")]
```

ruromero (changes requested):
I have added a suggestion that would fix the openapi.yaml after running again the `cargo xtask precommit`
A side issue is that by mistake the path is `/recommend/report` and should be part of the `/purl/recommend/report`

gildub:
@ruromero, should be ready besides CI failing...

ruromero (approved):
LGTM
Thanks!

ruromero (approved):

ctron (approved):

PR #2668, opened 2026-09-17
feat(ingestor): accept CycloneDX 1.7 and ingest extended license details into licensing_infos

rhredhat:
## Summary

Two related changes to CycloneDX ingestion:

1. **Accept CycloneDX 1.7 SBOMs** (`TC-6061`) — widen the ingestion spec-version gate so 1.7 documents are detected and ingested instead of rejected as an unsupported format, and list 1.7 in the `UnsupportedFormat` error message.
2. **Ingest extended license details into `licensing_infos`** (`TC-6068`) — CycloneDX ingestion previously recorded only a component's license expression and never wrote `licensing_infos`, which until now was populated for SPDX only. As a result License Analysis could not surface license names, texts or URLs for CycloneDX SBOMs, and `LicenseRef`s never resolved.

### Accepting 1.7

1.7-only fields are ignored by the existing `v_1_6` deserialization path. Per ECMA-424 conformance, a consumer is not required to interpret all content in a BOM and must not error on a conforming BOM, so the additive gate is spec-correct. Adopting the `v_1_7` model is tracked separately.

### `licensing_infos`

Each named license object is recorded in `licensing_infos`, keyed on its `bom-ref`, falling back to the license `name`. That key is what `extract_license_ref_mappings` looks up from the parsed SPDX expression, so it is the identifier expansion needs.

- Licenses carrying only an SPDX `id` are skipped — they are self-describing and need no mapping.
- License text is an `attachment`, so it is stored verbatim or base64-decoded when `encoding` says so; text that fails to decode (or is not UTF-8) is skipped and logged rather than stored raw.
- `populate_expanded_license` already joins `licensing_infos` by `sbom_id` alone, so `LicenseRef` expansion starts working for CycloneDX with no change there.
- `licensing_infos` is inserted before `licenses` in `ComponentCreator::create`, matching the SPDX loader's table order — the comment there warns that all SBOM loaders must agree in order to avoid cross-table deadlocks during concurrent ingestion.

Worth noting: the `license` object is **unchanged between CycloneDX 1.6 and 1.7**. The only differences in the schema are editorial prose plus the hoisting of the inline `licensing` object into a named definition. So this needs no `v_1_7` model and applies to every ingested CycloneDX version, not just 1.7.

## Test plan

New:

- `sbom::cyclonedx::parse_cyclonedx_1dot7` and `detect_cyclonedx_1dot7`, with a `simple_1dot7.json` fixture.
- `license::service::test::test_cyclonedx_1dot7_extracted_licensing_infos`, with a `licenses_1dot7.json` fixture covering a `bom-ref` + plain text + URL license, a base64-encoded text license, a name-only license, and an SPDX-`id`-only license that must **not** be recorded.

Changed: the two `extracted_licensing_infos` assertions in `modules/fundamental/tests/sbom/license.rs` move from `0` to `4` — `application.cdx.json` carries four distinct named non-SPDX licenses that are now recorded.

Verified locally:

- `cargo test -p trustify-module-fundamental license` — 33 lib + 4 integration tests pass.
- `cargo test -p trustify-module-fundamental cyclonedx` — 5 lib + 16 integration tests pass.
- `cargo clippy --workspace --all-targets -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -W clippy::absolute_paths` — clean.
- `cargo fmt --all --check` and `cargo xtask precommit` — clean, no generated-file drift.

## Summary by Sourcery

Support CycloneDX 1.7 ingestion and persist extended license details for CycloneDX components.

New Features:
- Accept CycloneDX 1.7 documents during format detection and ingestion.
- Capture extended CycloneDX license metadata, including names, text, URLs, and identifiers, in `licensing_infos` for license analysis and `LicenseRef` expansion.

Bug Fixes:
- Enable CycloneDX license details and non-SPDX license references to appear in license analysis results.

Build:
- Add the base64 dependency required to decode encoded CycloneDX license text.

Tests:
- Add CycloneDX 1.7 detection, ingestion, and extended-license coverage, including plain-text and base64-encoded attachments and identifier-only exclusions.


rh-jfuller (commented):
add @guacsec/trustify-maintainers as reviewers if this is ready for review and rebase ;) also need to identify which branches this might be backported to (I will assume 0.6.z but that needs confirming).

rhredhat:
@rh-jfuller Can't seem to add maintainers or labels but the release is 0.6.z

rh-jfuller:
checked with @ptomanRH this is not for 0.6.z ;) I added maintainers for review

ctron (approved):

rh-jfuller:
scale test seems to indicate license endpoints sped up by quite a bit https://guacsec.github.io/trustify-scale-test-runs/reports/report-2026-09-21T16-33-15.html

PR #2670, opened 2026-09-17
fix(ingestor): create_doc to update timestamp when reingesting

gildub:
Fix https://redhat.atlassian.net/browse/TC-2695

Reload an already uploaded document should be reported as a duplicate to be able to inform the end user accordingly.

(The initial attempt approach by updating the timestamp was wrong because the `ingestor`  is idempotent.)

- model.rs : Added pub duplicate: bool to IngestResult. Serialized only when true (skipped when false to keep the API non-breaking for existing clients).

- graph/mod.rs :  Reverted the timestamp refresh in create_doc. Re-upload of the same document is now a true no-op: no DB mutation, just returns the existing entity. Also removed the now-unused Expr and ColumnTrait/QueryFilter imports.

- csaf/loader.rs : The Outcome::Existed branch from ingest_advisory now returns immediately with duplicate: true. No separate early digest check is needed since create_doc (inside ingest_advisory) handles deduplication.

- spdx.rs, cyclonedx.rs, cve/loader.rs, osv/loader.rs : Same pattern: short-circuit on Outcome::Existed with duplicate: true. This also makes CVE and OSV loaders skip unnecessary re-linking work on duplicate uploads.

All other loaders (NVD, KEV, CWE, ClearlyDefined) : Added duplicate: false to their IngestResult constructions.
-  CVE loader — moved ingest_advisory before vuln_creator.create. Now on a duplicate, the function returns before any vulnerability upsert or other mutations occur.

- NVD loader — same reorder: ingest_advisory (with Outcome::Existed early return) now runs before vuln_creator.create, link_to_vulnerability, scores, CPEs, and descriptions. Added Outcome to imports.

- ClearlyDefined — was already detecting Outcome::Existed correctly (no mutations before ingest_sbom), but was returning duplicate: false. Now returns duplicate: true and exits early via the same pattern as the other loaders.

- ClearlyDefinedCuration — same fix as ClearlyDefined.

The UI can now check `result.duplicate === true` to show "this document was already uploaded.

## Summary by Sourcery

Report duplicate document uploads explicitly while preserving idempotent ingestion behavior.

New Features:
- Expose whether an ingestion request was a duplicate through the optional `duplicate` result field and OpenAPI schema.

Bug Fixes:
- Ensure reingesting an existing document is reported as a duplicate while remaining an idempotent no-op.
- Avoid unnecessary vulnerability, SBOM, and advisory processing when duplicate documents are uploaded.

Enhancements:
- Standardize duplicate handling across advisory, SBOM, KEV, CWE, and ClearlyDefined loaders while preserving backward compatibility for existing API clients.

Documentation:
- Document the new duplicate ingestion result field in the OpenAPI specification.


rh-jfuller (commented):
CSAF needs to be looked at because its loader has a format-specific early duplicate check:
```
if let Some(found) = self.graph.get_advisory_by_digest(&sha256, tx).await? {
    return Ok(existing_result);
}
```

so **create_doc** only refreshes timestamps when its insert gets a duplicate-key error:
1. Upload enters CsafLoader::load
2. Lines 99-107 query by digest
3. If the advisory already exists, the loader immediately returns its ID
4. ingest_advisory is never called!

**create_doc** /w new timestamp update never executes

gildub:
Thank you @rh-jfuller for catching this.

Adding timestamp refresh logic inside create_doc allows a duplicate-key DB error detection on insert to work correctly for CVE loader and SBOM loaders but the CSAF loader.

After analysis (with Claude's help), it seems https://github.com/guacsec/trustify/blob/main/modules/ingestor/src/service/advisory/csaf/loader.rs#L98-L107 early `get_advisory_by_digest` check gets neither the deduplication guarantee nor the timestamp update.
Thus the ingest_advisory is never called and therefore create_doc never runs so the ingested timestamp is never updated for CSAF documents.

By removing the above, the following self.graph.ingest_advisory call gets the same Outcome::Existed path via create_doc, which includes the timestamp refresh.

ctron:
Why would we want to do this? This should be a no-op.

gildub:
> Why would we want to do this? This should be a no-op.

@ctron, you're right since the ingestor is designed to be idempotent the right approach would be to inform the user with a warning like "Document already exists, no new data was ingested".

I've modified this PR accordingly.

WDYT

gildub:
@sourcery-ai summary.

rh-jfuller (commented):
looking at latest code I think dupe uploads still report duplicate: false for NVD and ClearlyDefined also  CVE duplicates still mutate the shared vulnerability before deduplication.  

FWIW intended no-op (?@ctron) is logical ingestion once source doc digest already exists eg. my interpretation of no op is:
* do not create / update advisory/SBOM
* do not upsert shared vulnerabilities
* do not recreate links, scores, descriptions, packages, statuses, or licenses ...
* do not emit Added change event (or reload analysis cache)
* and should return existing entity ID with duplicate: true

I guess the content addressed storage write still occurs because hashing currently happens during storage.store but theoretically should not change state

gildub:
@sourcery-ai summary.

ctron (approved):
Thanks!

PR #2683, opened 2026-09-21
fix(bulk-delete): single and bulk delete endpoints to behave identically

gildub:
Fix https://redhat.atlassian.net/browse/TC-5382 

 Changed: modules/fundamental/src/sbom/endpoints/mod.rs — replaced Id::from_str(&id)? (which propagated a 404 on invalid IDs) with an else branch that returns 204 directly.

Test added: modules/fundamental/src/sbom/endpoints/test.rs — added a case in delete_sbom asserting that a nonsensical ID ("blabla") returns 204, not 404.

The single delete endpoint now behaves identically to the bulk delete: any invalid or non-existent ID is treated as "already gone" and returns 204, consistent with REST idempotency semantics and the team's stated intent.

Claude assisted


ctron (approved):
Good catch!

PR #2684, opened 2026-09-21
feat(cyclonedx): adopt the CycloneDX 1.7 model

rhredhat:
Migrate every `serde_cyclonedx::cyclonedx::v_1_6` binding to `v_1_7` and move the sbom-walker preflight to `Sbom::V1_7`, so 1.7-specific fields are available to the mapper instead of being dropped.

Two model changes needed adapting:

* Licenses moved the union from array level to item level. In 1.6 `LicenseChoice` was a single enum over the whole array — either all license objects or all expressions. In 1.7 it is `Vec<LicenseChoiceItemUrl>`, so one `licenses` array may mix both. `Information::data_licenses` and `ComponentCreator::add_license` now match per item; the id -> name -> skip precedence is unchanged.
* `HashAlg` gained `Streebog-256` and `Streebog-512`.

`serde-cyclonedx` and `sbom-walker` point at forks until 1.7 support is released to crates.io; `csaf-walker` and `walker-common` move to the same rev so a single `walker-common` is in the graph, and both git sources are allow-listed in deny.toml. All three pins are temporary.

## Summary by Sourcery

Adopt CycloneDX 1.7 across the SBOM pipeline while preserving mixed license data and newly supported hash algorithms.

New Features:
- Adopt the CycloneDX 1.7 model throughout SBOM parsing, ingestion, migration, detection, UI, and tests.
- Support CycloneDX 1.7 license arrays containing mixed license objects and expressions.
- Recognize the CycloneDX 1.7 Streebog-256 and Streebog-512 hash algorithms.

Enhancements:
- Preserve CycloneDX 1.7-specific fields by running SBOM preflight checks against the 1.7 model.
- Update license extraction while retaining identifier, name, and skip precedence.

Build:
- Temporarily pin CycloneDX and walker dependencies to git revisions pending upstream crate releases.

Tests:
- Update CycloneDX ingestion tests and benchmarks to use the 1.7 model.

Chores:
- Allow the temporary git dependency sources in dependency policy configuration and align walker dependencies to a single revision.


rh-jfuller:
we have not released this func yet, so no need for a migration @rhredhat can you rebase and lets see if we can get this landed

rh-jfuller (approved):
LGTM

Cargo.toml
```diff
@@ -115,15 +115,15 @@ ring = "0.17.8"
 roxmltree = "0.21.1"
 rstest = "0.26.1"
 sanitize-filename = "0.6.0"
-sbom-walker = { version = "0.18.1", default-features = false, features = ["crypto-openssl", "serde-cyclonedx", "spdx-rs"] }
+sbom-walker = { git = "https://github.com/rhredhat/csaf-walker.git", rev = "9ad8958ee" , default-features = false, features = ["crypto-openssl", "serde-cyclonedx", "spdx-rs"] }
```
ctron: Stuff like this must go into the patch section!

ctron (changes requested):

